Issue #3, opened 2020-09-18
AFLplusplus Support


domenukk:
Great project!
As development on afl itself doesn't seem to be high priority anymore, it would be nice to see out of the box support for additional afl-based fuzzers, such as afl++.

bmc-msft:
User-provided custom fuzzers is a first-class supported feature.  We have a few examples, including [honggfuzz](https://github.com/microsoft/onefuzz/blob/main/src/cli/examples/honggfuzz.py), [domato](https://github.com/microsoft/onefuzz/blob/main/src/cli/examples/domato.py), and even [compiling an libfuzzer targets from oss-fuzz integrated projects](https://github.com/microsoft/onefuzz/blob/main/src/cli/examples/oss-fuzz-target.py).

That said, including support for AFL++ as a built-in path should be straight forwards.

Given the large number of ongoing research efforts based on AFL, we should make it stream-lined to let users pick from an arbitrary set of AFL implementations, including making it easy to add new builds, versions, or forks of AFL.

PR #7, opened 2020-09-19
afl++ addition

vanhauser-thc:

Adding afl++ as a default fuzzer, see #3 

## Info on Pull Request

adding afl++, basically just copying afl.* files to aflpp.* and making a few changes plus updating documentation

## Validation Steps Performed

I am unsure if this is all that is needed. if so, sweet :)


bmc-msft:
Thanks for the contribution!  With a few adjustments, this would be a beneficial contribution:
1. The additional build script you added, `src/ci/aflpp.sh`, needs to be added to `.github/workflows/ci.yml` to execute.
2. We pin all redistributed third-party tools needs to specific revisions.  As AFL++ does regular releases, can you update to use a specific release, such as https://github.com/AFLplusplus/AFLplusplus/archive/2.68c.tar.gz?

Rather than adding another CLI subcommand to launch afl++, it would be useful to expose another command line argument to the existing 'afl' subcommand that allows the user to specify which AFL binary to execute.  The auto-generation of CLI using python typing will need an update to support this easily.  I'm happy to do that part for you.

bmc-msft:
Looking into the AFL template, we already allow the user to specify alternate implementations using 'fuzzer_setup_container', though it would be useful to enable providing a list of existing containers as possible options.

```onefuzz template afl basic myproj target build-1 linux --fuzzer_setup_container aflpp-linux```

**Comment by vanhauser-thc** [comment]
1. is done.
2. we have a stable branch which is always the newest afl++ with current additions, to which we push every 2-4 weeks. currently it is 2.68c. This would prevent needing to push a new version every 1-2 months. (the checkout in this PR is to the stable branch). Would that be fine? 
3. this is a difficult question. afl and then sub commands for specific spin-offs or the specific spin-offs at the root command. IMHO It depends how compatible they are. e.g. afl++ has command line options that vanilla afl does not understand, also some environment variables. what do you think?

**Comment by bmc-msft** [comment]
1. I saw the update.  Thanks!  The `llvm_mode` does not compile.  Based on the output, I think the github build VMs does not include the llvm-config prerequisite.  In the short term, I think adding `apt-get install llvm-config` or `apt-get install llvm` will address this issue.
2. Keeping third-party versions up-to-date is a requirement of the Onefuzz team.  We are required to keep track of the version and license of every component that gets integrated.
3. The need for researchers to specify custom arguments & environment variables is entirely reasonable and supported.   I don't think we will want to try to keep up with fuzzing researcher's novel arguments & environment variables.  Rather, our goal is to make it easy to specify anything custom, while providing basic sensible defaults as needed.

   In the afl commandline plugin, we provide a basic set of arguments to pass to afl-fuzz via `--supervisor_options`.  By default for AFL, these are `-d -i {input_corpus} -o {runtime_dir} -- {target_exe} {target_options}`, where the [onefuzz-agent replaces the '{str}' at runtime](https://github.com/microsoft/onefuzz/blob/main/docs/command-replacements.md).  Users can similarly specify environment variables via `--supervisor_env KEY=VALUE KEY2=VALUE2`.

File: src/cli/onefuzz/templates/aflpp.py [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+from onefuzztypes.enums import OS, ContainerType, StatsFormat, TaskType
+from onefuzztypes.models import NotificationConfig
+from onefuzztypes.primitives import Container, Directory, File
+
+from onefuzz.api import Command
+
+from . import JobHelper
+
+
+class AFL(Command):
```
ranweiler, line 17: This class isn't used or exposed anywhere yet. FYI, to actually expose it as a subcommand of `template`, it would need to instantiated and [bound to `self`](https://github.com/microsoft/onefuzz/blob/31f4c51b87b542ef958e162f1ff55cb8c3016bc8/src/cli/onefuzz/template.py#L24) in `.template.Template.__init__()`.

ranweiler: Also, we'd probably want to name this something like `AFLPlusPlus` to prevent ambiguity in the existing use site linked above.

docs/tasks.md
```diff
@@ -56,7 +56,7 @@ include:
 * supervisor_env: User specified environment variables for the supervisor
 * supervisor_options: User specified command line options for the supervisor
 * supervisor_input_marker: Marker to specify the path to the filename for
-  supervisors (Example: for AFL, this should be '@@')
+  supervisors (Example: for AFL and AFL++, this should be '@@')
```
ranweiler, line 59: :+1: thanks for actually updating the docs!

ranweiler:
To note another option here: we could add a `plusplus` (sub-)subcommand to the existing `AFL` template. This would just mean adding a `plusplus` method to the existing `.afl.AFL` class.

Exposing AFL++ as a method is compatible with _arbitrary_ API differences between AFL and AFL++. It does not require any actual tool similarity. Sharing the `afl` part really just provides namespacing, and _if desired_, a convenient way to share setup code between templates (via the `AFL` class).

With that approach, invoking would look like:
```
onefuzz template afl plusplus <...>
```

vanhauser-thc:
I renamed the class, pinned the checkout to v2.68c and installed the llvm components.

If there is any python stuff to do ... I can't do that. I can do C, C++, Assembly, even Perl ... but I stayed away from Python and anything beyond simple script changes is not something I feel comfortable to do :)

docs/tasks.md
vanhauser-thc: current docs are key.

bmc-msft:
> I renamed the class, pinned the checkout to v2.68c and installed the llvm components.

Can you prefix the 'apt' calls with sudo?

> If there is any python stuff to do ... I can't do that. I can do C, C++, Assembly, even Perl ... but I stayed away from Python and anything beyond simple script changes is not something I feel comfortable to do :)

That's fine.  Can you delete the python code?  I'll take on exposing afl++ via CLI in a meaningful fashion.

vanhauser-thc:
sure! done.

src/ci/aflpp.sh
```diff
@@ -0,0 +1,19 @@
...
+sudo apt-get install -y llvm llvm-dev clang
+
+git clone https://github.com/AFLplusplus/AFLplusplus
+cd AFLplusplus
+# checkout v2.68c
+git checkout ee206da3897fd2d9f72206c3c5ea0e3fab109001
+make
+(test -d llvm_mode && cd llvm_mode && make)
+#(test -d gcc_plugin && cd gcc_plugin && make)
+(cd examples/libdislocator && make)
```
bmc-msft, line 17: libdislocator is in AFLplusplus, not in examples.   Additionally, the resulting file is in AFLplusplus, not AFLplusplus/examples/libdislocator

This line should be:

 `(cd libdislocator && make)`

src/ci/aflpp.sh
```diff
@@ -0,0 +1,19 @@
...
+
+cp -rf afl-* *.so *.a *.o dictionaries examples/libdislocator/libdislocator.so LICENSE ../artifacts/third-party/aflpp-linux
```
bmc-msft, line 19: This line should be:

`cp -rf afl-* *.so *.a *.o dictionaries libdislocator.so LICENSE ../artifacts/third-party/aflpp-linux`

bmc-msft (changes requested):

src/ci/aflpp.sh
vanhauser-thc: you would think I would know that ;)

src/ci/aflpp.sh
vanhauser-thc: done

bmc-msft (approved):

PR #10, opened 2020-09-21
Enable documentation of CLI sub-command arguments

bmc-msft:

This PR enables adding documentation to CLI sub-command arguments via docstring_parser.

## PR Checklist
* [X] Applies to work item: #7 
* [X] Tests added/passed

## Info on Pull Request

1. This includes adding docstring_parser as a prereq to the CLI.
2. Adds  support for parsing the docstrings of functions, and optionally passing the help strings as argument documentation if the documentation doesn't already exist.

## Validation Steps Performed

Run `onefuzz template afl basic --help |grep -i afl`

Previously, the argument 'fuzzer_setup_container' wasn't documented beyond "CONTAINER".  Now, afl_container says "Specify the AFL container to use in the job"

### DIFF
--- a/src/cli/onefuzz/templates/afl.py
+++ b/src/cli/onefuzz/templates/afl.py
@@ -47,7 +47,7 @@ class AFL(Command):
         tags: Optional[Dict[str, str]] = None,
         wait_for_running: bool = False,
         wait_for_files: Optional[List[ContainerType]] = None,
-        afl_version: Optional[Container] = None,
+        afl_container: Optional[Container] = None,
         existing_inputs: Optional[Container] = None,
         dryrun: bool = False,
         notification_config: Optional[NotificationConfig] = None,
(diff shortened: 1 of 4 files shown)


src/cli/onefuzz/templates/afl.py
```diff
@@ -106,13 +110,16 @@ def basic(
...
 
-        if fuzzer_setup_container is None:
-            fuzzer_setup_container = Container(
+        if afl_container is None:
+            afl_container = Container(
                 "afl-linux" if helper.platform == OS.linux else "afl-windows"
             )
 
+        # verify the AFL container exists
+        self.onefuzz.containers.get(afl_container)
```
ranweiler, line 119: It'd be nice to catch the `ValidationError` on failure and emit a specific and useful message to the user.

ranweiler (approved):

src/cli/onefuzz/templates/afl.py
bmc-msft: By using `onefuzz.containers.get` to do the verification, we get an Exception of 'invalid container' which comes to the user as:
 
`ERROR:cli:command failed: request did not succeed: HTTP 400 - {"code": 450, "errors": ["invalid container"]}`

The error handler in the SDK should get updated to do something better, but that's outside of the scope for this PR.

PR #19, opened 2020-09-22
finish scaleset state refactor

bmc-msft:

Finishes the pre-1.0 scaleset state refactor

## Info on Pull Request

1. Only executes `scaleset.update_configs` once every 24 hours, instead of once every 90 seconds.
2. Removes the ORM state-based event queuing that was replaced by `scaleset_events`

## Validation Steps Performed

1. Create a new pool
2. Create a scaleset as part of the new pool
3. Assign a task to the new pool, verify it gets executed

### DIFF
--- a/src/api-service/__app__/daily_events/__init__.py
+++ b/src/api-service/__app__/daily_events/__init__.py
@@ -9,6 +9,7 @@ import azure.functions as func
 from onefuzztypes.enums import VmState
 
 from ..onefuzzlib.proxy import Proxy
+from ..onefuzzlib.pools import Scaleset
 
 
 def main(mytimer: func.TimerRequest) -> None:  # noqa: F841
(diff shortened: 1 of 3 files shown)


src/api-service/__app__/daily_events/__init__.py
```diff
@@ -17,3 +18,8 @@ def main(mytimer: func.TimerRequest) -> None:  # noqa: F841
             logging.info("stopping proxy")
             proxy.state = VmState.stopping
             proxy.save()
+
+    scalesets = Scaleset.search()
+    for scaleset in scalesets:
+        logging.info("updating scaleset configs: %s", scaleset.scaleset_id)
+        scaleset.update_configs()
```
ranweiler, line 25: As part of a general incremental effort to make things more approachable: an we add a comment here and/or to `Scaleset.update_configs()` to make it more obvious what this is doing, and why it is reasonable to do on this schedule?

ranweiler (approved):

Issue #18, opened 2020-09-22
Full path of the input file


tunz:
In the command replacement, `{input}` provides a relative path of the input file such as `generator_tmp/input`. But, some binaries requires a full path of the file.

I think `{input}` needs to be replaced to the absolute path rather than a relative path. What do you think?

bmc-msft:
I'm in support.

Using the full, canonicalized paths would address multiple issues beyond this one, including targets that change their directory prior to opening the input file.  

We probably want to update all of the Path variables in [Expand](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz/src/expand.rs) should be updated to support the full canonicalized path.

PR #22, opened 2020-09-22
Use absolute path for command replacements

tunz:

Onefuzz will replace command replacement values such as `{input}` to absolute paths instead of relative path.

## PR Checklist
* [x] Applies to work item: #18
* [x] Tests added/passed

## Info on Pull Request

this will get canonicalized paths when they are evaluated. @bmc-msft this is what I described. This also handles the case that `target_options` has `{input}` well.

## Validation Steps Performed

Updated `test_expand` testcase.


src/agent/onefuzz/src/expand.rs
```diff
@@ -85,7 +86,7 @@ impl<'a> Expand<'a> {
 
     fn extract_file_name_no_ext(&self, _format_str: &str) -> Option<ExpandedValue<'a>> {
         match self.values.get(&PlaceHolder::Input.get_string()) {
-            Some(ExpandedValue::Scalar(fp)) => {
+            Some(ExpandedValue::Path(fp)) | Some(ExpandedValue::Scalar(fp)) => {
```
bmc-msft, line 89: Since you set `PlaceHolder::Input` to an `ExpandedValue::Path`, you shouldn't need the or case for Scalar here.

bmc-msft (approved):

Issue #13, opened 2020-09-21
On task setup script failure, tasks should be marked as "failed" and nodes should be re-imaged


bmc-msft:
## Information

- Onefuzz version: 1.0.0
- OS: Ubuntu 18.04

## Provide detailed reproduction steps (if any)

1. Launch a task with a custom setup script (setup.sh) with a failure (say, 'exit 1')

### Expected result

1. Task state set to failed.  Stderr/stdout from the setup script should be added to `task.error`
2. Node is appropriately reimaged (`/onefuzz/<task_id>` no longer exists)

### Actual result

1. Task state is `scheduled`
2. Verify `/onefuzz/<task_id>` on the node the task was assigned

PR #23, opened 2020-09-22
Emit setup script output from `ISetupRunner`

ranweiler:

Change the internal `ISetupRunner` interface to return a serializable representation of the script process output. In a future PR, we will consume this to reflect script failure in a new scheduler state transition.

## PR Checklist
* [x] Applies to work item: #13 
* [ ] Tests added/passed

## Validation Steps Performed

- [x] Tested manually


src/agent/onefuzz-supervisor/src/process.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        Self {
+            exit_status,
+            stderr,
+            stdout,
+        }
+    }
+}
+
+/// Serializable representation of a worker exit status.
```
ranweiler, line 23: Update doc comment.

src/agent/onefuzz-supervisor/src/setup.rs
```diff
@@ -75,31 +78,33 @@ impl SetupRunner {
...
                 );
                 info!("setup script succeeded");
             } else {
                 error!(
                     "setup script failed:\nstdout = {}\nstderr = {}",
-                    String::from_utf8_lossy(&output.stdout),
-                    String::from_utf8_lossy(&output.stderr),
+                    &output.stdout, &output.stderr,
                 );
-                anyhow::bail!("setup script failed:\nstdout = {}\nstderr = {}");
```
ranweiler, line 96: This without the upcoming changes results in us silently ignoring a script failure. Rather than impl-ing a `bail!` where this is used, I'm just going to aim the next PR at this one for easy holistic review.

src/agent/onefuzz-supervisor/src/process.rs
```diff
@@ -0,0 +1,51 @@
+/// Serializable representation of a process output.
```
ranweiler, line 1: Add license.

src/agent/onefuzz-supervisor/src/setup.rs
ranweiler: Just reverted for now, so we can freely merge this.

src/agent/onefuzz-supervisor/src/scheduler.rs
```diff
@@ -131,9 +131,15 @@ impl State<SettingUp> {
     pub async fn finish(self, runner: &mut dyn ISetupRunner) -> Result<SetupDone> {
         let work_set = self.ctx.work_set;
 
-        let reboot = runner.run(&work_set).await?;
+        let output = runner.run(&work_set).await?;
 
-        let done = if reboot {
+        if let Some(output) = output {
```
ranweiler, line 136: A non-success exit status is no longer an error within the `ISetupRunner`, so to preserve existing behavior, we need to scrutinize the script output here (if it exists).

src/agent/onefuzz-supervisor/src/scheduler.rs
```diff
@@ -131,9 +131,15 @@ impl State<SettingUp> {
...
+            if !output.exit_status.success {
+                anyhow::bail!("setup script failed: {:?}", output);
```
chkeita, line 138: Should we really be throwing an exception here? I think this should be encoded as a specific state like SetupDone::ScriptFailed.
IMHO throwing an exception is supposed to indicate that something went wrong with this code which is not the case.

ranweiler: This just preserves the existing behavior; this is the first in a series of PRs to change that behavior. For example, we add new detail to the `Done` state in a forthcoming PR, and then respond to it in the service event handler.

chkeita (approved):

PR #24, opened 2020-09-22
Transition to `Done` state on setup error, trigger node reimaging

ranweiler:

On target setup failure, transition the node to an explicit state, and emit and an event in the service.

## PR Checklist
* [x] Applies to work item: #13

## Validation Steps Performed

- [ ] Valid jobs work
- [ ] If target setup script exits nonzero, task is stopped (on all nodes!), nodes reimaged


bmc-msft:
This should tie back to the WorkSet that was part of the failure.  We need to use the Done messages to mark each of the tasks as failed.

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -82,6 +88,17 @@ def on_state_update(
...
+                # Model-validated.
+                done_data = cast(NodeDoneEventData, state_update.data)
+
+                if done_data.error:
+                    logging.error(
+                        "node done with error disposition: machine_id = %s, data = %s",
+                        machine_id,
+                        done_data,
+                    )
+                    # TODO: update task states linked to node.
```
ranweiler, line 101: Other than logging do we want to do something here, or let the normal `done` machinery pick up the failure?

src/agent/onefuzz-supervisor/src/scheduler.rs
```diff
@@ -129,17 +145,36 @@ impl State<Free> {
...
-        let output = runner.run(&work_set).await?;
-
-        if let Some(output) = output {
-            if !output.exit_status.success {
-                anyhow::bail!("setup script failed: {:?}", output);
+        let output = runner.run(&work_set).await;
+
+        match output {
+            Ok(Some(output)) => {
+                if !output.exit_status.success {
```
ranweiler, line 159: With this change, if the setup script fails, we transition to `Done`, with context.

ranweiler: This branch is for when the setup script ran, but had a nonzero exit.

src/agent/onefuzz-supervisor/src/scheduler.rs
```diff
@@ -129,17 +145,36 @@ impl State<Free> {
...
+                        script_output: Some(output),
+                    };
+                    let ctx = Done { cause };
+                    return Ok(SetupDone::Done(ctx.into()));
+                }
+            }
+            Ok(None) => {
+                // No script was executed.
+            }
+            Err(err) => {
```
ranweiler, line 171: This branch is for when we failed to run the setup script at all.

src/pytypes/onefuzztypes/models.py
```diff
@@ -500,28 +506,43 @@ class WorkerEvent(EnumModel):
...
-    ) -> Optional[SettingUpEventData]:
-        if data:
-            state = values.get("state")
-            if state and state != NodeState.setting_up:
-                raise ValueError(
-                    "data for node state update event does not match state = %s" % state
-                )
+    data: Optional[NodeStateData]
+
+    @root_validator(pre=False, skip_on_failure=True)
```
ranweiler, line 525: I haven't actually tested this yet.

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -82,6 +88,16 @@ def on_state_update(
                             state=NodeTaskState.setting_up,
                         )
                         node_task.save()
+            elif state == NodeState.done:
+                # Model-validated.
+                done_data = cast(NodeDoneEventData, state_update.data)
```
ranweiler, line 93: This should also be guarded on `state_update.data`, since we let it be optional.

ranweiler: Fixed.

src/api-service/__app__/agent_events/__init__.py
ranweiler: If we want to indicate task + node-task failure immediately, this is where we'd do it. Because we've established the node-task link in the `setting_up` event, we can query table storage for the tasks that were assigned to the node.

Question: do we want to also make sure we stop _other_ occurrences of the task on other nodes (if `vm_count` > 1), or even the job? Probably, because we've discussed "for tasks with multiple `vm_count`, the state should be the most-progressed or most-failed state".

src/pytypes/onefuzztypes/models.py
ranweiler: We expect the `state_update` events to be adjacently tagged.

For now, `state_update.data` is optional, even for events which have a case in `NodeEventData`. This is for back compat.

In the future, we can make the `data` field mandatory for states that have data.

Examples:
Initial `free` start transition:
```json
{
    "state": "free"
}
```
Non-legacy enriched `setting_up` transition:
```json
{
    "state": "setting_up",
    "data": {
        "tasks": ["c6927d49-479d-4eda-a0f0-7084a31459da"]
    }
}
```
Legacy `done` transition, with no data:
```json
{
    "state": "done",
    "data": null
}
```

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -82,6 +92,20 @@ def on_state_update(
...
                         )
                         node_task.save()
+            elif state == NodeState.done:
+                # Model-validated.
+                #
+                # This field will be required in the future.
+                # For now, it is optional for back compat.
+                done_data = cast(Optional[NodeDoneEventData], state_update.data)
+
+                if done_data:
```
ranweiler, line 102: This check is what lets us detect server-side that a node went to `done` because the setup script failed. Legacy `done` events won't have a `data` payload, so we can't tell why they transitioned. But they'll also never have transitioned due to a failed setup script (that's new to this PR), so it doesn't matter.

bmc-msft (approved):

PR #29, opened 2020-09-23
Refactor internal node event schemas

ranweiler:

Refactor the schema of events emitted by the agent. The new schema is meant to be more extensible, and anticipates upcoming changes where we will include a data payload with node state updates.

### DIFF
--- a/src/agent/onefuzz-supervisor/src/coordinator.rs
+++ b/src/agent/onefuzz-supervisor/src/coordinator.rs
@@ -18,10 +18,10 @@ pub struct StopTask {
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
-#[serde(tag = "command_type")]
+#[serde(rename_all = "snake_case")]
 pub enum NodeCommand {
-    #[serde(alias = "stop")]
     StopTask(StopTask),
+    Stop {},
 }
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
@@ -33,6 +33,17 @@ pub struct NodeCommandEnvelope {
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct PendingNodeCommand {
     envelope: Option<NodeCommandEnvelope>,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct PollCommandsRequest {
+    machine_id: Uuid,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ClaimNodeCommandRequest {
+    machine_id: Uuid,
+    message_id: String,
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
@@ -178,7 +189,9 @@ impl Coordinator {
         let pending: PendingNodeCommand = serde_json::from_slice(&data)?;
 
         if let Some(envelope) = pending.envelope {
-            // TODO: DELETE dequeued command via `message_id`.
+            let request = RequestType::ClaimCommand(envelope.message_id);
+            self.send_with_auth_retry(request).await?;
+
             Ok(Some(envelope.command))
         } else {
             Ok(None)
@@ -216,7 +229,7 @@ impl Coordinator {
         &mut self,
         request_type: RequestType<'a>,
     ) -> Result<Response> {
-        let request = self.build_request(request_type)?;
+        let request = self.build_request(request_type.clone())?;
         let mut response = self.client.execute(request).await?;
 
         if response.status() == StatusCode::UNAUTHORIZED {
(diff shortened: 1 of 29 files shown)


src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -54,21 +54,49 @@ pub struct NodeEventEnvelope {
...
+        NodeEvent::WorkerEvent(event)
     }
 }
 
-impl From<WorkerEvent> for NodeEvent {
-    fn from(event: WorkerEvent) -> Self {
-        NodeEvent::WorkerEvent { event }
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case", tag = "state")]
+pub enum StateUpdateEvent {
```
ranweiler, line 71: This may look a little redundant, but in an upcoming PR, we will add additional data to some of these events.

src/api-service/__app__/onefuzzlib/task_event.py
```diff
@@ -36,12 +31,20 @@ def get_summary(cls, task_id: UUID) -> List[TaskEventSummary]:
...
-    def get_event_data(cls, worker_event: WorkerEvent) -> str:
-        event = worker_event.event
-        if isinstance(event, WorkerDoneEvent):
-            return "exit status: %s" % event.exit_status
-        elif isinstance(event, WorkerRunningEvent):
-            return ""
-        else:
-            return "Unrecognized event: %s" % event
+
+def get_event_data(event: WorkerEvent) -> str:
```
ranweiler, line 35: These didn't actually need class access, so I just made them functions.

src/pytypes/onefuzztypes/models.py
```diff
@@ -476,15 +494,18 @@ class WorkerDoneEvent(BaseModel):
...
+    running: Optional[WorkerRunningEvent]
 
 
 class NodeStateUpdate(BaseModel):
     state: NodeState
 
 
-NodeEvent = Union[WorkerEvent, NodeStateUpdate]
+class NodeEvent(EnumModel):
+    worker_event: Optional[WorkerEvent]
```
ranweiler, line 507: Now, types like `NodeEvent` have a field for each variant of the (logical) discriminated union, instead of being represented by a `Union` (which is not a pydantic model, or even an instantiable class). A root validator in the superclass ensures that exactly one of the optional fields is non-`None`.

The motivation for moving away from this we can't do checks like `isinstance(foo, Union[A, B])`. We would need do this when we are scrutinizing a type that we would know to have type `Union[Union[A, B], C]`. So, any type that is logically just a `Union` previously needed to be wrapped in a model with a single artificial field, whose value is a union. This was very strange on the Rust side.

ranweiler: Eventually we expect pydantic to grow internally-tagged discriminated unions, and there will be much less mismatch between the types.

src/pytypes/onefuzztypes/models.py
```diff
@@ -32,6 +32,24 @@
...
 
 
+class EnumModel(BaseModel):
+    @root_validator(pre=True)
+    def exactly_one(cls: Any, values: Any) -> Any:
+        some = []
+
+        for field, val in values.items():
+            if val is not None:
+                some.append(field)
```
chkeita, line 42: no need to create and intermediate array you can just count the number of some(field) and validate the based on that number.

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -54,21 +54,49 @@ pub struct NodeEventEnvelope {
 }
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-#[serde(rename_all = "snake_case", untagged)]
+#[serde(rename_all = "snake_case")]
 pub enum NodeEvent {
-    StateUpdate { state: NodeState },
-    WorkerEvent { event: WorkerEvent },
+    StateUpdate(StateUpdateEvent),
```
chkeita, line 59: 'StateUpdateEvent' to be consistent with the other case WorkerEvent

chkeita (approved):

src/pytypes/onefuzztypes/models.py
ranweiler: True, but in this case I'm collecting the values on purpose so I can include them in the error message when validation fails. It's already helped me once!

src/agent/onefuzz-supervisor/src/coordinator.rs
ranweiler: I'm explicitly leaving the variant name as `StateUpdate` so the external tag when serialized is `state_update`. I only include `Event` in `WorkerEvent` because it truly contains a nested, entirely different event type as its event data.

At one point I had `NodeEvent::WorkerEvent` as just `NodeEvent::Worker`, but it sort of had confusingly less context at the use site (because you'd get at it as something like `node_event.worker`).

bmc-msft (approved):

Issue #28, opened 2020-09-23
Add a heartbeat to the supervisor agent


chkeita:
The supervisor should periodically send a heartbeat to the service 
The service should periodically check that the agent is responsive based on the heartbeat data.
The service should automatically restart the agent when it becomes unresponsive

PR #30, opened 2020-09-24
implementing heartbeat for the supervisor

chkeita:

_What is this about?_

## PR Checklist
* [x] Applies to work item: #28 
* [x] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-agent/src/tasks/heartbeat.rs
```diff
@@ -27,114 +24,52 @@ pub enum HeartbeatData {
...
-        HeartbeatClient::init_with_period(queue_url, task_id, DEFAULT_HEARTBEAT_PERIOD)
-    }
+pub type TaskHeartbeatClient = HeartbeatClient<HeartbeatData>;
 
-    pub fn init_with_period(queue_url: Url, task_id: Uuid, heartbeat_period: Duration) -> Self {
-        let messages = Arc::new(Mutex::new(HashSet::new()));
-        let cancelled = Arc::new(Notify::new());
-        let _heartbeat_process = HeartbeatClient::start_background_process(
-            task_id,
+pub fn init_task_heartbeat(
```
ranweiler, line 36: I think it make the most sense to have this be an additional constructor for `HeartbeatClient`, and make it shareable between the agent and supervisor.

src/agent/onefuzz-agent/src/tasks/heartbeat.rs
```diff
@@ -27,114 +24,52 @@ pub enum HeartbeatData {
...
+    queue_url: Url,
+    task_id: Option<Uuid>,
+) -> HeartbeatClient<HeartbeatData> {
+    let heartbeat_period = DEFAULT_HEARTBEAT_PERIOD;
+    let messages = Arc::new(Mutex::new(HashSet::<HeartbeatData>::new()));
+    let process_messages = Arc::clone(&messages);
+    let cancelled = Arc::new(Notify::new());
+    let process_cancelled = Arc::clone(&cancelled);
+    let heartbeat_process = task::spawn(async move {
+        let machine_id = get_machine_id().await.unwrap();
```
ranweiler, line 46: We should remove all the `unwrap` calls, so we can make sure to have a chance to log the error and flush telemetry.

src/agent/onefuzz-agent/src/tasks/heartbeat.rs
```diff
@@ -146,15 +81,15 @@ pub trait HeartbeatSender {
     }
 }
 
-impl HeartbeatSender for HeartbeatClient {
+impl HeartbeatSender for HeartbeatClient<HeartbeatData> {
     fn send(&self, data: HeartbeatData) -> Result<()> {
         let mut messages_lock = self.messages.lock().unwrap();
```
ranweiler, line 86: This isn't new, but is there a reason these are calling `unwrap()`, when they return a `Result`?

src/agent/onefuzz-agent/src/tasks/heartbeat.rs
```diff
@@ -27,114 +24,52 @@ pub enum HeartbeatData {
...
-        queue_url: Url,
-        messages: Arc<Mutex<HashSet<HeartbeatData>>>,
-        cancelled: Arc<Notify>,
-        heartbeat_period: Duration,
-    ) -> JoinHandle<Result<()>> {
-        let queue_client = QueueClient::new(queue_url);
-        task::spawn(async move {
-            let machine_id = get_machine_id().await?;
-            let machine_name = get_machine_name().await?;
+            |queue_client: Arc<QueueClient>, m: Arc<Mutex<HashSet<HeartbeatData>>>| async move {
```
ranweiler, line 52: Double-checking: was it actually necessary to specify the exact types here?

src/agent/onefuzz/src/heartbeat.rs
```diff
@@ -0,0 +1,86 @@
...
+    TaskAlive,
+    MachineAlive,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+struct Heartbeat {
+    task_id: Option<Uuid>,
+    machine_id: Uuid,
+    machine_name: String,
+    data: Vec<HeartbeatData>,
```
ranweiler, line 32: Is it correct that this is a queue of pending/not-yet-flushed heartbeat messages?

Maybe we should name it `pending` instead of the more generic `data`.

src/agent/onefuzz/src/heartbeat.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl<T> HeartbeatClient<T>
+where
+    T: Clone,
+    T: Sync,
+{
+    pub fn drain_current_messages(messages: Arc<Mutex<HashSet<T>>>) -> Vec<T> {
+        let lock = messages.lock();
+        let mut messages = lock.unwrap();
```
ranweiler, line 62: Again, I think we should replace all `unwrap` occurrences with `Result`.

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -84,7 +84,7 @@ impl<'a> Tester<'a> {
                 .map(|f| f.to_string())
                 .collect();
 
-            let crash_site = if let Some(frame) = call_stack.iter().next() {
+            let crash_site = if let Some(frame) = call_stack.get(0) {
```
ranweiler, line 87: :+1:

src/api-service/__app__/heartbeat/__init__.py
```diff
@@ -5,20 +5,44 @@
...
-from onefuzztypes.models import HeartbeatEntry
+from onefuzztypes.models import NodeHeartbeatEntry, TaskHeartbeatEntry
+from pydantic import ValidationError
 
 from ..onefuzzlib.dashboard import get_event
-from ..onefuzzlib.heartbeat import Heartbeat
+from ..onefuzzlib.heartbeat import NodeHeartbeat, TaskHeartbeat
+
+
+def try_add_task_heartbeat(raw: Dict) -> bool:
```
ranweiler, line 18: How about making these methods named `try_add()` on `TaskHeartbeat` and `NodeHeartbeat`, respectively?

src/agent/onefuzz-supervisor/src/heartbeat.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(Debug, Deserialize, Serialize, Clone)]
+struct Heartbeat {
+    node_id: Uuid,
+    machine_name: String,
+    data: Vec<HeartbeatData>,
+}
+
+pub type AgentHeartbeatClient = HeartbeatClient<HeartbeatData>;
+
+pub fn init_agent_heartbeat(queue_url: Url) -> HeartbeatClient<HeartbeatData> {
```
ranweiler, line 33: Apply earlier comments re unwraps and code sharing.

src/agent/onefuzz-agent/src/tasks/heartbeat.rs
ranweiler, line 52: Also: this callback body, and the specific type of `HeartbeatData`, are the only things specific to this particular heartbeat client, right? You also have `task_id` in one, but not the other. But, the task ID only gets used in the callback, which closes over it by move.

If it is doable, I'd love to see if if we could make it so users / call sites only have to specify the callback, period.

chkeita:
image

bmc-msft (approved):

PR #36, opened 2020-09-28
fix deleting repro VMs

bmc-msft:

As is, repro VMs are not properly deleted due to a logic bug.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

Fixes the logic when deleting repro VMs.

## Validation Steps Performed

Create a repro, note the VM ID.  Then delete it.  Wait ~5 minutes.  Check the Azure portal.  Resources related to $VM_ID


chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/vm.py
```diff
@@ -211,7 +211,7 @@ class VM(BaseModel):
     auth: Authentication
 
     def is_deleted(self) -> bool:
-        return has_components(str(self.name))
+        return not has_components(str(self.name))
```
ranweiler, line 214: While we're in here, maybe it'd be good to add a comment explaining why this check tells us the information we think it does.

ranweiler (approved):

PR #40, opened 2020-09-28
adding dev-deploy.ps1 for dev deployment on windows

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/api-service/dev-deploy.ps1
```diff
@@ -0,0 +1,19 @@
+param (
+    [Parameter()]
+    [string] $app_dir="$PSScriptRoot",
+    [Parameter(Mandatory=$true)]
+    [string] $target
+)
+
+Set-Location "$app_dir\..\pytypes"
```
lzybkr, line 8: If you use Linux paths, it will still work on Windows and the script might also work on Linux.

lzybkr, line 8: Scripts generally shouldn't change the current directory without restoring it, e.g. use Push-Location and in a finally block use Pop-Location.

src/api-service/dev-deploy.ps1
```diff
@@ -0,0 +1,19 @@
...
+python setup.py sdist bdist_wheel 
+Copy-Item dist/*.whl "$app_dir\__app__"
+Set-Location $app_dir
+Set-Location __app__
+New-Guid | ForEach-Object{$_.Guid} > onefuzzlib/build.id
```
lzybkr, line 13: This could be:

```powershell
(New-Guild).Guid > onefuzzlib/build.id
```

src/api-service/dev-deploy.ps1
```diff
@@ -0,0 +1,19 @@
...
+(Get-Content -path "requirements.txt") -replace 'onefuzztypes==0.0.0', './onefuzztypes-0.0.0-py3-none-any.whl' | Out-File "requirements.txt"
+func azure functionapp publish $target --python
+(Get-Content -path "requirements.txt") -replace './onefuzztypes-0.0.0-py3-none-any.whl', 'onefuzztypes==0.0.0' | Out-File "requirements.txt"
```
lzybkr, line 16: It seems safer and more reliable to backup the file and restore.

chkeita: Keeping the same logic as the dev-deploy.sh to make the maintenance easier. we can address the comment in both script in a sperate PR

src/api-service/dev-deploy.ps1
```diff
@@ -0,0 +1,24 @@
...
+)
+
+try { 
+    Push-Location
+    Set-Location "$app_dir/../pytypes"
+    python setup.py sdist bdist_wheel 
+    Copy-Item dist/*.whl "$app_dir/__app__"
+    Set-Location $app_dir
+    Set-Location __app__
+    (New-Guild).Guid > onefuzzlib/build.id
```
lzybkr, line 15: New-Guild sounds like a fun command but it should be New-Guid.

bmc-msft (approved):

PR #41, opened 2020-09-28
can schedule endpoint

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/api-service/__app__/agent_can_schedule/__init__.py
+++ b/src/api-service/__app__/agent_can_schedule/__init__.py
@@ -44,7 +44,7 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
 
     task = Task.get_by_task_id(request.task_id)
 
-    work_stopped = isinstance(task, Error) or (task.state == TaskState.scheduled)
+    work_stopped = isinstance(task, Error) or (task.state != TaskState.scheduled)
     return ok(CanSchedule(allowed=allowed, work_stopped=work_stopped))
 
 
--- a/src/api-service/__app__/agent_can_schedule/__init__.py
+++ b/src/api-service/__app__/agent_can_schedule/__init__.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import logging
+import azure.functions as func
+from onefuzztypes.enums import ErrorCode, TaskState
+from onefuzztypes.models import Error, NodeCommand, StopNodeCommand
+from onefuzztypes.requests import CanScheduleRequest
+from onefuzztypes.responses import CanSchedule
+
+from ..onefuzzlib.agent_authorization import verify_token
+from ..onefuzzlib.pools import Node, NodeMessage
+from ..onefuzzlib.request import not_ok, ok, parse_uri
+from ..onefuzzlib.tasks.main import Task
+
+
+def post(req: func.HttpRequest) -> func.HttpResponse:
+    request = parse_uri(CanScheduleRequest, req)
+    if isinstance(request, Error):
+        return not_ok(request, context="CanScheduleRequest")
+
+    node = Node.get_by_machine_id(request.machine_id)
+    if not node:
+        return not_ok(
+            Error(code=ErrorCode.UNABLE_TO_FIND, errors=["unable to find node"]),
+            context=request.machine_id,
+        )
+
+    allowed = True
+    work_stopped = False
+    if node.is_outdated:
+        logging.info(
+            "received can_schedule request from outdated node '%s' version '%s'",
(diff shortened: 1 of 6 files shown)


src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    node = Node.get_by_machine_id(request.machine_id)
+    if not node:
+        return not_ok(
+            Error(code=ErrorCode.UNABLE_TO_FIND, errors=["unable to find node"]),
+            context=request.machine_id,
+        )
+
+    allowed = True
+    work_stopped = False
+    if node.version != current_version:
```
bmc-msft, line 34: Can this be a function of Node?  I imagine we'll likely want to do this in multiple places.

src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+from ..onefuzzlib.request import not_ok, ok, parse_uri
+from ..onefuzzlib.tasks.main import Task
+
+
+def get(req: func.HttpRequest) -> func.HttpResponse:
+    request = parse_uri(CanScheduleRequest, req)
+    if isinstance(request, Error):
+        return not_ok(request, context="CanScheduleRequest")
+
+    current_version = ""
```
bmc-msft, line 23: onefuzzlib.__version__.__version__

src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    if not node:
+        return not_ok(
+            Error(code=ErrorCode.UNABLE_TO_FIND, errors=["unable to find node"]),
+            context=request.machine_id,
+        )
+
+    allowed = True
+    work_stopped = False
+    if node.version != current_version:
+        allowed = False
```
ranweiler, line 35: It'd probably be nice to have some logging here

src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        stop_message.save()
+
+    task = Task.get_by_task_id(request.task_id)
+
+    work_stopped = isinstance(task, Error) or (task.state == TaskState.scheduled)
+    return ok(CanSchedule(allowed=allowed, work_stopped=work_stopped))
+
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
+    if req.method == "GET":
```
ranweiler, line 49: We should probably make this a `POST`. `GET` is really meant for requests with no body. I'll update my PR too.

src/api-service/__app__/agent_can_schedule/function.json
```diff
@@ -0,0 +1,22 @@
+{
+	"scriptFile": "__init__.py",
+	"bindings": [
+		{
+			"authLevel": "anonymous",
+			"type": "httpTrigger",
+			"direction": "in",
+			"name": "req",
+			"methods": [
```
ranweiler, line 9: For now, I suggest just not routing the unused methods (I assume these are just here from copy/paste).

src/api-service/__app__/agent_can_schedule/__init__.py
ranweiler: (done)

src/api-service/__app__/agent_can_schedule/__init__.py
chkeita: moves logic to Node class

src/api-service/__app__/agent_can_schedule/__init__.py
chkeita: moves logic to Node class

src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+        allowed = False
+        stop_message = NodeMessage(
+            agent_id=node.machine_id, message=NodeCommand(stop=StopNodeCommand()),
+        )
+        stop_message.save()
+
+    task = Task.get_by_task_id(request.task_id)
+
+    work_stopped = isinstance(task, Error) or (task.state == TaskState.scheduled)
```
ranweiler, line 47: Just double-checking: if the task isn't found, that is represented as an `Error`, not a `None`?

Also, I think this should be `task.state != TaskState.scheduled`.

chkeita: yup here is the def get_by_task_id currently returns an error when the task is not found

ranweiler (approved):

PR #43, opened 2020-09-29
Link VMSS nodes and tasks when setting up

ranweiler:

Add a (backwards-compatible) data field to node state update events. Use this to link nodes and tasks as soon as the tasks have been claimed. Add a new task-level `setting_up` state to encode this.

### DIFF
--- a/src/api-service/__app__/agent_events/__init__.py
+++ b/src/api-service/__app__/agent_events/__init__.py
@@ -42,8 +42,8 @@ def get_node_checked(machine_id: UUID) -> Node:
 
 
 def on_state_update(
-        machine_id: UUID,
-        state_update: NodeStateUpdate,
+    machine_id: UUID,
+    state_update: NodeStateUpdate,
 ) -> func.HttpResponse:
     state = state_update.state
     node = get_node_checked(machine_id)
@@ -130,9 +130,7 @@ def on_worker_event(machine_id: UUID, event: WorkerEvent) -> func.HttpResponse:
 
     task.save()
     node.save()
-    task_event = TaskEvent(
-        task_id=task_id, machine_id=machine_id, event_data=event
-    )
+    task_event = TaskEvent(task_id=task_id, machine_id=machine_id, event_data=event)
     task_event.save()
     return ok(BoolResult(result=True))
 
--- a/src/api-service/__app__/agent_events/__init__.py
+++ b/src/api-service/__app__/agent_events/__init__.py
@@ -60,6 +60,10 @@ def on_state_update(
                     for task_id in state_update.data.tasks:
                         task = get_task_checked(task_id)
                         task.state = TaskState.setting_up
+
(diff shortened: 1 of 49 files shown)


src/pytypes/onefuzztypes/models.py
```diff
@@ -494,20 +494,43 @@ class WorkerDoneEvent(BaseModel):
...
+    data: Optional[SettingUpEventData]
+
+    @validator("data")
+    def check_data(cls, data: Optional[SettingUpEventData], values: Any) -> Optional[SettingUpEventData]:
+        if data:
+            state = values.get("state")
+            if state and state != NodeState.setting_up:
+                raise ValueError("data for node state update event does not match state = %s" % state)
+
+        data
```
ranweiler, line 517: Need a `return`.

earthmehuichapan: Fix all the state.(Erendira-_$ all needs\%._ofcrditnow.

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -41,13 +41,37 @@ def get_node_checked(machine_id: UUID) -> Node:
...
+
+            if state == NodeState.setting_up:
+                # This field will be required in the future.
+                # For now, it is optional for back compat.
+                if state_update.data:
+                    for task_id in state_update.data.tasks:
+                        task = get_task_checked(task_id)
+                        task.state = TaskState.setting_up
+
+                        # We don't yet call `on_start()` for the task.
```
bmc-msft, line 64: 👍

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -41,13 +41,37 @@ def get_node_checked(machine_id: UUID) -> Node:
...
             node.state = state
             node.save()
+
+            if state == NodeState.setting_up:
+                # This field will be required in the future.
+                # For now, it is optional for back compat.
+                if state_update.data:
+                    for task_id in state_update.data.tasks:
+                        task = get_task_checked(task_id)
+                        task.state = TaskState.setting_up
```
chkeita, line 62: won't this conflict with line 95 when we set the state to running in the case where we have multiple nodes running the same task

ranweiler: Yes. I guess the worst problem is that we could go from `setting_up`, to `running`, and back to `setting_up`, if we have e.g. a task with `vm_count == 2`. Also, I think that problem is not unique to this new state.

Regardless, maybe we should just remove the `Task` state, but leave `setting_up` as a `NodeTaskState`.

chkeita: my thoughts exactly

ranweiler: Handled for now by only updating if we aren't `running`.

chkeita (approved):

PR #46, opened 2020-09-29
add end-to-end integration testing of fuzzing pipelines

bmc-msft:

Migrates the majority of internal CICD tooling for integration testing into something consumable outside of Azure Devops.

## PR Checklist
* [X] Tests added/passed

## Info on Pull Request

* Each of the job templates now returns the Job (in the case of --dryrun, returns None)
* Uses the onefuzz-samples released artifacts of basic fuzzing binaries for testing.

The process followed:
1. Creates pools & scalesets needed (optionally uses existing pools if the user specifies their pools on the cmdlien)
2. Launches the templates specifying the above pools
3. Monitors each of the launched jobs until either the tasks fail or a report is created in the job specific container
4. Launches repro VMs for all of the successful tasks using one of the reports report
5. Monitors each of the repro VMs until either it fails or verifies the `rip` register works as expected.
6. Cleans up all created resources

This assumes you have the platform specific debugger (cdb for windows, gdb for linux) in your path.   If you're launching this script within WSL, this is possible.  Otherwise, use `--skip_repro` or `--os_list` to limit to the OS.

## Validation Steps Performed

`cd src/cli; examples/integration-test.py test ~/onefuzz-samples/`

### DIFF
--- a/src/cli/examples/integration-test.py
+++ b/src/cli/examples/integration-test.py
@@ -499,6 +499,7 @@ class Run(Command):
         region: Optional[str] = None,
         os_list: List[OS] = [OS.linux, OS.windows],
         targets: List[str] = list(TARGETS.keys()),
+        skip_repro: bool = False,
     ) -> None:
         tester = TestOnefuzz(
             self.onefuzz,
@@ -514,8 +515,11 @@ class Run(Command):
             tester.setup(region=region, user_pools=user_pools)
             tester.launch(samples)
             tester.check_jobs()
-            tester.launch_repro()
-            tester.check_repro()
+            if skip_repro:
+                self.logger.warn("not testing crash repro")
+            else:
+                tester.launch_repro()
+                tester.check_repro()
         except Exception as e:
             self.logger.error("testing failed: %s", repr(e))
             error = e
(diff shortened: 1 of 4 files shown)


src/cli/examples/integration-test.py
```diff
@@ -0,0 +1,555 @@
...
+        logger: logging.Logger,
+        *,
+        pool_size: int,
+        os_list: List[OS],
+        targets: List[str],
+    ) -> None:
+        self.of = onefuzz
+        self.logger = logger
+        self.pools: Dict[OS, Pool] = {}
+        self.project = str(uuid4()).split("-")[0]
```
chkeita, line 119: I suggest adding a suffix like "test" to make the data that depends on this easily identifiable when debugging

src/cli/examples/integration-test.py
```diff
@@ -0,0 +1,555 @@
...
+                tester.check_repro()
+        except Exception as e:
+            self.logger.error("testing failed: %s", repr(e))
+            error = e
+            success = False
+        except KeyboardInterrupt:
+            self.logger.error("interrupted testing")
+            success = False
+
+        if not tester.cleanup(user_pools=user_pools):
```
chkeita, line 531: Can we add an option for keeping things around for investigation

chkeita (approved):

PR #58, opened 2020-09-30
set more detailed version information during builds

bmc-msft:

Unless Onefuzz is built in Github from a 'refs/tags', append the SHA and potentially .localchanges to the version string.

This provides an easy marker for identifying if the system has local changes included or not.

## Info on Pull Request

* Updates the each of the `build.rs` to to check GITHUB_REF and `git diff --quiet` to identify local changes
* Adds a util to reset setting versions (reset-versions.sh)
* Adds a util to get the version string without setting it (get-version.sh)
* Updates the service API deploy-dev.sh to package onefuzztypes & the api with the stamped in build strings

## Validation Steps Performed

Manual inspection of built artifacts.  Locally built should be:
* 1.0.0-HASH or 1.0.0-HASH.localchanges in the case of modifying code locally without checking it in.
* CI builds should be 1.0.0-HASH
* Tag builds should be 1.0.0

### DIFF
--- a/src/proxy-manager/build.rs
+++ b/src/proxy-manager/build.rs
@@ -1,8 +1,8 @@
+use std::env;
 use std::error::Error;
 use std::fs::File;
 use std::io::prelude::*;
 use std::process::Command;
-use std::env;
 
 fn run_cmd(args: &[&str]) -> Result<String, Box<dyn Error>> {
     let cmd = Command::new(args[0]).args(&args[1..]).output()?;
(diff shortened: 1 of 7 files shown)


src/proxy-manager/build.rs
```diff
@@ -16,21 +17,43 @@ fn read_file(filename: &str) -> Result<String, Box<dyn Error>> {
...
-        ""
-    };
-    println!("cargo:rustc-env=GIT_VERSION={}{}", sha, with_changes);
 
-    let version = read_file("../../CURRENT_VERSION")?;
+    if include_sha {
+        version.push('-');
+        version.push_str(&sha);
+
+        // if we're a non-release buil, check to see if git has
```
chkeita, line 33: build

chkeita (approved):

PR #63, opened 2020-10-01
Set log levels in Azure Functions by hand for 3rd party libraries

bmc-msft:

Reduces the log levels to `WARN` for multiple third-party libraries early in processing.

## Info on Pull Request

There doesn't appear to be a supported mechanism to reset logging levels on a per logger method for Azure Functions.  As such, this makes sets the log level for the most verbose of the 3rd party libraries we use.

Hopefully, the python Azure Functions runtime will provide a better mechanism to reduce these log levels in the future.

https://github.com/Azure/azure-functions-python-worker/issues/743

## Validation Steps Performed

Install, note we don't see all of the INFO logging from APIs like msrest in app insights.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/monkeypatch.py
+++ b/src/api-service/__app__/onefuzzlib/azure/monkeypatch.py
@@ -7,6 +7,7 @@ import inspect
 import logging
 
 WORKERS_DONE = False
+REDUCE_LOGGING = False
 
 
 def allow_more_workers() -> None:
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/azure/monkeypatch.py
```diff
@@ -22,3 +23,32 @@ def allow_more_workers() -> None:
...
+        "aiohttp",
+        "asyncio",
+        "adal-python",
+    ]
+
+    for name in logging.Logger.manager.loggerDict:  # type: ignore
+        logger = logging.getLogger(name)
+        for prefix in to_quiet:
+            if logger.name.startswith(prefix):
+                logger.level = logging.WARN
```
chkeita, line 54: REDUCE_LOGGING should be set to true at the end of the function

chkeita (approved):

PR #64, opened 2020-10-01
only set stating to stopping

bmc-msft:


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -225,9 +225,10 @@ def get_or_create(cls, region: Region) -> Optional["Proxy"]:
         proxy = Proxy.get(region)
         if proxy is not None:
             if proxy.version != __version__:
-                # If the proxy is out-of-date, delete and re-create it
-                proxy.state = VmState.stopping
-                proxy.save()
+                if proxy.state != VmState.stopping:
```
chkeita, line 228: we should check if it is not stopped too

bmc-msft: Proxy doesn't go to 'stopped', it gets deleted rather than queuing the stopped callback.

PR #69, opened 2020-10-01
log the TokenData and URL on rejecting tokens

bmc-msft:
This logs a TokenData, which includes the application_id and object_id, as well as the URL with URL params.


src/api-service/__app__/onefuzzlib/agent_authorization.py
```diff
@@ -66,6 +67,7 @@ def verify_token(
         return not_ok(token, status_code=401, context="token verification")
 
     if not is_authorized(token):
+        logging.error("rejecting token - %s - %s", req.url, token)
```
chkeita, line 70: "Rejecting request". 
The token is valid (azure is doing the verification) but the request is not.
We should also log the body of the request too.

chkeita (approved):

PR #81, opened 2020-10-02
pre-create instance-specific-setup directory

bmc-msft:


src/runtime-tools/linux/setup.sh
```diff
@@ -6,7 +6,7 @@
 find .
 set -x
 
-INSTANCE_SETUP="/onefuzz/instance-specific-setup/setup.sh"
+INSTANCE_SETUP="/onefuzz/instance-specific-setup/linux/setup.sh"
```
chkeita, line 9: do we need this for the windows script too ?

chkeita (approved):

src/runtime-tools/linux/setup.sh
bmc-msft: Not sure.  I'll check that tomorrow and open another PR if need be

PR #86, opened 2020-10-02
use a semaphore to prevent agent restart during 'done' to reset early

bmc-msft:


src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -76,6 +77,11 @@ fn run(opt: RunOpt) -> Result<()> {
         env!("GIT_VERSION")
     );
 
+    if done::is_agent_done()? {
+        verbose!("agent is done, remove lock to continue");
```
chkeita, line 81: printing the path to the lock file could be helpful when debugging

chkeita (approved):

PR #87, opened 2020-10-02
fixing input marker expansion in the supervisor task

chkeita:
input marker was set as a path instead of a string value

_What is this about?_

## PR Checklist
* [x] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


bmc-msft (approved):
LGTM.

👍 to renaming `input` to `input_path`, which clearly separates it from `input_marker`.

PR #90, opened 2020-10-02
refactor node state to fully put the agent in charge

bmc-msft:
This moves the node state management in the hands of the agent, and we tell the agent when to shut down, etc.

### DIFF
--- a/src/api-service/__app__/agent_events/__init__.py
+++ b/src/api-service/__app__/agent_events/__init__.py
@@ -56,7 +56,13 @@ def on_state_update(
         node.stop()
         return
 
-    if state == NodeState.init or node.state not in NodeState.ready_for_reset():
+    if state == NodeState.init:
+        if node.delete_requested:
+            node.stop()
+            return
+        node.reimage_requested = False
+        node.save()
+    elif node.state not in NodeState.ready_for_reset():
         if node.state != state:
             node.state = state
             node.save()
--- a/src/api-service/__app__/agent_events/__init__.py
+++ b/src/api-service/__app__/agent_events/__init__.py
@@ -51,10 +51,16 @@ def on_state_update(
     state = state_update.state
     node = get_node_checked(machine_id)
 
-    if state == NodeState.free and (node.reimage_requested or node.delete_requested):
-        logging.info("stopping free node with reset flags: %s", node.machine_id)
-        node.stop()
-        return
+    if state == NodeState.free:
+        if node.reimage_requested or node.delete_requested:
+            logging.info("stopping free node with reset flags: %s", node.machine_id)
+            node.stop()
+            return
+
+        if node.could_shrink_scaleset():
+            logging.info("stopping free node to resize scaleset: %s", node.machine_id)
+            node.set_halt()
+            return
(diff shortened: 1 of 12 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -154,16 +160,87 @@ def stop_task(cls, task_id: UUID) -> None:
...
+        )
+        stop_message.save()
+
+    def to_reimage(self, done: bool = False) -> None:
+        if done:
+            if self.state not in NodeState.ready_for_reset():
+                self.state = NodeState.done
+
+        if not self.reimage_requested and not self.delete_requested:
+            self.reimage_requested = True
```
chkeita, line 227: Is this reset after the reimage ?

bmc-msft: It is now that #86 has been merged, I was able to enable it.

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -47,11 +47,22 @@ def get_node_checked(machine_id: UUID) -> Node:
...
+        logging.info("stopping free node with reset flags: %s", node.machine_id)
+        node.stop()
+        return
+
+    if state == NodeState.init:
+        if node.delete_requested:
+            node.stop()
+            return
+        node.reimage_requested = False
+        node.save()
```
chkeita, line 64: I think this logic should be in the node class.

src/pytypes/onefuzztypes/models.py
```diff
@@ -399,6 +399,8 @@ class Node(BaseModel):
     scaleset_id: Optional[UUID] = None
     tasks: Optional[List[Tuple[UUID, NodeTaskState]]] = None
     version: str = Field(default="1.0.0")
+    reimage_requested: bool = Field(default=False)
+    delete_requested: bool = Field(default=False)
```
chkeita, line 403: could both of these have been part of the node states

chkeita (approved):

src/pytypes/onefuzztypes/models.py
bmc-msft: We did that previously, but then got into issues where we were out of sync with the supervisor.  This treats the supervisor as having ground truth knowledge of node.state, but still lets us remember when a node should be reimaged/deleted.

src/api-service/__app__/agent_events/__init__.py
bmc-msft: probably, but I will do that in another PR

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -99,7 +105,7 @@ def search_outdated(
         # azure table query always return false when the column does not exist
         # We write the query this way to allow us to get the nodes where the
         # version is not defined as well as the nodes with a mismatched version
-        version_query = "not (version ne '%s')" % __version__
+        version_query = "not (version eq '%s')" % __version__
```
chkeita, line 108: 👍

PR #92, opened 2020-10-02
Scaleset size update

anshuman-goel:
Size is not increased for the scaleset when it is resized, thus leaving it in a hanging state of resize when scaling up.

_What is this about?_
The condition mentioned at https://github.com/microsoft/onefuzz/blob/3b3572d0b364df0bc9e354804b92f45672259e3e/src/api-service/__app__/onefuzzlib/pools.py#L684:L690

```
logging.info(
                    "resize is finished, waiting for nodes to check in: "
                    "%s (%d of %d nodes checked in)",
                    self.scaleset_id,
                    node_count,
                    self.size,
                )
```

is never fulfilled when I am trying out server side self-scaling especially when I have to scale up. I cannot see the scaleset size increasing when a new node is registered. This condition will add only if number of nodes get greater than scaleset size specified.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed
Also added logging steps, which were helpful in my debugging. let me know if it is not needed.

Tested manually with pool scaling work.

### DIFF
--- a/src/api-service/__app__/agent_registration/__init__.py
+++ b/src/api-service/__app__/agent_registration/__init__.py
@@ -3,6 +3,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+import logging
 from uuid import UUID
 
 import azure.functions as func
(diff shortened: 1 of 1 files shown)


bmc-msft:
Scaleset.size is intended to be the "how big does the user want this scaleset".   This PR converts this to "how many nodes have checked in".

In #90, I refactored how resizing works such that we'll maintain Scaleset.size during grow and shrink events in a simpler fashion.

bmc-msft (changes requested):
This is OBE by #90

anshuman-goel:
I will wait to check my changes after #90 lands. Thanks for looking into it.

PR #101, opened 2020-10-05
canonicalize the paths being monitored

bmc-msft:


src/agent/onefuzz-agent/src/tasks/utils.rs
```diff
@@ -156,7 +157,9 @@ pub fn file_uploader_monitor(synced_dir: SyncedDir) -> Result<impl Future> {
 /// The intent of this is to support use cases where we usually want a directory
 /// to be initialized, but a user-supplied binary, (such as AFL) logically owns
 /// a directory, and may reset it.
-pub async fn monitor_result_dir(synced_dir: SyncedDir) -> Result<()> {
+pub async fn monitor_result_dir(mut synced_dir: SyncedDir) -> Result<()> {
+    synced_dir.path = dunce::canonicalize(&synced_dir.path)?;
```
ranweiler, line 161: This requires the path to exist. But this function must not require that, since it is used dirs that another process may create. This is why we have the inner loop we we check `metadata().is_err()`.

One solution would be to use a successful canonicalization as proof of existence instead of `metadata()`.

bmc-msft:
The underlying issue I was trying to enable debugging is more complicated than this will address.

Issue #44, opened 2020-09-29
Enable auto-export from appinsights


bmc-msft:
By enabling continuous export from app insights, they can use existing tools to parse logs including better import into Power BI.

Unfortunately, this [isn't possible via ARM templates](https://feedback.azure.com/forums/357324-application-insights/suggestions/35020774-add-arm-template-support-for-continuous-export).

This will likely need to be added to `deploy.py` using [azure.mgmt.applicationinsights](https://docs.microsoft.com/en-us/python/api/azure-mgmt-applicationinsights/azure.mgmt.applicationinsights.applicationinsightsmanagementclient?view=azure-python#export-configurations)

PR #102, opened 2020-10-05
add the ability to configure app-insights export

bmc-msft:

Adds the ability to enable automatic log export from the instance specific appinsights

## PR Checklist
* [x] Applies to work item: #44 
* [ ] Tests added/passed

## Info on Pull Request

Adds the `--export_appinsights` to the deployment script to setup app insight exports to the `app-insights` container on the `func` storage account.  Note, it will automatically replace any existing app-insights exports to the same container prior to creating the export config.

## Validation Steps Performed

Deploy with `--export_appinsights`, run for a while.  Note new files are created in `app-insights` container on `func` storage account.


src/deployment/deploy.py
```diff
@@ -383,6 +389,68 @@ def create_eventgrid(self):
...
+
+        logger.info("adding appinsight log export")
+        account_name = self.results["deploy"]["func-name"]["value"]
+        key = self.results["deploy"]["func-key"]["value"]
+        account_url = "https://%s.blob.core.windows.net" % account_name
+        client = BlobServiceClient(account_url, credential=key)
+        if container_name not in [x["name"] for x in client.list_containers()]:
+            client.create_container(container_name)
+
+        expiry = datetime.utcnow() + timedelta(days=2 * 365)
```
ranweiler, line 407: Can we accomplish this with something other than an ad hoc (thus irrevocable) SAS URL?

ranweiler: Alternatively: use a much shorter-lived URL, and have a mechanism for refreshing within an instance (though that feels a bit roundabout).

bmc-msft: The API unfortunately requires SAS urls.

Doing it through the instance requires granting more privileges to the service.

ranweiler: Okay. Given the restrictions on perms, and the fact that none of our code touches it (it is only given directly to an Azure API, here), this seems okay to me. Can we just document our usage assumptions? In particular, note that this SAS URL must not "leave this script", except to the extent that it implicitly does in its use in the export request.

ranweiler (approved):

PR #107, opened 2020-10-06
Managing Pool Resizing at service side

anshuman-goel:

Managing Pool Resizing at service side using Time triggered AF

_What is this about?_

Currently, the client has to manage the scaling for large scale deployment of continuous fuzzing. Bring the same to service is ideal and will help others in reusing it.
 
## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request
It transfers the parameters needed by scaleset creation to pool. The pool will now handle a group of scalesets with same configuration. The pool_resize af will manage creation and deletion of tasks as required. Currently it assumes that only one workset is deployment per VM.

It takes one more additional argument which is maximum number of instances in scaleset.

_What does this include?_
It creates an Time triggered Azure function, which will check number of tasks and allocate the number of instances across a group of scalesets. It also manages creation and deletion of scalesets as required.

## Validation Steps Performed

Deploy the package, create the pool from cli. create bunch of fuzzing jobs and monitor `Node` and `Scaleset` table. It will create scalesets with maximum allowed number of instances in scaleset. The Node will assign tasks to themselves. Delete some jobs, over the time, Nodes will get deleted as soon as they become free.

## Next:
- Removing scaleset creation and deletion from cli.
- Including workset knowledge as needed.


src/api-service/__app__/onefuzzlib/tasks/main.py
```diff
@@ -153,6 +153,27 @@ def get_by_task_id(cls, task_id: UUID) -> Union[Error, "Task"]:
         task = tasks[0]
         return task
 
+    @classmethod
+    def get_tasks_by_pool_name(
+        cls, pool_name: str
+    ) -> Optional[Union[Error, List["Task"]]]:
+        tasks = cls.search()
```
bmc-msft, line 160: This will return all tasks regardless of state, then post-filter the results.  `search_states(states=TaskState.available())` will likely provide what you want, which is "all tasks that are not stopping/stopped"

src/pytypes/onefuzztypes/requests.py
```diff
@@ -91,6 +91,11 @@ class PoolCreate(BaseRequest):
     arch: Architecture
     managed: bool
     client_id: Optional[UUID]
+    max_size: int
```
bmc-msft, line 94: These need to be in an optional sub-struct for a few reasons.

1. We want to use pools for compute resources that are not Azure VM Scalesets
2. This forces auto-resizing where there are multiple cases where manually managed scalesets are preferred
3. Backwards compatibility

src/pytypes/onefuzztypes/models.py
```diff
@@ -419,11 +419,16 @@ class Pool(BaseModel):
     pool_id: UUID = Field(default_factory=uuid4)
     os: OS
     managed: bool
+    max_size: int
```
bmc-msft, line 422: In addition to max size, can you add min_size?  While we don't use it yet, we want to make use of pools of VMs for repro  & regression testing, and having an option for those to always available would be beneficial for user interaction.

src/pytypes/onefuzztypes/enums.py
```diff
@@ -281,6 +281,15 @@ def available(cls) -> List["ScalesetState"]:
         unavailable = [cls.shutdown, cls.halt, cls.creation_failed]
         return [x for x in cls if x not in unavailable]
 
+    @classmethod
+    def is_resizing(cls) -> List["ScalesetState"]:
+        """ set of states that indicate scaleset is resizing """
```
bmc-msft, line 286: nit: neither halt or shutdown is 'resizing'.  Can a different verb be used here?

src/api-service/__app__/onefuzzlib/tasks/main.py
bmc-msft: Since the result of an empty list and Error is handled the same by the caller, this could be simplified to return List["Task"].

src/api-service/__app__/onefuzzlib/tasks/main.py
```diff
@@ -153,6 +153,27 @@ def get_by_task_id(cls, task_id: UUID) -> Union[Error, "Task"]:
...
+        if not tasks:
+            return Error(code=ErrorCode.INVALID_REQUEST, errors=["unable to find task"])
+
+        pool_tasks = []
+
+        for task in tasks:
+            if not task.config.pool:
```
bmc-msft, line 167: please use `task.get_pool()` to determine the pool used by the task.

Our pre-release API specified pools via task.config.vm.sku/task.config.vm.image.  We need to support tasks from the old configuration method until we can get all of our internal users fully upgraded.

bmc-msft: (the same change needs to occur in the pool resizing function)

src/api-service/__app__/pool_resize/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+            vm_sku=pool.vm_sku,
+            image=pool.image,
+            region=pool.region,
+            size=max_nodes_scaleset,
+            spot_instances=pool.spot_instances,
+            tags={"pool": pool.name},
+        )
+        scaleset.save()
+        # don't return auths during create, only 'get' with include_auth
+        scaleset.auth = None
```
bmc-msft, line 63: this isn't needed.  scale_up doesn't return scalesets to the user.

src/api-service/__app__/pool_resize/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+
+def scale_down(scalesets: List[Scaleset], nodes_to_remove: int) -> None:
+    for scaleset in scalesets:
+        nodes = Node.search_states(
+            scaleset_id=scaleset.scaleset_id, states=[NodeState.free]
+        )
+        if nodes and nodes_to_remove > 0:
+            max_nodes_remove = min(len(nodes), nodes_to_remove)
+            if max_nodes_remove >= scaleset.size and len(nodes) == scaleset.size:
+                scaleset.state = ScalesetState.halt
```
bmc-msft, line 75: Using shutdown here instead of halt will let nodes that pick up work during your resize process to finish their existing work before getting removed.

bmc-msft (changes requested):
In general, this looks to be a very positive contribution and I can see it becoming our recommended method for managing VMs.   👍

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -125,6 +126,8 @@ def on_state_update(
                             machine_id,
                             done_data,
                         )
+        else:
+            logging.info("No change in Node state")
```
bmc-msft, line 130: This needs to include information about the node in question.  Given this is about a NOOP, it can also be made a `debug`

src/api-service/__app__/agent_registration/__init__.py
```diff
@@ -76,6 +77,7 @@ def get(req: func.HttpRequest) -> func.HttpResponse:
 
 def post(req: func.HttpRequest) -> func.HttpResponse:
     registration_request = parse_uri(AgentRegistrationPost, req)
+    logging.info(f"request: {registration_request}")
```
bmc-msft, line 80: please use `%s`, following the style of the rest of the code.

src/api-service/__app__/onefuzzlib/tasks/main.py
```diff
@@ -153,6 +153,23 @@ def get_by_task_id(cls, task_id: UUID) -> Union[Error, "Task"]:
         task = tasks[0]
         return task
 
+    @classmethod
+    def get_tasks_by_pool_name(cls, pool_name: str) -> Union[Error, List["Task"]]:
+        tasks = cls.search_states(states=TaskState.available())
+        if not tasks:
+            return Error(code=ErrorCode.INVALID_REQUEST, errors=["unable to find task"])
```
bmc-msft, line 160: Since the caller of this function turns Error into an empty list, you don't need to return an error. 

Removing `if not tasks: return Error(..)` makes this effectively the same.

src/pytypes/onefuzztypes/models.py
```diff
@@ -456,11 +456,22 @@ class NodeTasks(BaseModel):
     state: NodeTaskState = Field(default=NodeTaskState.init)
 
 
+class AutoScaleConfig(BaseModel):
+    image: str
+    max_size: Optional[int]  # max size of pool
```
bmc-msft, line 461: The level of data validation of this struct done in `pools/__init__.py` are reasonable, but should be done via pydantic validation `@validator` and `@root_validator`.

See JobConfig and NodeStateUpdate in this file for examples.

While the SDK doesn't make use of the validation available in the pydantic models yet, it would be great to make the same validation used at the service level happen at the SDK with minimal changes.

bmc-msft: Note, you won't be able to validate the image or region here.  The SDK doesn't have the ability to query for the underlying data.  Those still need to be done at the functions level.

src/cli/onefuzz/api.py
```diff
@@ -811,9 +811,32 @@ def create(
         *,
         unmanaged: bool = False,
         arch: enums.Architecture = enums.Architecture.x86_64,
+        min_size: Optional[int] = 0,  # min size of pool
```
bmc-msft, line 814: A recent change to the CLI makes it such that we can  document arguments and have it show up in the CLI.

This is currently only used in the `template afl basic' component.

If you add the following as a function level pydoc:

```
        """
        Create a worker pool

        :param str name: Name of the worker-pool
        :param Optional[int] min_size: Specify the minimum size of the pool (when autoscaled)
        :param Optional[int] max_size: Specify the maximum size of the  pool (when autoscaled)
        """
```

Then arguments will show up like this:
```bash
$ onefuzz pools create --help |grep Name
  name                  Name of the worker-pool
$
```

bmc-msft: Note: this change isn't required to merge the pr, but for your FYI, given you are taking the time to explain the arguments.

src/api-service/__app__/pool_resize/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+        for scaleset in scalesets:
+            if scaleset.state in ScalesetState.modifying():
+                pool_resize = True
+                break
+            nodes_needed = nodes_needed - scaleset.size
+
+        if pool_resize:
+            continue
+
+        logging.info(f"#Nodes Needed: {nodes_needed}")
```
bmc-msft, line 140: This log should include the name of the pool

src/api-service/__app__/pool_resize/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+    logging.info("Scaling up")
+    autoscale_config = pool.autoscale
+    if not isinstance(autoscale_config, AutoScaleConfig):
+        return
+
+    for scaleset in scalesets:
+        if scaleset.state == ScalesetState.running:
+
+            max_size = min(scaleset.max_size(), autoscale_config.scaleset_size)
+            logging.info(f"Scaleset size: {scaleset.size}, max_size: {max_size}")
```
bmc-msft, line 28: please include the scaleset_id in this log

bmc-msft:
As you mentioned, this uses tasks instead of worksets and will need to get updated in the near future as we work towards running multiple tasks on the same machines.

bmc-msft (changes requested):
There are a few relatively minor changes that will help maintainability in the future, but this is over-all a positive step forwards and I'm looking forward to including it.

src/pytypes/onefuzztypes/models.py
anshuman-goel: updated it to use validators

src/cli/onefuzz/api.py
anshuman-goel: added it. this is amazing one.

src/api-service/__app__/onefuzzlib/__version__.py
```diff
@@ -2,4 +2,4 @@
 # Licensed under the MIT License.
 
 # pylint: disable=W0612,C0111
-__version__ = "0.0.0"
+__version__ = "1.1.0-7cd15c13623262bf9e8a07bc94c08413e24a5d5c"
```
bmc-msft, line 5: I think you accidentally committed this?

bmc-msft (changes requested):
With the exception of something I think was accidentally committed, this looks good.  I want to do extensive testing of this early next week and think it's probably ready to merge.

anshuman-goel:
> With the exception of something I think was accidentally committed, this looks good. I want to do extensive testing of this early next week and think it's probably ready to merge.

fyi: I have observed some bugs related to node registration when job is deleted or when scaleset is scaling down.

bmc-msft (approved):

PR #125, opened 2020-10-09
Filter agent app-insights logs based on log::max_level

bmc-msft:

As is, we send all levels of logging from the agent to app-insights, which is too verbose.  This aligns app-insights with log levels. 

On linux, this was previously set to 'info' and not set on windows.  This PR includes setting RUST_LOG to 'info' on windows.

### DIFF
--- a/src/agent/onefuzz/src/telemetry.rs
+++ b/src/agent/onefuzz/src/telemetry.rs
@@ -341,25 +341,26 @@ macro_rules! event {
 #[macro_export]
 macro_rules! log {
     ($level: expr, $msg: expr) => {{
-        use appinsights::telemetry::SeverityLevel;
-        use SeverityLevel::*;
+        use appinsights::telemetry::SeverityLevel::{
+            Critical, Error, Information, Verbose, Warning,
+        };
 
-        {
-            let log_level = match $level {
-                SeverityLevel::Verbose => log::Level::Debug,
-                SeverityLevel::Information => log::Level::Info,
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz/src/telemetry.rs
```diff
@@ -341,25 +341,26 @@ macro_rules! event {
...
+            Warning => log::Level::Warn,
+            Error => log::Level::Error,
+            Critical => log::Level::Error,
+        };
 
-        if let Some(client) = $crate::telemetry::client($crate::telemetry::ClientType::Instance) {
-            client.track_trace($msg, $level);
+        // while log::log will filter based on log level, the telemetry
+        // client does *not*.
+        if log_level <= log::max_level() {
```
ranweiler, line 358: My only uncertainty with this is how it interacts with module-specific `RUST_LOG` settings like `RUST_LOG=onefuzz_supervisor=trace`. During dev, I almost never set e.g. `RUST_LOG=trace`, because that makes us emit log messages from dependencies.

I think this is fine for now, but it would be nice to be able to tune the telemetry log level separately.

ranweiler (approved):

src/agent/onefuzz/src/telemetry.rs
bmc-msft: I think the next step would be to move to a more configurable logging mechanism, perhaps [fern](https://crates.io/crates/fern).

PR #129, opened 2020-10-09
Fixes typing error identified by a new mypy release

bmc-msft:
This fixes a build failure from a typing issue where we were sorting an Optional[Datetime], which is not directly sortable.

This was discovered during an update to mypy, which included a new typeshed that supports checks if list.sort arguments are sortable.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/task_event.py
+++ b/src/api-service/__app__/onefuzzlib/task_event.py
@@ -3,6 +3,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+import datetime
 from typing import List, Optional, Tuple
 from uuid import UUID
 
@@ -13,19 +14,10 @@ from .orm import ORMMixin
 
 
 class TaskEvent(BASE_TASK_EVENT, ORMMixin):
-    # handle sorting TaskEvents even if one or more of Timestamps are not set yet
-    def __lt__(self, other: "TaskEvent") -> bool:
-        if self.Timestamp is None:
-            return False
-        elif other.Timestamp is None:
-            return True
-        else:
-            return self.Timestamp < other.Timestamp
-
     @classmethod
     def get_summary(cls, task_id: UUID) -> List[TaskEventSummary]:
         events = cls.search(query={"task_id": [task_id]})
-        events.sort()
+        events.sort(key=lambda e: e.Timestamp or datetime.datetime.max)
 
         return [
             TaskEventSummary(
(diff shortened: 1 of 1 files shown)


bmc-msft:
The new feature to typeshed was introduced in this PR: https://github.com/python/typeshed/pull/4155

src/api-service/__app__/onefuzzlib/task_event.py
```diff
@@ -13,10 +13,19 @@
 
 
 class TaskEvent(BASE_TASK_EVENT, ORMMixin):
+    # handle sorting TaskEvents even if one or more of Timestamps are not set yet
+    def __lt__(self, other: "TaskEvent") -> bool:
+        if self.Timestamp is None:
+            return False
```
ranweiler, line 19: Why are we handling `None` asymmetrically?

ranweiler: Clarifying: we seem to be treating `None` as `+infinity`, sort of, so we can always compare `TaskEvent` instances, even if one or both have a `None` timestamp.

Why is this choice the right one? What about if we have multiple task event instances with `None` timestamps?

If it's actually always an error to have a `None` timestamp, should we throw an error instead of just picking some convention?

src/api-service/__app__/onefuzzlib/task_event.py
```diff
@@ -13,10 +13,19 @@
...
+        elif other.Timestamp is None:
+            return True
+        else:
+            return self.Timestamp < other.Timestamp
+
     @classmethod
     def get_summary(cls, task_id: UUID) -> List[TaskEventSummary]:
         events = cls.search(query={"task_id": [task_id]})
-        events.sort(key=lambda e: e.Timestamp)
+        events.sort()
```
ranweiler, line 28: Instead of implementing `__lt__`, what if we did `events.sort(key= lambda e: e.Timestamp or MAX_TIMESTAMP)`?

src/api-service/__app__/onefuzzlib/task_event.py
bmc-msft: This convention makes it such that the sorting is stable if we don't have a Timestamp and doesn't cause an Exception in the case we don't have Timestamp.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/task_event.py
```diff
@@ -16,7 +17,7 @@ class TaskEvent(BASE_TASK_EVENT, ORMMixin):
     @classmethod
     def get_summary(cls, task_id: UUID) -> List[TaskEventSummary]:
         events = cls.search(query={"task_id": [task_id]})
-        events.sort(key=lambda e: e.Timestamp)
+        events.sort(key=lambda e: e.Timestamp or datetime.datetime.max)
```
ranweiler, line 20: Maybe add a comment noting that we expect to never reach the rhs of the `or`, and that it doesn't matter if we do.

PR #133, opened 2020-10-12
start work towards enabling launching fuzzing within docker

bmc-msft:
Enables launching fuzzing jobs from within a docker container, which will go a long way towards bring-your-own-compute that isn't on Azure.

TODO:
* [ ] - Docker containers can't reboot.  Tasks that require rebooting should be rejected from running in containers


bmc-msft:
This PR also lost it's reference back to my fork.  Closing and recreating.

PR #135, opened 2020-10-12
add migration to fix notifications from pre-1.0.0 upgrade

bmc-msft:

In pre-1.0.0, notifications mapped notification_id to RowKey and Container to PartitionKey.  In the 1.0.0 release, these got swapped causing pre-1.0.0 upgraded systems to fail.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

A migration for pre-1.0.0 keys in the notifications table

## Validation Steps Performed

### DIFF
--- a/src/deployment/data_migration.py
+++ b/src/deployment/data_migration.py
@@ -3,11 +3,15 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+import argparse
+from uuid import UUID
 import json
 from typing import Callable, Dict, List
 
 from azure.cosmosdb.table.tablebatch import TableBatch
 from azure.cosmosdb.table.tableservice import TableService
+from azure.mgmt.storage import StorageManagementClient
+from azure.common.client_factory import get_client_from_cli_profile
 
 
 def migrate_task_os(table_service: TableService) -> None:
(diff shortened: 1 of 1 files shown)


chkeita (approved):

src/deployment/data_migration.py
```diff
@@ -48,3 +82,26 @@ def migrate(table_service: TableService, migration_names: List[str]) -> None:
         print("applying migration '%s'" % name)
         migrations[name](table_service)
         print("migration '%s' applied" % name)
+
+
+def main():
```
ranweiler, line 87: It'd be nice to have some user-facing INFO-level logging so it is clear what's going on.

src/deployment/data_migration.py
```diff
@@ -38,8 +42,38 @@ def migrate_task_os(table_service: TableService) -> None:
...
+def migrate_notification_keys(table_service: TableService) -> None:
+    table_name = "Notification"
+    notifications = table_service.query_entities(
+        table_name, select="PartitionKey,RowKey,config"
+    )
+    partitionKey = None
+
+    count = 0
+    for entry in notifications:
+        try:
```
ranweiler, line 54: Could you add a comment about why this a non-exceptional UUID parse means we don't have to do anything?

src/deployment/data_migration.py
```diff
@@ -38,8 +42,38 @@ def migrate_task_os(table_service: TableService) -> None:
     print("migrated %s rows" % count)
 
 
+def migrate_notification_keys(table_service: TableService) -> None:
```
ranweiler, line 45: We should maybe have much more explicit names and/or comments, including version info, for migration functions.

ranweiler (approved):

src/deployment/data_migration.py
bmc-msft: Something to think about for the future.

PR #143, opened 2020-10-14
radamsa built for windows

bmc-msft:
Adds radamsa for windows, built with msys


src/ci/radamsa-windows.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -ex
+
+export PATH=/usr/bin:$PATH
+pacman -S gcc gzip --noconfirm
+
+mkdir -p artifacts/third-party/radamsa-win64
+git clone https://gitlab.com/akihe/radamsa
+cd radamsa
+git checkout 8121b78fb8f87e869cbeca931964df2b32435eb7
```
ranweiler, line 16: Maybe worth just flagging what this commit is expected to correspond to, if there's anything important in that choice. Looks like a snapshot of latest `develop` at time of writing?

ranweiler (approved):

PR #146, opened 2020-10-14
Cancel tasks for assigned to a node that re-registers

bmc-msft:

There are a few cases where a node re-registers while it has work claimed.  While those cases should get fixed, this PR fails each of the tasks in these cases.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/api-service/__app__/agent_registration/__init__.py
```diff
@@ -107,6 +107,9 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
         )
     node.save()
 
+    # if any tasks were running during an earlier instance of this node, clear them out
+    node.mark_tasks_stopped_early()
```
ranweiler, line 111: If the supervisor rebooted for target setup, and registers again, does this cancel its task? IOW, does this break user-requested reboots?

ranweiler: Concerns addressed by #150, so this makes sense to merge after that.

ranweiler (approved):

Issue #149, opened 2020-10-14
Add the ability to limit job / task display 


bmc-msft:
With large numbers of concurrent tasks, it would be useful to limit the jobs / tasks that are displayed.

Possible options:
* job_id
* job project
* job project/name
* job project/name/build

PR #152, opened 2020-10-14
add view filter for jobid, project, and/or name

bmc-msft:

This adds --job_id, --project, and --name to limit the display of jobs and tasks from the 'status top' display to only things that match those values, if provided.

## PR Checklist
* [x] Applies to work item: #149 
* [ ] Tests added/passed

## Validation Steps Performed
Have multiple jobs currently running.

Run `onefuzz jobs list` and get information for a running job

Then iteratively test each of these:
1. `onefuzz status top --name NAME` 
2. `onefuzz status top --project PROJECT` 
3. `onefuzz status top --job_id JOB_ID` 
3. `onefuzz status top --job_id JOB_ID --project PROJECT --name NAME`


src/cli/onefuzz/status/top_view.py
```diff
@@ -34,6 +34,17 @@ def now() -> datetime:
 EXTRA_LINES_PER_WIDGET = 3
 
 
+def column_config(fields: Optional[List[str]]) -> List[Union[int, str]]:
```
ranweiler, line 37: 👍

src/cli/onefuzz/status/top_view.py
```diff
@@ -78,30 +89,32 @@ def __init__(self, screen: Any, cache: TopCache, show_details: bool):
...
-            "status": {"height": 1, "setup": [None, [1, "100%"]]},
+            "status": {"height": 1},
         }
 
         for name in ["status", "pools", "jobs", "tasks", "messages"]:
             if name == "messages" and not self.show_details:
                 continue
 
-            if dimensions[name]["setup"][0]:
+            titles = dimensions[name].get("setup")
```
ranweiler, line 101: I'm assuming this is typechecked, so the use sites below work as expected.

But even so, when reading, the possible values can be hard to keep straight because we are doing truthy checks on what I'm guessing is an `Optional[List[T]]`, for some `T`.

All that said, should we maybe just default `titles` to an empty list?

src/cli/onefuzz/status/cache.py
```diff
@@ -271,10 +290,31 @@ def add_job(self, job_id: UUID, state: JobState, job: Optional[Job] = None) -> N
             except Exception:
                 logging.debug("unable to find job: %s", job_id)
 
+    def should_render_job(self, job: Job) -> bool:
+        if self.job_filters.job_id is not None:
+            if job.job_id not in self.job_filters.job_id:
+                logging.info("skipping:%s", job)
```
ranweiler, line 296: Should this maybe be at a higher verbosity level? We'd only want to see this if debugging, right?

src/cli/onefuzz/status/cache.py
```diff
@@ -64,29 +65,34 @@ def fmt(data: Any) -> Any:
...
+        "Name",
+        "Files",
+        "Pool",
+        "Time left",
+    ]
+    POOL_FIELDS = ["Updated", "Pool", "Name", "OS", "State", "Nodes"]
+
+    def __init__(
+        self,
+        onefuzz: "Onefuzz",
```
ranweiler, line 91: Not new to this PR, but how about we edit this to `"OneFuzz"`?

ranweiler (approved):

src/cli/onefuzz/status/cache.py
bmc-msft: The logs are not actually printed here.  the TUI hides the logging output

src/cli/onefuzz/status/cache.py
bmc-msft: It needs to be the type name, which is Onefuzz.   I don't want to change the name of the type to OneFuzz, as that's a pretty large breaking change for the SDK.

ranweiler: Ahh, of course. I was totally mis-reading this as a default value.

bmc-msft: One of the difficulties in the 'late type binding as strings'

Issue #148, opened 2020-10-14
Reset unresponsive nodes based on the heartbeat


chkeita:
We need to reset nodes that have not received any heartbeats after some amount of time. 
This is continuation of the work done in #30

PR #154, opened 2020-10-15
Reimage dead nodes

chkeita:

_What is this about?_

## PR Checklist
* [x] Applies to work item: #148
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -722,6 +727,13 @@ def cleanup_nodes(self) -> bool:
                 else:
                     to_reimage.append(node)
 
+        idle_nodes_ids = NodeHeartbeat.get_dead_nodes(
```
ranweiler, line 730: Maybe just choose one of `idle`/`dead` for consistency and ease of reading.

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -722,6 +727,13 @@ def cleanup_nodes(self) -> bool:
...
+            [node_id for node_id in azure_nodes.keys()], datetime.timedelta(hours=1)
```
ranweiler, line 731: The `hours=1` determines what counts as "dead", right? Probably worth giving a constant in this module.

ranweiler, line 731: This call seems potentially expensive. Do we have a sense of how long we expect it to take? Should this be its own scheduled task?

chkeita: It not more expensive than the existing calls to the database. For reference 
- the target throughput is [2000 entity per sec](https://docs.microsoft.com/en-us/azure/storage/tables/scalability-targets#scale-targets-for-table-storage) assuming a 1-KiB entity size. 
- the node heartbeat record is about 258 bytes using these [numbers](https://docs.microsoft.com/en-us/rest/api/storageservices/Understanding-the-Table-Service-Data-Model?redirectedfrom=MSDN#property-types)
- the maximum number nodes under a scaleset is [1000](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#virtual-machine-scale-sets-limits). 
I think we can expect a query time well under a seconds in the worst case.

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -743,6 +757,10 @@ def cleanup_nodes(self) -> bool:
                     # only add nodes that are not already set to reschedule
                     to_reimage.append(node)
 
+        dead_nodes = Node.get_dead_nodes(self.scaleset_id, NODE_EXPIRATION_TIME)
+        for node in dead_nodes:
```
bmc-msft, line 761: Adding `node.set_halt()` would be good to add here, just in case the node comes back alive while we're trying to reset it.

bmc-msft (approved):

Issue #124, opened 2020-10-08
Manage Scalesets within pool


anshuman-goel:
Currently client has to manage all the scaling up and down by creating scalesets and managing its size. Adding it in service side will automatically enable scaling for clients w/o managing for it. The pool will be able to manage scalesets and #instances inside scalesets.

PR #158, opened 2020-10-15
Bug fixes for Pool Resize

anshuman-goel:
Mainly 2 bug fixes:
- No need to check for task state as we are already fetching tasks from available states.
- Not to delete all the free nodes. Instead delete only how much is determined before.

_What is this about?_
Other fixes:
- Sometimes nodes go into halt and delete themselves w/o reducing size of scaleset, so making sure scaleset maintains the size as required.
- Take into the account for the nodes which are marked as delete_requested.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request
There is one more weird state that I face which I look at later on is, node is in init state but also requested for delete.

_What does this include?_

## Validation Steps Performed

Modify api.py to enable autoscaling. Then create and pool and jobs and then delete them.


src/api-service/__app__/onefuzzlib/autoscale.py
```diff
@@ -1,8 +1,3 @@
-#!/usr/bin/env python
```
bmc-msft, line 1: please re-add the copyright lines

bmc-msft (changes requested):

bmc-msft:
Testing in-progress

bmc-msft:
Existing tests pass.  Once the copyright lines are added, I will merge.

src/api-service/__app__/onefuzzlib/autoscale.py
anshuman-goel: added

bmc-msft (approved):

PR #160, opened 2020-10-15
Remove update_event as a single event loop for the system

bmc-msft:
Builds on #159 

TODO: 
* [x] Move major components into their own timer loops (task/job in one, etc)
* [ ] Update Task timer loop to pull from the task heartbeat queue (requires Heartbeat updates)
* [ ] Update Node timer loop to pull from the node heartbeat queue (requires Heartbeat updates)
* [x] Merge pool resize & scaleset timer loops into a single loop
* [x] Ensure tasks stop after their duration
* [x] Ensure jobs stop after their duration
* [x] Ensure repro stop after their duration


src/api-service/__app__/onefuzzlib/autoscale.py
```diff
@@ -1,151 +1,141 @@
...
+
+
+def scale_up(pool: Pool, scalesets: List[Scaleset], nodes_needed: int) -> None:
+    logging.info("Scaling up")
+    autoscale_config = pool.autoscale
+    if not isinstance(autoscale_config, AutoScaleConfig):
+        return
+
+    for scaleset in scalesets:
+        if scaleset.state == [ScalesetState.running, ScalesetState.resize]:
```
anshuman-goel, line 17: `in` instead of `==`

bmc-msft:
The remaining todo items will be done in a separate PR once #164 is finished.

src/api-service/__app__/onefuzzlib/autoscale.py
```diff
@@ -1,151 +1,143 @@
-#!/usr/bin/env python
-#
-# Copyright (c) Microsoft Corporation.
```
ranweiler, line 3: We still want the (c) part of this, right?

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -66,6 +67,26 @@
 HOURS = 60 * 60
 
 
+class HasState(Protocol):
+    # TODO: this should be bound tighter than Any
```
ranweiler, line 71: Could you document that, in particular, we expect `state` to have a `name` property, so it must not be a bare `str`?

Incidentally, can we at least say it must an `Enum`?

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -66,6 +67,26 @@
...
+    state: Any
+
+
+def process_update(obj: HasState) -> None:
+    """ process through the state machine for an object """
+
+    # process the state machine up to 5 times unless it's stopped changing
+    for _ in range(0, 5):
```
ranweiler, line 79: 1. Why this choice of 5? 

2. The name is singular, but this might process many updates, right? It would be nice to have separate methods for (1) a single step of updating, returning a bool if the state changed, and (2) a wrapper, equivalen to this funciton, that calls the singel-step function in a loop, with e.g. a `max_updates` kwarg that defaults to 5.

3. More generally, does it may sense to just have this be part of one of the ORM classes?

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -195,6 +214,12 @@ def can_process_new_work(self) -> bool:
             self.stop()
             return False
 
+        if self.state in NodeState.ready_for_reset():
+            logging.info(
+                "can_schedule should be recycled.  machine_id:%s", self.machine_id
```
ranweiler, line 219: What does "recycled" mean here?

Do we really want to have this message be identical to the one below?

Do we need to call `stop()` here, too?

src/api-service/__app__/timer_workers/__init__.py
```diff
@@ -6,43 +6,44 @@
...
-
-        if hasattr(scaleset, scaleset.state.name):
-            getattr(scaleset, scaleset.state.name)()
-        return
+    process_update(scaleset)
 
 
 def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
+    Node.mark_oudated_nodes()
+    nodes = Node.search_states(states=NodeState.needs_work())
```
chkeita, line 30: These two lines is result is two queries to the node table. There will be another query to get the dead nodes too. 
it think it is worth getting all the nodes in one query and process them here in the service once.

src/api-service/__app__/onefuzzlib/autoscale.py
chkeita, line 3: was it intentional to remove this ?

src/api-service/__app__/onefuzzlib/tasks/scheduler.py
```diff
@@ -7,17 +7,34 @@
...
 from ..azure.containers import blob_exists, get_container_sas_url, save_blob
 from ..azure.creds import get_func_storage
+from ..pools import Pool
 from .config import build_task_config, get_setup_container
 from .main import Task
 
 HOURS = 60 * 60
 
 
+def schedule_workset(workset: WorkSet, pool: Pool, count: int) -> bool:
```
ranweiler, line 22: This `count` is `vm_count`, right? Can we make that explicit?

src/api-service/__app__/timer_tasks/__init__.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
ranweiler, line 1: Is the shebang needed?

src/api-service/__app__/timer_tasks/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
+    expired_tasks = Task.search_expired()
+    for task in expired_tasks:
+        logging.info("stopping task: %s", task.job_id)
+        task.stopping()
+
+    expired_jobs = Job.search_expired()
+    for job in expired_jobs:
+        logging.info("stopping job: %s", job.job_id)
```
ranweiler, line 26: Can we make this `"stopping expired job"`?

src/api-service/__app__/timer_workers/__init__.py
```diff
@@ -6,43 +6,44 @@
...
-        )
-
-        if hasattr(scaleset, scaleset.state.name):
-            getattr(scaleset, scaleset.state.name)()
-        return
+    process_update(scaleset)
 
 
 def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
+    Node.mark_oudated_nodes()
```
ranweiler, line 29: Typo either here and/or in the method name: want `outdated`.

src/api-service/__app__/timer_workers/__init__.py
```diff
@@ -6,43 +6,44 @@
...
-            scaleset.state,
-        )
-
-        if hasattr(scaleset, scaleset.state.name):
-            getattr(scaleset, scaleset.state.name)()
-        return
+    process_update(scaleset)
 
 
 def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
```
ranweiler, line 28: This function seems to be doing quite a bit of work. We may want to chop it up more in the future.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/autoscale.py
bmc-msft: Yes.  It's now in onefuzzlib/autoscale.py

src/api-service/__app__/timer_tasks/__init__.py
bmc-msft: This is common across the entire project.

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: 1. updated the messages to say reset or deletion as appropriate.
2. no, we don't want it to say stop here, as ready_for_reset() indicates the nodes have already seen the stop message.

src/api-service/__app__/timer_workers/__init__.py
bmc-msft: I will leave this alone for now, and will address this when adding the heartbeat checking.  (See discussion in Teams regarding this)

src/api-service/__app__/onefuzzlib/tasks/scheduler.py
ranweiler: That sounds reasonable.

PR #163, opened 2020-10-15
split node and task heartbeats in two nodes

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


bmc-msft (commented):
Please rename these to `queue_node_heartbeat` and `queue_task_heartbeat`

bmc-msft (changes requested):
This leaves the existing `queue_heartbeat` and the `heartbeat` queue as they exist.  Is that intended?

bmc-msft (approved):

src/api-service/__app__/queue_task_heartbeat/function.json
```diff
@@ -0,0 +1,18 @@
+{
+	"scriptFile": "__init__.py",
+	"bindings": [
+		{
+			"name": "msg",
+			"type": "queueTrigger",
+			"direction": "in",
+			"queueName": "node-heartbeat",
```
bmc-msft, line 8: This is the config for the task heartbeat, but pulling from the node-heartbeat.

bmc-msft (changes requested):
(see comment about pulling from the queue)

bmc-msft (approved):

PR #153, opened 2020-10-14
Resilient connection

chkeita:
## Summary of the Pull Request

Using an exponential backoff strategy for the http calls in the agent.

## PR Checklist
* [x] Applies to work item: #79 
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.
* [ ] Tests added/passed
* [ ] Requires documentation to be updated
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

_How does someone test & validate?_

PR #166, opened 2020-10-16
fail louder when uploads fail

bmc-msft:

Makes upload failures more verbose and fail fast.

## PR Checklist
* [x] Applies to work item: #153 
* [ ] Tests added/passed

## Info on Pull Request

1. Standardizes on a single SyncDir based director monitor.
2. Update said uploader to use `while ... monitor.next()` rather than `monitor.for_each` such that failures can be reported
3. Update azcopy wrapper to include stdout & stderr on failure, as some (but not all) errors in azcopy are printed to stdout.

## Validation Steps Performed

Run standard integration tests

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/utils.rs
+++ b/src/agent/onefuzz-agent/src/tasks/utils.rs
@@ -6,11 +6,12 @@ use std::time::Duration;
 
 use anyhow::Result;
 use async_trait::async_trait;
-use futures::{future::Future, stream::StreamExt};
+use futures::stream::StreamExt;
 use onefuzz::{
     az_copy,
     monitor::DirectoryMonitor,
-    telemetry::{Event::new_result, EventData},
+    telemetry::{Event, EventData},
+    uploader::BlobUploader,
 };
 use reqwest::Url;
 use tokio::{fs, io};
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz/src/az_copy.rs
```diff
@@ -49,11 +51,13 @@ pub async fn copy(src: impl AsRef<OsStr>, dst: impl AsRef<OsStr>, recursive: boo
 
     let output = cmd.spawn()?.wait_with_output().await?;
     if !output.status.success() {
+        let stdout = String::from_utf8_lossy(&output.stdout);
         let stderr = String::from_utf8_lossy(&output.stderr);
         anyhow::bail!(
-            "copy failed '{:?}' ==> '{:?}' : {}",
+            "sync failed src:{:?} dst:{:?} stdout:{:?} stderr:{:?}",
```
chkeita, line 57: copy failed

src/agent/onefuzz-agent/src/tasks/utils.rs
```diff
@@ -113,37 +114,23 @@ impl CheckNotify for tokio::sync::Notify {
...
-            let mut uploader = onefuzz::uploader::BlobUploader::new(url.url());
-            let result = uploader.upload(item.clone()).await;
-
-            if let Err(err) = result {
-                error!("couldn't upload item = {}, error = {}", item.display(), err);
-            } else {
-                verbose!("uploaded item = {}", item.display());
-            }
+    while let Some(item) = monitor.next().await {
+        event!(event.clone(); EventData::Path = item.display().to_string());
```
chkeita, line 127: The event should be fired after the upload is successful

bmc-msft: I'd prefer to know that the event happened, even if the upload fails.

chkeita (approved):

PR #170, opened 2020-10-17
limit size of error log saved as a Task Error

bmc-msft:

Limits stdout/stderr logged as part of the Task.error upon task failure to the last 4k of each.

### DIFF
--- a/src/api-service/__app__/agent_events/__init__.py
+++ b/src/api-service/__app__/agent_events/__init__.py
@@ -168,8 +168,8 @@ def on_worker_event(machine_id: UUID, event: WorkerEvent) -> None:
                     code=ErrorCode.TASK_FAILED,
                     errors=[
                         "task failed. exit_status:%s" % exit_status,
-                        event.done.stdout,
-                        event.done.stderr,
+                        event.done.stdout[-4096:],
+                        event.done.stderr[-4096:],
                     ],
                 )
             )


src/api-service/__app__/agent_events/__init__.py
```diff
@@ -168,8 +168,8 @@ def on_worker_event(machine_id: UUID, event: WorkerEvent) -> None:
                     code=ErrorCode.TASK_FAILED,
                     errors=[
                         "task failed. exit_status:%s" % exit_status,
-                        event.done.stdout,
-                        event.done.stderr,
+                        event.done.stdout[-4096:],
```
ranweiler, line 171: Nit, I'd find it slightly clearer to express this as `data[:4096]`.

I was worried 4096 might be a bit much, but I just checked the value against the overall [AI limits](https://docs.microsoft.com/en-us/azure/azure-monitor/app/pricing#limits-summary), and it seems reasonable to me.

ranweiler (approved):

src/api-service/__app__/agent_events/__init__.py
bmc-msft: data[:4096] and data[-4096:] are conceptually different things.  The first is the first 4096 bytes, the second is the last 4096 bytes.

ranweiler: Oops, of course.

PR #175, opened 2020-10-19
only emit events on a change of state

bmc-msft:

This PR updates the supervisor to only emit events when there is a change in state.

## Validation Steps Performed

1. Standard integration tests will verify it doesn't break the state machine
2. Check app-insights to verify we don't see sequential "Free" or "Busy" events from a given node

### DIFF
--- a/src/agent/onefuzz-supervisor/src/agent.rs
+++ b/src/agent/onefuzz-supervisor/src/agent.rs
@@ -107,10 +107,10 @@ impl Agent {
         match (&event, self.previous_state) {
             (StateUpdateEvent::Free, NodeState::Free)
             | (StateUpdateEvent::Busy, NodeState::Busy)
-            | (StateUpdateEvent::SettingUp{..}, NodeState::SettingUp)
+            | (StateUpdateEvent::SettingUp { .. }, NodeState::SettingUp)
             | (StateUpdateEvent::Rebooting, NodeState::Rebooting)
             | (StateUpdateEvent::Ready, NodeState::Ready)
-            | (StateUpdateEvent::Done{..}, NodeState::Done) => {}
+            | (StateUpdateEvent::Done { .. }, NodeState::Done) => {}
             _ => {
                 self.coordinator.emit_event(event.into()).await?;
             }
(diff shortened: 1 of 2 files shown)


src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -100,8 +104,10 @@ impl Agent {
     }
 
     async fn free(&mut self, state: State<Free>) -> Result<Scheduler> {
-        let event = StateUpdateEvent::Free.into();
-        self.coordinator.emit_event(event).await?;
+        if !matches!(self.previous_state, Some(StateValue::Free)) {
```
ranweiler, line 107: Maybe we should have a method like `emit_state_update_if_changed(new_state: NodeState)`, and just call that in each place we're currently calling `emit`.

src/agent/onefuzz-supervisor/src/scheduler.rs
```diff
@@ -35,6 +35,28 @@ impl fmt::Display for Scheduler {
     }
 }
 
+pub enum StateValue {
```
ranweiler, line 38: I think this is redundant. We can just nose [`NodeState`](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-supervisor/src/coordinator.rs#L52).

src/agent/onefuzz-supervisor/src/agent.rs
ranweiler: (and I suggest we just call it everywhere we update the state, not only the states that may currently transition back to themselves)

src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -21,6 +21,7 @@ pub struct Agent {
     work_queue: Box<dyn IWorkQueue>,
     worker_runner: Box<dyn IWorkerRunner>,
     _heartbeat: Option<AgentHeartbeatClient>,
+    previous_state: Option<StateValue>,
```
ranweiler, line 24: We could make this non-`Option`-al by setting it to `NodeState::Init`.

src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -92,16 +95,33 @@ impl Agent {
...
+        self.previous_state = NodeState::from(&next);
         let done = matches!(next, Scheduler::Done(..));
 
         self.scheduler = Some(next);
 
         Ok(done)
     }
 
+    async fn emit_state_update_if_changed(&mut self, event: StateUpdateEvent) -> Result<()> {
+        match (&event, self.previous_state) {
```
ranweiler, line 107: 👍 nice!

ranweiler (approved):

PR #176, opened 2020-10-19
Debugger improvements

lzybkr:

Refine the parameters of debugger callbacks to be more Rust friendly and pass the module name to dll load/process create.
Also improve symbolic breakpoint handling.

## Validation Steps Performed

Tested locally against a crashing target and other private testing.


src/agent/debugger/src/debugger.rs
```diff
@@ -205,15 +201,15 @@ pub trait DebugEventHandler {
...
         DBG_EXCEPTION_NOT_HANDLED
     }
-    fn on_create_thread(&mut self, _debugger: &mut Debugger, _info: &CREATE_THREAD_DEBUG_INFO) {}
-    fn on_create_process(&mut self, _debugger: &mut Debugger, _info: &CREATE_PROCESS_DEBUG_INFO) {}
-    fn on_exit_thread(&mut self, _debugger: &mut Debugger, _info: &EXIT_THREAD_DEBUG_INFO) {}
-    fn on_exit_process(&mut self, _debugger: &mut Debugger, _info: &EXIT_PROCESS_DEBUG_INFO) {}
-    fn on_load_dll(&mut self, _debugger: &mut Debugger, _info: &LOAD_DLL_DEBUG_INFO) {}
-    fn on_unload_dll(&mut self, _debugger: &mut Debugger, _info: &UNLOAD_DLL_DEBUG_INFO) {}
+    fn on_create_process(&mut self, _debugger: &mut Debugger, _name: &Path, _base_address: u64 ) {}
+    fn on_create_thread(&mut self, _debugger: &mut Debugger) {}
```
ranweiler, line 205: Isn't there any thread info we'd maybe want to pass here? An ID? Or even just the `hThread`?

ranweiler: Use case: I can imagine using this as a site to initialize per-thread symbolic context.

ranweiler (approved):

src/agent/debugger/src/debugger.rs
lzybkr: I decided to defer that decision until the first use case came up.

src/agent/debugger/src/debugger.rs
```diff
@@ -205,15 +201,15 @@ pub trait DebugEventHandler {
...
+    fn on_exit_process(&mut self, _debugger: &mut Debugger, _exit_code: u32) {}
+    fn on_exit_thread(&mut self, _debugger: &mut Debugger, _exit_code: u32) {}
+    fn on_load_dll(&mut self, _debugger: &mut Debugger, _name: &Path, _base_address: u64) {}
```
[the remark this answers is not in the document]
lzybkr: Good idea since the module name might at some point be derived from other parts such as the base address in the uncommon case of loading multiple dlls with the same base name.

src/agent/debugger/src/debugger.rs
ranweiler: Works for me.

src/agent/win-util/src/handle.rs
```diff
@@ -0,0 +1,40 @@
+use winapi::{
+    shared::minwindef::{FALSE, LPHANDLE},
+    um::{
+        handleapi::{CloseHandle, DuplicateHandle, INVALID_HANDLE_VALUE},
+        processthreadsapi::GetCurrentProcess,
+        winnt::{DUPLICATE_SAME_ACCESS, HANDLE},
+    },
+};
+
+pub struct Handle(pub HANDLE);
```
ranweiler, line 10: Nice, I like how you use this to scope the `Drop` impl.

PR #177, opened 2020-10-19
Adding version as param in script

anshuman-goel:

_What is this about?_
I need to modify `__version__.py` for every change and then build using `build_cli.ps1`. Introducing this change I can invoke it like `build_cli.ps1 -version "<version_number>`

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request
https://www.cyberciti.biz/faq/bash-shell-find-out-if-a-variable-has-null-value-or-not/

_What does this include?_

## Validation Steps Performed

Can be tested by running `.\build_cli.ps1 -version "1.3.1-78b655501368ce5e78bda632834f2e5ecdfd1bbb"` and `.\build_cli.ps1`


src/ci/unset-versions.sh
```diff
@@ -7,7 +7,12 @@ set -ex
 
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 GET_VERSION=${SCRIPT_DIR}/get-version.sh
```
anshuman-goel, line 9: it looks like we don't need `GET_VERSION` and `VERSION` during unset. will remove it.

anshuman-goel: removed

src/ci/set-versions.sh
```diff
@@ -7,7 +7,12 @@ set -ex
 
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 GET_VERSION=${SCRIPT_DIR}/get-version.sh
-VERSION=$(${GET_VERSION})
```
bmc-msft, line 10: Please use:

```bash
VERSION=${1:-$(${GET_VERSION})}
```

src/ci/unset-versions.sh
```diff
@@ -6,8 +6,6 @@
 set -ex
 
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
-GET_VERSION=${SCRIPT_DIR}/get-version.sh
```
bmc-msft, line 9: 👍

bmc-msft (changes requested):

bmc-msft (approved):

bmc-msft:
Testing in progress.

PR #181, opened 2020-10-20
allows hardcoding the version on build

bmc-msft:


src/agent/onefuzz-agent/build.rs
```diff
@@ -38,13 +46,18 @@ fn print_version(include_sha: bool, include_local: bool) -> Result<(), Box<dyn E
...
-    println!("cargo:rustc-env=ONEFUZZ_VERSION={}", version);
-
+    print_values(version, sha);
     Ok(())
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
+    let hardcode_version = env::var("ONEFUZZ_SET_VERSION");
+    let hardcode_sha = env::var("ONEFUZZ_SET_SHA");
+    if hardcode_version.is_ok() && hardcode_sha.is_ok() {
```
ranweiler, line 56: Suggestions:
- Replace `_SET_SHA` with something that explicitly mentions git, e.g. `_GIT_COMMIT` or `GIT_SHA`
- Allow default the checksum to the actual checksum; I'd expect there to be cases where I'd want to force a stamped-in build version to pass certain back-compat checks, but I wouldn't actually care about the checksum.

ranweiler (approved):

src/agent/onefuzz-agent/build.rs
bmc-msft: Yeah, having thought about it more, I am just removing the _SET_SHA all together.  The _SET_VERSION is enough for the actual needs (let me specify a version such that I can thunk a debug build into a running system)

PR #198, opened 2020-10-23
Add Onefuzz.container.reset to the SDK

anshuman-goel:

To reduce overload for scaling and for testing I need to delete particular set of container. Adding `container reset` cli argument.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

Also, 
- changing `reset --containers` to call `containers reset`
- sorting types in `ContainerType`

## Validation Steps Performed

Manually tested.

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -274,6 +274,74 @@ class Containers(Endpoint):
         self.logger.debug("list containers")
         return self._req_model_list("GET", responses.ContainerInfoBase)
 
+    def reset(
+        self,
+        analysis: bool = False,
+        coverage: bool = False,
+        crashes: bool = False,
+        inputs: bool = False,
+        no_repro: bool = False,
+        readonly_inputs: bool = False,
+        reports: bool = False,
+        setup: bool = False,
+        unique_reports: bool = False,
+    ) -> None:
+        """Reset all Conatiners unless specified.
+        [Caution]: It can lead to unexpected results.
+
+        :param bool analysis: Delete only analysis containers.
+        :param bool coverage: Delete only coverage containers.
+        :param bool crashes: Delete only crashes containers.
+        :param bool inputs: Delete only inputs containers.
+        :param bool no_repro: Delete only no_repro containers.
+        :param bool readonly_inputs: Delete only readonly_inputs containers.
+        :param bool reports: Delete only reports containers.
+        :param bool setup: Delete only setup containers.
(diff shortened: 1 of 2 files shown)


src/cli/onefuzz/api.py
```diff
@@ -274,6 +274,61 @@ def list(self) -> List[responses.ContainerInfoBase]:
         self.logger.debug("list containers")
         return self._req_model_list("GET", responses.ContainerInfoBase)
 
+    def reset(
```
bmc-msft, line 277: This should use:

`container_type: Optional[List[ContainerType]] = None`

bmc-msft: Can you add a helper method to ContainerType that indicates the types that are reset by default?

src/cli/onefuzz/api.py
```diff
@@ -274,6 +274,28 @@ def list(self) -> List[responses.ContainerInfoBase]:
...
+        self, container_types: Optional[List[enums.ContainerType]] = None
+    ) -> None:
+        """Reset all Conatiners unless specified.
+        [Caution]: It can lead to unexpected results.
+        """
+        if not container_types:
+            container_types = enums.ContainerType.__dict__["_member_map_"].values()
```
bmc-msft, line 284: It's easier to just iterate over the enum directoy.  `[x for x in enums.ContainerType]`

However, this erases *all* container types, including those we probably don't want to erase, like the various tools containers (such as `afl-linux`)

src/cli/onefuzz/api.py
anshuman-goel: updated and used same default as defined in earlier in `reset --container`

src/cli/onefuzz/api.py
```diff
@@ -274,6 +274,29 @@ def list(self) -> List[responses.ContainerInfoBase]:
         self.logger.debug("list containers")
         return self._req_model_list("GET", responses.ContainerInfoBase)
 
+    def reset(
+        self, container_types: Optional[List[enums.ContainerType]] = None
+    ) -> None:
+        """Reset containers unless specified.
+        [Caution]: It can lead to unexpected results.
```
anshuman-goel, line 281: @bmc-msft The cli is not able to read the lines after first one. Some bug exists over there.

src/cli/onefuzz/api.py
anshuman-goel: using `reset_defaults` with suggested iterator.

bmc-msft (approved):

PR #201, opened 2020-10-23
expose the ability manually override node reset

bmc-msft:


src/cli/onefuzz/debug.py
```diff
@@ -169,6 +170,19 @@ def _get_node(
 
         raise Exception("unable to find scaleset node running on task")
 
+    def disable_node_reset(
+        self, task_id: UUID_EXPANSION, *, node_id: Optional[UUID] = None
+    ) -> None:
+        while True:
+            try:
+                scaleset_id, node_id = self._get_node(task_id=task_id, node_id=node_id)
```
chkeita, line 178: How about moving those two operation on the server side to minimize the the chances that a node stops running the task before we update it

bmc-msft: Let's talk about this offline.  I wasn't happy with any of the endpoint APIs I prototyped to expose this.

chkeita (approved):

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -173,8 +179,21 @@ def on_worker_event(machine_id: UUID, event: WorkerEvent) -> None:
                     ],
                 )
             )
+            if task.config.debug and (
+                TaskDebugFlag.keep_node_on_failure in task.config.debug
+                or TaskDebugFlag.keep_node_on_completion in task.config.debug
+            ):
+                node.manual_reset_override = True
```
ranweiler, line 186: Can you revisit the name of this property? I think `manual` and `override` throw me off. How about something in the ballpark of `debug_keep_node`, to tie it to the other new property names.

src/api-service/__app__/node/__init__.py
```diff
@@ -55,6 +73,9 @@ def delete(req: func.HttpRequest) -> func.HttpResponse:
         )
 
     node.set_halt()
+    if node.manual_reset_override:
```
ranweiler, line 76: I'm guessing we guard on this to avoid the `save()` if it isn't needed?

src/cli/onefuzz/cli.py
```diff
@@ -241,19 +254,27 @@ def parse_annotation_class(
...
+                    "help": "(Default: True.  Sets value to False)",
+                }
+            elif default is None:
+                return {
+                    "type": arg_bool,
+                    "optional": True,
+                    "help": "Provide 'true' to set to true and 'false' to set to false",
+                }
+            else:
+                raise Exception("oh no! %s", repr(default))
```
ranweiler, line 277: How about `"Argument parsing error: %s"`, or similar

ranweiler (approved):

PR #206, opened 2020-10-24
address multiple issues found by pylint

### LINKED ISSUE #205: Exception in resetting VMs
The logging format used in onefuzzlib.azure.vm.get_extensionis wrong.

https://github.com/microsoft/onefuzz/blob/d363a22e39eb88e787e75e80325e515ef15cee28/src/api-service/__app__/onefuzzlib/azure/vm.py#L121-L126

bmc-msft:
1. overwriting module imports (`func` and `os`)
2. not using lazy-formatting in logging
3. attempting to use lazy-formatting in exceptions
4. invalid format string params (fixes #205 )
5. returning values on funcs declared to return None

### DIFF
--- a/src/api-service/__app__/onefuzzlib/pools.py
+++ b/src/api-service/__app__/onefuzzlib/pools.py
@@ -923,7 +923,7 @@ class Scaleset(BASE_SCALESET, ORMMixin):
 
     def update_nodes(self) -> None:
         # Be in at-least 'setup' before checking for the list of VMs
-        if self.state == self.init:
+        if self.state == ScalesetState.init:
             return
 
         nodes = Node.search_states(scaleset_id=self.scaleset_id)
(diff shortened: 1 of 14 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -923,7 +923,7 @@ def search_states(
 
     def update_nodes(self) -> None:
         # Be in at-least 'setup' before checking for the list of VMs
-        if self.state == self.init:
+        if self.state == ScalesetState.init:
```
ranweiler, line 926: So this evaluated `False` 100% of the time, right? Why didn't mypy catch this?

ranweiler (approved):

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Not sure.  It's something to investigate offline.

PR #211, opened 2020-10-26
Adds Radamsa & Coverage collection to integration tests

bmc-msft:


src/cli/examples/integration-test.py
```diff
@@ -333,10 +335,15 @@ def clear() -> None:
                 if job_id in self.containers:
                     del self.containers[job_id]
 
+        msg = "waiting on: %s" % ",".join(
+            sorted(x.config.name for x in self.jobs.values())
+        )
+        if len(msg) > 80:
```
ranweiler, line 341: This (and below) might be clearer (and not important, but more efficient) as a full if/else statement, rather than a mutating `if`.

ranweiler (approved):

src/cli/examples/integration-test.py
bmc-msft: I'm not sure that would make it clearer. It's more complex to compute the expected length of msg, rather than build it and to check if it's too long.

Given where this is used, the efficiency isn't a concern.

Issue #213, opened 2020-10-27
Minor updates to deployment script


anslutsk:
## Information

- Onefuzz version: 1.4.0
- OS: Windows

Updates:

(1) I hit the 25 second timeout that added to create_password() even though I was sitting at my desk on campus. I think Azure AAD can get backed up at certain times. I suggest increasing the timeout to 60 seconds.

(2) Unified command line args "--create_pool_registration" to match "--export_appinsights" since no value is used for the value given on the command line, e.g. False/True.

(3) Added logging for "Service principal password" since it can be used later with an automated deployment azure pipeline. I was experimenting with Anshuman's 'ado-deployment' branch.

PR #214, opened 2020-10-27
Minor updates to deployment script

anslutsk:

(1)
I hit the 25 second timeout that was recently added to create_password() even though I was sitting at my desk on campus. I think Azure AAD can get backed up at certain times. I suggest increasing the timeout to 60 seconds. I also noticed it typically takes at least 15 seconds after you do az login, for some reason, so maybe the combination of being busy and a fresh session causes AAD to lag. 

(2)
Unified command line args "--create_pool_registration" to match "--export_appinsights" since no value is used for the value given on the command line, e.g. False/True.

(3)
Added logging for "service principal password" since it can be used later on in an automated azure pipeline. I was experimenting with Anshuman's 'ado-deployment' branch. If I do one deployment manually, then I can just reuse these credentials instead of using AZ CLI to make another one.

## PR Checklist
* [X] Applies to work item: #213
* [X] CLA signed. 
* [X] Tests passed (see output below)
* [X] Does NOT require documentation to be updated
* [X] I haven't checked with anyone. If there are any issues we can discuss it during office hours.

## Info on Pull Request

Minor updates to deployment script

## Validation Steps Performed

I tested the deployment changes on Windows, here is the output:

$  python deploy.py westus2 example01 example01 "email@microsoft.com" --export_appinsights --create_pool_registration
INFO:deploy:service principal password: <PASSWORD_REMOVED>
INFO:deploy:deploying arm template: azuredeploy.json
INFO:deploy:creating eventgrid destination queue
INFO:deploy:creating eventgrid subscription
INFO:deploy:uploading tools from tools
INFO:deploy:uploading instance-specific-setup from instance-specific-setup
INFO:deploy:uploading third-party tools from third-party
INFO:deploy:deploying function app api-service.zip
INFO:deploy:adding appinsight log export
INFO:deploy:replacing existing export: <STRING REMOVED>
INFO:deploy:Updating application registration
DEBUG:msrest.async_paging:Paging async iterator protocol is not available for ApplicationPaged
DEBUG:msrest.async_paging:Paging async iterator protocol is not available for ApplicationPaged
DEBUG:msrest.async_paging:Paging async iterator protocol is not available for ApplicationPaged
INFO:deploy:Registration complete
INFO:deploy:These generated credentials are valid for a year
INFO:deploy:client_id: <CLIENT ID REMOVED>
INFO:deploy:client_secret: <SECRET REMOVED>
INFO:deploy:Telemetry collection on stats and OneFuzz failures are sent to Microsoft. To disable, delete the ONEFUZZ_TELEMETRY application setting in the Azure Functions instance
INFO:deploy:Update your CLI config via: onefuzz config --endpoint https://<NAME_REMOVED>.azurewebsites.net --authority <AUTHORITY_REMOVED> --client_id <CLIENT_ID_REMOVED>


src/deployment/deploy.py
```diff
@@ -306,6 +306,8 @@ def setup_rbac(self):
 
         (password_id, password) = self.create_password(app.object_id)
 
+        logger.info('service principal password: %s', password)
```
bmc-msft, line 309: This needs to include the app ID as well as the password.  Note, this is already being done later in this function, but at the 'debug' level.  

Can you add the command line option "--log_service_principal", and use existing logging at the end of the function at the 'info' level, and 'debug' if the command line is not set?
 
https://github.com/microsoft/onefuzz/blob/10267a3cb4dfd71895aaeb917110b7abecdd4503/src/deployment/deploy.py#L330-L333

src/deployment/deploy.py
```diff
@@ -217,15 +217,15 @@ def check_region(self):
...
+        # be created yet. Timeout after 60 seconds.
         count = 0
         while True:
             time.sleep(5)
             count += 1
             try:
                 return add_application_password(object_id)
             except CLIError as err:
-                if count > 5:
+                if count > 12:
```
bmc-msft, line 228: Can this be updated to use a constant?  It would be useful when we come back later to understand why it's 12.

bmc-msft (changes requested):

src/deployment/deploy.py
anslutsk: Accepted and updated my code in the PR

src/deployment/deploy.py
anslutsk: Accepted and updated my code in the PR. Let me know if you had a different solution in mind. I'm open to suggestions.

anslutsk (commented):
Updated PR

anslutsk:
I have tested the deploy.py script with and without the new arg "--log_service_principal"

bmc-msft (approved):

bmc-msft:
Performing integration testing

PR #217, opened 2020-10-27
Set _NT_SYMBOLS_PATH to local symbols unless otherwise specified

bmc-msft:
Processing inputs loads symbols for each module, which is excessively expensive.  This hard-codes the path for symbol loading to the setup container unless the user has set the symbol path explicitly via target_env

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
@@ -106,7 +106,6 @@ impl CoverageTask {
             fs::remove_dir_all(&dir.path).await?;
         }
         self.config.coverage.sync_push().await?;
-        processor.report_total().await?;
 
         info!(
             "recorded coverage for {} containers in `readonly_inputs`",
--- a/src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
@@ -106,6 +106,8 @@ impl CoverageTask {
             fs::remove_dir_all(&dir.path).await?;
         }
         self.config.coverage.sync_push().await?;
+        processor.report_total().await?;
+
 
         info!(
             "recorded coverage for {} containers in `readonly_inputs`",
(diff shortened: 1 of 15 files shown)


src/agent/onefuzz-agent/src/tasks/coverage/recorder.rs
```diff
@@ -138,18 +138,13 @@ impl CoverageRecorder {
 
         // Unless users explicitly set symbol path, don't perform external symbols
         // lookup during sancov extraction
-        if !self.config.target_env.contains_key("_NT_SYMBOL_PATH") {
-            cmd.arg("-sins");
-        }
-
         let mut target_env = self.config.target_env.clone();
-
+        let cache_path = format!("cache*{}", self.config.target_exe.parent().ok()?.display());
```
lzybkr: The cache will create directories `exe` and `dll`. Those are unlikely to conflict with user provided files/directories, but it might be safer and more idiomatic to put the symbols in a `symbols` directory under the target_exe directory.

src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
```diff
@@ -106,6 +106,7 @@ impl CoverageTask {
             fs::remove_dir_all(&dir.path).await?;
         }
         self.config.coverage.sync_push().await?;
+        processor.report_total().await?;
```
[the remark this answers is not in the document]
bmc-msft: Merge I thought was on another branch.  Removed.

src/agent/onefuzz-agent/src/tasks/coverage/recorder.rs
```diff
@@ -136,6 +136,25 @@ impl CoverageRecorder {
...
+        // Unless users explicitly set symbol path, don't perform external symbols
+        // lookup during sancov extraction
+        let mut target_env = self.config.target_env.clone();
+        let cache_path = format!(
+            "cache*{}",
+            self.config
+                .target_exe
+                .parent()
+                .ok_or_else(|| format_err!("invalid path parent"))?
+                .join("symbols")
```
ranweiler, line 148: What happens if the setup container has a pre-existing `symbols` dir that isn't well-formed as a WinDBG cache directory?

More generally, I thought we just wanted to set the `setup` dir as one of the _search_ paths, not the root of a _cache_ directory. Do we actually want the `cache*` part of this?

ranweiler (approved):

bmc-msft:
OBE by #222

Issue #209, opened 2020-10-26
Enable User assigned managed identity for the scaleset


chkeita:
Change the current scaleset deployment to use user assigned managed identity and assign it to a role at deployment.

PR #219, opened 2020-10-28
Enable User assigned managed identity for scalesets

chkeita:

Create a User Assigned Managed Identity at deplyoment 
Assign the idenity to scalesets deployed by the service

## PR Checklist
* [x] Applies to work item: #209 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

- Deployed
- Enabled "User assignment required" on the entreprise application of the instance
- Verified that job can run to completion


src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -123,3 +124,22 @@ def is_member_of(group_id: str, member_id: str) -> bool:
             CheckGroupMembershipParameters(group_id=group_id, member_id=member_id)
         ).value
     )
+
+
+def get_scaleset_idenity_resource_path() -> str:
```
bmc-msft, line 129: If `get_scaleset_principal_id` is safe to cache, we should cache this too.

bmc-msft:
Running integration tests

bmc-msft:
Integration tests pass

bmc-msft (approved):

src/api-service/__app__/onefuzzlib/agent_authorization.py
```diff
@@ -54,6 +55,8 @@ def try_get_token_auth_header(request: func.HttpRequest) -> Union[Error, TokenDa
 
 @cached(ttl=60)
 def is_authorized(token_data: TokenData) -> bool:
+    if get_scaleset_principal_id() == token_data.object_id:
```
ranweiler, line 58: Let's add a comment describing what situation this is checking, at a higher level. Is this an optimization? Or does the check below miss something?

src/api-service/__app__/onefuzzlib/agent_authorization.py
```diff
@@ -54,6 +55,8 @@ def try_get_token_auth_header(request: func.HttpRequest) -> Union[Error, TokenDa
 
 @cached(ttl=60)
 def is_authorized(token_data: TokenData) -> bool:
```
ranweiler, line 57: Not new to this PR, but let's add a comment emphasizing why we trust the `object_id` in the `TokenData`.

src/api-service/__app__/onefuzzlib/azure/creds.py
ranweiler, line 129: Typo, should be `.._identity_..`.

ranweiler, line 129: Is there documentation we can link to describing what path spec we are trying to satisfy? Maybe something in the managed identity docs?

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -123,3 +124,22 @@ def is_member_of(group_id: str, member_id: str) -> bool:
...
+    return "%s/Microsoft.ManagedIdentity/userAssignedIdentities/%s" % (
+        resource_group_path,
+        scaleset_id_name,
+    )
+
+
+@cached
+def get_scaleset_principal_id() -> UUID:
+    client = mgmt_client_factory(ResourceManagementClient)
+    uid = client.resources.get_by_id(get_scaleset_idenity_resource_path(), "2018-11-30")
```
ranweiler, line 144: Maybe this API version should be a constant.

Is there a specific ARM template valuable it needs to match, that we should document at its definition site?

src/deployment/deploy.py
```diff
@@ -395,6 +396,10 @@ def deploy_template(self):
             sys.exit(1)
         self.results["deploy"] = result.properties.outputs
 
+        assign_scaleset_role(
```
ranweiler, line 399: How about a log message here?

src/deployment/registration.py
```diff
@@ -298,28 +298,38 @@ def assign_scaleset_role(onefuzz_instance_name: str, scaleset_name: str):
...
         % scaleset_service_principal["id"],
-        body={
-            "principalId": scaleset_service_principal["id"],
-            "resourceId": onefuzz_service_principal["id"],
-            "appRoleId": lab_machine_role["id"],
-        },
     )
 
+    # check if the role is already assigned
+    if not seq(assignments["value"]).find(
```
ranweiler, line 319: For readability, can we maybe hoist this into a variable like:
```python
role_assigned = seq(assignments["value"]).find(lambda a: a["appRoleId"] == managed_node_role["id"])
if not role_assigned:
    query_microsoft_graph(..)
```

src/api-service/__app__/onefuzzlib/azure/creds.py
ranweiler: Generally: I think relating this code to the ARM template and what we're doing conceptually will be useful in the future.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/agent_authorization.py
chkeita: Already have a comment on line 50 for that

Issue #221, opened 2020-10-28
{setup_dir} needs to be added to variable expansion


bmc-msft:
We need to add `{setup_dir}` to the variable expansion options provided by Onefuzz.  This is important to provide target information about the setup path.  At the moment, this can *mostly* be calculated by the `basename {target_exe}`, but is not reliable.

Note, management of the underlying `{setup_dir}` is currently handled as part of the supervisor, not the agent.

See [docs/command-replacements.md](https://github.com/microsoft/onefuzz/blob/main/docs/command-replacements.md) for more information.

Issue #220, opened 2020-10-28
libfuzzer_coverage doesn't perform variable expansion


bmc-msft:
`libfuzzer_coverage` needs to include include support for variable expansion similar to `libfuzzer_fuzz`.

See [docs/command-replacements.md](https://github.com/microsoft/onefuzz/blob/main/docs/command-replacements.md) for more information.

bmc-msft:
@ranweiler , I believe this to be OBE as you're looking to remove this task.   Can you confirm?

PR #222, opened 2020-10-28
disable sympath in libfuzzer_coverage extraction unless it's set via target_env

bmc-msft:

Disables sympath used in cdb unless it's explicitly set via target_env.  This significantly improves the runtime performance of sancov extraction.

## Validation Steps Performed

Ran with `onefuzz-agent.exe debug libfuzzer-coverage`, once with `--target_env _NT_SYMBOL_PATH=c:\symbols` and once without.  Note that without is significantly faster.

## Follow on work
* Issue #220 tracks updating `libfuzzer_coverage` to support variable expansion appropriate.
* Issue #221 tracks updating variable expansion to support `{setup_dir}` such that the setup dir can be explicitly set.

### DIFF
--- a/src/agent/script/win64/libfuzzer-coverage/DumpCounters.js
+++ b/src/agent/script/win64/libfuzzer-coverage/DumpCounters.js
@@ -83,7 +83,14 @@ function processModule(module, results_dir) {
     return true;
 }
 
-function dumpCounters(results_dir, sample_name) {
+function dumpCounters(results_dir, should_disable_sympath) {
+    if (should_disable_sympath == true) {
+        logln(`disabling sympath`);
+        execute('.sympath ""');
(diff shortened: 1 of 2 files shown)


src/agent/script/win64/libfuzzer-coverage/DumpCounters.js
```diff
@@ -83,7 +83,14 @@ function processModule(module, results_dir) {
     return true;
 }
 
-function dumpCounters(results_dir, sample_name) {
```
bmc-msft, line 86: Note, `sample_name` was not provided by any of the users of this script and was only used to pass to a function that didn't have it as an argument.

lzybkr (approved):

PR #224, opened 2020-10-28
Cache SAS tokens used for container file management

anshuman-goel:

During scaling this operation is adding lot of calls to storage container.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

manually


src/cli/onefuzz/api.py
```diff
@@ -146,6 +147,7 @@ class Files(Endpoint):
 
     endpoint = "files"
 
+    @cached
```
anshuman-goel, line 150: @bmc-msft is there any suggested ttl for it?

bmc-msft: Given the common usage of the Files endpoint, I think adding a cache of 1 hour is fine.

Since ttl for memoization.cached is specified in seconds, please add a constant ONE_HOUR_IN_SECONDS and use that.

src/cli/onefuzz/api.py
```diff
@@ -162,7 +166,7 @@ def delete(self, container: str, filename: str) -> None:
         client = self._get_client(container)
         client.delete_blob(filename)
 
-    def get(self, container: str, filename: str) -> bytes:
+    def get(self, container: str, filename: str) -> Any:
```
anshuman-goel, line 169: not fully sure if it right way to do. if there are any suggestions, let me know.

bmc-msft: Instead of using Any, this is a case where `cast` should be used.

At the end of this function, replace:
```python
        return client.download_blob(filename)
```

with:
```
        downloaded = cast(bytes, client.download_blob(filename))
        return downloaded
```

Additionally, you will need to import `cast` from typing.

bmc-msft (approved):

bmc-msft:
Running integration tests

PR #228, opened 2020-10-29
Premium storage for fuzz data

bmc-msft:

Draft PR to use a premium account for fuzzing data

TODO:
* [ ] Validate this actually improves IOPs in a meaningful fashion
* [ ] Handle the upgrade case.

Upgrading will likely be challenging.  We will need to either require all tasks stop and migrate all of the data (which can be *extremely* time consuming) OR support a naïve form of multiplexing across multiple storage accounts.

This is primarily an experiment to identify if premium storage accounts should be used for the fuzzing data.  In the end, multiplexing across multiple storage accounts is likely going to be required either way.


bmc-msft:
This will not immediately solve the scaling needs.  We'll need to do the multiplexing work regardless.

PR #229, opened 2020-10-29
enable setting ensemble sync duration timer

bmc-msft:

This PR allows enables a --ensemble_sync_delay, which sets the duration for the ensemble sync duration for fuzzing tasks.  Note, this additionally provides the ability to disable ensemble syncing at large by setting the value to 0.

## Validation Steps Performed

1. Launch a fuzzing task with `--ensemble_sync_delay` of 5.  SSH into the node running the fuzzing task.  Note azcopy runs every 30 seconds.
2. Launch a fuzzing task with `--ensemble_sync_delay` of 0.  SSH into the node running the fuzzing task.  Note azcopy pull from the inputs never runs.

### DIFF
--- a/src/agent/onefuzz/src/syncdir.rs
+++ b/src/agent/onefuzz/src/syncdir.rs
@@ -77,7 +77,7 @@ impl SyncedDir {
             return Ok(());
         }
         let delay = Duration::from_secs(delay_seconds);
-    
+
         loop {
             self.sync(operation).await?;
             delay_with_jitter(delay).await;
@@ -143,7 +143,6 @@ pub async fn continuous_sync(
     }
 
     let delay = Duration::from_secs(delay_seconds);
-
 
     loop {
         for dir in dirs {
--- a/src/agent/onefuzz/src/syncdir.rs
+++ b/src/agent/onefuzz/src/syncdir.rs
@@ -73,6 +73,10 @@ impl SyncedDir {
         delay: Option<Duration>,
     ) -> Result<()> {
         let delay = delay.unwrap_or(DEFAULT_CONTINUOUS_SYNC_DELAY);
+        if delay.as_secs() == 0 {
+            return Ok(());
+        }
+
         loop {
             self.sync(operation).await?;
             delay_with_jitter(delay).await;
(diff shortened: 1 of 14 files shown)


src/agent/onefuzz/src/syncdir.rs
```diff
@@ -70,9 +70,14 @@ impl SyncedDir {
     pub async fn continuous_sync(
         &self,
         operation: SyncOperation,
-        delay: Option<Duration>,
+        delay_seconds: Option<u64>,
     ) -> Result<()> {
-        let delay = delay.unwrap_or(DEFAULT_CONTINUOUS_SYNC_DELAY);
+        let delay_seconds = delay_seconds.unwrap_or(DEFAULT_CONTINUOUS_SYNC_DELAY_SECONDS);
+        if delay_seconds == 0 {
```
chkeita, line 76: This is confusing. Without reading the implementation I would have expected that a delay of Some(0) continuously sync without any delay between syncs. 
I think the intent would be clearer if we don't call this function at all when sync is disabled

bmc-msft: The issue is backwards compatability.  Existing jobs don't have this defined, but instead don't have a value set and use ensemble.

We need to differentiate existing "no value set, therefore use ensemble with a 60s delay" from "no value set, disable ensemble".

I will add help to the sdk which will be printed during - -help that documents how to disable it.

chkeita: i was suggesting to do the check for the delay before we even call this function 
basically replacing calls like this
``` rust
let continuous_sync_task = inputs.continuous_sync(Pull, config.ensemble_sync_delay);
```
with something line 
``` rust
let continuous_sync_task = 
  if delay_seconds != 0{  
     inputs.continuous_sync(Pull, config.ensemble_sync_delay);
   } else {
    async{Ok()}
}
```
This makes it clearer at call site that we disable the sync when we receive 0.

chkeita (approved):

src/agent/onefuzz/src/syncdir.rs
bmc-msft: Ah, I understand.

My reason for putting it here was to reduce the number of places that the logic had to be specified.

PR #238, opened 2020-10-30
Autoscaling Azure Function

anshuman-goel:

While scaling to 1000 nodes, azure function starts to hang and needs more instances to be in working condition.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

When creating multiple jobs or creating a scaleset with a big size, there is a sudden surge in number of requests. Therefore, waiting for high cpu usage and then auto scaling does do much good. I am scaling rapidly in multiples of 5 and then scaling down slowly to make sure we have required number of instance to work with.

Sorting out the variable names as well because it was hard to find what I was looking for.

Ref. for work is an [azure sample repository](https://github.com/Azure/azure-quickstart-templates/blob/master/monitor-autoscale-webappserviceplan-simplemetricbased/azuredeploy.json) and azure portal.

## Validation Steps Performed

Tested manually on an existing subscription w/o the feature avail, creating a new resource, updating the existing one with the feature avail.

### DIFF
--- a/src/deployment/azuredeploy.json
+++ b/src/deployment/azuredeploy.json
@@ -37,26 +37,27 @@
             }
         }
     },
-    "variables": {
-        "scaleset_identity": "[concat(parameters('name'), '-scalesetid')]",
-        "telemetry": "d7a73cf4-5a1a-4030-85e1-e5b25867e45a",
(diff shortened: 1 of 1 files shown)


bmc-msft:
Performing integration tests

bmc-msft (approved):

PR #239, opened 2020-10-30
Adding version as a param to dev-deploy script

anshuman-goel:

Adding version as a param to dev-deploy script

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

I have to modify `__version__.py` many times and sometimes goes into commit. Adding version in script will give more flexibility for dev working on windows and function app.

Also, I have to delete bunch of `onefuzztypes-*-py3-none-any.whl`. Therefore, using regex now to delete them.

## Validation Steps Performed

manually tested.

### DIFF
--- a/src/api-service/dev-deploy.ps1
+++ b/src/api-service/dev-deploy.ps1
@@ -7,10 +7,10 @@ param (
     # Specifying version for function
     [Parameter(Mandatory = $false)]
     [string]
-    $version
+    $version = "0.0.0"
 )
 
-try { 
+try {
     Push-Location
     Set-Location "$app_dir/../pytypes"
     python setup.py sdist bdist_wheel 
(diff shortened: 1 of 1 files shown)


src/api-service/dev-deploy.ps1
```diff
@@ -14,9 +19,15 @@ try {
     Set-Location __app__
     (New-Guid).Guid | Out-File onefuzzlib/build.id -Encoding ascii
     (Get-Content -path "requirements.txt") -replace 'onefuzztypes==0.0.0', './onefuzztypes-0.0.0-py3-none-any.whl' | Out-File "requirements.txt"
+    if ($version -ne "$null") {
```
chkeita, line 22: if $version

chkeita, line 22: you can set the default value of $version to be "0.0.0" and remove the  version check here and below.
also the check should be $null not "$null"

chkeita (approved):

PR #243, opened 2020-11-02
Tests for scaleset size validation

anshuman-goel:

Adding test for the change #156 for scaleset size validation

## PR Checklist
* [x] Tests added/passed


src/pytypes/tests/test_models.py
```diff
@@ -28,5 +28,41 @@ def test_model(self) -> None:
             NotificationCreate.parse_obj(missing_container)
 
 
+class TestScaleset(unittest.TestCase):
+    def test_scaleset_size_lt_0(self) -> None:
+        self.assertRaises(
```
bmc-msft, line 33: please use `with self.assertRaises(ValueError):` here.  See `test_alnum_filter.py` for an example.

anshuman-goel: updated.

bmc-msft (changes requested):

bmc-msft (approved):

Issue #234, opened 2020-10-29
Add a unique per-install GUID to telemetry


bmc-msft:
In order to understand the growth of the number of OneFuzz installations, we should add a unique install identifier to the telemetry.

Note, as with all telemetry, this should be easy to disable and clearly explained in the telemetry documentation [docs/telemetry.md](https://github.com/microsoft/onefuzz/tree/main/docs/telemetry.md).

PR #245, opened 2020-11-02
add instance_id generated at install time

bmc-msft:

Adds a randomly generated UUID that is unique to each install that is included in telemetry.

## PR Checklist
* [x] Applies to work item: #234 
* [ ] Tests added/passed

## Info on Pull Request

This adds an `instance_id` that is generated at install time, and add it to the supervisor and agent configs to be used if telemetry is enabled.

NOTE: Including a per-install randomly generated UUID as an instance ID was approved during the initial internal privacy review.

## Validation Steps Performed

Standard integration tests

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/containers.py
+++ b/src/api-service/__app__/onefuzzlib/azure/containers.py
@@ -6,7 +6,7 @@
 import datetime
 import os
 import urllib.parse
-from typing import Any, Dict, Optional, Union, cast
+from typing import Dict, Optional, Union, cast
 
 from azure.common import AzureHttpError, AzureMissingResourceHttpError
 from azure.storage.blob import BlobPermissions, ContainerPermissions
@@ -129,11 +129,11 @@ def save_blob(
 
 def get_blob(
     container: str, name: str, account_id: Optional[str] = None
-) -> Optional[Any]:  # should be bytes
+) -> Optional[bytes]:
     service = get_blob_service(account_id)
     try:
         blob = service.get_blob_to_bytes(container, name).content
-        return blob
+        return cast(bytes, blob)
     except AzureMissingResourceHttpError:
         return None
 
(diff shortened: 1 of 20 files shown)


src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -129,11 +129,11 @@ def save_blob(
 
 def get_blob(
     container: str, name: str, account_id: Optional[str] = None
-) -> Optional[Any]:  # should be bytes
+) -> Optional[bytes]:
     service = get_blob_service(account_id)
     try:
         blob = service.get_blob_to_bytes(container, name).content
-        return blob
+        return cast(bytes, blob)
```
ranweiler, line 136: :+1:

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -102,6 +102,16 @@ def get_instance_url() -> str:
     return "https://%s.azurewebsites.net" % get_instance_name()
 
 
+@cached
+def get_instance_id() -> UUID:
+    from .containers import get_blob
+
+    blob = get_blob("base-config", "instance-id", account_id=get_func_storage())
+    if blob is None:
+        raise Exception("missing instance_id")
```
ranweiler, line 111: Should this be `instance-id` in the exception message? Thinking about greppability.

ranweiler: Oh, actually, is the `instance-id` string above just incorrect? Looks like we set the snake case version below.

src/deployment/deploy.py
```diff
@@ -464,6 +465,29 @@ def create_eventgrid(self):
                 % json.dumps(result.as_dict(), indent=4, sort_keys=True),
             )
 
+    def add_instance_id(self):
+        logger.info("setting instance_id log export")
+
+        container_name = "base-config"
+        blob_name = "instance_id"
```
ranweiler, line 472: Should be changed here or above.

ranweiler (changes requested):

ranweiler (approved):

PR #252, opened 2020-11-02
setting nodes for shutdown in scaleset shutdown

anshuman-goel:

I have observed that shutting down scaleset doesn't actually removes node. using similar strategy as in pools for it.

## PR Checklist
* [ ] Applies to work item: #124
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

manually


bmc-msft:
I understand the issue you're trying to address but I think aligning `Scaleset.shutdown` and `Pool.shutdown` is probably in order rather than this PR.

I'll work on that tomorrow.

bmc-msft:
In the short term, going forward with this  PR.  We can address cleaning up pool/scaleset/node shutdown cleanup in a later effort.

bmc-msft (approved):

bmc-msft:
Performing integration tests

PR #261, opened 2020-11-04
refactor agent_events handler

bmc-msft:

This PR refactors the agent_events function handler.

While debugging an agent that appeared to stay in the "init" state, we identified that a 'node state' event that should have set the state to 'free' was seen at the service, but somehow not saved.  It isn't clear why the update was not saved.  This PR attempts to refactor the agent_events handler in a handful of ways to make it easier to trace.

Note, it may be useful to review the individual commits, as GitHub is calling agent_events.py a "new file", rather than a renamed file.

## Info on Pull Request

1. Moves most of the actual state handling into onefuzzlib/agent_events.py.  Testing the function code is unfortunately complex.  This makes it such that we can call the individual methods from a repl shell or unit testing without the burdon of Azure Functions handlers
2. Simplifies processing NodeEventEnvelope by directly operating on the objects, rather than potentially casting into NodeEvents
3. Moves to explicit error handling rather than exceptions.
4. For everything that is a Union of complex tasks, moved to separate methods for the underling context (example, on_worker_event now calls on_worker_event_running and on_worker_event_done)
5. Always log in each of the primary on_state_update branches.  There were multiple that we didn't.
6. Since nodes now only send state updates on transition, we always save the node state as well as log the transition details in on_state_update.
7. Unless the node is marked for deletion, always save the state in when the NodeState is init.
8. The on_worker_event no longer updates node state.  This must come from on_state_update now

## Validation Steps Performed

1. Standard integration tests.
2. Have a scaleset with 3 nodes.  Submit a libfuzzer template jobs.  Once it's fully scheduled, submit another libfuzzer template job with a duration of 1 hour.  Stop the first one  (`onefuzz jobs delete JOBID`).  Verify the second job eventually gets fully scheduled.  Verify after 1 hour, the job stops and all the nodes get reimaged.

## TODO

* [x] Re-add task level updating node state.  Apparently, the agent isn't sending the state updates as we expect when a task state update also happens.  While that should be addressed, that's out of scope for this PR.  This PR needs to re-add that functionality back for the time being.


src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+        # This field will be required in the future.
+        # For now, it is optional for back compat.
+        setting_up_data = cast(
+            Optional[NodeSettingUpEventData],
+            state_update.data,
+        )
+
+        if setting_up_data:
+            if not setting_up_data.tasks:
+                raise Exception("setup without tasks.  machine_id: %s", machine_id)
```
chkeita, line 85: this should be an Error instance

bmc-msft: I used an exception on purpose here, as something is totally screwed up if we have a "setup tasks" with no tasks.

During our standup tomorrow, let's talk about this.  I think there is value in differentiating between errors that we should gracefully handle verses "ACK, THIS IS ALL WRONG" style issues.

chkeita (approved):

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+    WorkerEvent,
+    WorkerRunningEvent,
+)
+
+from ..onefuzzlib.pools import Node, NodeTasks
+from ..onefuzzlib.task_event import TaskEvent
+from ..onefuzzlib.tasks.main import Task
+
+
+def get_node(machine_id: UUID) -> Union[Node, Error]:
```
ranweiler, line 32: Can we a define a generic alias `Result[T] = Union[T, Error]`? Then this could be `Result[Node]`, and we could use it in return type of the `on_` handlers, too, which would become `Result[None]`.

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+            node.stop()
+            return None
+
+        if node.could_shrink_scaleset():
+            logging.info("stopping free node to resize scaleset: %s", node.machine_id)
+            node.set_halt()
+            return None
+
+    if state == NodeState.init:
+        if node.delete_requested:
```
ranweiler, line 60: Should we check for `reimage_requested`, too?

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+            state_update.data,
+        )
+
+        if setting_up_data:
+            if not setting_up_data.tasks:
+                raise Exception("setup without tasks.  machine_id: %s", machine_id)
+
+            for task_id in setting_up_data.tasks:
+                task = Task.get_by_task_id(task_id)
+                if isinstance(task, Error):
```
ranweiler, line 89: If we hit this error case, what should happen to the node? Right now, we would have already committed the state update, and we don't attempt to explicitly fail it or roll back.

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+                    machine_id=machine_id,
+                    task_id=task_id,
+                    state=NodeTaskState.setting_up,
+                )
+                node_task.save()
+
+    elif state == NodeState.done:
+        # if tasks are running on the node when it reports as Done
+        # those are stopped early
+        node.mark_tasks_stopped_early()
```
ranweiler, line 116: Double-checking: this only impacts _running_ tasks?

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+        return node
+
+    node_task = NodeTasks(
+        machine_id=machine_id, task_id=event.task_id, state=NodeTaskState.running
+    )
+    node_task.save()
+
+    if task.state in TaskState.shutting_down():
+        logging.info(
+            "ignoring task start from node.  machine_id:%s %s:%s (state: %s)",
```
ranweiler, line 152: 👍

src/api-service/__app__/onefuzzlib/agent_events.py
ranweiler: Also, this seems exceptional and a sign of a bug. The node is done, but there are tasks running. Does this imply error logging? Is there a normal path where this could happen that I'm not thinking of?

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if task.config.debug and (
+            TaskDebugFlag.keep_node_on_failure in task.config.debug
+            or TaskDebugFlag.keep_node_on_completion in task.config.debug
+        ):
+            node.debug_keep_node = True
+            node.save()
+
+    else:
+        logging.error(
```
ranweiler, line 214: This shouldn't be error; we're in the `not not exit_status.success` branch.

Maybe we should condition on `if exit_status.success:`, instead.

ranweiler (approved):

src/api-service/__app__/agent_events/__init__.py
```diff
@@ -70,184 +58,164 @@ def on_state_update(
 
     if state == NodeState.init:
         if node.delete_requested:
```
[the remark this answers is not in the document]
ranweiler: (ignore this comment, add comment to the latest code)

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: Good suggestion.  Implemented.

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: No.  I added comments to explain why.

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: Yes, this only impacts running tasks, though the logic is a bit convoluted.  Refactoring that to be more clear is out of scope for what this PR is focused on.

mark_tasks_stopped_early iterates over the NodeTasks, calling `task.mark_failed()` for each task.  `Task.mark_failed` checks if the task state is already in Stopping/Stopped.

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: I swapped the check around, and moved that to 'info'

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: Moved to an error

src/api-service/__app__/onefuzzlib/agent_events.py
bmc-msft: We moved to node state management to solely be at the purview of the supervisor.  Even with the exception in the weird case, the node state still is set to whatever the node declared.

src/pytypes/onefuzztypes/models.py
```diff
@@ -58,6 +58,10 @@ class Error(BaseModel):
     errors: List[str]
 
 
+ErrType = TypeVar("ErrType")
```
ranweiler, line 61: Suggestion: give this type var a different name, since it generically represents the _non_-`Err` case (to use the Rust analogy).

How about `OkType`?

src/api-service/__app__/onefuzzlib/agent_events.py
ranweiler: Okay, I buy that. That means the user-facing node state is what the supervisor last claimed it to be, whether or not that is in some sense inaccurate.

If we accept that, then my only complaint would be that if a node is running 2 tasks, and we can't `get_by_id()` the 1st one, but we _could_ for the 2nd one, we'll fail to update the 2nd task because we return early, even though something has only gone wrong with the 1st `task`. But that's not unique to this PR, and that case is actually currently unreachable, so shouldn't be addressed now.

PR #262, opened 2020-11-04
Comment update to register app as owner

anshuman-goel:

For better instruction updating the comment to indicate adding application as owner in resource group.

## PR Checklist
* [ ] Tests added/passed


chkeita (approved):

bmc-msft (changes requested):
While the `Owner` role is sufficient to deploy onefuzz, we should recommend using the minimum required roles.

Currently, [deployment-role.json](https://github.com/microsoft/onefuzz/blob/main/src/deployment/deployment-role.json) shows most of the roles required, but needs to be expanded given the recent change to user assigned managed identity creation on install.

anshuman-goel:
> While the `Owner` role is sufficient to deploy onefuzz, we should recommend using the minimum required roles.
> 
> Currently, [deployment-role.json](https://github.com/microsoft/onefuzz/blob/main/src/deployment/deployment-role.json) shows most of the roles required, but needs to be expanded given the recent change to user assigned managed identity creation on install.

Currently, user assigned managed identity don't work with service principal credential deployment. I will return back to this PR after the fix.

chkeita:
@anshuman-goel the deployment issue is caused by the fact that your deployment credential is missing the permission [AppRoleAssignment.ReadWrite.All](https://docs.microsoft.com/en-us/graph/permissions-reference#delegated-permissions-4) which require admin approval currently

anshuman-goel:
> While the `Owner` role is sufficient to deploy onefuzz, we should recommend using the minimum required roles.
> 
> Currently, [deployment-role.json](https://github.com/microsoft/onefuzz/blob/main/src/deployment/deployment-role.json) shows most of the roles required, but needs to be expanded given the recent change to user assigned managed identity creation on install.

It turn that I cannot create because there is a limit of 2000 custom roles per tenant.

bmc-msft:
A few thoughts.

1. Recommending owner is still probably not preferred.   Would it be acceptable to you to adjust it to "Please see X link for the roles required"?
2. With the change in #271, the additional role of `AppRoleAssignment.ReadWrite.All` is only required for install, not upgrade.

anshuman-goel:
> A few thoughts.
> 
> 1. Recommending owner is still probably not preferred.   Would it be acceptable to you to adjust it to "Please see X link for the roles required"?
> 2. With the change in #271, the additional role of `AppRoleAssignment.ReadWrite.All` is only required for install, not upgrade.

That looks reasonable. I will update the comment shortly.

bmc-msft (approved):

ryanspletzer:
> @anshuman-goel the deployment issue is caused by the fact that your deployment credential is missing the permission [AppRoleAssignment.ReadWrite.All](https://docs.microsoft.com/en-us/graph/permissions-reference#delegated-permissions-4) which require admin approval currently

To clarify is this the `AppRoleAssignment.ReadWrite.All` a delegated or application permission? Former is OK but latter is much higher privilege for the service principal... We have someone trying to set this up at our org and we're reviewing it.

> While the `Owner` role is sufficient to deploy onefuzz, we should recommend using the minimum required roles.
> 
> Currently, [deployment-role.json](https://github.com/microsoft/onefuzz/blob/main/src/deployment/deployment-role.json) shows most of the roles required, but needs to be expanded given the recent change to user assigned managed identity creation on install.

I'm assuming that of the common roles `Contributor` vs. `Owner`, you need to be in the `Owner` role on the sub for this to work, correct? Unless you use one of the more specific roles beyond these that includes the necessary permissions.

chkeita:
> To clarify is this the AppRoleAssignment.ReadWrite.All a delegated or application permission? Former is OK but latter is much higher privilege for the service principal... We have someone trying to set this up at our org and we're reviewing it.

the delegated permission should suffice according to the [documentation](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-approleassignedto?view=graph-rest-1.0&tabs=http#permissions)

> I'm assuming that of the common roles Contributor vs. Owner, you need to be in the Owner role on the sub for this to work,

Correct

> Unless you use one of the more specific roles beyond these that includes the necessary permissions (or if you make a custom role for this).

Yes, and the json file linked contains the definition for the custom role

PR #275, opened 2020-11-09
removing nodes whose ground truth is not avail

anshuman-goel:

This PR is more focused on making node more fault tolerant to network and availability issues. During my test runs I have observed that node exists in table but not in scaleset. Removing node if the ground truth doesn't contain `machine_id` should work.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

Let me know if it is incorrect way to address it.

## Validation Steps Performed

Manually. Create a fake entry in Node table and during time triggered runs it deletes that entry.


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -759,24 +759,28 @@ def cleanup_nodes(self) -> bool:
...
-        )
-
-        if not nodes:
-            logging.info("no nodes need updating: %s", self.scaleset_id)
-            return False
-
         # ground truth of existing nodes
         azure_nodes = list_instance_ids(self.scaleset_id)
 
+        nodes = Node.search_states(scaleset_id=self.scaleset_id)
```
bmc-msft, line 765: I expect a significantly small portion the Nodes to be in the ready_for_reset states at any given time.  With early exiting based on this expectation (previously on line 768), it made sense to not retrieve all of the nodes up front.

This PR removes the early exit, which is fine.  However, as a byproduct, we should not perform two queries against Node.

Instead, this PR should do something like this:

```python
nodes = Node.search_states(scaleset_id=self.scaleset_id)
nodes_to_reset = [x for x in nodes if x.state in NodeState.ready_for_reset()]
```

bmc-msft (changes requested):
see above comment

bmc-msft (approved):

bmc-msft:
Integration tests queued

PR #278, opened 2020-11-09
Autoscale test: I

anshuman-goel:

Adding a small test set to see if works well with pipeline and get more understanding on Onefuzz tests semantics.

## PR Checklist
* [x] Tests added/passed


bmc-msft (changes requested):
The unittests themselves look fine, but please put these in `src/api-service/tests/` directly instead of `src/api-service/tests/onefuzzlib`.

Multiple existing service-level tests that primarily focus on onefuzzlib live there.

bmc-msft (approved):

PR #280, opened 2020-11-10
Add uninstrumented block coverage recording

ranweiler:
Update the internal `coverage` crate to enable recording approximate block coverage for uninstrumented targets on Windows and Linux. This will let us add a new `generic_coverage` task for targets, with enough context for flexible output formats. We currently support multi-threading on both platforms, but not multi-processing.

We require some amount of symbol support: a local PDB for Windows targets, and an unstripped ELF symbol table for Linux targets.

This PR introduces a redundant type for recording block coverage. Future work will merge this and the existing internal block coverage format.

### DIFF
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -1,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub mod windows;
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -17,10 +17,7 @@ pub struct ModuleCov {
 }
 
 impl ModuleCov {
-    pub fn new(
-        module: impl Into<PathBuf>,
-        blocks: impl IntoIterator<Item = (u64, u32)>,
-    ) -> Self {
+    pub fn new(module: impl Into<PathBuf>, blocks: impl IntoIterator<Item = (u64, u32)>) -> Self {
         let module = module.into();
         let blocks = blocks
             .into_iter()
@@ -40,6 +37,10 @@ pub struct BlockCov {
 
 impl BlockCov {
     pub fn new(offset: u64, size: u32) -> Self {
-        Self { offset, size, count: 0 }
+        Self {
+            offset,
+            size,
+            count: 0,
+        }
     }
 }
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -47,6 +47,14 @@ pub struct BlockCov {
 
     /// Number of times a block was seen to be executed, relative to some input
     /// or corpus.
+    ///
+    /// Right now, we only set one-shot breakpoints, so the max `count` for a
+    /// single input is 1. In this usage, if we measure corpus block coverage
+    /// with `sum()` as the aggregation function, then `count` / `corpus.len()`
+    /// tells us the proportion of corpus inputs that cover a block.
+    ///
+    /// If we reset breakpoints and recorded multiple block hits per input, then
+    /// the corpus semantics would depend on the aggregation function.
     pub count: u32,
 }
 
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -12,9 +12,15 @@ pub mod windows;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
 
+/// Block coverage for a module.
+///
+/// May describe coverage relative to a single input or corpus of inputs.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct ModuleCov {
+    /// Absolute path to the module on the filesystem where coverage was recorded.
     pub module: PathBuf,
+
+    /// Mapping from basic block module-relative offsets to block coverage info.
     pub blocks: BTreeMap<u64, BlockCov>,
 }
 
@@ -30,10 +36,17 @@ impl ModuleCov {
     }
 }
 
+/// Coverage info for a specific block, identified by its offset.
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct BlockCov {
+    /// Offset of the block, relative to the module base load address.
     pub offset: u64,
+
+    /// Size of the block in bytes.
     pub size: u32,
+
+    /// Number of times a block was seen to be executed, relative to some input
+    /// or corpus.
     pub count: u32,
 }
 
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -1,5 +1,45 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#[cfg(target_os = "linux")]
+pub mod linux;
+
 #[cfg(target_os = "windows")]
 pub mod windows;
+
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ModuleCov {
+    pub module: PathBuf,
+    pub blocks: BTreeMap<u64, BlockCov>,
+}
+
+impl ModuleCov {
+    pub fn new(
+        module: impl Into<PathBuf>,
+        blocks: impl IntoIterator<Item = (u64, u32)>,
+    ) -> Self {
+        let module = module.into();
+        let blocks = blocks
+            .into_iter()
+            .map(|(o, sz)| (o, BlockCov::new(o, sz)))
+            .collect();
+
+        Self { module, blocks }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct BlockCov {
+    pub offset: u64,
+    pub size: u32,
+    pub count: u32,
+}
+
+impl BlockCov {
+    pub fn new(offset: u64, size: u32) -> Self {
+        Self { offset, size, count: 0 }
+    }
+}
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -25,11 +25,11 @@ pub struct ModuleCov {
 }
 
 impl ModuleCov {
-    pub fn new(module: impl Into<PathBuf>, blocks: impl IntoIterator<Item = (u64, u32)>) -> Self {
+    pub fn new(module: impl Into<PathBuf>, blocks: impl IntoIterator<Item = u64>) -> Self {
         let module = module.into();
         let blocks = blocks
             .into_iter()
-            .map(|(o, sz)| (o, BlockCov::new(o, sz)))
+            .map(|o| (o, BlockCov::new(o)))
             .collect();
 
         Self { module, blocks }
@@ -41,9 +41,6 @@ impl ModuleCov {
 pub struct BlockCov {
     /// Offset of the block, relative to the module base load address.
     pub offset: u64,
-
-    /// Size of the block in bytes.
-    pub size: u32,
 
     /// Number of times a block was seen to be executed, relative to some input
     /// or corpus.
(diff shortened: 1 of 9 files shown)


src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    for block in &mut blocks {
+        block.offset -= load_va;
+    }
+
+    Ok(blocks)
+}
+
+pub fn find_symbol_blocks(section: ElfSection, sym: ElfSymbol) -> Result<Vec<Block>> {
+    use iced_x86::*;
```
bmc-msft, line 407: Why are you importing here, instead of at a the file level?

ranweiler: Dev artifact. I'll do a style pass on any local/wildcard imports.

ranweiler: Done, kept a few local enum imports.

src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,499 @@
...
+            if let Err(err) = ptracer.restart(tracee, Restart::Syscall) {
+                log::error!("unable to restart tracee: {}", err);
+            }
+        }
+
+        Ok(())
+    }
+
+    fn update_images(&mut self, tracee: &mut Tracee) -> Result<()> {
+        let events = self.images.as_mut().unwrap().update()?;
```
ranweiler, line 92: These unwraps are assume the method is private and preconditions are met. Convert to `Result` values and use `?` instead.

src/agent/coverage/src/block.rs
```diff
@@ -0,0 +1,46 @@
...
+            .map(|(o, sz)| (o, BlockCov::new(o, sz)))
+            .collect();
+
+        Self { module, blocks }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct BlockCov {
+    pub offset: u64,
```
lzybkr, line 33: The fields could use comments.

src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,512 @@
...
+    let mut inst = Instruction::default();
+    while decoder.can_decode() {
+        decoder.decode_out(&mut inst);
+
+        if let Some(target) = branch_target(&inst) {
+            // The branch target is a leader.
+            leaders.insert(target);
+
+            // The next instruction is a leader, if it exists.
+            if decoder.can_decode() {
```
lzybkr, line 491: The Windows and Linux code should be unified.

This condition points out to me distinct (but probably harmless) bugs in both platforms:

* Linux (here) treats the fall thru instruction of an unconditional branch as a leader which isn't necessarily correct.
* The very similar Windows code incorrectly assumes the fall thru instruction of a conditional branch is reachable.

src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+fn branch_target(inst: &iced_x86::Instruction) -> Option<u64> {
+    use iced_x86::FlowControl;
+
+    match inst.flow_control() {
+        FlowControl::ConditionalBranch | FlowControl::UnconditionalBranch => {
+            Some(inst.near_branch_target())
+        }
+        _ => None,
```
lzybkr, line 510: I do have a slight preference to specify the cases explicitly here (as in the Windows equivalent code) because those cases do feel like a TODO (low priority) instead of a don't care.

src/agent/coverage/src/block/windows.rs
```diff
@@ -0,0 +1,175 @@
...
+            log::warn!(
+                "cannot record coverage for module = {}, err = {}",
+                module.path().display(),
+                err,
+            );
+            return;
+        }
+
+        // Won't panic, due to above check.
+        let bitset = bitset.unwrap();
```
lzybkr, line 74: Maybe use `and_then`/`or_else` or`match` instead of `unwrap?

src/agent/coverage/src/block/windows.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub fn pc(&self, dbg: &mut Debugger) -> Result<(u64, Option<SymInfo>)> {
+        let rip = iced_x86::Register::RIP;
+        let pc = dbg.read_register_u64(rip)?;
+        let sym = dbg.get_symbol(pc).ok();
+
+        Ok((pc, sym))
+    }
+
+    fn add_module(&mut self, dbg: &mut Debugger, module: &Module) {
+        let bitset = crate::pe::process_image(module.path(), false);
```
lzybkr, line 60: I think you'll want to cache the module processing somehow. Disassembly is pretty fast, but we'll be doing it thousands of times without a cache and that may be noticeable.

src/agent/coverage/src/block/windows.rs
```diff
@@ -0,0 +1,175 @@
...
+            bp_to_block,
+            coverage,
+            max_duration,
+            started,
+            timed_out,
+        }
+    }
+
+    pub fn pc(&self, dbg: &mut Debugger) -> Result<(u64, Option<SymInfo>)> {
+        let rip = iced_x86::Register::RIP;
```
lzybkr, line 52: Any reason to not inline this, perhaps with `use`?

src/agent/coverage/src/lib.rs
```diff
@@ -1,13 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-
-#![cfg(windows)]
 #![allow(clippy::as_conversions)]
 #![allow(clippy::new_without_default)]
 
+#[macro_use]
+extern crate serde;
```
lzybkr, line 7: I don't think this is still necessary.

src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,512 @@
...
+        while let Some(mut tracee) = ptracer.wait()? {
+            match tracee.stop {
+                Stop::SyscallEnterStop(..) => log::trace!("syscall-enter: {:?}", tracee.stop),
+                Stop::SyscallExitStop(..) => {
+                    self.update_images(&mut tracee)?;
+                }
+                Stop::SignalDeliveryStop(_pid, Signal::SIGTRAP) => {
+                    self.on_breakpoint(&mut tracee)?;
+                }
+                Stop::Clone(pid, tid) => {
```
ranweiler, line 75: We need to ensure that this was from a `clone()` with `CLONE_VM`, i.e. a new thread.

src/agent/coverage/src/block/linux.rs
```diff
@@ -0,0 +1,512 @@
...
+    // Slice symbol's data within section.
+    let lo = (sym.address() - section.address()) as usize;
+    let hi = lo + (sym.size() as usize);
+    let data = section.data()?;
+    let data = &data[lo..hi];
+
+    // Find basic block leaders, as VAs relative to the section load addr.
+    let leaders = find_function_leaders(data, sym);
+
+    let mut decoder = Decoder::new(64, data, DecoderOptions::NONE);
```
lzybkr, line 427: Comments elsewhere may address this, but raising again here as I see decoding every function twice which seems inefficient for our typical use case.

How will `offset` and `size` be used?

src/agent/coverage/src/block.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub size: u32,
+    pub count: u32,
```
lzybkr, line 35: Is this `count` of how many times a single input hits the block, or of how many inputs hit the block at least once?

lzybkr (changes requested):

src/agent/coverage/src/lib.rs
ranweiler: I was using it to bring the `derive` proc macros in scope throughout. That said, I just switched to module-local imports instead.

src/agent/coverage/src/block/windows.rs
ranweiler: I'm assuming you mean cached across multiple command invocations, yeah?

ranweiler: (and agreed)

src/agent/coverage/src/block/windows.rs
ranweiler: Sure, I think that'd be more idiomatic.

src/agent/coverage/src/block.rs
ranweiler: It is meant to support either, depending on context.

Right now, all breakpoints can only be hit once per input, on both platforms, so `count > 1` only when aggregating runs. In this case, for a corpus of `N` runs, `count / N` would tell us what proportion of the corpus covers the block.

If we added support for recording multiple hits per input, then the meaning would be dependent on the aggregation function used when merging (sum, avg, max, etc).

src/agent/coverage/src/block/windows.rs
ranweiler: Here's the `match` version:
```rust
let bitset = match bitset {
    Ok(bitset) => bitset,
    Err(err) => {
        log::warn!(
            "cannot record coverage for module = {}, err = {}",
            module.path().display(),
            err,
        );
        return;                
    }
};
```
I think it's often a bit more surprising to see an early return when using a `match` _expression_, vs. statement.

In the `if let`, the early return pattern felt more blatant. But maybe it's still better to not have an unreachable `unwrap()` to justify (and whose unreachability we'd need to preserve).

Which do you prefer?

src/agent/coverage/src/block/windows.rs
lzybkr: Mostly yes, but on Windows at least, a dll could be unloaded and loaded again.

src/agent/coverage/src/block.rs
lzybkr: Great, do add a comment.

Note that if breakpoint overhead were slow, we might only want to note hit/not hit so we can set breakpoints on unhit blocks only, though that might not be a big win if we're detecting coverage in every loaded module.

src/agent/coverage/src/block/windows.rs
lzybkr: I honestly prefer to read the `if let` version, but from a future maintainability point of view, prefer the match.

src/agent/coverage/src/block/windows.rs
ranweiler: Yeah, we could have the same thing happening on Linux. Opened #285 to capture this.

src/agent/coverage/src/block.rs
ranweiler: Yeah, "only break on uncovered" would be a nice mode for a hot path use case. For now I'll just make sure the meaning is clear in the comment.

src/agent/coverage/src/block/linux.rs
ranweiler: I think we can drop `size`. The original goal was to minimize the work (and file access) required to generate `lcov` traceinfo line coverage, but that probably properly belongs to a fully separate later pass, as it may need to (yet again) disassemble the block anyway.

src/agent/coverage/src/block/linux.rs
ranweiler: The subtlety here is that the leader is doing two things:
1. Explicitly determining the start of a basic block
2. Implicitly delimiting the _end_ of any (linearly) preceding block.

For use (1), the addr of the bytes after an unconditional jump may be incorrect in some corner cases. In particular, if they are not actually code. For (2), we want the non-fallthrough next instruction as a leader, _even if_ it is unreachable. If we didn't treat it as a leader, then the second sweep, which takes us from leaders to blocks, would give us "blocks" with unconditional jumps in the middle of the instruction stream!

I think this distinction is muddled by algorithms which (in an abstracted setting) assume non-mixing of code and data.

src/agent/coverage/src/block/linux.rs
ranweiler: At least on Ubuntu 18.04 (Linux 4.15), we only see `Clone` events if `VM_CLONE` was set. Added comment.

lzybkr (approved):

bmc-msft:
Performing integration tests

PR #281, opened 2020-11-10
Add application-insights debug cli

bmc-msft:
Quite frequently, I find myself needing to perform queries against an instance's application insights log.  Rather than continually going into the portal, this exposes the query API via the CLI.

TODO:
* [x] Add recording the instance specific `appId` for application insights to the ARM template for the Azure Function
* [x] Add said `appId` to the `onefuzz info get`
* [x] Update the debug code to use said "appId" from `info.get()` instead of `os.environ['APP_ID']`


bmc-msft:
The following example use searches for the keyword "scaleset".  Then using jmespath, prints the first message.

```
❯ onefuzz --endpoint https://pr-check-e250f3223ada4b2fb84a807c33063dfb.azurewebsites.net debug logs keyword scaleset --query 'PrimaryResult[0].message'
"Executed 'Functions.scaleset' (Succeeded, Id=bc1cb88e-9508-4e1f-b4b9-15d18ded3439, Duration=23ms)"
❯
```

src/cli/onefuzz/debug.py
```diff
@@ -262,6 +266,88 @@ def download_files(self, job_id: UUID_EXPANSION, output: Directory) -> None:
...
+        :param bool raw: Do not simplify the data result
+        """
+        creds, _ = get_azure_cli_credentials(
+            resource="https://api.applicationinsights.io"
+        )
+        client = ApplicationInsightsDataClient(creds)
+
+        app_id = self.onefuzz.info.get().insights_appid
+        if app_id is None:
+            raise Exception("instance does not have an insights_appid")
```
ranweiler, line 305: Maybe note the full env var name that we expect to be set.

src/cli/onefuzz/debug.py
```diff
@@ -262,6 +266,88 @@ def download_files(self, job_id: UUID_EXPANSION, output: Directory) -> None:
...
+    ) -> Any:
+        """
+        Perform an Application Insights keyword query akin to "Transaction Search"
+
+        :param str value: Keyword to query Application Insights
+        :param str timespan: ISO 8601 duration format
+        :param int limit: Limit the number of records returned
+        :param bool raw: Do not simplify the data result
+        """
+        components = ["union isfuzzy=true exceptions, traces, customEvents"]
```
ranweiler, line 333: Great, I was hoping this is how you'd approach the query building.

ranweiler: I suggest documenting that you are building KQL, which should help make it clearer what `|` means below, for example.

ranweiler (approved):

src/cli/onefuzz/debug.py
bmc-msft: At this stage, I'm not concerned about someone deleting the variable from the service.  

This provides context for the failures that could pop up from trying to use this feature against a previous release.

PR #282, opened 2020-11-10
Adding option to merge all inputs at once

chkeita:


bmc-msft:
Can 'merge' be it's on template, rather than part of the existing libfuzzer template?

tunz:
What if I want to use "merge" to minimize corpus, what would be the best approach?

I'm running these commands to minimize corpus.
```bash
mkdir empty_dir
azcopy sync {container} ./corpus
./fuzzer -merge=1 ./empty_dir ./corpus
azcopy sync ./empty_dir {container} --deleteDestination=true
```

I'm wondering if I can use this new merge task for this purpose. I imagine this might be one way:
1. Create a new container
2. Run merge with inputs=old container and unique_inputs=new container
3. Remove the old container
4. Copy corpus from the new container to the old container
5. Remove the new container

But then, some other process should wait for the task to be completed and run the copy step.

chkeita:
> What if I want to minimize corpus using "merge", what would be the best approach?
> 
> I'm running these commands to minimize corpus.
> 
> ```shell
> mkdir empty_dir
> azcopy sync {container} ./corpus
> ./fuzzer -merge=1 ./empty_dir ./corpus
> azcopy sync ./empty_dir {container} --deleteDestination=true
> ```
> 
> I'm wondering if I can use this new merge task for this purpose. I imagine this might be one way:
> 
> 1. Create a new container
> 2. Run merge with inputs=old container and unique_inputs=new container
> 3. Remove the old container
> 4. Copy corpus from the new container to the old container
> 5. Remove the new container
> 
> But then, some other process should wait for the task to be completed and run the copy step.

@tunz you will be able to use this task for that purpose. i am adding a new option to the libfuzzer template that will allow you to do this with this command. 
```
onefuzz template libfuzzer merge  <project> <name> <build> <pool_name> --inputs <input_dir>
```

tunz:
@chkeita Thanks for the new option :) This will helps us a lot.
But, I was also wondering if I can minimize the corpus within the same container (i.e. pushing results to the same container as the "inputs" container) because I want to run continuous fuzzing keeping the same container. If I understood correctly, I still have to copy the results to the inputs container back.

chkeita:
@tunz you are correct. You will need to manually copy the result of the merge from the unique_inputs container. 
if we output the result of the merge to the inputs container, it can result in some files being deleted. This will prevent other tasks like coverage from functioning properly.

bmc-msft:
I think it's reasonable to optionally support merging into the inputs container, especially if we're talking about performing in a bulk fashion.

bmc-msft:
Something important to remember, `merge` provides a minimum set of inputs that provide the same coverage.  By default, we should save the results to an minimized output container.  

However, it's common enough to say "ok, this is a known good binary, minimize my corpus such that I can use the minimized set moving forwards" without having manual effort

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -233,3 +234,113 @@ def basic(
         self.logger.info("done creating tasks")
         helper.wait()
         return helper.job
+
+    def merge(
```
chkeita, line 238: Will add an option to specify the output directory which could be the initial input directory

bmc-msft:
Removing auto-merge until integration tests have passed.

src/agent/onefuzz-agent/src/tasks/merge/libfuzzer_merge.rs
```diff
@@ -113,6 +101,41 @@ async fn process_message(config: Arc<Config>) -> Result<()> {
     }
 }
 
+async fn sync_and_merge(
+    config: Arc<Config>,
+    input_dirs: Vec<impl AsRef<Path>>,
+) -> Result<LibFuzzerMergeOutput> {
+    config.unique_inputs.sync_pull().await?;
+    match merge_inputs(config.clone(), input_dirs).await {
```
bmc-msft, line 109: Does this support deleting files?

A common use case for merging is to reduce the set of inputs that provide the minimal number of files with the maximal coverage.

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -233,3 +234,122 @@ def basic(
...
+        crash_report_timeout: Optional[int] = None,
+        tags: Optional[Dict[str, str]] = None,
+        wait_for_running: bool = False,
+        wait_for_files: Optional[List[ContainerType]] = None,
+        extra_files: Optional[List[File]] = None,
+        existing_inputs: Optional[List[Container]] = None,
+        dryrun: bool = False,
+        notification_config: Optional[NotificationConfig] = None,
+        debug: Optional[List[TaskDebugFlag]] = None,
+        overwrite_unique_inputs: bool = False,
```
chkeita, line 263: testing in progress

src/agent/onefuzz/src/az_copy.rs
```diff
@@ -15,7 +15,9 @@ pub async fn sync(src: impl AsRef<OsStr>, dst: impl AsRef<OsStr>) -> Result<()>
         .stderr(Stdio::piped())
         .arg("sync")
         .arg(&src)
-        .arg(&dst);
+        .arg(&dst)
+        .arg("--delete_destination")
```
ranweiler, line 19: Should this be:
```suggestion
        .arg("--delete-destination")
```

src/agent/onefuzz/src/az_copy.rs
```diff
@@ -15,7 +15,9 @@ pub async fn sync(src: impl AsRef<OsStr>, dst: impl AsRef<OsStr>) -> Result<()>
...
+        .arg(delete_dst.to_string());
```
ranweiler, line 20: Does this work the same as passing `--delete-destination=$bool`? (i.e. with an `=` instead of space)?

src/agent/onefuzz/src/az_copy.rs
chkeita: 👍

src/agent/onefuzz/src/az_copy.rs
chkeita: updated the logic to only add --delete-destination if delete_dst is true

bmc-msft (approved):

bmc-msft:
Standard integration tests pass, though I've not tested the new merge job.

docs/webhook_events.md
```diff
@@ -265,6 +265,10 @@ Each event will be submitted via HTTP POST to the user provided URL.
                 "ensemble_sync_delay": {
                     "title": "Ensemble Sync Delay",
                     "type": "integer"
+                },
+                "overwrite_output_container": {
```
ranweiler, line 269: I don't think this makes clear that this setting causes us to _delete missing_ files, which is very different from just "it's okay to overwrite".

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -233,3 +234,122 @@ def basic(
...
+        crash_report_timeout: Optional[int] = None,
+        tags: Optional[Dict[str, str]] = None,
+        wait_for_running: bool = False,
+        wait_for_files: Optional[List[ContainerType]] = None,
+        extra_files: Optional[List[File]] = None,
+        existing_inputs: Optional[List[Container]] = None,
+        dryrun: bool = False,
+        notification_config: Optional[NotificationConfig] = None,
+        debug: Optional[List[TaskDebugFlag]] = None,
+        overwrite_output_container: bool = False,
```
ranweiler, line 263: I think this should change to emphasize that we will _delete missing_ from the remote.

ranweiler (changes requested):

src/cli/onefuzz/templates/libfuzzer.py
ranweiler: (cc @chkeita)

chkeita: @ranweiler  is **delete_missing** you suggestion for the new name ?

bmc-msft: Perhaps 'minimize_result_set'

chkeita: How about, 'preserve_existing_outputs' and flip the logic

ranweiler: I like `preserve_existing_results` or `keep_redundant_results`. Or to flip it back to the original polarity, `delete_redundant_results`?

bmc-msft: I think the default should be to delete,and give the option to preserve.

ranweiler: Agreed. I guess if this ends up as a flag arg on the CLI, it would look like `--keep_redundant_results`, with a default of `false`. If we did `--delete_redundant_results` with a default of `true`, then disabling it would have to be specified as `--delete_redundant_results=false`. So maybe `--keep_..` is a better UX, given the desired default.

ranweiler: (I'm happy to go with whatever you two decide on, just wanted us to put our heads together on the name to emphasize that something may get _deleted_ in some cases, not just mutated)

chkeita: it looks like we have two votes (me and brian) for `preserve_existing_results` and one (joe) for `keep_redundant_results`.
I will go ahead and rename to `preserve_existing_results`

ranweiler (approved):

PR #283, opened 2020-11-10
Make supervisor heartbeat only fire on main loop progress

### LINKED ISSUE #277: Make supervisor heartbeat only fire on main loop progress
The supervisor heartbeat is currently a background thread that emits heartbeats on a timer, and whose liveliness is unrelated to any business logic / progress on the main agent loop.

Instead, make it so the heartbeat is **only** emitted on (for example) each iteration of the main loop [here](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-supervisor/src/agent.rs#L69-L84).

Note, we don't want to `await` on a heartbeat request, but just queue a heartbeat to be sent by a background thread. I believe the shared heartbeat code already fully supports that usage.

chkeita:

sends the supervisor heartbeat in the main loop
removed the default machine alien heartbeat from the tasks

## PR Checklist
* [x] Applies to work item: closes #277
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
+++ b/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
@@ -68,7 +68,7 @@ pub trait HeartbeatSender {
 
     fn alive(&self) {
         if let Err(error) = self.send(HeartbeatData::TaskAlive) {
-            error!("{}", error);
+            error!("failed to send heartbeat: {}", error);
         }
     }
 }
--- a/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
+++ b/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
@@ -48,8 +48,7 @@ pub async fn init_task_heartbeat(queue_url: Url, task_id: Uuid) -> Result<TaskHe
             let machine_id = context.state.machine_id;
             let machine_name = context.state.machine_name.clone();
 
-            let data =
-                HeartbeatClient::<TaskContext, _>::drain_current_messages(context.clone());
+            let data = HeartbeatClient::<TaskContext, _>::drain_current_messages(context.clone());
             let _ = context
                 .queue_client
                 .enqueue(Heartbeat {
--- a/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
+++ b/src/agent/onefuzz-agent/src/tasks/heartbeat.rs
@@ -64,26 +64,24 @@ pub async fn init_task_heartbeat(queue_url: Url, task_id: Uuid) -> Result<TaskHe
 }
 
 pub trait HeartbeatSender {
-    fn send(&self, data: HeartbeatData) -> Result<()>;
+    fn send(&self, data: HeartbeatData);
 
     fn alive(&self) {
-        self.send(HeartbeatData::TaskAlive).unwrap()
+        self.send(HeartbeatData::TaskAlive)
     }
 }
 
 impl HeartbeatSender for TaskHeartbeatClient {
-    fn send(&self, data: HeartbeatData) -> Result<()> {
+    fn send(&self, data: HeartbeatData) {
         let mut messages_lock = self.context.pending_messages.lock().unwrap();
         messages_lock.insert(data);
-        Ok(())
     }
 }
 
 impl HeartbeatSender for Option<TaskHeartbeatClient> {
-    fn send(&self, data: HeartbeatData) -> Result<()> {
-        match self {
-            Some(client) => client.send(data),
-            None => Ok(()),
+    fn send(&self, data: HeartbeatData) {
+        if let Some(client) = self {
+            client.send(data)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -44,7 +44,7 @@ impl Agent {
             setup_runner,
             work_queue,
             worker_runner,
-            _heartbeat: heartbeat,
+            heartbeat: heartbeat,
```
ranweiler, line 47: Since the arg and field names now match, you can replace `heartbeat: heartbeat` with `heartbeat`.

src/agent/onefuzz-supervisor/src/heartbeat.rs
```diff
@@ -54,3 +53,28 @@ pub async fn init_agent_heartbeat(queue_url: Url) -> Result<AgentHeartbeatClient
...
+    fn send(&self, data: HeartbeatData) -> Result<()>;
+
+    fn alive(&self) {
+        self.send(HeartbeatData::MachineAlive).unwrap()
+    }
+}
+
+impl HeartbeatSender for AgentHeartbeatClient {
+    fn send(&self, data: HeartbeatData) -> Result<()> {
+        let mut messages_lock = self.context.pending_messages.lock().unwrap();
```
ranweiler, line 67: How about a `?` here, instead, since you're already returning a `Result`?

ranweiler: You could also permit failure, and maybe only return an error if lock acquisition fails `N` times, where `N` is compared against a counter value stored in `self`/`AgentHeartbeatClient`.

ranweiler (approved):

src/agent/onefuzz-supervisor/src/heartbeat.rs
chkeita: ? does not work here  i get this error
```
 --> onefuzz-supervisor\src\heartbeat.rs:67:69
   |
67 |         let mut messages_lock = self.context.pending_messages.lock()?;
   |                                                                     ^ `std::sync::MutexGuard<'_, std::collections::HashSet<heartbeat::HeartbeatData>>` cannot be sent between threads safely
   |
   = help: within `std::sync::PoisonError<std::sync::MutexGuard<'_, std::collections::HashSet<heartbeat::HeartbeatData>>>`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, std::collections::HashSet<heartbeat::HeartbeatData>>`
   = note: required because it appears within the type `std::sync::PoisonError<std::sync::MutexGuard<'_, std::collections::HashSet<heartbeat::HeartbeatData>>>`
   = note: required because of the requirements on the impl of `std::convert::From<std::sync::PoisonError<std::sync::MutexGuard<'_, std::collections::HashSet<heartbeat::HeartbeatData>>>>` for `anyhow::Error`
   = note: required by `std::convert::From::from`
```
I removed the result return type. the expected behavior is to wait until the lock is free so no need to return a result.

bmc-msft: You could use map_err to convert the error to something that is `Send` here.

src/agent/onefuzz-supervisor/src/heartbeat.rs
```diff
@@ -54,3 +54,32 @@ pub async fn init_agent_heartbeat(queue_url: Url) -> Result<AgentHeartbeatClient
     );
     Ok(hb)
 }
+
+pub trait HeartbeatSender {
+    fn send(&self, data: HeartbeatData) -> Result<()>;
+
+    fn alive(&self) {
+        self.send(HeartbeatData::MachineAlive).unwrap()
```
ranweiler, line 62: I suggest _not_ unwrapping here, or anywhere else in the heartbeat code. If heartbeat sending fails, we don't want it to kill the agent.

src/agent/onefuzz-supervisor/src/heartbeat.rs
```diff
@@ -54,3 +54,32 @@ pub async fn init_agent_heartbeat(queue_url: Url) -> Result<AgentHeartbeatClient
...
+    }
+}
+
+impl HeartbeatSender for AgentHeartbeatClient {
+    fn send(&self, data: HeartbeatData) -> Result<()> {
+        let mut messages_lock = self
+            .context
+            .pending_messages
+            .lock()
+            .map_err(|_| Error::new(ErrorKind::Other, "Unable to acquire the lock"))?;
```
ranweiler, line 72: You could also just use `anyhow::format_err!("Unable to acquire heartbeat lock")`, if you just want to create a new error value with a string message.

src/agent/onefuzz-agent/src/tasks/heartbeat.rs
```diff
@@ -68,7 +67,9 @@ pub trait HeartbeatSender {
     fn send(&self, data: HeartbeatData) -> Result<()>;
 
     fn alive(&self) {
-        self.send(HeartbeatData::TaskAlive).unwrap()
+        if let Err(error) = self.send(HeartbeatData::TaskAlive) {
+            error!("{}", error);
```
ranweiler: :+1: perfect

ranweiler (approved):

Issue #241, opened 2020-10-30
Possible race when reimaging nodes, deleting node registration


ranweiler:
The day after an upgrade to 1.5.0, an internal user observed a task fail with the following error:
```
        "error": {
            "code": 468,
            "errors": [
                "node reimaged during task execution"
            ]
        },
```
Investigating turned up a huge number of trace log events at `error` level, like so:
```
error running supervisor agent: HTTP status client error (404 Not Found) for url 
(https://<..>.azurewebsites.net/api/agents/registration?machine_id=<..>&pool_name=<..>&version=1.5.0&scaleset_id=<..>)
``` 
This 404 can only happen when a VMSS node makes a `GET` to the agent registration endpoint, but no registration exists (i.e. there is no entry for the machine in the `Node` table). This HTTP call in turn only happens when the agent calls `renew()` on a registration, which occurs if one the SAS URLs in a registration/dynamic config body is stale, or when loading a cached registration on startup (from `/onefuzz/etc/dynamic-config.json`).

There were an improbably large number of these, suggesting there may be race between node reimaging and deleting the node registration entry from the `Node` table.

bmc-msft:
This could be also triggered by Azure Functions.  If azure functions doesn't think the endpoint exists, it will also generate a 404.  On Teams,  you mentioned that the instance was being upgraded.  There is a real possibility that there is a race condition on Azure Functions upgrade.

At a minimum, we should not return 404 from the agent registration endpoint.

ranweiler:
Thank you @bmc-msft, I didn't realize that!

I looked at the telemetry again with that in mind. The 404s all occur in appx. 1 min clusters. **Each** cluster is preceded by an `info` trace event "Initializing Warmup Extension".

- `Category` = `Host.Triggers.Warmup`
- `SDK version` = `azurefunctions: 3.0.14785.0`

It equally correlates with `Starting JobHost` messages, from `Microsoft.Azure.WebJobs.Hosting.JobHostService`.

ranweiler:
IIUC, the warm trigger being correlated is particularly interesting because we don't define one, it's only available to apps running on a Premium plan, and it is only called during scale-out operations ([ref](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=python#trigger---limits)).

bmc-msft:
[...]

PR #287, opened 2020-11-11
Don't use HTTP 404 on agent refresh error

bmc-msft:
As discussed in #241, we should not use 404 status code on agent registration failure.

This change will cause agent registration failure to use the default status code for `not_ok`, which is 400.


bmc-msft:
Passed integration tests

ranweiler (approved):

PR #288, opened 2020-11-11
Render envelope.event as json

bmc-msft:
This PR updates the agent event logging to use a json representation, rather than repr, of the event.

By using pydantic's json renderer, we can specify exclude_none, which removes 'error=None' from showing up in the logs.


bmc-msft:
Passed integration tests

ranweiler (approved):

PR #289, opened 2020-11-11
reduce calls to containers.list()

bmc-msft:
Once an instance has massively large numbers of jobs, `containers.list()` becomes unacceptably slow.  

This removes `container.list()` in two places.

1. Template helper.  As creating containers is effectively idempotent, it's more efficient to always create the container, rather than check if the container exists and create it otherwise.
2. Template deletion.  As `container.delete()` returns a `BoolResult` of "true if deleted, false if it didn't exist", we can still indicate if the deletion occurred, without the overhead of checking if the container exists prior to deletion.


bmc-msft:
passed integration tests

ranweiler (approved):

PR #290, opened 2020-11-11
include deleting unique_inputs on containers.reset

bmc-msft:
For `onefuzz containers reset`, we filter which containers to delete and only delete based on context.    This specifically allows us to not delete tools containers while removing all of the fuzzing metadata.

This PR adds unique_inputs as a containertype that gets reset, along with all of the other fuzzing metadata containers.


bmc-msft:
passed integration tests

ranweiler (approved):

PR #291, opened 2020-11-11
Add support for bring-your-own-compute for fuzzing via docker agents

bmc-msft:
Enables launching fuzzing jobs from within a docker container, which will go a long way towards bring-your-own-compute that isn't on Azure.

TODO:
* [ ] Docker containers can't reboot. Tasks that require rebooting should be rejected from running in containers


mgreisen:
This is off the road map for now.

PR #294, opened 2020-11-11
Windows debugger updates

lzybkr:

Set execution context before calling callbacks so the callbacks see the correct state.
Implements something like `.bpsync` so no breakpoints are missed when restoring a breakpoint.
Also simplify debug tracing (which helped understand issues in missing breakpoints).

## Validation Steps Performed

Tested locally against private targets.


src/agent/debugger/src/debug_event.rs
```diff
@@ -41,62 +41,66 @@ impl<'a> Display for DebugEventInfo<'a> {
...
+                        "second_chance"
+                    } else {
+                        "first_chance"
+                    }
                 )?;
             }
-            CreateThread(_info) => {
-                write!(formatter, "event=CreateThread")?;
+            CreateThread(info) => {
+                write!(formatter, "CreateThread handle={:x}", info.hThread as u64)?;
```
ranweiler, line 55: Probably want `handle=0x{:x}` to avoid confusion.

src/agent/debugger/src/target.rs
```diff
@@ -150,17 +212,24 @@ impl Target {
...
-        self.thread_handles.insert(thread_id, thread_handle);
+        self.thread_info.insert(
+            thread_id,
+            ThreadInfo::new(thread_id, thread_handle, self.wow64),
+        );
     }
 
     pub fn set_current_thread(&mut self, thread_id: DWORD) {
-        self.current_thread_handle = *self.thread_handles.get(&thread_id).unwrap();
+        self.current_thread_handle = self.thread_info.get(&thread_id).unwrap().handle;
```
ranweiler, line 228: Not new, but should we really be `unwrap()`-ing the `get()`, here?

ranweiler (approved):

src/agent/debugger/src/debug_event.rs
lzybkr: It's actually more useful without the `0x` when searching by word, e.g. in vim, typing `*`, it will match the `process.thread` text used in other messages.

src/agent/debugger/src/target.rs
lzybkr: It does seem panic worthy, as in we or the the Win32 debug api has a serious logic error, but I'll consider adding a log message in another PR.

PR #296, opened 2020-11-11
Event based webhooks

bmc-msft:
This PR adds support for arbitrary third-party integrations via webhooks, modeled after [Github Webhooks](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/about-webhooks).

Users will be able to specify URL that OneFuzz will call and a set of events that will hit said URL.

Currently, the event types supported are:
* task_created (sends a WebhookEventTaskCreated)
* task_stopped (sends a WebhookEventTaskStopped)
* task_failed (sends a WebhookEventTaskFailed)

The webhooks will atempt up to 5 times, then be marked as failed.

Features:
* `onefuzz webhook logs UUID` retrieves a log of all of the messages sent to the webhook for the last 7 days.
* `onefuzz webhook ping UUID` sends a special webhook with a generated GUID to verify the webhooks are setup.
* Webhook logs are automatically removed after 7 days.
* Webhook schema (using openapi spec) have documentation & example payloads provided via code generation.  (CI verifies that said documentation is up-to-date).
* Messages sent to a webhook can optionally be verified using [HMAC-SHA512](https://docs.microsoft.com/en-us/security/sdl/cryptographic-recommendations#machmackeyed-hash-algorithms).


src/api-service/__app__/onefuzzlib/webhooks.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+
+class Webhook(BASE_WEBHOOK, ORMMixin):
+    @classmethod
+    def key_fields(cls) -> Tuple[str, Optional[str]]:
+        return ("webhook_id", "url")
+
+    @classmethod
+    def send_event(cls, event_type: WebhookEventType, event: WebhookEvent) -> None:
```
chkeita, line 133: can't we just infer the event_type form the event type ?

bmc-msft: We can, and it's probably reasonable for the service side to do that.  That said, I wanted to specifically include the event type in the messages that we send.

bmc-msft:
Until this is merged, this link should work to view the generated webhook documentation.

https://github.com/bmc-msft/onefuzz/blob/event-based-webhooks/docs/webhook_events.md

src/api-service/__app__/onefuzzlib/webhooks.py
chkeita: I agree to keep the event type on the message the suggestion is only for the service. basically changing the signature to 
`def send_event(cls, event: WebhookEvent) -> None` and use the type of the event to infer the event type in the message

src/api-service/__app__/onefuzzlib/webhooks.py
```diff
@@ -0,0 +1,222 @@
...
+    event_id: UUID
+
+
+class WebhookMessageLog(BASE_WEBHOOK_MESSAGE_LOG, ORMMixin):
+    @classmethod
+    def key_fields(cls) -> Tuple[str, Optional[str]]:
+        return ("webhook_id", "event_id")
+
+    @classmethod
+    def search_expired(cls) -> List["WebhookMessageLog"]:
```
chkeita, line 38: Any thought on using a an azure queue instead of a table to store the message. we can take advantage of the message visibility for the retrying logic and the message expiration for and removing expired messages.

bmc-msft: There is a queue used for the actual orchestration.  It uses the message visibility to delay the retry.

However, by using a table for storing the underlying message log, we enable getting a list of recent deliveries.

Inspiration; [Github Webhook Events - Recent Deliveries](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/testing-webhooks#listing-recent-deliveries)

chkeita (approved):

PR #297, opened 2020-11-11
Change prepare_test_result signature

lzybkr:

This small change isn't currently used in OneFuzz but helps a dependent project.


src/agent/input-tester/src/tester.rs
```diff
@@ -206,7 +206,7 @@ impl Tester {
             // and return a collection of results to possibly be reported.
             let mut test_results = vec![];
             for result in results {
-                match self.prepare_test_result(result) {
+                match self.prepare_test_result(&result) {
```
ranweiler, line 209: If this function really requires ownership, maybe we should have the caller invoke `clone()` at the use site? Or maybe iterate with `results.drain(..)`?

lzybkr: It doesn't always require ownership, it clones what it needs if it needs to.

ranweiler (approved):

bmc-msft:
Performing integration tests

PR #300, opened 2020-11-12
prevent users from enabling public access to containers

anshuman-goel:

Though it needs some configuration and extra steps to access the data. Ideally it will be better to not allow public access at all.

Ref: https://docs.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-configure?tabs=portal#about-anonymous-public-read-access

## PR Checklist
* [ ] Tests added/passed

At configuration page of storage account
Before:
image

After:
image

tested on redeployment. Gave a test run with a fuzzer as well.

### DIFF
--- a/src/deployment/azuredeploy.json
+++ b/src/deployment/azuredeploy.json
@@ -526,7 +526,8 @@
             },
             "properties": {
                 "supportsHttpsTrafficOnly": true,
-                "accessTier": "Hot"
+                "accessTier": "Hot",
+                "allowBlobPublicAccess": false
             },
             "tags": {
                 "OWNER": "[parameters('owner')]"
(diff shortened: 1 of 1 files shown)


bmc-msft (approved):

bmc-msft:
To emphasize for those reading along: this is not about "default access policies for items in a storage account", but "the space of possible access policies".
 
In particular: OneFuzz already never creates publicly-accessible storage containers/queues/tables. it also does not, in any of its APIs, enable making such data public (accessible without auth).
 
By default, in Azure, all storage resources require authentication (are non-public). however, also by (storage account) default, a user can manually make specific containers public, as a container-level setting. this can be done via the Azure portal, az-cli, REST APIs, &c.

PR #303, opened 2020-11-12
Add User Info to created tasks

bmc-msft:
This PR makes user information from JWT tokens available as part of a Task.

Included changes:
* Renamed `verify_token` to `call_if_agent`, since this function is specific to agent token verification
* Renames `is_authorized` to `is_agent`, since this function checks if the token is an agent
* Adds support for unmanaged nodes in `is_agent` (see #133 for information) 
* Saves the user information from the JWT token on task create as part of `TaskConfig`

Note, `TaskConfig` is what is provided to notification templates.  This enables Github issues and ADO work items to tie back to the user that created the task.

Note, while `upn` _usually_ means email for AAD user tokens.  If we were going to make use of the email address, we should perform a graph lookup based on the `oid`, but we're not.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/agent_authorization.py
+++ b/src/api-service/__app__/onefuzzlib/agent_authorization.py
@@ -4,77 +4,50 @@
 # Licensed under the MIT License.
 
 import logging
-from typing import Callable, Union
+from typing import Callable
 from uuid import UUID
 
 import azure.functions as func
-import jwt
 from memoization import cached
 from onefuzztypes.enums import ErrorCode
-from onefuzztypes.models import Error
-from pydantic import BaseModel
+from onefuzztypes.models import Error, UserInfo
 
 from .azure.creds import get_scaleset_principal_id
-from .pools import Scaleset
+from .pools import Pool, Scaleset
 from .request import not_ok
-
-
-class TokenData(BaseModel):
-    application_id: UUID
-    object_id: UUID
-
-
-def try_get_token_auth_header(request: func.HttpRequest) -> Union[Error, TokenData]:
-    """ Obtains the Access Token from the Authorization Header """
-    auth: str = request.headers.get("Authorization", None)
-    if not auth:
-        return Error(
-            code=ErrorCode.INVALID_REQUEST, errors=["Authorization header is expected"]
-        )
-    parts = auth.split()
-
-    if parts[0].lower() != "bearer":
-        return Error(
-            code=ErrorCode.INVALID_REQUEST,
-            errors=["Authorization header must start with Bearer"],
-        )
-
-    elif len(parts) == 1:
-        return Error(code=ErrorCode.INVALID_REQUEST, errors=["Token not found"])
-
-    elif len(parts) > 2:
-        return Error(
-            code=ErrorCode.INVALID_REQUEST,
-            errors=["Authorization header must be Bearer token"],
-        )
-
-    # This token has already been verified by the azure authentication layer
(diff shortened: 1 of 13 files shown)


chkeita (approved):

src/api-service/__app__/onefuzzlib/agent_authorization.py
```diff
@@ -4,77 +4,50 @@
...
-            errors=["Authorization header must be Bearer token"],
-        )
+@cached(ttl=60)
+def is_agent(token_data: UserInfo) -> bool:
 
-    # This token has already been verified by the azure authentication layer
-    token = jwt.decode(parts[1], verify=False)
-    return TokenData(application_id=UUID(token["appid"]), object_id=UUID(token["oid"]))
+    if token_data.object_id:
+        # backward compatibility case for scalesets deployed before the migration
```
ranweiler, line 25: Can you add more detail to this comment, emphasizing when we can remove this?

We should probably have something greppable for this back-compat checks.

src/api-service/__app__/onefuzzlib/user_credentials.py
```diff
@@ -0,0 +1,45 @@
...
+
+def parse_jwt_token(request: func.HttpRequest) -> Result[UserInfo]:
+    """ Obtains the Access Token from the Authorization Header """
+
+    auth: str = request.headers.get("Authorization", None)
+    if not auth:
+        return Error(
+            code=ErrorCode.INVALID_REQUEST, errors=["Authorization header is expected"]
+        )
+    parts = auth.split()
```
ranweiler, line 22: If `auth` were somehow a nonempty but blank string, e.g. `"         "` or `"\t"`, then `"    ".split() == []`, which could cause `parts[0]` to throw an `IndexError`.

It's unlikely, and maybe already impossible, depending on how `request.headers` is constructed. That said, we can guarantee this will never happen if we bind `request.headers.get("Authorization", "").strip()` to `auth` on line 17.

src/api-service/__app__/onefuzzlib/user_credentials.py
```diff
@@ -0,0 +1,45 @@
...
+        return Error(
+            code=ErrorCode.INVALID_REQUEST,
+            errors=["Authorization header must be Bearer token"],
+        )
+
+    # This token has already been verified by the azure authentication layer
+    token = jwt.decode(parts[1], verify=False)
+
+    application_id = UUID(token["appid"])
+    object_id = UUID(token["oid"]) if "oid" in token else None
```
ranweiler, line 43: Can we use `token.get("oid")` here, too?

ranweiler: Oh, misread. We want to apply the `UUID` ctor to the result of a non-`None` indexing op.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/agent_authorization.py
bmc-msft: I don't actually know what's required here.  This was just lifted entirely from the agent authorization code.

While that's reasonable to do for the future, out of scope for this PR.

src/api-service/__app__/onefuzzlib/user_credentials.py
bmc-msft: That's a non-issue, as the auth headers are validated by Azure Functions.  That said, I simplified the checks which perform what you're after.

PR #271, opened 2020-11-05
Make the role assignment step optional in the deployment

chkeita:
## Summary of the Pull Request
Adding an option to make the role assignment optional. 
The deployment credentials do not have the permissions to assign roles.
Those permission require an admin approval.
 

## PR Checklist
* [x] Applies to work item: closes #268
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.
* [ ] Tests added/passed
* [ ] Requires documentation to be updated
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

_How does someone test & validate?_

PR #304, opened 2020-11-12
Add `--upgrade` to the Azure Devops deployment pipeline

anshuman-goel:
## PR Checklist
* [ ] Tests added/passed


bmc-msft (approved):

anshuman-goel:
@anslutsk this should help fix the break as per #271

PR #306, opened 2020-11-12
add symbols for windows agents

### LINKED ISSUE #260: include symbols for the windows agent & supervisor
When debugging the windows agent & supervisor, it would be extremely valuable to have the symbols for the rust code.

bmc-msft:
This PR adds PDBs for the supervisor & agent binaries.

Fixes #260

### DIFF
--- a/src/ci/agent.sh
+++ b/src/ci/agent.sh
@@ -27,6 +27,6 @@ cargo test --release --manifest-path ./onefuzz/Cargo.toml
 cp target/release/onefuzz-agent* ../../artifacts/agent
 cp target/release/onefuzz-supervisor* ../../artifacts/agent
 
-if [ -f target/release/onefuzz_agent.pdb ]; then
-    cp target/release/*.pdb ../../artifacts/agent
-fi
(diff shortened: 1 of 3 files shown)


src/ci/agent.sh
```diff
@@ -26,3 +26,7 @@ cargo test --release --manifest-path ./onefuzz/Cargo.toml
 
 cp target/release/onefuzz-agent* ../../artifacts/agent
 cp target/release/onefuzz-supervisor* ../../artifacts/agent
+
+if [ -f target/release/onefuzz_agent.pdb ]; then
```
ranweiler, line 30: Why are we using this as a condition, specifically?

ranweiler: (IOW, why not use `onefuzz_supervisor.pdb`, or unconditionally do `cp target/release/onefuzz*.pdb [...]` or similar?)

bmc-msft: Updated to use for loop with globbing

ranweiler (approved):

PR #266, opened 2020-11-05
Declarative templates

bmc-msft:
Provide the ability to maintain job templates, akin to `onefuzz template libfuzzer basic` at the service level. The templates include a job definition, an arbitrary set of tasks, and an arbitrary set of notification configs. The templates are managed at the service level, with job-submission time updates using a declarative syntax based on `jsonpatch`.

The SDK makes use of template configs, provided by the service, to dynamically build python methods for each template. This process enables the automatic argument parser generated by type signatures to automatically create the CLI subcommands for the template.

PR #313, opened 2020-11-16
Add SDK Feature Flags

bmc-msft:

This enables feature flags for the SDK, which enables gating access to preview features to those that have specifically asked for them.  This is intended to be used within #266.

Note, this change also moves to using a `pydantic` model for the config, rather than hand-crafted JSON dicts.


src/cli/onefuzz/backend.py
```diff
@@ -45,16 +45,24 @@ def _temporary_umask(new_umask: int) -> Generator[None, None, None]:
             os.umask(prev_umask)
 
 
+class BackendConfig(BaseModel):
+    authority: str
+    client_id: str
+    client_secret: Optional[str]
+    endpoint: Optional[str]
+    features: Dict[str, bool] = Field(default_factory=dict)
```
bmc-msft, line 53: Note, this uses str, rather than the enum used by the API.  Eventually, feature flags should be removed.  We don't want to fail in parsing if users have old feature flags added.  Rather, we should just ignore them.

chkeita, line 53: a set to represent the list of enabled feature might be easier to read when dumping the config with onefuzz config

chkeita (approved):

src/cli/onefuzz/backend.py
bmc-msft: Good idea.

I was thinking about this from underlying JSON, which doesn't directly support `set`.    However, pydantic maps `Set[str]` to `List[str]` with appropriate checking/deduplication.

PR #317, opened 2020-11-17
Move more run-time actions to setup-time

bmc-msft:
This script moves more of the run-time actions to setup-time.  This is important for running fuzzing within docker containers, such that installing llvm & gdb is done as part of the container, rather than on each launch.


src/runtime-tools/linux/run.sh
```diff
@@ -7,53 +7,35 @@ set -ex
...
-        fi
-        export RUST_BACKTRACE=full
         onefuzz-supervisor run --config /onefuzz/config.json
     ;;
     "repro")
         logger "onefuzz: starting repro"
-        if type apt > /dev/null 2> /dev/null; then
-            sudo apt install -y gdb gdbserver
-        fi
+        echo repro
```
ranweiler, line 38: I forget, are these `echo` calls load-bearing?

src/runtime-tools/linux/setup.sh
```diff
@@ -84,6 +85,17 @@ fi
 
 chmod -R a+rx /onefuzz/tools/linux
 
+if type apt > /dev/null 2> /dev/null; then
+    sudo apt update
+    sudo apt install -y gdb gdbserver
```
ranweiler, line 90: Does this mean we're now installing `gdbserver` on all `apt` VMs?

ranweiler (approved):

src/runtime-tools/linux/setup.sh
bmc-msft: Yes.  We'll need to do this anyways once we move to performing live-repro in managed scalesets.

src/runtime-tools/linux/run.sh
bmc-msft: Nope, purely informative.

bmc-msft:
Integration tests passed

PR #318, opened 2020-11-17
Enable unmanaged registrations and configuration by environment variables

bmc-msft:
This feature is cherry-picked #291, split off into more consumable components.

The intent is to enable bring-your-own-compute for fuzzing via docker containers.


src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -123,13 +129,20 @@ async fn run_agent(config: StaticConfig) -> Result<()> {
     telemetry::set_property(EventData::InstanceId(config.instance_id));
     telemetry::set_property(EventData::MachineId(get_machine_id().await?));
     telemetry::set_property(EventData::Version(env!("ONEFUZZ_VERSION").to_string()));
-    if let Ok(scaleset) = get_scaleset_name().await {
-        telemetry::set_property(EventData::ScalesetId(scaleset));
+    let scaleset_result = get_scaleset_name().await;
```
ranweiler, line 132: Instead of using the `Result` variant to signal info about our environment, maybe `get_scaleset_name()` should return a `Result<Option<String>>`.

Even better, you could introduce a new function that returns a enum like:
```rust
enum NodeType {
    Managed { scaleset_name: String },
    Unmanaged,
}

impl NodeType {
    pub fn get() -> Result<Self> { ... }
}
```

ranweiler (approved):

src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -59,25 +60,29 @@ fn main() -> Result<()> {
...
         Opt::Debug(opt) => debug::debug(opt)?,
         Opt::Licenses => licenses()?,
+        Opt::Version => version()?,
     };
 
     Ok(())
 }
 
+fn version() -> Result<()> {
+    println!(
```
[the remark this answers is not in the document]
bmc-msft: The version is part of the telemetry properties.  See line 131.

src/agent/onefuzz-supervisor/src/main.rs
bmc-msft: Added the former.

PR #320, opened 2020-11-17
Use Storage Account types, rather than account_id

bmc-msft:
We need to move to supporting data sharding.

One of the steps towards that is stop passing around `account_id`, rather we need to specify the type of storage we need.


src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -6,76 +6,108 @@
...
+
+
+class StorageType(Enum):
+    corpus = "corpus"
+    config = "config"
+
+
+def get_account_id_by_type(storage_type: StorageType) -> str:
+    if storage_type == StorageType.corpus:
+        account_id = get_fuzz_storage()
```
ranweiler, line 26: Is `corpus` -> `fuzz..` and `config` -> `func..` a relationship we expect to always hold? If so, we should consider explicitly documenting it.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: As I continue to address the issues relating to storage account sharding, I expect this to change, which is why I have not described this in developer-focused documentation yet.

chkeita (approved):

Issue #215, opened 2020-10-27
Nodes not appearing in the pool to take on jobs even though they are running


maxnoka:
## Information

- Onefuzz version: 1.3.1
- OS: Windows client, Windows VMs

## Provide detailed reproduction steps (if any)

Follow the getting started guide, with these changes:
- I'm using a trial subscription
- Use Windows as pool OS
- I create a scaleset with size 1
- I use the `simple-libfuzzer` example

### Expected result

When invoking `onefuzz status top` I should see the singular node and the pool and it should take on the libfuzzer tasks.

### Actual result

There are no nodes in the pool, and jobs remain `scheduled` and `waiting`, even after hours of waiting (see below for `onefuzz status top`).

![nodes_not_there_to_take_jobs](https://user-images.githubusercontent.com/16231006/97298654-a8e31400-184b-11eb-8dc8-00f279cf0cf8.jpg)

`onefuzz nodes list` also returns empty `[]`. 

I can see the node as "running" both in the azure portal and also when I check `onefuzz scalesets get <scaleset id here>` (see attached txt for the output to a bunch of list and get commands of the pool and the scaleset).

[onefuzz_pools_scalesets_list_and_get.txt](https://github.com/microsoft/onefuzz/files/5445090/onefuzz_pools_scalesets_list_and_get.txt)

bmc-msft:
Can you check the application insights log for the `machine_id` of your node?  It would be good to see if there was something preventing the agent from starting.

From the Azure Portal, go to your Application Insights instance for `OnefuzzTestInstance`, click on "Transaction search" under "Investigate", from there search for the `machine_id` reported in the response from `onefuzz scalesets get`.  You may need to increase your time range for your query.

kevinmoro:
I have the same issue as @maxnoka described above. When I checked the application insights log for the machine_id i found this reoccuring message: 
``Registration failed: HTTP status client error (401 Unauthorized) for url (https://<websitename>.azurewebsites.net/api/agents/registration?machine_id=<machine_id>&pool_name=pool_windows&version=1.5.0&scaleset_id=<scaleset_id>)
 retrying in 60 seconds``
And sometimes this message is occuring, but less often:
``error running supervisor agent: Unable to register agent``

bmc-msft:
@chkeita, can you take a look at this please?

chkeita:
@maxnoka can you please confirm that you still have the issue an seeing the messages in appinsight as @kevinmoro 

[...]

PR #321, opened 2020-11-17
Include the body when receiving a registration error

chkeita:
* Include the body when receiving a registration error to help debug issued like #215 
* increase the agent registration timeout to 20 min


bmc-msft (approved):

bmc-msft:
Performing integration tests

PR #322, opened 2020-11-17
Address typing issues hidden by memoization.caching

bmc-msft:
The caching library used by OneFuzz does not preserve type signatures.  

This PR includes the following changes:
* Scripts to enable & disable the use of the `memoization` library
* Uses the disable script to enable type checking without the `cached` function wrappers.
* Fixes two issues exposed once the type signatures were exposed.

See https://github.com/lonelyenvoy/python-memoization/issues/16 for more information.

### DIFF
--- a/src/ci/disable-py-cache.sh
+++ b/src/ci/disable-py-cache.sh
@@ -3,6 +3,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+
+# Temporary work-around to improve the efficacy of static analysis of functions
+# decorated with memoization.cached.
+#
+# For more information:
+# https://github.com/lonelyenvoy/python-memoization/issues/16
+
 set -ex
 
 SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
--- a/src/ci/disable-py-cache.sh
+++ b/src/ci/disable-py-cache.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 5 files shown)


src/ci/disable-py-cache.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+set -ex
+
+SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
+
+sed -i "s/^from memoization import cached/##### from memoization import cached/" $(find . -name '*.py' -not -path .python_packages)
```
ranweiler, line 10: (1) whatever we do here really would really benefit from docs that make it very clear what is going on
(2) what if, instead of rewriting every file that used memoization, we introduce a memoization module in `onefuzzlib` that could be programmatically disabled, or at worst, just rewritten in one place? IOW, in one mode, it would export a typed identity decorator, and in the other, the actually `memoization.cached` decorator?

ranweiler:
Nice catch, btw.

src/ci/disable-py-cache.sh
bmc-msft: I am working on upstreaming a patch to python-memoization that will keep the type signatures for decorated functions, which will remove the need for this work-around.

I will add documentation to these scripts.  These are intended as a stop-gap only.

ranweiler: Cool, this is okay a stopgap, and the comments will help.

ranweiler (approved):

PR #325, opened 2020-11-18
Expose coverage/exec_sec for libfuzzer targets via CLI

bmc-msft:
Adds debug subcommands to the SDK/CLI that simplify querying Application Insights for libfuzzer telemetry.  

Querying for the latest execs_sec for a job, by job_id fragment.
```
$ onefuzz debug job libfuzzer_execs_sec 88 --limit 1
[
    {
        "execs_sec": "191035",
        "machine_id": "b2dbe720-4fd8-4342-957a-6cb0979d2187",
        "timestamp": "2020-11-18T00:08:53.98Z",
        "worker_id": "0"
    }
]
```

Querying for the latest coverage for a job, by job_id fragment.
```
$ onefuzz debug job libfuzzer_coverage 88 --limit 1
[
    {
        "covered": "10",
        "features": "21",
        "rate": "0.47619047619047616",
        "timestamp": "2020-11-18T00:09:40.793Z"
    }
]
```

### DIFF
--- a/src/cli/onefuzz/debug.py
+++ b/src/cli/onefuzz/debug.py
@@ -28,6 +28,7 @@ from .ssh import ssh_connect
 
 EMPTY_SHA256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 ZERO_SHA256 = "0" * len(EMPTY_SHA256)
+DAY_TIMESPAN = "PT24H"
 
 
 class DebugRepro(Command):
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/debug.py
```diff
@@ -28,6 +28,7 @@
 
 EMPTY_SHA256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 ZERO_SHA256 = "0" * len(EMPTY_SHA256)
+DAY_TIMESPAN = "PT24H"
```
ranweiler, line 31: Are there docs we can link to for this?

Checking my understanding: this ends up in the generated KQL, so in general needs to be an ISO 8601 time interval. The default value is only really specifies a duration on its own, but it gets interpreted as a time interval in context as "the last <duration> since now()".

src/cli/onefuzz/debug.py
```diff
@@ -337,11 +383,17 @@ def query(
             return raw_data
         return self._convert(raw_data)
 
+    def _query_parts(
+        self, parts: List[str], timespan: str, *, raw: bool = False
+    ) -> Any:
```
ranweiler, line 388: Can we return `Union` instead of `Any`?

ranweiler (approved):

src/cli/onefuzz/debug.py
bmc-msft: Application Insights doesn't provide us any typing hints, so we'd have to cast it.  Given that the response is arbitrarily nested json, it wouldn't be very meaningful.

For the libfuzzer specific queries where we are projecting specific fields, I can see that being valuable.  I will investigate that as future work.

src/cli/onefuzz/debug.py
bmc-msft: It's included in the help for the debug.logs.query.  I'll add it for these new commands.

PR #326, opened 2020-11-18
Storing secrets in azure keyvault

chkeita:


src/api-service/__app__/onefuzzlib/notifications/main.py
```diff
@@ -88,6 +91,37 @@ def create(
         )
         return entry
 
+    def save(self, new: bool = False, require_etag: bool = False) -> Optional[Error]:
```
chkeita, line 94: Validation in progress

bmc-msft:
Should we hold off on reviewing this while you're working on #349?

chkeita:
> Should we hold off on reviewing this while you're working on #349?

yes i will convert this back into a draft to avoid any confusions

chkeita:
This is what the creation of template looks like.
```
C:\temp\Onefuzz_10> onefuzz notifications create_teams oft-unique-reports-6596118fa36b5fc9ad21a5507b3daaa0 "http://test2" -vv
DEBUG:onefuzz:create teams notification integration: oft-unique-reports-6596118fa36b5fc9ad21a5507b3daaa0
DEBUG:nsv-backend:request POST https://chkeitaonefuzz.azurewebsites.net/api/notifications {'config': {'url': {'secret': 'http://test2'}}, 'container': 'oft-unique-reports-6596118fa36b5fc9ad21a5507b3daaa0'}
{
    "config": {
        "url": {
            "secret": {
                "url": "https://chkeitaonefuzz-vault.vault.azure.net/secrets/1d93f5a3-0a45-4b78-a4d2-ce31eb082711/02b6d296f84f46b1b77727eee16fc1e3"
            }
        }
    },
    "container": "oft-unique-reports-6596118fa36b5fc9ad21a5507b3daaa0",
    "notification_id": "b52b24d1-eec6-46c9-b06a-818a997da43c"
}
```

bmc-msft:
Very cool.  I am going to look into how this interplays with the job templates component.

bmc-msft:
This doesn't work with the nested data types used by job_templates.  

I followed the example [using the CLI to create a template that created Teams messages]( https://github.com/microsoft/onefuzz/blob/main/docs/managing-templates.md#using-the-cli--modifying-json),
then fetched the template gain using

`onefuzz job_templates manage get libfuzzer_linux > libfuzzer_linux.json`

What I get is: 
```json
    "notifications": [
        {
            "container_type": "unique_reports",
            "notification": {
                "config": {
                    "url": {
                        "secret": "https://contoso.com/webhook-url-here"
                    }
                }
            }
        }
    ]
```

What I expected is:
```
    "notifications": [
        {
            "container_type": "unique_reports",
            "notification": {
                "config": {
                    "url": {
                        "secret": {
                            "url": "https://some-keyvault-url-here",
                        }
                    }
                }
            }
        }
    ]
```

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -281,18 +286,44 @@ def get_keys(self) -> Tuple[KEY, KEY]:
...
+    @classmethod
+    def hide_secrets(
+        cls,
+        model: BaseModel,
+        hider: Callable[["SecretData"], None],
+        visited: Set[int] = set(),
+    ) -> None:
+        model_type = type(model)
+        if Type_has_secrets.get(model_type, True):
+            for field in model.__fields__:
```
bmc-msft, line 298: If a field is List[Field], which is common, this doesn't iterate over those nested fields.

bmc-msft (changes requested):

src/api-service/__app__/onefuzzlib/orm.py
bmc-msft: Probably needs to support isinstance of `list` and the values of `dict`

bmc-msft (dismissed):

bmc-msft:
The changes now work with the new job templates as implemented.

I've tagged this as impacts-backcompat, as it will require new permissions to any service-principal being used during deployment.   As such, we will want to roll this into a major release.  This will probably be the impetus for the 2.0.0 release.

bmc-msft:
This shouldn't be merged until #389 is merged.

bmc-msft:
In testing this with Teams integration, I get the following exception:
```
Exception while executing function: Functions.queue_file_changes Result: Failure
Exception: InvalidSchema: No connection adapters were found for "{'url': 'https://KEYVAULT_INSTANCE.vault.azure.net/secrets/GUID_HERE/GUID_HERE'}"
Stack:   File "/azure-functions-host/workers/python/3.7/LINUX/X64/azure_functions_worker/dispatcher.py", line 357, in _handle__invocation_request
    self.__run_sync_func, invocation_id, fi.func, args)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/azure-functions-host/workers/python/3.7/LINUX/X64/azure_functions_worker/dispatcher.py", line 542, in __run_sync_func
    return func(**params)
  File "/home/site/wwwroot/queue_file_changes/__init__.py", line 34, in main
    file_added(event)
  File "/home/site/wwwroot/queue_file_changes/__init__.py", line 22, in file_added
    new_files(container, path)
  File "/home/site/wwwroot/onefuzzlib/notifications/main.py", line 138, in new_files
    notify_teams(notification.config, container, filename, report)
  File "/home/site/wwwroot/onefuzzlib/notifications/teams.py", line 129, in notify_teams
    send_teams_webhook(config, title, facts, text)
  File "/home/site/wwwroot/onefuzzlib/notifications/teams.py", line 50, in send_teams_webhook
    response = requests.post(config_url, json=message)
  File "/home/site/wwwroot/.python_packages/lib/site-packages/requests/api.py", line 119, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/home/site/wwwroot/.python_packages/lib/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/site/wwwroot/.python_packages/lib/site-packages/requests/sessions.py", line 530, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/site/wwwroot/.python_packages/lib/site-packages/requests/sessions.py", line 637, in send
    adapter = self.get_adapter(url=request.url)
  File "/home/site/wwwroot/.python_packages/lib/site-packages/requests/sessions.py", line 730, in get_adapter
    raise InvalidSchema("No connection adapters were found for {!r}".format(url))
```

src/deployment/azuredeploy.json
```diff
@@ -101,6 +102,35 @@
             "apiVersion": "2018-11-30",
             "location": "[resourceGroup().location]"
         },
+        {
+            "type": "Microsoft.KeyVault/vaults",
+            "apiVersion": "2016-10-01",
+            "name": "[variables('keyVaultName')]",
+            "location": "[resourceGroup().location]",
+            "properties": {
+                "enabledForDeployment": true,
```
ranweiler, line 111: Where do we use this enablement?

src/deployment/azuredeploy.json
```diff
@@ -101,6 +102,35 @@
             "apiVersion": "2018-11-30",
             "location": "[resourceGroup().location]"
         },
+        {
+            "type": "Microsoft.KeyVault/vaults",
+            "apiVersion": "2016-10-01",
```
ranweiler, line 107: This seems to be a few versions behind the latest. Is there a specific reason we're using this version?

src/deployment/azuredeploy.json
```diff
@@ -101,6 +102,35 @@
...
+                        "permissions": {
+                            "secrets": ["get", "list", "set", "delete"]                        
+                        }
+                    }
+                ],
+                "sku": {
+                    "name": "standard",
+                    "family": "A"
+                },
+                "networkAcls": {
```
ranweiler, line 128: Does this property actually exist in the ARM resource API version we are using? 

Compare: https://docs.microsoft.com/en-us/azure/templates/microsoft.keyvault/2016-10-01/vaults#vaultproperties-object

Vs: https://docs.microsoft.com/en-us/azure/templates/microsoft.keyvault/2018-02-14/vaults#vaultproperties-object

src/api-service/__app__/notifications/__init__.py
```diff
@@ -15,9 +15,6 @@
 
 def get(req: func.HttpRequest) -> func.HttpResponse:
     entries = Notification.search()
-    for entry in entries:
-        entry.config.redact()
```
ranweiler, line 19: We no longer need to do this, because it is handled by the new `SecretData` class.

ranweiler: (Er, this is a comment to myself as I review, not an assertion that you need to do change anything)

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -212,6 +214,9 @@ def raw(
         return result
 
 
+Type_has_secrets: Dict[type, bool] = dict()
```
ranweiler, line 217: I'm not sure what's going on here.

It looks like we have a module-scoped mutable variable. It starts out empty, and when we call `hide_secrets()` on some particular class, the global gets updated.

This seems very tricky. I'm not really sure how I'm supposed to tell if it is correct. Could you talk a little about what's going on here?

ranweiler: Less, important, but so I don't forget: stylistically, assuming this is intended to be a private "constant", we'd probably want to name this like:
```
_TYPE_HAS_SECRETS: Dict[type, bool] = dict()
```
Leading `_` to indicate "intended private", all caps to indicate "constant-ish".

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+
+from azure.keyvault.secrets import KeyVaultSecret
+from onefuzztypes.models import SecretAddress, SecretData
+from pydantic import BaseModel
+
+from .azure.creds import get_instance_name, get_keyvault_client
+
+
+def save_to_keyvault(secret_data: SecretData) -> None:
+    if isinstance(secret_data.secret, SecretAddress):
```
ranweiler, line 19: This is very surprising to me. Why don't we save instances of `SecretAddress`?

ranweiler: Is this because it's supposed to be idempotent?

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+    secret_name = str(uuid4())
+    if isinstance(secret_data.secret, str):
+        secret_value = secret_data.secret
+    elif isinstance(secret_data.secret, BaseModel):
+        secret_value = secret_data.secret.json()
+    else:
+        raise Exception("invalid secret data")
+
+    kv = store_in_keyvault(get_keyvault_address(), secret_name, secret_value)
+    secret_data.secret = SecretAddress(url=kv.id)
```
ranweiler, line 31: So, when we save `SecretData` to the keyvault, we actually overwrite the secret data with the newly-created address of the secret, throwing away (this copy of) the raw secret.

This helps explain some of what's going on with `SecretData`'s definition.

What if we only allowed the existence of `SecretData` instances that were already saved to the keyvault? Then it would have fewer valid states and be easier to think about. If needed, we could have a `RedactedData` class for something not yet saved, and then this function could return a `SecretData`/`KeyvaultData` after saving. The runtime difference is then encoded in a type distinction.

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+def get_secret_string_value(self: SecretData[str]) -> str:
+    if isinstance(self.secret, SecretAddress):
+        secret = get_secret(self.secret.url).value
+        return cast(str, secret.value)
+    else:
+        return self.secret
+
+
+def get_keyvault_address() -> str:
+    return f"https://{get_instance_name()}-vault.vault.azure.net"
```
ranweiler, line 43: Not a big deal, but could we link to docs for this URL format?

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+    elif isinstance(secret_data.secret, BaseModel):
+        secret_value = secret_data.secret.json()
+    else:
+        raise Exception("invalid secret data")
+
+    kv = store_in_keyvault(get_keyvault_address(), secret_name, secret_value)
+    secret_data.secret = SecretAddress(url=kv.id)
+
+
+def get_secret_string_value(self: SecretData[str]) -> str:
```
ranweiler, line 34: This really seems like it should be a getter on the `SecretData` class. Is it not because you want it only to be defined for the concrete type `SecretData[str]`, vs. the generic `SecretData[T]`?

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+) -> KeyVaultSecret:
+    keyvault_client = get_keyvault_client(keyvault_url)
+    kvs: KeyVaultSecret = keyvault_client.set_secret(secret_name, secret_value)
+    return kvs
+
+
+def parse_secret_url(secret_url: str) -> Tuple[str, str]:
+    u = urlparse(secret_url)
+    vault_url = f"{u.scheme}://{u.netloc}"
+    secret_name = u.path.split("/")[2]
```
ranweiler, line 57: Could you add a comment about what we expect this to look like before we split it?

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -0,0 +1,70 @@
...
+def get_secret(secret_url: str) -> KeyVaultSecret:
+    (vault_url, secret_name) = parse_secret_url(secret_url)
+    keyvault_client = get_keyvault_client(vault_url)
+    return keyvault_client.get_secret(secret_name)
+
+
+def delete_secret(secret_url: str) -> None:
+    (vault_url, secret_name) = parse_secret_url(secret_url)
+    keyvault_client = get_keyvault_client(vault_url)
+    keyvault_client.begin_delete_secret(secret_name)
```
ranweiler, line 70: It looks like returning from this function does not imply that the secret is deleted, since `begin_delete_secret` returns a poller, which we throw away. Should we call `wait()` on the poller, blocking until we are sure the secret is deleted?

src/api-service/__app__/onefuzzlib/secrets.py
chkeita: if the SecreSata instance already has a secretAddress it means that it was already saved to the keyvault. so no need to save it again

src/api-service/__app__/onefuzzlib/secrets.py
chkeita: agree but I did not want to add a dependency on the keyvault library to the onefuzztypes.
I tried extending the type here but i struggled to find the right syntax (if you have a suggestion let me know)

ranweiler: Does the `SecretData` class even make sense without keyvault methods enabled? It seems like the semantics basically require it. Could we just have it NOT live in the `pytypes` package?

src/api-service/__app__/onefuzzlib/secrets.py
chkeita: `SecretData` needs to be backward compatible for the existing notification code. So we need to allow secret not yet saved in the keyvault. new classes can use `SecretAddress` as a backing field and never allow secret not saved to keyvault as you suggested

src/pytypes/onefuzztypes/models.py
```diff
@@ -41,6 +42,42 @@ class UserInfo(BaseModel):
...
+class SecretAddress(BaseModel):
+    url: str
+
+
+T = TypeVar("T")
+
+
+@dataclass
+class SecretData(Generic[T]):
+    secret: Union[T, SecretAddress]
```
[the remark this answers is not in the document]
chkeita: Not sure what is the goal of the suggestion here.
`SecretData` allows us to store some data that will eventually be saved in keyvault. The union here encoded the state of the instance. the `T` case is for data not yet saved to keyvault, `SecretData` is for data saved in keyvault. This allows me to know how to handle the data in other places.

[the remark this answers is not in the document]
ranweiler: Yep, I didn't understand the API at first, because I started by reading the class definition.

src/api-service/__app__/onefuzzlib/secrets.py
chkeita: Secret data is used to post data to the service and show the information to the user inthe console. The keyvault api is only needed by the backend to save and retrieve the secret
We still need to reference SecretData in the notification template classes defined in `pytypes`.

src/pytypes/onefuzztypes/models.py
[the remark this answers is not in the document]
ranweiler: (which suggests to me that it would benefit from some doc comments)

src/api-service/__app__/onefuzzlib/secrets.py
chkeita: added expected format as comment

src/pytypes/onefuzztypes/models.py
```diff
@@ -41,6 +42,42 @@ class UserInfo(BaseModel):
...
+        else:
+            return "[REDACTED]"
+
+    def __repr__(self) -> str:
+        if isinstance(self.secret, SecretAddress):
+            return str(self.secret)
+        else:
+            return "[REDACTED]"
+
+    def try_get_value(self) -> Optional[T]:
```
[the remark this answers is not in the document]
chkeita: removed this was unused

src/pytypes/onefuzztypes/models.py
```diff
@@ -223,23 +260,39 @@ class ADODuplicateTemplate(BaseModel):
...
     unique_fields: List[str]
     comment: Optional[str]
     ado_fields: Dict[str, str]
     on_duplicate: ADODuplicateTemplate
 
-    def redact(self) -> None:
-        self.auth_token = "***"
+    @validator("auth_token", pre=True, always=True)
+    def validate_auth_token(cls, v: Any) -> SecretData:
+        if isinstance(v, str):
```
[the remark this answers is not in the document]
chkeita: `isinstance(v, str)` will return false for a SecretData[str]

The base class syntax here is just how you define a [generic type parameter in python](https://docs.python.org/3/library/typing.html#user-defined-generic-types). Like any other python type this information does not affect the runtime.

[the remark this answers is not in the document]
ranweiler: Ah, great!

src/deployment/azuredeploy.json
chkeita: removed

src/deployment/azuredeploy.json
chkeita: updated to use the latest

src/deployment/azuredeploy.json
chkeita: updated the version to the latest which has this property

src/api-service/__app__/onefuzzlib/orm.py
chkeita: commented and renamed

src/pytypes/onefuzztypes/models.py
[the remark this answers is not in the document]
chkeita: added comments

src/api-service/__app__/onefuzzlib/secrets.py
ranweiler: Got it. This is easy to understand, so let's keep it, and we can maybe revisit later. We can imagine defining a subclass of `SecretData` that is local to `onefuzzlib`, and which extends the base class to add these "live" methods that interact with keyvault, but I see that as more of a syntactic difference.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -281,18 +291,56 @@ def get_keys(self) -> Tuple[KEY, KEY]:
...
+        hider: Callable[["SecretData"], None],
+        visited: Set[int] = set(),
+    ) -> None:
+        model_type = type(model)
+        if id(model) in visited:
+            return
+
+        visited.add(id(model))
+
+        if _TYPE_HAS_SECRETS.get(model_type, True):
```
bmc-msft, line 307: I'm wondering what happens in this case:

```python
class B(BaseModel):
    c: SecretData[str]

class A(BaseModel):
    b: Optional[List[B]]
```

Consider hiding the secrets for these iteratively:
```python
A()
A(b=[B(c="hi")])
```

I think that `getattr` will return None on the first instance, which falls through the later results.  This will result in _TYPE_HAS_SECRETS falling through as False, even though the type has nested secrets.

bmc-msft:
Testing with Teams integration results in the following error:

```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.queue_file_changes
 ---> Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException: Result: Failure
Exception: AttributeError: 'str' object has no attribute 'value'
Stack:   File "/azure-functions-host/workers/python/3.7/LINUX/X64/azure_functions_worker/dispatcher.py", line 357, in _handle__invocation_request
    self.__run_sync_func, invocation_id, fi.func, args)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/azure-functions-host/workers/python/3.7/LINUX/X64/azure_functions_worker/dispatcher.py", line 542, in __run_sync_func
    return func(**params)
  File "/home/site/wwwroot/queue_file_changes/__init__.py", line 34, in main
    file_added(event)
  File "/home/site/wwwroot/queue_file_changes/__init__.py", line 22, in file_added
    new_files(container, path)
  File "/home/site/wwwroot/onefuzzlib/notifications/main.py", line 138, in new_files
    notify_teams(notification.config, container, filename, report)
  File "/home/site/wwwroot/onefuzzlib/notifications/teams.py", line 130, in notify_teams
    send_teams_webhook(config, title, facts, text)
  File "/home/site/wwwroot/onefuzzlib/notifications/teams.py", line 50, in send_teams_webhook
    config_url = get_secret_string_value(config.url)
  File "/home/site/wwwroot/onefuzzlib/secrets.py", line 39, in get_secret_string_value
    return cast(str, secret.value)
```

My notification config is:

```
[
    {
        "config": {
            "url": {
                "secret": {
                    "url": "https://INSTANCE_NAME_HERE.vault.azure.net/secrets/3b26a03b-cb51-4722-9ad4-fa94f9ee8c7c/1833f0a96d504574b8b614a69f6c57c1"
                }
            }
        },
        "container": "oft-unique-reports-fc60e3f7cb2f58a692e7ee189f6d113c",
        "notification_id": "479ed0b4-473c-4501-aba1-99f21372ff02"
    }
]
```

PR #330, opened 2020-11-18
Support retry during function deploy

bmc-msft:
Starting earlier today, I saw roughly 1 in 3 deployments fail with the error `Azure.Functions.Cli.Common.CliException: Timed out waiting for SCM to update the Environment Settings`.  Redeploying the application resolves the issue.  New builds and past releases alike hit this exception.

According to https://github.com/Azure/azure-functions-core-tools/issues/1863, function app deployments may fail due to timeouts related to cold-start.

This PR executes the deploy in a loop with a delay in the case of failure.

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -680,6 +680,7 @@ class Client:
                     except subprocess.CalledProcessError as err:
                         error = err
                         if i + 1 < max_tries:
+                            logger.debug("func failure error: %s", err)
                             logger.warning(
                                 "function failed to deploy, waiting 60 seconds and trying again"
                             )
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -16,6 +16,7 @@ import time
 import uuid
 import zipfile
 from datetime import datetime, timedelta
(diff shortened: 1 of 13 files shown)


bmc-msft:
Integration tests passed

src/deployment/deploy.py
```diff
@@ -658,19 +659,33 @@ def deploy_app(self):
...
+                                self.application_name,
+                                "--python",
+                                "--no-build",
+                            ],
+                            env=dict(os.environ, CLI_DEBUG="1"),
+                            cwd=tmpdirname,
+                        )
+                        return
+                    except subprocess.CalledProcessError as err:
+                        error = err
```
chkeita, line 681: Can we log the error here. So we have a record of all the errors in case it is failing for different reasons

bmc-msft: I will update this to add it to debug logging, but I'd prefer not to include it by default.  The logs are extremely verbose.  I'd prefer that the output be as simple as possible during expected operations.

chkeita (approved):

PR #332, opened 2020-11-19
Add linting to deployment tools

bmc-msft:
Adds black, isort, and mypy to the deployment tools.


bmc-msft:
A handful of the type signatures are underspecified.  The under-specifications are almost always where we're getting back an arbitrary nested json from Azure and should verify the soundness before use, but that is future work beyond this PR.

ranweiler (approved):

PR #334, opened 2020-11-20
support multiple corpus accounts

bmc-msft:
Add support for sharding across multiple storage accounts for blob containers used for corpus management.

Things to note:

1. Additional storage accounts must be in the same resource group, support the "blob" endpoint, and have the tag `storage_type` with the value `corpus`.  A utility is provided (`src/utils/add-corpus-storage-accounts`), which adds storage accounts. 
2. If any secondary storage accounts exist, they are used by default for containers.
3. Storage account names are cached in memory the Azure Function instance forever.   Upon adding new storage accounts, the app needs to be restarted to pick up the new accounts.


bmc-msft:
This PR needs to add `proxy-configs` to the base setup.

src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -4,81 +4,128 @@
...
-    else:
-        raise NotImplementedError
-    return account_id
-
-
-@cached(ttl=5)
-def get_blob_service_by_type(storage_type: StorageType) -> Any:
-    account_id = get_account_id_by_type(storage_type)
-    return get_blob_service(account_id)
+def get_service_by_container(
```
ranweiler, line 33: Would it be useful to cache this, too?

src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -4,81 +4,128 @@
...
+def get_account_by_container(
+    container: Container, storage_type: StorageType
 ) -> Optional[str]:
-    try:
-        get_blob_service_by_type(storage_type).create_container(name, metadata=metadata)
-    except AzureHttpError:
-        # azure storage already logs errors
+    accounts = get_accounts(storage_type)
+
+    # check secondary accounts first by searching in reverse
```
ranweiler, line 65: What are the sorting assumptions that imply a reverse search checks secondary accounts first? Can we link to any justification, or maybe elaborate?

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
```diff
@@ -0,0 +1,101 @@
...
+        filter=base.filter,
+        retry_policy=base.retry_policy,
+    )
+
+    topic_id = uuid.uuid5(STORAGE_GUID_NAMESPACE, src_account_id).hex
+
+    result = client.event_subscriptions.create_or_update(
+        src_account_id, "corpus" + topic_id, event_subscription_info
+    ).result()
+    if result.provisioning_state != "Succeeded":
```
chkeita, line 54: They should a retry loop when the state is [Creating](https://docs.microsoft.com/en-us/python/api/azure-mgmt-eventgrid/azure.mgmt.eventgrid.models.eventsubscriptionprovisioningstate?view=azure-python)

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
```diff
@@ -0,0 +1,101 @@
...
+        access_tier=AccessTier.hot,
+        allow_blob_public_access=False,
+        minimum_tls_version="TLS1_2",
+    )
+
+    client = get_client_from_cli_profile(StorageManagementClient)
+    account = client.storage_accounts.create(
+        resource_group, account_name, params
+    ).result()
+    if account.provisioning_state != "Succeeded":
```
chkeita, line 76: we should handle the creating state here too

src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -4,81 +4,128 @@
...
-    return get_container_sas_url(
-        name,
-        storage_type,
+    return container_metadata(container, account)
+
+
+def create_container(
+    container: Container,
+    storage_type: StorageType,
+    metadata: Optional[Dict[str, str]],
```
ranweiler, line 105: It looks like `metadata` is used only if the container does not already exist. If the caller provides `metadata` that does not match the existing container, the difference can't be directly observed. Should we indicate "container already exists, and was not updated" in some way, to help avoid a subtle future misuse?

src/api-service/__app__/onefuzzlib/azure/storage.py
```diff
@@ -0,0 +1,101 @@
...
+
+def choose_account(storage_type: StorageType) -> str:
+    accounts = get_accounts(storage_type)
+    if not accounts:
+        raise Exception(f"no storage accounts for {storage_type}")
+
+    if len(accounts) == 1:
+        return accounts[0]
+
+    # Use a random secondary storage account if any are available
```
ranweiler, line 70: This means we _never_ choose a primary account if secondaries exist, right? Why is that desirable?

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
```diff
@@ -0,0 +1,101 @@
+import argparse
```
ranweiler, line 1: Needs license header; double-check other new files.

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
```diff
@@ -0,0 +1,101 @@
...
+from azure.mgmt.storage import StorageManagementClient
+from azure.mgmt.storage.models import (
+    AccessTier,
+    Kind,
+    Sku,
+    SkuName,
+    StorageAccountCreateParameters,
+)
+
+# This was generated randomly and should be preserved moving forwards
```
ranweiler, line 17: 👍

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_base_event(
+    client: EventGridManagementClient, resource_group: str, location: str
+) -> EventSubscription:
+    for entry in client.event_subscriptions.list_regional_by_resource_group(
+        resource_group, location
+    ):
+        if (
+            entry.name == "onefuzz1"
```
ranweiler, line 28: Where does this (`"onefuzz1"`) come from?

ranweiler (approved):

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
bmc-msft: `.result()` polls until the results are either success or failure.  We don't need to check for creating here.

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
bmc-msft: `.result()` polls until the results are either success or failure. We don't need to check for creating here.

src/api-service/__app__/onefuzzlib/azure/storage.py
bmc-msft: The primary account account provides both containers as well as queues.  Secondary accounts only support containers, but at significantly higher performance rates.

Offloading container use to only secondary accounts reduces  IOP contention for the queues.

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: It would add difficulties relating to having to expire the cache when containers are deleted.  Caching at the account level, by `get_blob_service`, provides the majority of the benefit caching here would provide.

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
chkeita: 👍

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/storage.py
ranweiler: Ah, that's a very good reason! Could we document it here (or elsewhere, if more appropriate).

src/utils/add-corpus-storage-accounts/add-corpus-storage-account.py
ranweiler: (Is this the expected name for the primary account? If so, can we document that, and/or express it via a named constant?)

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: Added explanation.

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: Perhaps, but the that's existing behavior for creating containers and out of scope for this effort.

.github/workflows/ci.yml
```diff
@@ -214,7 +214,7 @@ jobs:
         vulture --min-confidence 61 __app__ 
 
         ../ci/disable-py-cache.sh
-        mypy __app__ ./tests
+        mypy __app__ ./tests --no-warn-unused-ignores --no-warn-redundant-casts
```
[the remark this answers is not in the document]
bmc-msft: Will it make you 👎 that I'm about to remove this, as the `disable @cache` integration that landed in main makes this no longer required?  😀

PR #337, opened 2020-11-20
Fix state management in the scheduler

chkeita:


bmc-msft:
Given @ranweiler's experience with the state machine code, I added them to review this PR.

bmc-msft:
Note, this passed the integration tests.

ranweiler (approved):
Can we add some unit tests to validate this going forward?

It should just require adding some test cases [here](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-supervisor/src/agent/tests.rs) to ensure that the right events get pushed (or not) to the [`events` field](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-supervisor/src/coordinator/double.rs#L9) of `CoordinatorDouble`, depending on the initial state.

ranweiler:
@chkeita, is it correct that you can now manually observe _all_ expected events being received by the service handler? I.e. in a normal job with no setup script, transitions from `init` to `free` to `setting_up` to `pending` to `busy` to `done`?

chkeita:
validating

src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -86,7 +86,7 @@ impl Agent {
 
     async fn update(&mut self) -> Result<bool> {
         let last = self.scheduler.take().ok_or_else(scheduler_error)?;
-
+        let previous_state = NodeState::from(&last);
```
chkeita, line 89: validating

src/agent/onefuzz-supervisor/src/worker/tests.rs
```diff
@@ -67,6 +67,22 @@ pub struct ChildDouble {
     killed: bool,
 }
 
+impl Default for ChildDouble {
+    fn default() -> Self {
+        Self {
+            id: 0,
+            exit_status: Some(ExitStatus {
```
ranweiler, line 74: This should be an impossible value for `exit_status`: either `code` or `signal` is always set.

src/agent/onefuzz-supervisor/src/worker/tests.rs
```diff
@@ -67,6 +67,22 @@ pub struct ChildDouble {
     killed: bool,
 }
 
+impl Default for ChildDouble {
+    fn default() -> Self {
```
ranweiler, line 71: Idea: instead of making a different impl for `Default`, which was already being derived, just add another ctor to the impl block for `ChildDouble`, like:
```rust
impl ChildDouble {
    fn exited_success() -> Self { 
        // Use the derived `Default` impl here, then update it to give it a non-`None` exit status
    }
}
```

chkeita: removed, it was not needed for my changes

src/agent/onefuzz-supervisor/src/worker/tests.rs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 use crate::work::WorkUnit;
+use crate::worker::double::ChildDouble;
```
ranweiler, line 5: 👍 awesome!

src/agent/onefuzz-supervisor/src/setup/double.rs
```diff
@@ -7,12 +7,16 @@ use super::*;
...
     pub script: SetupOutput,
+    pub error_message: Option<String>,
 }
 
 #[async_trait]
 impl ISetupRunner for SetupRunnerDouble {
     async fn run(&mut self, work_set: &WorkSet) -> Result<SetupOutput> {
         self.ran.push(work_set.clone());
+        if let Some(error) = self.error_message.clone(){
+            anyhow::bail!(error);
```
ranweiler, line 18: 👌  excellent, I was hoping something like this would be easy to add.

ranweiler: (I also like how this introduces fault injection at the level of the setup runner itself, vs. just the child)

src/agent/onefuzz-supervisor/src/agent/tests.rs
```diff
@@ -64,7 +66,7 @@ impl Fixture {
         WorkSet {
             reboot: false,
             setup_url: self.setup_url(),
-            script: false,
+            script: true,
```
ranweiler, line 69: Was this necessary, or a dev artifact? I suggesting reverting, if so.

ranweiler (approved):

PR #338, opened 2020-11-20
Re-enable the retry logic for App Password creation

bmc-msft:
PR #332 inadvertently broke app password creation retry logic.

This re-adds it in a fashion that calling registration.py directly benefits from the retry attempts.


bmc-msft:
Note, this PR also addresses issues from flake8.  I will look to enabling flake8 in the cicd pipeline in the near future.

src/deployment/registration.py
```diff
@@ -354,7 +376,8 @@ def assign_scaleset_role(onefuzz_instance_name: str, scaleset_name: str) -> None
 
     if not managed_node_role:
         raise Exception(
-            "ManagedNode role not found int the onefuzz application registration. Please redeploy the instance"
+            "ManagedNode role not found int the onefuzz application "
```
ranweiler, line 379:
```suggestion
            "ManagedNode role not found in the OneFuzz application "
```

src/deployment/data_migration.py
```diff
@@ -47,7 +47,6 @@ def migrate_notification_keys(table_service: TableService) -> None:
     notifications = table_service.query_entities(
         table_name, select="PartitionKey,RowKey,config"
     )
-    partitionKey = None
```
ranweiler, line 50: Just unused, I assume?

src/deployment/registration.py
```diff
@@ -211,6 +211,28 @@ def create_application_registration(
 
 
 def add_application_password(app_object_id: UUID) -> Tuple[str, str]:
+    # Work-around the race condition where the app is created but passwords cannot
```
ranweiler, line 214: Do we have any generic exponential backoff code we could use in this function?

ranweiler (approved):

src/deployment/data_migration.py
bmc-msft: Correct.

src/deployment/registration.py
bmc-msft: We don't at this time.  We could use something like that. I'll work on that in the future.

PR #339, opened 2020-11-23
Updating yml file to run config endpoint command with tenant/authority ID. 

nharper285:

Originally, the yml file printed out a semi-generalized _onefuzz config --endpoint_ comman. This command did have a specified _--authority_ and so it used the Microsoft id by default. To enable users to work with OneFuzz on tenants other than the standard Microsoft tenant, we have added a _--authority_ parameter that is printed out at the end of the deployment. 

## PR Checklist

## Info on Pull Request

Changes to the yml file. 

## Validation Steps Performed

We have made this change to our local automation repository and tested an automated deployment pipeline with this change.


contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
```diff
@@ -121,6 +121,6 @@ stages:
           targetType: 'inline'
           script: |
             set -ex
-            ./onefuzz-cli-$(version).exe config --endpoint $(ONEFUZZ_SERVICE_URL) --client_id "$(AZURE_CLIENT_ID)" --client_secret "$(AZURE_CLIENT_SECRET)"
+            ./onefuzz-cli-$(version).exe config --endpoint $(ONEFUZZ_SERVICE_URL) --client_id "$(AZURE_CLIENT_ID)" --client_secret "$(AZURE_CLIENT_SECRET)" --authority "https://login.microsoftonline.com/$(AZURE_TENANT_ID)"
```
bmc-msft, line 124: This is a breaking change for anyone that has deployed this.

Can you make this optional?

bmc-msft (changes requested):

contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
nharper285: Hi Brian, we can absolutely make an update that makes this optional -- but if I might ask -- how is this breaking? 
The yml file has always asked for the AZURE_TENANT_ID, so how is this not semi-redundant for most users?

anslutsk: If the concern is about hardcoding login.microsoftonline.com, it's not a breaking change since that is the default value used by the CLI if none is provided. I also don't understand why this is a breaking change either.

bmc-msft: @nharper285 I missed that the script already had the tenant id.

bmc-msft (approved):

contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
nharper285: Thanks Brian!

PR #341, opened 2020-11-24
ignore nodes already scheduled for re-imaging in outdated check

bmc-msft:
If a node is already scheduled to be reimaged/deleted, we should not bother checking if it's outdated.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/pools.py
+++ b/src/api-service/__app__/onefuzzlib/pools.py
@@ -99,6 +99,7 @@ class Node(BASE_NODE, ORMMixin):
         scaleset_id: Optional[UUID] = None,
         states: Optional[List[NodeState]] = None,
         pool_name: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -118,6 +118,10 @@ def search_outdated(
     def mark_outdated_nodes(cls) -> None:
         outdated = cls.search_outdated()
         for node in outdated:
+            if node.reimage_requested or node.delete_requested:
```
ranweiler, line 121: Instead of doing this when iterating, would it be better to have the `search_outdated()` method only return such nodes, or if it is used elsewhere, add a new method for querying just those nodes?

ranweiler (approved):

chkeita (approved):

ranweiler (approved):

PR #342, opened 2020-11-24
clarify prefix-expansion errors

bmc-msft:
Updates the error upon prefix-expansion issues from:
```
ERROR:cli:command failed: name does not expand to a value: linux
```

to:
```
ERROR:cli:command failed: Unable to find pool name based on prefix: linux
```


ranweiler (approved):
Nice UX touch-up.

PR #345, opened 2020-11-25
Adding -DD parameter to unzip command. Forces file replacement irrespective of timestamp.

nharper285:

_What is this about?_
When running automated deployments, Andrew and I found that 'tools' were not being properly replaced with the updated versions if the deployment was created _prior_ to the original instance deployment. 
To explain the scenario, I've provided our own experience as an example. 

Let's look at the following scenario:
- MSR released version 1.9.0 at 2:55pm on 11/20/2020. 
- A 1.7.0 instance of OneFuzz was manually deployed by a user at 2:55pm on 11/23/2020. 
- The user then creates an Azure pipeline to automatically update the instance and triggers an update at 2:55pm on 11/24/2020

What ends up happening is that when the automated deployment occurs, it unzips the 1.9.0 deployment zip and compares the timestamps within that zip (2:55pm on 11/20/2020) to the timestamps currently in the 'tools' container within the function app (2:55pm on 11/23/2020). Because the container timestamps are more recent than the zip timestamps, the files **are not** replaced. We are then stuck in a situation where some of the tools, like the onefuzz-agent, are version 1.7.0 while the rest of the api have been properly updated to 1.9.0. 

To mitigate this issue, we've added the -DD parameter to the unzip command. 
The duplicated option -DD forces suppression of timestamp restoration for all extracted entries (files and directories). This option results in setting the timestamps for all extracted entries to the current time.

## Info on Pull Request

_What does this include?_
Changes to the .yml deployment file. Specifically, we've added the -DD param to the unzip command. 

## Validation Steps Performed
We've tested this change in our deployment pipeline. 

One can test by following a similar pattern of events as the ones detailed above. Without the param, the files are not updated. With the param, the files are properly updated.

### DIFF
--- a/contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
+++ b/contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
@@ -69,7 +69,7 @@ stages:
             workingDirectory: "$(ONEFUZZ_DEPLOY_LOC)/$(artifact)"
             script: |
               set -ex
-              unzip -DD onefuzz-deployment-$(onefuzz_release.version).zip
+              unzip onefuzz-deployment-$(onefuzz_release.version).zip
               pip install -r requirements.txt
               wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
               sudo dpkg -i packages-microsoft-prod.deb
--- a/contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
+++ b/contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
@@ -121,6 +121,6 @@ stages:
           targetType: 'inline'
           script: |
             set -ex
-            ./onefuzz-cli-$(version).exe config --endpoint $(ONEFUZZ_SERVICE_URL) --client_id "$(AZURE_CLIENT_ID)" --client_secret "$(AZURE_CLIENT_SECRET)"
+            ./onefuzz-cli-$(version).exe config --endpoint $(ONEFUZZ_SERVICE_URL) --client_id "$(AZURE_CLIENT_ID)" --client_secret "$(AZURE_CLIENT_SECRET)" --authority "https://login.microsoftonline.com/$(AZURE_TENANT_ID)"
             ./onefuzz-cli-$(version).exe --version
             until ./onefuzz-cli-$(version).exe versions check --exact; do echo "waiting due to version mismatch"; sleep 1; done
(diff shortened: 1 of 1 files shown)


contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
```diff
@@ -69,7 +69,7 @@ stages:
             workingDirectory: "$(ONEFUZZ_DEPLOY_LOC)/$(artifact)"
             script: |
               set -ex
-              unzip onefuzz-deployment-$(onefuzz_release.version).zip
+              unzip -DD onefuzz-deployment-$(onefuzz_release.version).zip
```
ranweiler, line 72: IIUC, the issue is that (1) the new files _were_ getting correctly extracted on the build machine, but (2) indirect invocations of `azcopy sync` were not remotely updating some genuinely new files, because it takes an action/no-action based on local vs. remote timestamps.

If that's true, does that mean we should consider changing [the `azcopy` invocation](https://github.com/microsoft/onefuzz/blob/main/src/deployment/deploy.py#L579), e.g. to `azcopy copy`?

nharper285: I believe either solution would resolve the issue.

bmc-msft: We should use `azcopy copy`, as `unzip -DD` isn't portable.

anslutsk: There are two different places in deploy.py that invoke the 'sync' command and changing one and not the other could lead to confusion. Any objection to changing both calls? 

One is for 'tools' dir and one is for 'instance-specific-setup' dir.

bmc-msft: Both should be updated.

nharper285: Thanks Brian. I'm happy to go ahead and push the changes to deploy.py. Please let me know if you'd rather do it yourself, but otherwise, I will update.

bmc-msft: You're welcome to make the updates.

nharper285: There appears to be 3 places where azcopy "sync" is used. Should I make all 3 changes?

bmc-msft: Yes.

bmc-msft:
OBE by #347

Issue #343, opened 2020-11-24
ASAN log parsing fails to parse "AddressSanitizer CHECK failed" logs


bmc-msft:
In the linux-radamsa integration test using the ASAN instrumented "trivial-crash" binary, I saw the following log message:

```
unable to parse asan log /tmp/.tmpVZ5iqh/asan-log.13614:
=================================================================
==13614==AddressSanitizer CHECK failed: /build/llvm-toolchain-10-yegZYJ/llvm-toolchain-10-10.0.0/compiler-rt/lib/asan/asan_descriptions.cpp:80 "((0 && "Address is not in memory and not in shadow?")) != (0)" (0x0, 0x0)
    #0 0x49bbee in __asan::AsanCheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/onefuzz/blob-containers/oft-setup-3a9daf5d852453d4af5680a55b21d482/fuzz.exe+0x49bbee)
    #1 0x4b010f in __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) (/onefuzz/blob-containers/oft-setup-3a9daf5d852453d4af5680a55b21d482/fuzz.exe+0x4b010f)
    #2 0x423c3e in __asan::GetShadowAddressInformation(unsigned long, __asan::ShadowAddressDescription*) (/onefuzz/blob-containers/oft-setup-3a9daf5d852453d4af5680a55b21d482/fuzz.exe+0x423c3e)
    #3 0x424ca6 in __asan::AddressDescription::AddressDescription(unsigned long, unsigned long, bool) (/onefuzz/blob-containers/oft-setup-3a9d...<truncated>
```

There are two issues here.

1. We should differentiate `no error` and `unable to parse error` when handling ASAN log files. 
2. We should fix the support for CHECK failed testing to indicate there is something wrong with the test.

PR #351, opened 2020-12-01
Fail the task if parsing asan_log files fail

bmc-msft:
This differentiates parsing ASAN log parse failures from ASAN logs not existing, fixing the first part of #343.

### DIFF
--- a/src/agent/onefuzz/src/asan.rs
+++ b/src/agent/onefuzz/src/asan.rs
@@ -7,6 +7,8 @@ use regex::Regex;
 use std::{collections::HashMap, hash::BuildHasher, path::Path};
 use tokio::{fs, stream::StreamExt};
 
+const ASAN_LOG_TRUNCATE_SIZE: usize = 4096;
+
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz/src/asan.rs
```diff
@@ -136,8 +136,8 @@ pub async fn check_asan_string(mut data: String) -> Result<Option<AsanLog>> {
     if asan.is_some() {
         return Ok(asan);
     } else {
-        if data.len() > 1024 {
-            data.truncate(1024);
+        if data.len() > 4096 {
```
ranweiler, line 139: Shall we add a constant for this value, if you're changing it anyway?

I'm guessing the change is not necessary, but just made debugging easier and produces a better UX.

ranweiler (approved):

src/agent/onefuzz/src/asan.rs
bmc-msft: Added const.

It wasn't specifically needed, but as you mentioned, gives more context for the asan log, which makes it easier to debug.

Issue #141, opened 2020-10-13
Shut down agents running tasks before deleting task queues


bmc-msft:
When tasks are stopped, the related queues are deleted immediately.  This causes any agents currently using the queue to generate failures such as

```
[2020-10-13T22:06:54Z ERROR onefuzz_supervisor] error running supervisor agent: error sending request for url (https://HOST/messages?SAS_URL): connection error: Connection reset by peer (os error 104)
Error: error sending request for url (https://HOST/messages?SAS_URL): connection error: Connection reset by peer (os error 104)

Caused by:
    0: connection error: Connection reset by peer (os error 104)
    1: Connection reset by peer (os error 104)
[2020-10-13T22:06:54Z ERROR onefuzz_downloader] error running downloader: supervisor failed: Some(1)
Error: supervisor failed: Some(1)
```

While these errors are silenced at the service end as the task is in the stopping state, these look concerning to the user.

PR #356, opened 2020-12-01
Check application insights for errors generated during integration tests

bmc-msft:
In practice, Application Insights can take up to 3 minutes before something sent to it is available via KQL.

This PR logs a start and stop marker such that the integration tests only search for logs during the integration tests.  This reduces the complexity when using the integration tests during the development process.


src/integration-tests/integration-test.py
```diff
@@ -478,6 +507,45 @@ def stop_template(self, target: str, delete_containers: bool = True) -> None:
...
+            self.stop_log_marker, limit=1, timespan="PT1H"
+        )
+        return (
+            len(logs["PrimaryResult"]) > 0,
+            "waiting for application insight logs to flush",
+            True,
+        )
+
+    def check_logs_for_errors(self) -> bool:
+        # only check for errors that exist between the start and stop markers
```
ranweiler, line 523: It looks like the query has an additional condition: within the last 2 hours. Maybe worth emphasizing here, too?

ranweiler (approved):

bmc-msft:
In larger testing, I have found the functionality works well.  However, before this gets merged the integration tests should wait for the tasks to stop before deleting the containers.  

As is, the integration tests stop the tasks and immediately delete the related containers.  This deletion of containers as well as the queues getting deleted prior to stopping the tasks, as discussed in #141, leads to additional errors that we should cleanup ahead of time.

PR #358, opened 2020-12-01
handle asan check failures

### LINKED ISSUE #343: ASAN log parsing fails to parse "AddressSanitizer CHECK failed" logs

bmc-msft:
Handle parsing "CHECK failed" ASAN logs.  Fixes #343

### DIFF
--- a/src/agent/onefuzz/src/asan.rs
+++ b/src/agent/onefuzz/src/asan.rs
@@ -20,7 +20,11 @@ pub struct AsanLog {
 
 impl AsanLog {
     pub fn parse(text: String) -> Option<Self> {
-        let (summary, sanitizer, fault_type) = parse_summary(&text)?;
+        let (summary, sanitizer, fault_type) = match parse_summary(&text) {
+            Some(x) => x,
+            None => parse_failure(&text)?,
+        };
+
         let call_stack = parse_call_stack(&text).unwrap_or_else(Vec::default);
 
         let log = Self {
@@ -53,6 +57,16 @@ impl AsanLog {
     pub fn call_stack_sha256(&self) -> String {
         sha256::digest_iter(self.call_stack())
     }
+}
+
+fn parse_failure(text: &str) -> Option<(String, String, String)> {
+    let pattern = r"==\d+==((\w+) (CHECK failed): [^ \n]+)";
+    let re = Regex::new(pattern).ok()?;
+    let captures = re.captures(text)?;
+    let summary = captures.get(1)?.as_str().trim();
+    let sanitizer = captures.get(2)?.as_str().trim();
+    let fault_type = captures.get(3)?.as_str().trim();
+    Some((summary.into(), sanitizer.into(), fault_type.into()))
 }
 
 fn parse_summary(text: &str) -> Option<(String, String, String)> {
@@ -176,7 +190,7 @@ mod tests {
     use super::AsanLog;
 
     #[test]
-    fn test_asan_log_parse() {
+    fn test_asan_log_parse() -> anyhow::Result<()> {
         let test_cases = vec![
             (
                 "data/libfuzzer-asan-log.txt",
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz/src/asan.rs
```diff
@@ -20,7 +20,11 @@ pub struct AsanLog {
 
 impl AsanLog {
     pub fn parse(text: String) -> Option<Self> {
-        let (summary, sanitizer, fault_type) = parse_summary(&text)?;
+        let (summary, sanitizer, fault_type) = match parse_summary(&text) {
+            Some(x) => x,
+            None => parse_failure(&text)?,
```
ranweiler, line 25: In falling back to `parse_failure()`, aren't we throwing away detail?

Should `parse()` really be returning an enum that explicitly encodes the different cases of:
1. We parsed an ASAN log where the _target_ tripped a fault oracle
2. We parsed an ASAN log where there was an internal ASAN error
3. There was no ASAN-specific logging at all

src/agent/onefuzz/src/asan.rs
```diff
@@ -55,6 +59,16 @@ impl AsanLog {
     }
 }
 
+fn parse_failure(text: &str) -> Option<(String, String, String)> {
```
ranweiler, line 62: Can we rename this to `parse_asan_runtime_error()` or similar, to make the difference between this an "the target code failed" extra obvious?

src/agent/onefuzz/src/asan.rs
```diff
@@ -226,15 +240,28 @@ mod tests {
...
+            (
+                "data/asan-check-failure.txt",
+                "AddressSanitizer",
+                "CHECK failed",
+                12,
+            ),
+            (
+                "data/asan-check-failure-missing-symbolizer.txt",
+                "AddressSanitizer",
+                "CHECK failed",
```
ranweiler, line 252: Double-checking: this test case just tests that if the symbolizer is missing, our parser doesn't choke on that in the output, and can still extract that the problem was a `CHECK failed`, right?

ranweiler (approved):

src/agent/onefuzz/src/asan.rs
bmc-msft: Correct.

src/agent/onefuzz/src/asan.rs
bmc-msft: Discussed offline.

PR #359, opened 2020-12-01
Add support for ASAN print_scariness

bmc-msft:
This adds support for ASAN's "scariness" score and description.

The scariness index is exposed by the user providing ASAN_OPTIONS=print_scariness=1 for their target_env during crash reporting.

https://github.com/llvm/llvm-project/blob/master/compiler-rt/lib/asan/asan_scariness_score.h

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
+++ b/src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
@@ -36,6 +36,9 @@ pub struct CrashReport {
     pub task_id: Uuid,
 
     pub job_id: Uuid,
+
+    pub scariness_index: Optional<u32>,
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -36,6 +36,9 @@ pub struct CrashReport {
     pub task_id: Uuid,
 
     pub job_id: Uuid,
+
+    pub scariness_index: Option<u32>,
```
ranweiler, line 40: I suggest we call these `scariness_score` and `scariness_description`, to match the compiler-rt language.

ranweiler (approved):

PR #360, opened 2020-12-02
fix job_templates deletion

bmc-msft:
The `delete` endpoint did not actually delete job templates.

### DIFF
--- a/src/api-service/__app__/job_templates_manage/__init__.py
+++ b/src/api-service/__app__/job_templates_manage/__init__.py
@@ -58,6 +58,9 @@ def delete(req: func.HttpRequest) -> func.HttpResponse:
         return not_ok(request, context="JobTemplateDelete")
 
     entry = JobTemplateIndex.get(request.name)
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/job_templates_manage/__init__.py
```diff
@@ -54,6 +54,9 @@ def delete(req: func.HttpRequest) -> func.HttpResponse:
         return not_ok(request, context="JobTemplateDelete")
 
     entry = JobTemplateIndex.get(request.name)
+    if entry is not None:
+        entry.delete()
```
ranweiler, line 58: Do have a strategy for integration testing this?

ranweiler (approved):

src/api-service/__app__/job_templates_manage/__init__.py
bmc-msft: Not yet.  I've explored a few options, but will need time to explore first.

PR #378, opened 2020-12-10
Verify a workset only exists along with a reboot context

bmc-msft:
Adds the following:

1. Serializes a workset to disk during setup.
2. Upon deserializing a RebootContext, delete the file from disk (We support rebooting once and only once)
3. Check if a workset exists with a RebootContext
    1. If True, continuing processing
    2. if False, mark the tasks & node as "Done" with appropriate errors via:
        1. send WorkerEvent::Done events for each of the tasks in the work set
        2. send StateUpdateEvent::Done for the node

### DIFF
--- a/src/agent/onefuzz-supervisor/src/main.rs
+++ b/src/agent/onefuzz-supervisor/src/main.rs
@@ -12,12 +12,16 @@ extern crate serde;
 #[macro_use]
 extern crate clap;
 
-use crate::heartbeat::*;
+use crate::{
+    config::StaticConfig, coordinator::StateUpdateEvent, heartbeat::*, work::WorkSet,
+    worker::WorkerEvent,
+};
 use std::path::PathBuf;
 
 use anyhow::Result;
 use onefuzz::{
     machine_id::{get_machine_id, get_scaleset_name},
+    process::ExitStatus,
     telemetry::{self, EventData},
 };
 use structopt::StructOpt;
@@ -34,8 +38,6 @@ pub mod scheduler;
 pub mod setup;
 pub mod work;
 pub mod worker;
-
-use config::StaticConfig;
 
 #[derive(StructOpt, Debug)]
 enum Opt {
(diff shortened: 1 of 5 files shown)


bmc-msft:
This passed basic integration tests, but none of those do a reboot.

This PR should update the integration tests to one of the tests to enable `reboot_after_setup`.

src/agent/onefuzz-supervisor/src/work.rs
```diff
@@ -28,6 +30,43 @@ impl WorkSet {
     pub fn task_ids(&self) -> Vec<TaskId> {
         self.work_units.iter().map(|w| w.task_id).collect()
     }
+
+    pub fn context_path() -> Result<PathBuf> {
+        Ok(onefuzz::fs::onefuzz_root()?.join("workset_context.json"))
+    }
+
+    pub async fn load_context() -> Result<Option<Self>> {
```
ranweiler, line 38: In the future, we should try to decouple these kinds of ctors from the actual type. This impl is convenient, but bakes application-wide knowledge into a `WorkSet` ctor that might be better named `load_from_onefuzz_context()` or `load_from_fs_context()`.

ranweiler: Concretely, for now, how about we just rename this `load_from_context()`, since it loads a `WorkSet` from the filesystem context?

ranweiler: (Ah, I see, you were mostly just copying what we do in the reboot case. There "context == workset" is an implementation detail, which is why I'd choose different language here).

ranweiler (approved):

src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -125,6 +127,37 @@ fn load_config(opt: RunOpt) -> Result<StaticConfig> {
     Ok(config)
 }
 
+async fn check_existing_worksets(coordinator: &mut coordinator::Coordinator) -> Result<()> {
```
ranweiler, line 130: Could you add a comment describing what we intend this to do?

src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -125,6 +127,37 @@ fn load_config(opt: RunOpt) -> Result<StaticConfig> {
...
+            coordinator.emit_event(event.into()).await?;
+        }
+
+        let event = StateUpdateEvent::Done {
+            error: Some(failure.to_string()),
+            script_output: None,
+        };
+        coordinator.emit_event(event.into()).await?;
+
+        done::set_done_lock().await?;
```
ranweiler, line 154: Do we need to check for a pre-existing lock before starting this function?

PR #381, opened 2020-12-10
handle libfuzzer fuzzing non-zero exits better

bmc-msft:
When running libfuzzer in 'fuzzing' mode, we expect the following on exit.

If the exit code is zero, crashing input isn't required.  This happens if the user specifies '-runs=N'

If the exit code is non-zero, then crashes are expected.  In practice, there are two causes to non-zero exits.
1. If the binary can't execute for some reason, like a missing prerequisite
2. If the binary _can_ execute, sometimes the sanitizers are put in such a bad place that they are unable to record the input that caused the crash.

This PR enables handling these two non-zero exit cases.

1. Optionally verify the libfuzzer target loads appropriately using `target_exe -help=1`.  This allows failing faster in the common issues, such a missing prerequisite library.
2. Optionally allow non-zero exits without crashes to be a warning, rather than a task failure.


bmc-msft:
This passed integration tests.

src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -40,6 +40,35 @@ impl<'a> LibFuzzer<'a> {
         }
     }
 
+    pub async fn check_help(&self) -> Result<()> {
```
ranweiler, line 43: Could you add a doc comment for this?

src/api-service/__app__/onefuzzlib/job_templates/defaults/libfuzzer.py
```diff
@@ -217,6 +217,38 @@
...
+                ),
+                UserFieldLocation(
+                    op=UserFieldOperation.add,
+                    path="/tasks/2/task/check_fuzzer_help",
+                ),
+            ],
+        ),
+        UserField(
+            name="require_crash_on_failure",
+            help="Require crashes upon non-clean exits from libfuzzer",
```
ranweiler, line 242: Replace "non-clean" with "non-zero"?

docs/webhook_events.md
```diff
@@ -182,6 +182,14 @@ Each event will be submitted via HTTP POST to the user provided URL.
                     "title": "Check Retry Count",
                     "type": "integer"
                 },
+                "check_fuzzer_help": {
+                    "title": "Check Fuzzer Help",
```
ranweiler, line 186: We could just call this `check_fuzzer`, and potentially do other checks in the future. The fact that we invoke `-help=1` specifically seems like an implementation detail.

docs/webhook_events.md
```diff
@@ -182,6 +182,14 @@ Each event will be submitted via HTTP POST to the user provided URL.
...
+                    "type": "boolean"
+                },
+                "require_crash_on_failure": {
```
ranweiler, line 189: The word `require` here isn't wrong, but maybe `expect` is more appropriate?

ranweiler (approved):

docs/webhook_events.md
bmc-msft: The explicit setting is intentional.  This enables working towards libfuzzer-like binaries that don't use `-help=1` but have many other similar properties.

PR #387, opened 2020-12-11
add delayed start to heartbeats

### LINKED ISSUE #386: agent & supervisor heartbeat needs an initial jitter
The heartbeat async thread needs to start at a random interval within in the periodicity of the heartbeat.

As is, if we start ~3000 nodes at once, we have a recurring storm roughly once every 5 minutes when all of the nodes check in.

bmc-msft:
Adds a random initial jitter the size of the heartbeat periodicity to prevent heartbeats storming the service when we launch 3000 nodes roughly at the same time.

Fixes #386


src/agent/onefuzz/src/jitter.rs
```diff
@@ -13,3 +13,9 @@ pub fn jitter(value: Duration) -> Duration {
 pub async fn delay_with_jitter(value: Duration) {
     delay_for(jitter(value)).await
 }
+
+pub async fn delay_with_jitter_size(value: Duration) {
+    let random: u64 = thread_rng().gen_range(0, value.as_secs());
```
ranweiler, line 18: 1. Do we really want to allow a "delay" of 0?

2. It looks like this function is not really a delay with jitter, but more of a uniformly random delay, parametric in an upper bound. I suggest we just rename it to avoid confusion, given that right it looks misleadingly similar to a function where the user-provided value is a _lower_ bound.

ranweiler (approved):

src/agent/onefuzz/src/jitter.rs
bmc-msft: Sigh, resolve comment doesn't include feedback.

1. Some amount of delay of zero is fine here.  The intent is to reduce the initial heartbeat storm.   
2. renamed per feedback

PR #389, opened 2020-12-14
add keyvault management to administrative role

bmc-msft:
This is a prerequisite for #326


ranweiler (approved):

src/deployment/deployment-role.json
```diff
@@ -3,6 +3,7 @@
     "Description": "Permissions required for OneFuzz deployment",
     "Actions": [
         "Microsoft.Authorization/locks/*",
+        "Microsoft.Keyvault/vaults/*",
```
ranweiler, line 6: Is there someplace where we should add/update documentation about this new permission?

bmc-msft: At the moment, no.

When we integration keyvault support, we'll need to specifically call out in the release notes.  (Hence the breaking-change label)

chkeita (approved):

PR #391, opened 2020-12-14
filter out deleted nodes as to prevent them from being saved later

bmc-msft:
In `Scaleset.cleanup_nodes`, nodes that are no longer part of the scaleset should get deleted.  Without filtering the list, the nodes could get re-saved to the Node table later on.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/pools.py
+++ b/src/api-service/__app__/onefuzzlib/pools.py
@@ -792,7 +792,7 @@ class Scaleset(BASE_SCALESET, ORMMixin):
                     "no longer in scaleset: %s:%s", self.scaleset_id, node.machine_id
                 )
                 node.delete()
-        nodes = [x for x in nodes if x.machine_id not in nodes]
+        nodes = [x for x in nodes if x.machine_id not in azure_nodes]
 
         nodes_to_reset = [x for x in nodes if x.state in NodeState.ready_for_reset()]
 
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -783,13 +783,16 @@ def cleanup_nodes(self) -> bool:
...
+        removed = []
         # Nodes do not exists in scalesets but in table due to unknown failure
         for node in nodes:
             if node.machine_id not in azure_nodes:
+                removed.append(node.machine_id)
                 logging.info(
                     "no longer in scaleset: %s:%s", self.scaleset_id, node.machine_id
                 )
                 node.delete()
+        nodes = [x for x in nodes if x.machine_id not in nodes]
```
ranweiler, line 795: Should the filter condition be `if x.machine_id not in removed`?

Suggestion: consider not shadowing/rebinding `nodes`, but using a fresh name that describes what is important about this list of nodes.

ranweiler (changes requested):

src/api-service/__app__/onefuzzlib/pools.py
ranweiler: Also, as written, it seems like the condition here should arguably be a type error, since `nodes` is something like a `List[Node]`, not `List[UUID]`.

bmc-msft: Good catch.  This should have checked `x.machine_id in azure_nodes`

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -783,13 +783,16 @@ def cleanup_nodes(self) -> bool:
             logging.info("no nodes need updating: %s", self.scaleset_id)
             return False
 
+        removed = []
```
ranweiler, line 786: Why do we track `removed` if we don't do anything with it?

ranweiler (approved):

PR #392, opened 2020-12-14
add crash_report_created webhook event

bmc-msft:
This adds `crash_report_created`, which enables performing crash report notifications outside our existing integrations.


bmc-msft:
Note, this PR also validates that all entries in the webhook event enum & union are represented in the documentation.

bmc-msft:
OBE by #394

PR #394, opened 2020-12-15
Unify Dashboard & Webhook events

bmc-msft:
This change unifies the previously adhoc SignalR events and Webhooks into a single event format.


src/cli/onefuzz/status/cache.py
```diff
@@ -4,39 +4,75 @@
...
+MESSAGE = Tuple[datetime, EventType, str]
 
 MINUTES = 60
 HOURS = 60 * MINUTES
 DAYS = 24 * HOURS
 
 
-def fmt_delta(data: timedelta) -> str:
-    result = []
+class MiniNode(BaseModel):
```
ranweiler, line 49: What are these `Mini*` classes?

src/cli/onefuzz/status/cache.py
```diff
@@ -106,194 +139,195 @@ def __init__(
...
-            elif isinstance(v, (int, str)):
-                data[k] = v
-
-        as_str = fmt(data)
-        messages = [x for x in self.messages if (x[1:] != [name, as_str])][-99:]
-        messages += [(datetime.now(), name, as_str)]
-
+        messages = [x for x in self.messages][-99:]
+        messages += [
+            (datetime.now(), message.event_type, message.event.json(exclude_none=True))
```
ranweiler, line 176: This isn't new, but should we be explicitly specifying the TZ as UTC in all our calls to `now()`?

src/cli/onefuzz/status/cache.py
```diff
@@ -311,63 +345,60 @@ def should_render_job(self, job: Job) -> bool:
...
         return True
 
+    def job_stopped(self, event: EventJobStopped) -> None:
+        if event.job_id in self.jobs:
+            del self.jobs[event.job_id]
+
     def render_jobs(self) -> List[Tuple]:
         results: List[Tuple] = []
 
-        for (timestamp, job) in sorted(self.jobs.values(), key=lambda x: x[0]):
```
ranweiler, line 317: Why do we no longer sort by timestamp?

ranweiler (approved):

src/cli/onefuzz/status/cache.py
bmc-msft: Once I had a 3k concurrent tasks in the view, sorting on timestamp caused the data to jump around far too frequently to be useful.

src/cli/onefuzz/status/cache.py
bmc-msft: In general, yes.  However, in the case of displaying to users, it made sense to use `now()` here.

This only displays the local timestamp of when the message was received.

src/cli/onefuzz/status/cache.py
bmc-msft: We don't send the full context of the Node/Task/Job, etc via SignalR.  As such, these are slimmed down representations of the models specifically for the context of `status top`.

Added a comment to each of them to describe that.

Issue #372, opened 2020-12-04
Capture stdout/stderr from generators


bmc-msft:
Right now, we do not capture (and share) the stdout/stderr of generators. This makes debugging issues with the generators difficult.

PR #400, opened 2020-12-16
log stdout & stderr lines for supervisor & generator

### LINKED ISSUE #371: Capture stdout/stderr from supervisor
Right now, we do not capture (and share) the stdout/stderr of the supervisor.  This makes debugging issues with the supervisor difficult.

bmc-msft:
This fixes #371 and #372.

### DIFF
--- a/src/agent/onefuzz/src/process.rs
+++ b/src/agent/onefuzz/src/process.rs
@@ -13,6 +13,8 @@ use tokio::{
     sync::Notify,
 };
 
+const MAX_LOG_LINE_LENGTH: usize = 8192;
+
 /// Serializable representation of a process output.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct Output {
@@ -141,7 +143,12 @@ async fn monitor_stream(name: &str, context: &str, stream: impl AsyncRead + Unpi
         if bytes_read == 0 && buf.is_empty() {
             break;
         }
-        let line = String::from_utf8_lossy(&buf);
+        let mut line = String::from_utf8_lossy(&buf).to_string();
+        if line.len() > MAX_LOG_LINE_LENGTH {
+            line.truncate(MAX_LOG_LINE_LENGTH);
+            line.push_str("...<truncated>");
+        }
+
         info!("process ({}) {}: {}", name, context, line);
     }
     Ok(())
--- a/src/agent/onefuzz/src/process.rs
+++ b/src/agent/onefuzz/src/process.rs
@@ -7,6 +7,11 @@ use std::path::Path;
 use std::process::Command;
 use std::time::Duration;
 use std::{collections::HashMap, process::Stdio};
+use tokio::{
+    io::{AsyncBufReadExt, AsyncRead, BufReader},
+    process::Child,
+    sync::Notify,
+};
 
 /// Serializable representation of a process output.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz/src/process.rs
```diff
@@ -126,3 +131,76 @@ pub async fn run_cmd<S: ::std::hash::BuildHasher>(
     // convert processcontrol::Output into our Output
     runner.await?.map(|result| result.into())
 }
+
+async fn monitor_stream(name: &str, context: &str, stream: impl AsyncRead + Unpin) -> Result<()> {
+    let mut stream = BufReader::new(stream);
+    loop {
+        let mut buf = vec![];
```
lzybkr, line 138: Does this allocate a new buffer every time through the loop? If so, maybe move this out of the loop and clear the buffer here.

src/agent/onefuzz/src/process.rs
```diff
@@ -126,3 +131,76 @@ pub async fn run_cmd<S: ::std::hash::BuildHasher>(
...
+                .stderr
+                .take()
+                .ok_or_else(|| format_err!("stderr not captured"))?;
+
+            let stdout = process
+                .stdout
+                .take()
+                .ok_or_else(|| format_err!("stdout not captured"))?;
+
+            let stdout_log = monitor_stream("stdout", &context, stdout);
```
lzybkr, line 190: Can this deadlock? Say the stderr buffer fills up, so the child blocks waiting for us to read from it, but we're stuck waiting for a newline in stdout?

wait_with_output handles this case by waiting on both streams and reading from both as they signal that data is available.

bmc-msft: Looking at the underlying [read_until](https://github.com/tokio-rs/tokio/blob/aa171f2aa9593f64e78bde2c679a4c9bd9d9d163/tokio/src/io/util/read_until.rs#L55-L63), it polls and consumes what data is available.

Is this substantially different than what [wait_with_output](https://docs.rs/tokio/0.3.6/src/tokio/process/mod.rs.html#990-996) does, which is [read_to_end](https://github.com/tokio-rs/tokio/blob/8dbc3c79379f2243fc04d444239d009c1c610016/tokio/src/io/util/read_to_end.rs#L46-L53)?

bmc-msft: From offline discussion, I will test with a generator that interleaves printing 10MB of 'A' without newline to stdout & stderr, then prints a newline to each.

src/agent/onefuzz/src/process.rs
bmc-msft: From discussion, leaving as is.

src/agent/onefuzz/src/process.rs
```diff
@@ -7,6 +7,13 @@ use std::path::Path;
 use std::process::Command;
 use std::time::Duration;
 use std::{collections::HashMap, process::Stdio};
+use tokio::{
+    io::{AsyncBufReadExt, AsyncRead, BufReader},
+    process::Child,
+    sync::Notify,
+};
+
+const MAX_LOG_LINE_LENGTH: usize = 8192;
```
ranweiler, line 17: Could you comment on how this value was chosen? Even if arbitrary, helpful to keep track of.

src/agent/onefuzz/src/process.rs
```diff
@@ -126,3 +133,81 @@ pub async fn run_cmd<S: ::std::hash::BuildHasher>(
...
+            line.truncate(MAX_LOG_LINE_LENGTH);
+            line.push_str("...<truncated>");
+        }
+
+        info!("process ({}) {}: {}", name, context, line);
+    }
+    Ok(())
+}
+
+async fn monitor_process_child(
```
ranweiler, line 157: The name of this compared to the function below is not very illuminating, mostly because of the generality of the word "monitor". Also, `monitor_process_child()` seems to be the same conceptually as `monitor_process()`, since the `Child` is our handle on the process in question. The real difference is that in `monitor_process()`, we're sipping at the stdio streams, right? Could you add some brief item-level comments about intent, and consider an alternate name for one of them?

ranweiler (approved):

src/agent/onefuzz/src/process.rs
bmc-msft: This worked in my local testign.

PR #401, opened 2020-12-17
add 'onefuzz debug log tail'

bmc-msft:
Adds `onefuzz debug log tail <keyword>`, which enables performing the same component in `onefuzz debug log keyword <keyword>` in a loop.  

Optimizations:
* only returns the N records at a time (default 1000)
* each query only returns records that occur after the latest record received.
* If no results are returned, waits 10s before retrying
* Increases the wait time by 1.5x until the wait time is larger than 60s

Using `--filter` provides the ability to filter each record that comes back via jmespath.

Example uses:

Monitor any log messages (which ignores metrics) for a given job_id GUID
```
onefuzz debug logs tail bf4efdfd-685c-444a-81c5-d911477433ae --filter message
```

Log the job_id and task_id for each new unique report:
```
onefuzz debug logs tail new_unique_report --filter '[customDimensions.job_id, customDimensions.task_id]'
```

Log the job_id and task_id for each new unique report only for the specific job_id:
```
onefuzz debug logs tail "new_unique_report d5bcd4d2-4dab-49d5-a215-66db94fb0309" --filter '[customDimensions.job_id, customDimensions.task_id]'
```

### DIFF
--- a/src/cli/onefuzz/debug.py
+++ b/src/cli/onefuzz/debug.py
@@ -481,6 +481,7 @@ class DebugLog(Command):
         limit: int = 1000,
         indent: Optional[int] = None,
         filter: Optional[str] = "[message, name, customDimensions]",
+        timespan: Optional[str] = HOUR_TIMESPAN,
     ) -> None:
         """
         Perform an Application Insights keyword query akin to "Transaction Search"
@@ -502,11 +503,9 @@ class DebugLog(Command):
 
         while True:
             query = base_query.copy()
-            if last_seen is None:
-                results = self._query_parts(query, timespan=DAY_TIMESPAN)
-            else:
+            if last_seen is not None:
                 query.append(f'where timestamp > datetime("{last_seen}")')
-                results = self._query_parts(query, timespan=HOUR_TIMESPAN)
+            results = self._query_parts(query, timespan=timespan)
 
             if results:
                 last_seen = results[-1]["timestamp"]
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/debug.py
```diff
@@ -434,21 +470,57 @@ def keyword(
...
-            components.append(f"take {limit}")
+        expression = None
+        if filter:
+            expression = jmespath.compile(filter)
 
-        return self._query_parts(components, timespan=timespan, raw=raw)
+        base_query = self._build_keyword_query(value, limit=limit, desc=False)
+
+        last_seen: Optional[str] = None
+        wait = 10.0
```
ranweiler, line 502: Let's introduce a constant for the default `wait` value, which can use here, and when we reset it below.

src/cli/onefuzz/debug.py
```diff
@@ -434,21 +470,57 @@ def keyword(
...
+            results = self._query_parts(query, timespan=timespan)
+
+            if results:
+                last_seen = results[-1]["timestamp"]
+                for entry in results:
+                    if expression is not None:
+                        entry = expression.search(entry)
+                    if entry:
+                        print(json.dumps(entry, indent=indent, sort_keys=True))
+                wait = 10
```
ranweiler, line 517: This is an `int`, though `wait` is meant to be a `float`. This issue will disappear if we use a constant for resetting the `wait` value.

ranweiler (approved):

PR #407, opened 2020-12-31
remove redundant cast

bmc-msft:
ContainerWrapper.download_blob returns bytes, which makes this call redundant.

Found via the latest version of mypy.


bmc-msft:
This is merged into #408

PR #411, opened 2021-01-04
Add ssh keys to nodes on demand

bmc-msft:
Our existing model has a per-scaleset SSH key.  This update moves towards using user provided SSH keys when they need to connect to a given node.


bmc-msft:
Note, this isn't a breaking change yet.  However, when we remove the existing SSH key distribution mechanisms, it will be.

bmc-msft:
Feedback on user ergonomics would be useful.

As is, you need to specify the user on linux and not specify the user on Windows.

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct SshKeyInfo {
+    pub key: Secret<String>,
+    pub user: Option<String>,
+    set_permissions: bool,
+}
+
+#[cfg(target_os = "windows")]
+pub async fn add_ssh_key(key_info: SshKeyInfo) -> Result<()> {
+    if key_info.user.is_some() {
+        bail!("specifying a user is not supported on windows at this time");
```
ranweiler, line 25:
```suggestion
        bail!("specifying a user is not supported on Windows at this time");
```

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    let mut ssh_path =
+        PathBuf::from(env::var("ProgramData").unwrap_or_else(|_| "c:\\programdata".to_string()));
```
ranweiler, line 29: Ideally we'd try to grab `$SystemDrive` instead of hardcoding `c:`, but still fall back on `C:`.

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+    ssh_path.push("ssh");
+
+    let mut dsa_path = ssh_path.clone();
```
ranweiler, line 32: I think we can coalesce some of the copying/mutation via:
```suggestion
    let host_key_path = ssh_path.join("ssh_host_dsa_key");
    let admin_auth_keys_path = ssh_path.join("administrators_authorized_keys");
```
Note I also suggest some name changes.

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+    dsa_path.push("ssh_host_dsa_key");
+    key_path.push("administrators_authorized_keys");
+
+    {
+        let mut file = fs::OpenOptions::new().append(true).open(&key_path).await?;
+        file.write_all(key_info.key.expose_ref().as_bytes()).await?;
+    }
+
+    if key_info.set_permissions {
+        verbose!("removing Authenicated Users permissions to ssh key");
```
ranweiler, line 44:
```suggestion
        verbose!("removing Authenticated Users permissions from administrators_authorized_keys");
```

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    dsa_path.push("ssh_host_dsa_key");
+    key_path.push("administrators_authorized_keys");
+
+    {
+        let mut file = fs::OpenOptions::new().append(true).open(&key_path).await?;
+        file.write_all(key_info.key.expose_ref().as_bytes()).await?;
+    }
+
+    if key_info.set_permissions {
```
ranweiler, line 43: What's the use case for this?

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+            .await?;
+        if !result.status.success() {
+            bail!(
+                "set authorized_keys ({}) permissions failed: {:?}",
+                key_path.display(),
+                result
+            );
+        }
+
+        verbose!("removing inheritance");
```
ranweiler, line 63: Could you add comments describing why we need to do each of these things? Brief comments with documentation links would be great.

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+#[cfg(target_os = "linux")]
+pub async fn add_ssh_key(key_info: SshKeyInfo) -> Result<()> {
+    let user = match &key_info.user {
+        Some(user) => get_user_by_name(&user),
+        None => get_user_by_uid(get_current_uid()),
+    }
+    .ok_or_else(|| format_err!("unable to find user"))?;
+    info!("adding sshkey:{:?} to user:{:?}", key_info, user);
+
+    let mut ssh_path = user.home_dir().to_owned();
```
ranweiler, line 118: To avoid confusion and remove unnecessary mutability, how about:
```rust
let home_dir = user.home_dir();
// check if exists...
let ssh_path = home_dir.join(".ssh");
// create if needed, &c

File: src/agent/onefuzz-supervisor/src/commands.rs [hunk 8]
```diff
@@ -0,0 +1,176 @@
...
+            .await?;
+        if !result.status.success() {
+            bail!(
+                "set authorized_keys ({}) permissions failed: {:?}",
+                key_path.display(),
+                result
+            );
+        }
+
+        verbose!("copying ACL from ssh_host_data_key");
```
ranweiler, line 81:
```suggestion
        verbose!("copying ACL from ssh_host_dsa_key");
```

src/agent/onefuzz-supervisor/src/commands.rs
ranweiler, line 25: Is this not possible, or is it just a rabbit hole that we also consider not worth implementing, because all provided SSH users are admins?

If the latter, we should document that here.

src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[cfg(target_os = "windows")]
+use std::{env, path::PathBuf};
+
+#[cfg(target_os = "linux")]
+use users::{get_current_uid, get_user_by_name, get_user_by_uid, os::unix::UserExt};
+
+#[derive(Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct SshKeyInfo {
+    pub key: Secret<String>,
```
ranweiler, line 17: Maybe rename this `public_key`, though I think keeping it `Secret`-wrapped is fine.

src/pytypes/onefuzztypes/requests.py
```diff
@@ -233,3 +233,9 @@ class WebhookUpdate(BaseModel):
     event_types: Optional[List[WebhookEventType]]
     url: Optional[AnyHttpUrl]
     secret_token: Optional[str]
+
+
+class NodeAddSshKey(BaseModel):
+    machine_id: UUID
+    user: Optional[str]
```
ranweiler, line 240: Can we document the Windows vs. Linux gotcha here?

ranweiler (approved):

src/agent/onefuzz-supervisor/src/commands.rs
bmc-msft: When we support arbitrary locations, handle this then.

src/agent/onefuzz-supervisor/src/commands.rs
bmc-msft: Removed per discussion

ranweiler (approved):

chkeita (approved):

PR #417, opened 2021-01-06
adding {setup_dir} to variable expansion

chkeita:

Adds a new placeholder {setup_dir} for the setup directory 

## PR Checklist
* [x] Applies to work item: #221

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -33,6 +33,8 @@ pub struct CommonConfig {
     pub heartbeat_queue: Option<Url>,
 
     pub telemetry_key: Option<Uuid>,
+
+    pub setup_dir: Option<PathBuf>,
```
ranweiler, line 37: Is it ever valid for `setup_dir` to be absent? Doesn't every task require access to `target_exe`, which will always occur within a `setup_dir`?

src/agent/onefuzz-supervisor/src/work.rs
```diff
@@ -67,6 +67,13 @@ impl WorkSet {
 
         Ok(())
     }
+
+    pub fn setup_dir(&self) -> Result<PathBuf> {
```
ranweiler, line 71: Thinking about the future, we'll want some struct that represents the OneFuzz filesystem context, and this method should live on that, then compute the setup dir for some work parmeter (relative to itself). But this is good for now.

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -195,7 +202,12 @@ impl IWorkerRunner for WorkerRunner {
 
         let config_path = work.config_path()?;
 
-        fs::write(&config_path, work.config.expose_ref()).await?;
+        // injecting the setup_dir path in the agent config
+        let mut json_config: serde_json::Value = serde_json::from_str(work.config.expose_ref())?;
```
ranweiler, line 206: Do we really want this to be part of the task config?

I feel like it would be better for the supervisor to pass the container information that it manages to the worker agent when invoking, and then the agent would then use that container info in its context.

ranweiler: So, even if we don't want to do a full "supervisor-managed container registry" right now, would it still be better to pass just the `setup_dir` explicitly to the agent binary?

src/agent/onefuzz-agent/src/tasks/config.rs
chkeita: I was thinking about keeping the backward compatibility with the way we currently run the tasks by not requiring an additional parameter. 
I still have to make the default value being the parent directory of target_exe

src/agent/onefuzz-supervisor/src/worker.rs
chkeita: how would you suggest that we pass the setup_dir to the individual tasks without modifying the task config? I get the idea of sending the info from the supervisor_agent to the worker_agent via a parameter. But that info needs to eventually get to the tasks from the config.

I am ok to switch to using a paramter to use a paramter to pass the setup_dir but it will sill be part of the config.

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -72,7 +79,7 @@ impl<'a> Tester<'a> {
     ) -> Result<Option<Crash>> {
         const IGNORE_FIRST_CHANCE_EXCEPTIONS: bool = true;
         let report = input_tester::crash_detector::test_process(
-            self.exe_path,
+            self.exe_path.clone(),
```
bmc-msft, line 82: Why is this cloning now?

src/agent/onefuzz-agent/src/main.rs
```diff
@@ -39,6 +39,12 @@ fn main() -> Result<()> {
                 .short("c")
                 .takes_value(true),
         )
+        .arg(
+            Arg::with_name("setup_dir")
+                .long("setup_dir")
+                .short("sd")
```
ranweiler, line 45: FYI, only the `s` will get used, and `-sd` will not work as expected. See the `NOTE` section [here](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.short).
> only the first non `-` character will be used as the short version

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -79,9 +81,16 @@ pub enum Config {
...
 
 impl Config {
-    pub fn from_file(path: impl AsRef<Path>) -> Result<Self> {
-        let json = std::fs::read_to_string(path)?;
-        Ok(serde_json::from_str(&json)?)
+    pub fn from_file(path: impl AsRef<Path>, setup_dir: Option<impl AsRef<Path>>) -> Result<Self> {
+        let file = std::fs::File::open(path)?;
+        let reader = std::io::BufReader::new(file);
+        let mut json_config: serde_json::Value = serde_json::from_reader(reader)?;
+        if let Some(setup_dir) = setup_dir {
```
bmc-msft, line 88: A few things.  This appears to be optional, but the CommonConfig entry is not.

Since this uses the command line arguments, instead of inserting into the json prior to deserialization, why not make setup_dir optional, de-serialize the json as is, then set the value to the object?

This could be replaced with:

```
let mut config: Self = serde_json::from_value(json_config)?;
config.setup_dir = setup_dir;
```

src/agent/onefuzz-agent/src/tasks/config.rs
bmc-msft: Yes.  In the existing debugging mode as well as the upcoming `local fuzzing` mode, it means having to specify an additional argument that is often not useful locally.

src/agent/onefuzz-agent/src/tasks/config.rs
chkeita: The idea is to fail fast if the setup_dir is not specified in the config or the command line. I wanted to avoid the situation where the setup_dir not specified and we only know it while the task is running.

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -79,9 +81,16 @@ pub enum Config {
 }
 
 impl Config {
-    pub fn from_file(path: impl AsRef<Path>) -> Result<Self> {
-        let json = std::fs::read_to_string(path)?;
-        Ok(serde_json::from_str(&json)?)
+    pub fn from_file(path: impl AsRef<Path>, setup_dir: Option<impl AsRef<Path>>) -> Result<Self> {
+        let file = std::fs::File::open(path)?;
+        let reader = std::io::BufReader::new(file);
```
ranweiler, line 86: I think it's both simpler and probably more efficient to just use `fs::read_to_string()` or `fs::read()` here. Those functions are meant for one bulk read from a file, pre-allocate a buffer based on file size, &c. Was there a specific reason you changed this?

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -79,9 +81,16 @@ pub enum Config {
...
+        let mut json_config: serde_json::Value = serde_json::from_reader(reader)?;
+        if let Some(setup_dir) = setup_dir {
+            json_config["setup_dir"] =
+                serde_json::Value::String(setup_dir.as_ref().to_string_lossy().into());
+        }
+
+        Ok(serde_json::from_value(json_config)?)
```
ranweiler, line 93: Nice use of `from_value()`.

src/agent/onefuzz-supervisor/src/worker/tests.rs
```diff
@@ -64,7 +64,9 @@ impl IWorkerRunner for RunnerDouble {
 async fn test_ready_run() {
     let mut runner = Fixture.runner(Fixture.child_running());
     let state = State {
-        ctx: Ready,
+        ctx: Ready {
+            setup_dir: PathBuf::from(""),
```
ranweiler, line 68: FYI, this can be achieved by either `PathBuf::new()` and `PathBuf::default()`.

That said, when testing, I suggest we use some fake but nonempty value.

src/agent/onefuzz-agent/src/tasks/config.rs
ranweiler: IIRC, this is a convenience for local fuzzing users, so they don't _have_ to pass the `setup_dir` as an arg. On a remote node, the `setup_dir` argument to this function will always be present.

I'm fine with removing this convenience, and making this `setup_dir` function arg non-optional too.

My thinking is that in the future, we want to move to a more unified description of all FS-local contextual info (local paths, abstract work queue "provider", &c), and that will subsume any current pain in the local use case.

ranweiler: Whatever we end up with here, I do think we should add  a doc comment for this `from_file()` ctor, explaining what's going on with the "stomped in" value.

ranweiler (approved):

src/agent/onefuzz/src/input_tester.rs
chkeita: removed

src/agent/onefuzz-agent/src/tasks/config.rs
chkeita: changed this be non optional

src/agent/onefuzz-agent/src/debug/generic_crash_report.rs
```diff
@@ -29,6 +29,7 @@ async fn run_impl(input: String, config: Config) -> Result<()> {
 
 pub fn run(args: &clap::ArgMatches) -> Result<()> {
     let target_exe = value_t!(args, "target_exe", PathBuf)?;
+    let setup_dir = value_t!(args, "setup_dir", PathBuf)?;
```
bmc-msft, line 32: `setup_dir` needs to be added to the arg builder for this helper.

src/agent/onefuzz-agent/src/debug/libfuzzer_coverage.rs
```diff
@@ -38,6 +38,7 @@ async fn run_impl(input: String, config: Config) -> Result<()> {
 
 pub fn run(args: &clap::ArgMatches) -> Result<()> {
     let target_exe = value_t!(args, "target_exe", PathBuf)?;
+    let setup_dir = value_t!(args, "setup_dir", PathBuf)?;
```
bmc-msft, line 41: `setup_dir` needs to be added to the arg builder for this helper.

src/agent/onefuzz-agent/src/debug/libfuzzer_crash_report.rs
```diff
@@ -30,6 +30,7 @@ async fn run_impl(input: String, config: Config) -> Result<()> {
 
 pub fn run(args: &clap::ArgMatches) -> Result<()> {
     let target_exe = value_t!(args, "target_exe", PathBuf)?;
+    let setup_dir = value_t!(args, "setup_dir", PathBuf)?;
```
bmc-msft, line 33: `setup_dir` needs to be added to the arg builder for this helper.

src/agent/onefuzz-agent/src/debug/libfuzzer_merge.rs
```diff
@@ -16,6 +16,7 @@ use uuid::Uuid;
 
 pub fn run(args: &clap::ArgMatches) -> Result<()> {
     let target_exe = value_t!(args, "target_exe", PathBuf)?;
+    let setup_dir = value_t!(args, "setup_dir", PathBuf)?;
```
bmc-msft, line 19: `setup_dir` needs to be added to the arg builder for this helper.

src/agent/onefuzz-agent/src/debug/libfuzzer_fuzz.rs
```diff
@@ -25,6 +25,7 @@ pub fn run(args: &clap::ArgMatches) -> Result<()> {
     let crashes_dir = value_t!(args, "crashes_dir", String)?;
     let inputs_dir = value_t!(args, "inputs_dir", String)?;
     let target_exe = value_t!(args, "target_exe", PathBuf)?;
+    let setup_dir = value_t!(args, "setup_dir", PathBuf)?;
```
bmc-msft, line 28: `setup_dir` needs to be added to the arg builder for this helper.

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -79,9 +81,16 @@ pub enum Config {
...
-    pub fn from_file(path: impl AsRef<Path>) -> Result<Self> {
+    pub fn from_file(path: impl AsRef<Path>, setup_dir: Option<impl AsRef<Path>>) -> Result<Self> {
         let json = std::fs::read_to_string(path)?;
-        Ok(serde_json::from_str(&json)?)
+        let mut json_config: serde_json::Value = serde_json::from_str(&json)?;
+        // override the setup_dir in the config file with the parameter value if specified
+        if let Some(setup_dir) = setup_dir {
+            json_config["setup_dir"] =
+                serde_json::Value::String(setup_dir.as_ref().to_string_lossy().into());
+        }
```
bmc-msft, line 91: My previous comment regarding setup_dir being optional had an additional component.

This takes a PathBuf and bounces in and out of json.  We shouldn't do that.  Instead, the structure should be parsed as is from file, then the setup_dir set.

src/agent/onefuzz-agent/src/tasks/coverage/libfuzzer_coverage.rs
```diff
@@ -104,6 +104,7 @@ impl CoverageTask {
                 &self.config.target_exe,
                 &self.config.target_options,
                 &self.config.target_env,
+                self.config.common.setup_dir.clone(),
```
bmc-msft, line 107: Why does this need to .clone?

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -72,6 +72,7 @@ impl LibFuzzerFuzzTask {
                 &self.config.target_exe,
                 &self.config.target_options,
                 &self.config.target_env,
+                self.config.common.setup_dir.clone(),
```
bmc-msft, line 75: Why does this need to clone?

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -156,6 +157,7 @@ impl LibFuzzerFuzzTask {
             &self.config.target_exe,
             &self.config.target_options,
             &self.config.target_env,
+            self.config.common.setup_dir.clone(),
```
bmc-msft, line 160: Why does this need to clone?

src/agent/onefuzz-agent/src/tasks/merge/libfuzzer_merge.rs
```diff
@@ -52,6 +52,7 @@ pub async fn spawn(config: Arc<Config>) -> Result<()> {
             &config.target_exe,
             &config.target_options,
             &config.target_env,
+            config.common.setup_dir.clone(),
```
bmc-msft, line 55: Why does this need to clone?

src/agent/onefuzz-agent/src/tasks/merge/libfuzzer_merge.rs
```diff
@@ -162,6 +163,7 @@ pub async fn merge_inputs(
         &config.target_exe,
         &config.target_options,
         &config.target_env,
+        config.common.setup_dir.clone(),
```
bmc-msft, line 166: Why does this need to clone?

bmc-msft (changes requested):
See comments

src/agent/onefuzz-agent/src/tasks/merge/libfuzzer_merge.rs
chkeita: these clone were there because i had changed the type of the parameter. I reverted those changes

PR #418, opened 2021-01-07
compare containers rather than SAS urls when building worksets

bmc-msft:
By comparing container names rather than SAS urls, this removes a race condition that prevented co-locatable tasks from being co-located.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/tasks/scheduler.py
+++ b/src/api-service/__app__/onefuzzlib/tasks/scheduler.py
@@ -9,6 +9,7 @@ from uuid import UUID, uuid4
 
 from onefuzztypes.enums import OS, PoolState, TaskState
 from onefuzztypes.models import WorkSet, WorkUnit
+from onefuzztypes.primitives import Container
 from pydantic import BaseModel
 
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/tasks/scheduler.py
```diff
@@ -102,7 +103,7 @@ def bucket_tasks(tasks: List[Task]) -> Dict[Tuple, List[Task]]:
 class BucketConfig(BaseModel):
     count: int
     reboot: bool
-    setup_url: str
+    setup_container: Container
```
ranweiler, line 106: Does `Container` include the storage account?

ranweiler: More to the point: do we lose important information by dropping the storage account, which is part of the SAS URL?

ranweiler: Out-of-band answer: `Container` does not include the storage account, but we don't need this information, and we still correctly the recover the SAS URL when building the work set.

ranweiler (approved):

PR #419, opened 2021-01-08
Clear node messages on deletion

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/api-service/__app__/node/__init__.py
```diff
@@ -73,6 +73,8 @@ def delete(req: func.HttpRequest) -> func.HttpResponse:
             context=request.machine_id,
         )
 
+    NodeMessage.clear_messages(node.agent_id)
```
ranweiler, line 76: Could new messages also arrive in the meantime?

Is there a time later on in the deletion lifecycle where we should clear the messages, instead?

bmc-msft: IMO, these should probably be cleared at the Node.delete() level, such that it's called when the service finalizes the delete.

https://github.com/microsoft/onefuzz/blob/46e84545696b1c69466ebaa1941648d085b98c95/src/api-service/__app__/onefuzzlib/pools.py#L318-L320

bmc-msft (changes requested):

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -258,10 +258,11 @@ def is_outdated(self) -> bool:
         return self.version != __version__
 
     def send_message(self, message: NodeCommand) -> None:
-        NodeMessage(
-            agent_id=self.machine_id,
-            message=message,
-        ).save()
+        if not self.delete_requested:
```
bmc-msft, line 261: We still need to send the messages here.  In the case we're halting the node, we need to tell it to stop processing any tasks immediately and report back that it has done so.

bmc-msft (changes requested):

src/api-service/__app__/node/__init__.py
chkeita: moved to Node.delete()

bmc-msft (approved):

PR #422, opened 2021-01-08
Fix condition for triggering new unique report event

chkeita:


src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -72,7 +72,7 @@ async fn upload_deduped(report: &CrashReport, container: &BlobContainerUrl) -> R
         .header("If-None-Match", "*")
         .send_retry_default()
         .await?;
-    if result.status() != StatusCode::NOT_MODIFIED {
```
chkeita, line 75: The expected return code here would have been a [409](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes) with a "BlobAlreadyExists" in the response headers. 
But i just just simplified the logic to just check that the blob was created instead.

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -72,7 +72,7 @@ async fn upload_deduped(report: &CrashReport, container: &BlobContainerUrl) -> R
         .header("If-None-Match", "*")
```
ranweiler, line 72: I should've done this originally, but in the comment above this, can we add a link to these docs: https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
ranweiler: Hmm, I think that's inconsistent with the docs I link to above. Or at least, it's not clear what behavior to expect, given our API version.

This raises a question: are we using a constant REST API version? It looks like we aren't.

With a SAS URL, if the API version is not otherwise specified, then the API version used is derived from [the version used in signature](
https://docs.microsoft.com/en-us/rest/api/storageservices/versioning-for-the-azure-storage-services#requests-authorized-with-a-shared-access-signature).

It looks like we don't set the API version in the `BlobClient` methods, and we visibly are not doing so here, so I'm wondering if the API changed and this code stopped working correctly for that reason. (Either way, your new check is better).

ranweiler (approved):

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
ranweiler: Opened #424 to capture the above concerns.

PR #427, opened 2021-01-11
update python prereqs

bmc-msft:
Updates the following libraries in the service:
* azure-core
* azure-functions
* azure-identity
* azure-keyvault-keys
* azure-keyvault-secrets
* azure-mgmt-compute
* azure-mgmt-core
* azure-mgmt-loganalytics
* azure-mgmt-network
* azure-mgmt-resource
* azure-mgmt-storage
* azure-mgmt-subscription
* azure-storage-blob
* azure-storage-queue
* pydantic
* requests
* jsonpatch

Removes the following libraries in the service:
* azure-cli-core
* azure-cli-nspkg
* azure-mgmt-cosmosdb
* azure-servicebus

Updates the following libraries in the CLI:
* requests
* semver
* asciimatics
* pydantic
* tenacity

Updates the following libraries in onefuzztypes:
* pydantic

The primary "legacy" libraries are [azure-graphrbac](https://pypi.org/project/azure-graphrbac/) and azure-cosmosdb-table.  The former has not been updated to use azure-identity yet. The later is being rewritten as [azure-data-tables](https://pypi.org/project/azure-data-tables/), but is still in early beta.


src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -19,47 +28,37 @@
...
-
-
-def get_service_by_container(
-    container: Container, storage_type: StorageType
-) -> Optional[BlockBlobService]:
-    account = get_account_by_container(container, storage_type)
-    if account is None:
-        return None
-    service = get_blob_service(account)
+    account_url = f"https://{account_name}.blob.core.windows.net/"
```
ranweiler, line 39: Is the `blob.core.windows.net` part exposed as a default via the SDK, in any way?

src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -152,49 +152,58 @@ def delete_container(container: Container, storage_type: StorageType) -> bool:
...
+            tag=tag,
+        ),
+        expiry=datetime.datetime.utcnow() + datetime.timedelta(days=30),
     )
 
-    url = service.make_container_url(container, sas_token=sas_token)
-    url = url.replace("?restype=container&", "?")
-    return str(url)
+    with_sas = ContainerClient(
+        f"https://{account_name}.blob.core.windows.net/",
```
ranweiler, line 184: It may make sense to provide a constant for the default Azure blob service domain, or maybe a function that translates an account name to the default URL.

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -23,24 +22,16 @@
 
 @cached
 def get_msi() -> MSIAuthentication:
+    allow_more_workers()
```
ranweiler, line 25: Why do we do this here, too?

ranweiler: If we have some shared setup we want to invoke in multiple places, maybe it should live in a single function, with some documentation explaining when we must invoke it, and why.

src/api-service/__app__/onefuzzlib/azure/monitor.py
```diff
@@ -4,17 +4,17 @@
...
 
-from .creds import get_base_resource_group, mgmt_client_factory
+from .creds import get_base_resource_group, get_identity, get_subscription
 
 
-@cached(ttl=60)
-def get_montior_client() -> Any:
-    return mgmt_client_factory(LogAnalyticsManagementClient)
+@cached
+def get_montior_client() -> LogAnalyticsManagementClient:
```
ranweiler, line 16: Not new, but `monitor` is misspelled.

src/api-service/__app__/onefuzzlib/azure/monitor.py
```diff
@@ -4,17 +4,17 @@
...
 from memoization import cached
 
-from .creds import get_base_resource_group, mgmt_client_factory
+from .creds import get_base_resource_group, get_identity, get_subscription
 
 
-@cached(ttl=60)
-def get_montior_client() -> Any:
-    return mgmt_client_factory(LogAnalyticsManagementClient)
+@cached
```
ranweiler, line 15: Why the TTL change? Why here and not below?

src/api-service/__app__/onefuzzlib/azure/vmss.py
```diff
@@ -35,39 +37,45 @@ def list_vmss(name: UUID) -> Optional[List[str]]:
...
     resource_group = get_base_resource_group()
-    compute_client = mgmt_client_factory(ComputeManagementClient)
+    compute_client = get_compute_client()
     try:
-        compute_client.virtual_machine_scale_sets.delete(resource_group, str(name))
-    except CloudError as err:
+        compute_client.virtual_machine_scale_sets.begin_delete(
+            resource_group, str(name)
+        )
+    except (ResourceNotFoundError, CloudError) as err:
```
ranweiler, line 53: Review TODO: what since we now catch these errors and map them to a value, what do we do with the resulting `bool`?

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -1070,8 +1081,11 @@ def halt(self) -> None:
         vmss = get_vmss(self.scaleset_id)
         if vmss:
             logging.info("scaleset deleting: %s", self.scaleset_id)
-            delete_vmss(self.scaleset_id)
-            self.save()
+            if delete_vmss(self.scaleset_id):
+                logging.info("scaleset deleted: %s", self.scaleset_id)
```
ranweiler, line 1085: Now, `delete_vmss` returns `True` if there is a `CloudError` or `ResourceError`, and internally logs at error level.

1. Doesn't that make this `info` log message incorrect?
2. Why do we `delete()`? That makes sense for a `ResourceNotFoundError`, but it seems like we also can reach this branch due to a `CloudError`, in which case the underlying VMSS may still exist, but we'd delete our `Scaleset` metadata which tracks it, leaving the Azure resource dangling. Am I misreading?

src/api-service/__app__/onefuzzlib/azure/vmss.py
ranweiler: Based on review of at least one of the use sites for this function, we may want to distinguish the outcomes, or document why `CloudError` and `ResourceNotFoundError` should be treated as equivalent.

src/pytypes/onefuzztypes/enums.py
```diff
@@ -85,10 +85,8 @@ class TaskFeature(Enum):
 class ContainerPermission(Enum):
     Read = "Read"
     Write = "Write"
-    Create = "Create"
```
ranweiler, line 88: Were these just removed and subsumed by `Write`? Any docs we can link to?

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: I looked and didn't see a viable method.  Note, this follows the SDK's examples & tests:

https://github.com/Azure/azure-sdk-for-python/blob/f8be630ec739324fbc540083f4aa3c4e71869f3f/sdk/storage/azure-storage-blob/samples/blob_samples_authentication.py#L31-L33

https://github.com/Azure/azure-sdk-for-python/blob/f8be630ec739324fbc540083f4aa3c4e71869f3f/sdk/storage/azure-storage-blob/tests/test_container.py#L1044-L1046

src/api-service/__app__/onefuzzlib/azure/creds.py
bmc-msft: It's our ugly hot-patching.  We're down to only a single instance of MSIAuthentication which will be going away soon.

src/api-service/__app__/onefuzzlib/azure/monitor.py
bmc-msft: fixed

src/api-service/__app__/onefuzzlib/azure/monitor.py
bmc-msft: We cache the client forever.  It handles renewing auth tokens automatically.

However, the settings for Azure Monitor could change, so we keep those such that they expire ever so often.

src/pytypes/onefuzztypes/enums.py
bmc-msft: Create doesn't exist as a container permission exposed by the SDK any longer.  

https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.containersaspermissions?view=azure-python

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Split the exceptions in half.  If it's ResourceNotFoundError, then treat it as it's already deleted, which was my original intent.

src/api-service/__app__/onefuzzlib/azure/vmss.py
bmc-msft: Addressed (see response in other comment)

src/api-service/__app__/onefuzzlib/pools.py
ranweiler: I think the business logic now correct, but we still want to update the logging here.

How about in the `True` branch, we log something like `"underlying scaleset resource already deleted"`, and `"scaleset deleted"` in the other branch (since that's the case where we actually performed a deletion)?

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: added

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Having rethought this, we can just always attempt the delete, rather than `get` then `delete`, saving a round trip to Azure.

If delete_vmss returns true, it's deleted.  otherwise, we save the scaleset which will cause the delete to get retried alter on.

ranweiler: Great. I'd only also suggest you note the expectation that `save()` assumes "we'll retry deletion failed", since that's important but not locally obvious in the source code.

ranweiler (approved):

chkeita (approved):

Issue #370, opened 2020-12-04
Capture stdout/stderr of the applications under test


bmc-msft:
Right now, we do not capture the stdout or stderr or the applications under test.  This makes debugging issues with the application difficult. 

AB#39994716

bmc-msft:
On linux, `pete` 0.4.0 moves away from a private `Command` to using `std::process::Command`, which means we can now use `std::process::Stdio`.

The change makes a few other things, like the custom kill on drop wrapper less ergonomic.

bmc-msft:
@ranweiler Can you take a look at this?

chkeita:
addressed in #400

PR #429, opened 2021-01-12
Update Linux tracer version

ranweiler:
Update `pete` to 0.4, which enables and requires us to use `std::process::Child` for spawning tracees.

Toward #370.

### DIFF
--- a/src/agent/coverage/src/block/linux.rs
+++ b/src/agent/coverage/src/block/linux.rs
@@ -4,12 +4,13 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
+use std::process::Command;
 
 use anyhow::{format_err, Result};
 use iced_x86::{Decoder, DecoderOptions, Instruction};
 use object::endian::LittleEndian as LE;
 use object::{read::elf, Object, ObjectSection, ObjectSegment, ObjectSymbol};
-use pete::{Command, Ptracer, Restart, Signal, Stop, Tracee};
+use pete::{Ptracer, Restart, Signal, Stop, Tracee};
 use procfs::process::{MMapPath, MemoryMap, Process};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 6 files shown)


src/agent/coverage/src/block/linux.rs
```diff
@@ -484,3 +474,15 @@ fn branch_target(inst: &iced_x86::Instruction) -> Option<(u64, bool)> {
         | FlowControl::XbeginXabortXend => None,
     }
 }
+
+fn continue_to_init_execve(tracer: &mut Ptracer) -> Result<Tracee> {
```
ranweiler, line 478: This is copy-pasted in our `triage` crate. We could factor it out, but upstream should probably provide a similar method for the common `spawn()` use case.

src/agent/onefuzz/Cargo.toml
```diff
@@ -45,7 +45,7 @@ input-tester = { path = "../input-tester" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 cpp_demangle = "0.3"
-nix = "0.17"
-pete = "0.3"
+nix = "0.19"
```
ranweiler, line 48: Required due to `pete` requiring at least `0.19`.

src/agent/onefuzz/src/triage.rs
```diff
@@ -375,3 +387,15 @@ mod se {
...
+fn continue_to_init_execve(tracer: &mut Ptracer) -> Result<Tracee> {
+    while let Some(tracee) = tracer.wait()? {
+        if let Stop::SyscallExitStop(..) = &tracee.stop {
+            return Ok(tracee);
+        }
+
+        tracer.restart(tracee, Restart::Continue)?;
+    }
+
+    anyhow::bail!("did not see initial execve() in tracee while recording coverage");
```
ranweiler, line 400: Fix this error message.

src/agent/onefuzz/src/triage.rs
```diff
@@ -23,15 +25,25 @@ pub struct TriageCommand {
 }
 impl TriageCommand {
     pub fn new(argv: Vec<String>, env: HashMap<String, String>) -> Result<Self> {
```
bmc-msft, line 27: Since the callers refer to the program as a PathBuf, can this be separated out, rather than converting it to a String in a vec?

ranweiler: Yep, we can probably just pass an already-constructed `Command`, and invoke `cmd.envs(env)` in the caller.

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -110,10 +110,12 @@ impl<'a> Tester<'a> {
...
     async fn test_input_debugger(
         &self,
-        mut argv: Vec<String>,
+        args: Vec<String>,
         env: HashMap<String, String>,
     ) -> Result<Option<Crash>> {
-        argv.insert(0, self.exe_path.display().to_string());
+        let mut cmd = std::process::Command::new(self.exe_path);
+        cmd.args(args);
+        cmd.envs(&env);
```
ranweiler, line 118: Now we set the tracee args here, instead of in the `TriageCommand` ctor.

bmc-msft (approved):

ranweiler:
Manually verified that the generic Linux crash reports are correct after the tracer changes.

PR #431, opened 2021-01-14
enable sccache & incremental builds for non-release builds

bmc-msft:


.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         path: |
           sccache
           src/agent/target
-        key: agent-${{ runner.os }}-${{ hashFiles('src/agent/Cargo.lock') }}
+        key: agent-${{ runner.os }}-${{ hashFiles('src/agent/Cargo.lock') }}-2021-01-14-01
```
ranweiler, line 35: Is there something important about this choice of date? Do we need to preserve it?

src/ci/agent.sh
```diff
@@ -9,7 +9,18 @@ exists() {
     [ -e "$1" ]
 }
 
-#export RUSTC_WRAPPER=$(which sccache)
+if sccache --help; then
+    export RUSTC_WRAPPER=$(which sccache)
+fi
+
+# only set CARGO_INCREMENTAL on non-release builds
```
ranweiler, line 16: Can we document why, maybe with a helpful link for future readers?

src/ci/agent.sh
```diff
@@ -9,7 +9,18 @@ exists() {
...
+if [ "${GITHUB_REF}" != "" ]; then
+    TAG_VERSION=${GITHUB_REF#refs/tags/}
+    if [ ${TAG_VERSION} == ${GITHUB_REF} ]; then
+        export CARGO_INCREMENTAL=1
+    fi
+fi
+
 #sccache --start-server
```
ranweiler, line 24: Should we drop this commented code?

ranweiler (approved):

.github/workflows/ci.yml
bmc-msft: It's not that we need to preserve it, but we need to differentiate it from previous cache keys to get the benefit.

bmc-msft: Unfortunately, there isn't an easy way to invalidate a cache entry.  If you've not used a cache entry in 7 days, then it will automatically get removed.  However, that's either read or write.  :/  There isn't an API to evict the cache early.

PR #433, opened 2021-01-14
regression template

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-agent/src/tasks/regression/generic.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    pub async fn get_crash_result(
+        &self,
+        input: impl AsRef<Path>,
+        input_url: Option<Url>,
+        is_asan: bool,
+    ) -> Result<CrashTestResult> {
+        if is_asan {
+            libfuzzer_report::test_input(
```
bmc-msft, line 170: The CrashReport having an `asan` entry doesn't identify if this is a libfuzzer target or not.  As an example, the radamsa template can optionally support ASAN instrumented binaries and checking for ASAN log files.

The crash reports don't have any indication as to the type of target.

src/api-service/__app__/onefuzzlib/job_templates/defaults/libfuzzer.py
```diff
@@ -330,6 +330,17 @@
                 ),
             ],
         ),
+        UserField(
```
bmc-msft, line 333: I don't think this is what you meant.

This modifies the first task in the libfuzzer template list, which is libfuzzer_fuzz, and adds the value `report_list`.  I thought `report_list` was only used during the new regression task.

bmc-msft: I think this will likely need to be it's on JobTemplate

bmc-msft (changes requested):
I think this should be two tasks, one for libfuzzer crash reports and one for generic crash reports.

The job templates don't support branching in the argument validation.  This makes doing something like asking the user to specify the type of target / crash report odd.  In the libfuzzer case, use_debugger and asan files don't make sense.

src/agent/onefuzz-agent/src/tasks/regression/libfuzzer.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub input_reports: Option<SyncedDir>,
+    pub crashes: Option<SyncedDir>,
+    pub report_list: Vec<String>,
+
+    pub no_repro: SyncedDir,
+    pub reports: SyncedDir,
+
+    pub target_timeout: Option<u64>,
+
+    #[serde(default)]
```
bmc-msft, line 43: libfuzzer regression doesn't need check_asan_log or check_debugger.  However, it does need check_fuzzer_help.

src/agent/onefuzz-agent/src/debug/generic_regression.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+use tokio::runtime::Runtime;
+use url::Url;
+use uuid::Uuid;
+
+async fn run_impl(input: String, config: Config) -> Result<()> {
```
bmc-msft, line 21: Where does `Config` come from?  I don't see it on an import

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -139,9 +146,57 @@ impl CrashTestResult {
...
+                        &RegressionReport {
+                            crash_result: Self::CrashReport(report),
+                            original_report,
+                        },
+                        &name,
+                        reports,
+                    )
+                    .await?
+                    {
+                        event!(new_unique_report; EventData::Path = name);
```
bmc-msft, line 172: Is new_unique_report the correct event here?

bmc-msft: I think this and new_unable_to_reproduce might not be the right result here.  These are not new crashes, these are reproducing the existing crashes

src/agent/onefuzz-agent/src/debug/generic_regression.rs
bmc-msft: Oh!  These are not included in `src/debug/mod.rs` or in `src/debug/cmd`.  Adding them exposed some compile errors

chkeita: removed the file. will add the debug/local capability in a separate PR

src/agent/onefuzz-agent/src/tasks/regression/libfuzzer.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub target_options: Vec<String>,
+
+    #[serde(default)]
+    pub target_env: HashMap<String, String>,
+
+    pub inputs: Option<SyncedDir>,
+
+    pub input_reports: Option<SyncedDir>,
+    pub crashes: Option<SyncedDir>,
+    pub report_list: Vec<String>,
```
ranweiler, line 36: Should this this be annotated with `#[serde(default)]`, too?

src/agent/onefuzz-agent/src/tasks/regression/libfuzzer.rs
```diff
@@ -0,0 +1,113 @@
...
+#[async_trait]
+impl RegressionHandler for LibFuzzerRegressionTask {
+    async fn get_crash_result(
+        &self,
+        input: PathBuf,
+        input_url: Option<Url>,
+    ) -> Result<CrashTestResult> {
+        libfuzzer_report::test_input(
+            input_url,
+            input.as_ref(),
```
ranweiler, line 65: You can just do `&input` here, I think.

src/agent/onefuzz-agent/src/tasks/regression/generic.rs
```diff
@@ -0,0 +1,116 @@
...
+#[async_trait]
+impl RegressionHandler for GenericRegressionTask {
+    async fn get_crash_result(
+        &self,
+        input: PathBuf,
+        input_url: Option<Url>,
+    ) -> Result<CrashTestResult> {
+        generic::test_input(
+            input_url,
+            input.as_ref(),
```
ranweiler, line 66:
```suggestion
            &input,
```

src/agent/onefuzz-agent/src/tasks/regression/libfuzzer.rs
ranweiler:
```suggestion
            &input,
```

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
```diff
@@ -0,0 +1,117 @@
...
+    ) -> Result<()>;
+}
+
+pub async fn run(
+    heartbeat_client: Option<TaskHeartbeatClient>,
+    input_reports: &Option<SyncedDir>,
+    crashes: &Option<SyncedDir>,
+    inputs: &Option<SyncedDir>,
+    report_list: &[String],
+    handler: &impl RegressionHandler,
```
ranweiler, line 35: I can imagine wanting this to be `&mut` in some cases. However, I think it is fine to use a shared ref for now, and we make it a mutable ref later, if actually needed.

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::tasks::{
+    heartbeat::*,
```
ranweiler, line 5: I suggest we avoid wildcard imports at the top level.

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
```diff
@@ -0,0 +1,117 @@
...
+    let mut report_files = tokio::fs::read_dir(&input_reports.path).await?;
+    while let Some(file) = report_files.next_entry().await? {
+        heartbeat_client.alive();
+        let crash_report_str = std::fs::read_to_string(file.path())?;
+        let crash_report: CrashReport = serde_json::from_str(&crash_report_str)?;
+        let input_url = crash_report
+            .input_blob
+            .clone()
+            .map(|b| b.name)
+            .and_then(|crash_name| crashes.url.clone().map(|u| u.blob(crash_name).url()));
```
ranweiler, line 107: I think these nested combinator calls can get a little hard to read. Consider breaking this whole chain of functions into a helper function whose name makes it obvious what you're doing. You could then use `?` for early returns, in some cases.

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
```diff
@@ -0,0 +1,117 @@
...
+        original_report: Option<CrashReport>,
+    ) -> Result<()>;
+}
+
+pub async fn run(
+    heartbeat_client: Option<TaskHeartbeatClient>,
+    input_reports: &Option<SyncedDir>,
+    crashes: &Option<SyncedDir>,
+    inputs: &Option<SyncedDir>,
+    report_list: &[String],
```
ranweiler, line 34: I think we need to document what a `report_list` is, and anything we're assuming about it.

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -139,9 +149,57 @@ impl CrashTestResult {
...
+                    if upload_or_save_local(
+                        &RegressionReport {
+                            crash_result: Self::CrashReport(report),
+                            original_report,
+                        },
+                        &name,
+                        reports,
+                    )
+                    .await?
+                    {
```
ranweiler, line 174: Is this block necessary for lifetime reasons?

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -139,9 +149,57 @@ impl CrashTestResult {
         }
         Ok(())
     }
+
+    pub async fn save_regression(
```
ranweiler, line 153: Could we add doc comments to all of these new public methods and types? It would help to know the intention at a glance, including where we expect things to get saved.

ranweiler: (and what's supposed to happen for the repro/no repro)

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+        crash_report_timeout: Optional[int] = None,
+        debug: Optional[List[TaskDebugFlag]] = None,
+        check_retry_count: Optional[int] = None,
+        check_fuzzer_help: bool = True,
+        fail_on_repro: bool = False,
+    ) -> Optional[Job]:
+
+        if not ((crashes and input_reports) or inputs):
+            self.logger.error(
+                "please specify either the 'crash' and 'input_reports' parameters or the inputs parameter"
```
ranweiler, line 155:
```suggestion
                "please specify either the 'crash' and 'input_reports' parameters or the 'inputs' parameter"
```

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+
+        if dryrun:
+            return None
```
ranweiler, line 159: Do we really have to stop here, or can we do more while passing around the `dryrun` param?

src/pytypes/tests/test_models.py
```diff
@@ -4,6 +4,7 @@
 # Licensed under the MIT License.
 
 import unittest
+from onefuzztypes.primitives import PoolName, Region
```
ranweiler, line 7: Is this used?

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+            check_retry_count=check_retry_count,
+            debug=debug,
+            check_fuzzer_help=check_fuzzer_help,
+            report_list=report_list,
+        )
+        helper.wait_for_stopping = fail_on_repro
+
+        self.logger.info("done creating tasks")
+        helper.wait()
+        if fail_on_repro:
```
bmc-msft, line 238: what does this do in the case that the task fails?

It would be useful to differentiate task failure from report files not being created.

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+            build,
+            duration,
+            pool_name=pool_name,
+            target_exe=target_exe,
+        )
+
+        helper.add_tags(tags)
+        if crashes:
+            if self.onefuzz.containers.get(crashes):
+                helper.define_containers(ContainerType.unique_inputs)
```
bmc-msft, line 177: This doesn't use the Crashes container as provided on the CLI.  Instead, this uses the name-spaced GUID generation to create a new container `unique_inputs` container.  I don't think this is what you're intending here.

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+            else:
+                self.logger.error(f"invalid crash container {crashes}")
+
+        if input_reports:
+            if self.onefuzz.containers.get(input_reports):
+                helper.define_containers(ContainerType.input_reports)
```
bmc-msft, line 183: Similar to above, this doesn't use the `input_reports` container specified by the user on the CLI, rather it uses the name-spaced GUID generation to identify the `input_reports` container to use.

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+        dryrun: bool = False,
+        duration: int = 24,
+        report_list: Optional[List[str]] = None,
+        crash_report_timeout: Optional[int] = None,
+        debug: Optional[List[TaskDebugFlag]] = None,
+        check_retry_count: Optional[int] = None,
+        check_fuzzer_help: bool = True,
+        fail_on_repro: bool = False,
+    ) -> Optional[Job]:
+
```
bmc-msft, line 152: This would be extremely useful to add the pydoc documentation here, as `inputs` is conceptually very different than `crashes` + `input_reports`

see `AFL.basic` from `cli/onefuzz/templates/afl.py` for an example use to specify the arguments.

src/cli/onefuzz/templates/regression.py
```diff
@@ -0,0 +1,250 @@
...
+            else:
+                self.logger.error(f"invalid crash container {crashes}")
+
+        if input_reports:
+            if self.onefuzz.containers.get(input_reports):
+                helper.define_containers(ContainerType.input_reports)
+            else:
+                self.logger.error(f"invalid crash container {input_reports}")
+
+        if inputs:
```
bmc-msft, line 187: I wonder if we should make the container used for this more specific to the effort at hand, and maybe even unique to this instance of the task.

Here's my thought.  I have a target `foo` that I've been fuzzing for for a while.  A 3rd party sends me a potentially crashing input for `foo`.  I want to see if this new input is crashing.

As is, this will upload said new input into the container "unique_inputs+UUID(project + name)", which pollutes the unqiue_inputs container with this potentially crashing input.

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
chkeita: this is the code block for the if clause on line 165

bmc-msft: When I first read it, I wondered the same thing.  Perhaps making RegressionReport a named variable would make this easier to understand.

ranweiler: Oh wow, yeah. I totally misread that!

Stylistically, I suggest as a team that we try to avoid pattern matching again complex RHS expressions. It fares particularly badly when combined with `rustfmt`.

src/cli/onefuzz/templates/regression.py
bmc-msft: The dry run functionally has all but been phased out of the existing templates and is not part  of the managed templates.

At this point, it's primarily used for command validation.

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
chkeita: I replaced the body of the last 'and_then' with a code block that should be easier to read

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
chkeita: Can we customize the rustfmt instead ? may just increase the max line length. 
The idea changing the way we write our code because of a tool that is supposed to help not think about formatting seems weird to me.

src/agent/onefuzz-agent/src/tasks/regression/regression.rs
chkeita: I usually default immutable and introduce mutability only if necessary

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
ranweiler: > The idea changing the way we write our code because of a tool that is supposed to help not think about formatting seems weird to me.

I agree, we should avoid doing that! The deeper reason is: I often find judiciously-named temporaries to be helpful for keeping less stuff in my head at one time. This is especially true if the temporary is going to be used in a conditional or RHS of an `if let`/`match`. This varies from case to case, of course, and I don't suggest we should be rigid / make a strict rule about it.

src/cli/onefuzz/templates/regression.py
chkeita: added a case for when we detect a job error

src/cli/onefuzz/templates/regression.py
chkeita: updated the code to generate a unique name for the input container

src/cli/onefuzz/templates/__init__.py
```diff
@@ -233,9 +247,58 @@ def wait_on(
...
+
+        for container in new:
+            if new[container] > self.to_monitor[container]:
+                del self.to_monitor[container]
+        return (
+            not self.to_monitor,
+            "waiting for new files: %s" % ", ".join(self.to_monitor.keys()),
+            None,
+        )
+
```
bmc-msft, line 296: These are now implemented in `JobMonitor`, which is used below.

```suggestion
```

src/cli/onefuzz/templates/__init__.py
```diff
@@ -7,9 +7,10 @@
 import os
 import tempfile
 import zipfile
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
```
bmc-msft, line 10: With the removal of the code now used by `JobMonitor`, this is not needed.

```suggestion
from typing import Any, Dict, List, Optional
```

src/cli/onefuzz/templates/__init__.py
```diff
@@ -7,9 +7,10 @@
...
+from uuid import uuid4
 
-from onefuzztypes.enums import OS, ContainerType
+from onefuzztypes.enums import OS, ContainerType, TaskState
```
bmc-msft, line 13: With the removal of the code now used by `JobMonitor`, this is not needed.

```suggestion
from onefuzztypes.enums import OS, ContainerType
```

src/cli/onefuzz/job_templates/job_monitor.py
```diff
@@ -78,11 +78,21 @@ def has_files(self) -> Tuple[bool, str, Any]:
...
+            if x.state not in TaskState.shutting_down()
+        ]
+        return (not stopping, "waiting on: %s" % ", ".join(sorted(stopping)), None)
+
     def wait(
         self,
         *,
         wait_for_running: Optional[bool] = False,
         wait_for_files: Optional[Dict[Container, int]] = None,
+        wait_for_stopping: Optional[bool] = False,
```
bmc-msft, line 95: This is a useful feature that I imagine we'll want to expose in the managed templates as well.  However, I think this is fine to leave for a follow-on PR

bmc-msft:
This failed for me using the `--inputs` version:

```
$ 1f template regression libfuzzer regression-test regression-test 2 linux --inputs ./oft-crashes-7900f84c60a657db8ecfd658e1cedb18
INFO:onefuzz:creating regression task from template
WARNING:onefuzz:ignoring compatibility check as the CLI was installed from git.  api: 2.5.0-f72dac5049359e6ed2a14ac6825e8415f67bcaa5 cli: 0.0.0
INFO:onefuzz:creating job (runtime: 24 hours)
INFO:onefuzz:created job: ced6de34-22d5-409a-b175-8fa1086f72a5
INFO:onefuzz:using container: oft-readonly-inputs-b0d88106fda34cd98f5c18f7e229e022
INFO:onefuzz:using container: oft-setup-72cb8ae73cf4552485d96d1b57e9f888
INFO:onefuzz:using container: oft-reports-7900f84c60a657db8ecfd658e1cedb18
INFO:onefuzz:using container: oft-no-repro-7900f84c60a657db8ecfd658e1cedb18
INFO:onefuzz:uploading inputs: `./oft-crashes-7900f84c60a657db8ecfd658e1cedb18`
INFO:onefuzz:uploading target exe `fuzz.exe`
INFO:onefuzz:creating regression task
ERROR:cli:command failed: request did not succeed: HTTP 400 - {"code": 450, "errors": ["container type input_reports: expected Equal 1, got 0"]}
```

bmc-msft:
Per discussion, OBE by #664

PR #435, opened 2021-01-15
Add missing prerequisites to the Getting Started guide

### LINKED ISSUE #434: suggestions for getting started guide
The getting started guide could use some more detail (onefuzz/docs/getting-started.md). 

Add Microsoft.SignalRService to the list of resource providers 

I also needed openSSL and azure-functions-core-tools-3 to be installed to run the deployment script

bmc-msft:
Fixes #434

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -16,7 +16,7 @@ registered:
 
 ## Deploying an instance of Onefuzz
 
-Ensure you have python3 (at least 3.7), [Azure Functions Core Tools
+Ensure you have Python with `python --version` >= 3.7, [Azure Functions Core Tools
 v3](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local),
 and OpenSSL installed.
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -12,9 +12,14 @@ registered:
 - `Microsoft.EventGrid`
 - `Microsoft.Network`
 - `Microsoft.Compute`
+- `Microsoft.SignalRService`
 
 ## Deploying an instance of Onefuzz
 
+Ensure you have python3 (at least 3.7), [Azure Functions Core Tools
```
ranweiler, line 19: It looks like the the deploy scripts (like `deply.py`) have an interpreter of `#!/usr/bin/env python`, which makes this potentially misleading on older systems, given the instructions below. We want `python` to alias to `python3.7` or greater, not any `python2`.

How about something like:
```suggestion
Ensure you have Python with `python --version` >= 3.7, [Azure Functions Core Tools
```

ranweiler (approved):

PR #445, opened 2021-01-20
always set SignalR events at the end of handlers

bmc-msft:
Addresses an issue where events meant for SignalR do not get sent to the service when an App Service instance spins down before a timer event fires.


src/api-service/__app__/agent_commands/__init__.py
```diff
@@ -42,7 +43,13 @@ def delete(req: func.HttpRequest) -> func.HttpResponse:
...
 
 
-def main(req: func.HttpRequest) -> func.HttpResponse:
+def main(req: func.HttpRequest, dashboard: func.Out[str]) -> func.HttpResponse:
     methods = {"DELETE": delete, "GET": get}
     method = methods[req.method]
-    return call_if_agent(req, method)
+    result = call_if_agent(req, method)
+
+    events = get_events()
```
chkeita, line 51: can we move this logic inside call_if_agent and call_if_user

src/api-service/__app__/tasks/__init__.py
```diff
@@ -3,14 +3,16 @@
...
-# import logging
 import azure.functions as func
 from onefuzztypes.enums import ErrorCode, JobState
 from onefuzztypes.models import Error, TaskConfig
 from onefuzztypes.requests import TaskGet, TaskSearch
 from onefuzztypes.responses import BoolResult
 
 from ..onefuzzlib.endpoint_authorization import call_if_user
+
+# import logging
```
chkeita, line 14: delete

src/api-service/__app__/agent_commands/__init__.py
bmc-msft: There are a handful of things that I'd like to wrap up into a  streamlined endpoint method.

I'd like to have a method wrapper that handles the req.method->python func mapping, call_if_*, the dashboard, the parsing of the model, and the rendering of a mode.

However, my thought is to do that as a follow-on development effort.

bmc-msft: Having a standardized wrapper would make it easier to build the swagger specs for the API we've discussed.

chkeita (approved):

PR #447, opened 2021-01-20
Update the registration logic to print manual steps when adal authentication fails

chkeita:
Mitigate the deployment issue related to the conditional access policy.
The registration logic is updated to use the old rbac python library when possible. 
The deployment will print some manual step for operations that cannot be automated


src/deployment/registration.py
```diff
@@ -336,71 +317,58 @@ def assign_scaleset_role(onefuzz_instance_name: str, scaleset_name: str) -> None
...
 
-    # check if the role is already assigned
-    role_assigned = seq(assignments["value"]).find(
-        lambda assignment: assignment["appRoleId"] == managed_node_role["id"]
+    query = (
+        "az rest --method post --url https://graph.microsoft.com/v1.0/servicePrincipals/%s/appRoleAssignedTo --body '%s' --headers \"Content-Type\"=application/json"
+        % (scaleset_service_principal.object_id, body)
+    )
+
+    logger.warning(
```
chkeita, line 369: image

bmc-msft (changes requested):

bmc-msft (approved):

PR #450, opened 2021-01-20
unify node resetting & deleting into delete/recreate

bmc-msft:


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -359,16 +359,17 @@ def set_state(self, state: NodeState) -> None:
         self.save()
 
     def delete(self) -> None:
+        self.mark_tasks_stopped_early()
         NodeTasks.clear_by_machine_id(self.machine_id)
+        NodeMessage.clear_messages(self.machine_id)
```
ranweiler, line 364: Is the change of order here (`clear_messages()`) significant?

ranweiler: I suggest adding a comment to that effect, if so.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: It's stylistic, but insignificant.

chkeita (approved):

PR #458, opened 2021-01-22
remove early-exit from cleanup_nodes that broke dead node cleanup

### LINKED ISSUE #456: reimaging of dead nodes only happens if there are other nodes to get reimaged
Because of an early-exit in `cleanup_nodes`, specifically:
https://github.com/microsoft/onefuzz/blob/c0a4b0dba484bec3c4e97a52b4f69bd62061f3ad/src/api-service/__app__/onefuzzlib/pools.py#L892-L894

The `get_dead_nodes` call later on never gets called:
https://github.com/microsoft/onefuzz/blob/c0a4b0dba484bec3c4e97a52b4f69bd62061f3ad/src/api-service/__app__/onefuzzlib/pools.py#L907-L910

bmc-msft:
Fixes #456

### DIFF
--- a/src/api-service/__app__/onefuzzlib/pools.py
+++ b/src/api-service/__app__/onefuzzlib/pools.py
@@ -881,9 +881,11 @@ class Scaleset(BASE_SCALESET, ORMMixin):
                     "no longer in scaleset: %s:%s", self.scaleset_id, node.machine_id
                 )
                 node.delete()
-        nodes = [x for x in nodes if x.machine_id in azure_nodes]
 
-        nodes_to_reset = [x for x in nodes if x.state in NodeState.ready_for_reset()]
+        existing_nodes = [x for x in nodes if x.machine_id in azure_nodes]
+        nodes_to_reset = [
+            x for x in existing_nodes if x.state in NodeState.ready_for_reset()
+        ]
 
         for node in nodes_to_reset:
             if node.delete_requested:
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -921,10 +913,11 @@ def cleanup_nodes(self) -> bool:
 
             if to_reimage:
                 self.reimage_nodes(to_reimage)
+            return bool(to_reimage) or bool(to_delete)
```
chkeita, line 916: isn't this over indented ?

src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -889,10 +885,6 @@ def cleanup_nodes(self) -> bool:
 
         nodes_to_reset = [x for x in nodes if x.state in NodeState.ready_for_reset()]
```
chkeita, line 886: i suggest rename `nodes` here to help the readability

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Yes.  Addressed.

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Not sure what you mean.  Can you provide more context?

chkeita (approved):

src/api-service/__app__/onefuzzlib/pools.py
bmc-msft: Handled after offline discussion

PR #466, opened 2021-01-26
add context to command failures

### LINKED ISSUE #465: Improve error message when user-provided file is not found
In instances where OneFuzz expects a file from the user and the file is missing, the recorded error message would ideally indicate which user file is missing. For example, if the `supervisor_exe` path is incorrect, the recorded error message is currently:

```
error running task: The system cannot find the file specified. (os error 2)
```

It would be nice if this recorded that the `supervisor_exe` could not be found and the user provided path to the executable. In my case, I had forgotten to provide the tools directory, i.e. `{tools_dir}\my_supervisor.exe`.

bmc-msft:
Fixes #465


src/agent/onefuzz/src/fs.rs
```diff
@@ -74,9 +74,11 @@ pub async fn set_executable(path: impl AsRef<Path>) -> Result<()> {
         .stdin(Stdio::null())
         .stdout(Stdio::null())
         .stderr(Stdio::piped())
-        .spawn()?
+        .spawn()
+        .context("chmod")?
```
lzybkr, line 78: This could be slightly more descriptive.

src/agent/onefuzz/src/fs.rs
```diff
@@ -74,9 +74,11 @@ pub async fn set_executable(path: impl AsRef<Path>) -> Result<()> {
...
         .wait_with_output()
-        .await?;
+        .await
+        .context("chmod")?;
```
lzybkr, line 81: This could be slightly more descriptive.

lzybkr (approved):

PR #474, opened 2021-01-27
expose supervisor tasks that are fully self-contained fuzzing tasks in the service

### LINKED ISSUE #439: Generic fuzzing task type
The only generic fuzzing tasks currently available are:

* generic_supervisor
* generic_generator

Both of these task types assume there are at least 2 executables: a fuzzing framework executable and the target executable. However there is not a generic task that accepts a single target executable with the fuzzing framework integrated therein (similar to LibFuzzer).

This limitation can be worked around by creating a shim executable that simply executes the target with the provided options. This shim can be uploaded to the tools container and used as the `supervisor_exe` in a generic supervisor task. However it requires additional setup and maintenance of a shim exe specifically for using the fuzzer in onefuzz, which isn't ideal.

bmc-msft:
Exposes the functionality added in #454 to the service & CLI.

Fixes #439

### DIFF
--- a/docs/webhook_events.md
+++ b/docs/webhook_events.md
@@ -1378,10 +1378,7 @@ Each event will be submitted via HTTP POST to the user provided URL.
             },
             "required": [
                 "type",
-                "duration",
-                "target_exe",
-                "target_env",
-                "target_options"
+                "duration"
             ],
             "title": "TaskDetails",
             "type": "object"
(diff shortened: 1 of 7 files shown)


chkeita (approved):

bmc-msft:
In my review of a different PR, I realized that this does not include reports, unique reports, and no repro containers which were all added in #454.

chkeita (approved):

PR #475, opened 2021-01-27
Update expansion documentation to describe nested options as implemented

bmc-msft:
This expands the documentation to demonstrate how nested lists are handled in practice, as well as to provide examples for how to deal with the nested list expansion.

### DIFF
--- a/docs/command-replacements.md
+++ b/docs/command-replacements.md
@@ -21,11 +21,18 @@ The following values are replaced with the specific values at runtime.
 
 Assume the following:
 
-* `supervisor_options` is: "a", "{target_options}", "b"
-* `target_options` is: "c", "{target_exe}"
-* `target_exe` is: "d"
+* `supervisor_options` is: ``"a", "{target_options}", "b"`
+* `target_options` is: `"c", "{target_exe}"`
+* `target_exe` is: `"d"`
 
-The resulting `supervisor_options` is: "a", "c", "d", "b"
+The resulting `supervisor_options` is: `"a", "c d", "b"`.
+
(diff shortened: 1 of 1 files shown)


docs/command-replacements.md
```diff
@@ -21,11 +21,18 @@ The following values are replaced with the specific values at runtime.
...
-* `target_exe` is: "d"
+* `supervisor_options` is: ``"a", "{target_options}", "b"`
+* `target_options` is: `"c", "{target_exe}"`
+* `target_exe` is: `"d"`
+
+The resulting `supervisor_options` is: `"a", "c d", "b"`.
+
+If you need `supervisor_options` to expand to: `"a", "c", "b", "d"`, you should use the following values:
+
+* `supervisor_options`: `"a", "c", "{target_exe}"`
```
chkeita, line 32:
```suggestion
* `supervisor_options`: `"a", "c", "b", "{target_exe}"`
```

bmc-msft: You're right in that the logic in the original PR is wrong.  I've updated the content to use A, B, C, D in alphabetic ordering to reduce confusion.  Can you re-review?

chkeita (approved):

docs/command-replacements.md
chkeita: I thought you had a non alphabetic order as a trick to force the reader to pay attention.
But i am ok with the alphabetical order

bmc-msft: Nope.  I didn't think about readability of the example when I wrote it.  Your feedback made me recognize that it was hard to grok.  The alphabetic example is less difficult to trace.

PR #482, opened 2021-01-28
Add Threat Model documentation

nharper285:
This PR includes a [Threat Model](https://aka.ms/tmt) for OneFuzz.


ranweiler (changes requested):
@nharper285, could you please update these file names and contents so they only refer to OneFuzz generically?

chkeita (approved):

PR #483, opened 2021-01-28
Expose Azure Storage container legal holds

bmc-msft:
TODO:
* [ ] - call clear_legal_hold on non-existant holds
* [ ] - add CLI calls
* [ ] - verify adding holds & metadata updates container in Azure Portal


resaisguvenligi:
Ty

stishkin:
Python is no longer supported

PR #488, opened 2021-01-29
Fix bug when checking if the default cli application exists

### LINKED ISSUE #486: OneFuzz 2.0 Initial Deployment Error "InvalidAppId"
## Information

- OneFuzz version: 2.0.0
- OS: Tried on both macOS and Azure console "Common Base Linux Delridge (quinault)"

1. Download the deployment artifact and install the requirements using `pipenv install`
2. Deploy OneFuzz using the command as shown in the section below.

### Expected result

For it to be deployed, the same steps (with different venv) could deploy version 1.11.0.

### Actual result

```
(onefuzz) user@Azure:~/onefuzz$ python deploy.py northeurope user_group <redacted>-onefuzz <redacted>@....com
  File "deploy.py", line 902, in <module>
    main()
  File "deploy.py", line 896, in main
    state[1](client)
  File "deploy.py", line 349, in setup_rbac
    authorize_application(onefuzz_cli_app_uuid, app.app_id)
  File "/home/user/onefuzz/registration.py", line 340, in authorize_application
    "preAuthorizedApplications": preAuthorizedApplications.to_list()
  File "/home/user/onefuzz/registration.py", line 79, in query_microsoft_graph
    response.status_code,
registration.GraphQueryError: request did not succeed: HTTP 400 - {
  "error": {
    "code": "InvalidAppId",
    "message": "Property PreAuthorizedApplication references applications 72f1562a-8c0c-41ea-beb9-fa2b71c80134 that cannot be found.",
    "innerError": {
      "date": "2021-01-29T16:24:51",
      "request-id": "9253d1ee-4992-4a39-8c22-6add7ff311eb",
      "client-request-id": "9253d1ee-4992-4a39-8c22-6add7ff311eb"
    },
    "details": [
      {
        "target": "api.preAuthorizedApplications.appId",
        "code": "InvalidValue"
      }
    ]
  }
}
```

Thank you!

chkeita:
closes #486

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -328,11 +328,9 @@ class Client:
 
         (password_id, password) = self.create_password(app.object_id)
 
-        cli_app = client.applications.list(filter="appId eq '%s'" % ONEFUZZ_CLI_APP)
+        cli_app = list(client.applications.list(filter="appId eq '%s'" % ONEFUZZ_CLI_APP))
 
-        onefuzz_cli_app_uuid = uuid.UUID(ONEFUZZ_CLI_APP)
-
-        if not cli_app:
+        if len(cli_app) == 0:
             logger.info(
                 "Could not find the default CLI application under the current "
                 "subscription, creating a new one"
@@ -346,7 +344,7 @@ class Client:
             }
 
         else:
-            authorize_application(onefuzz_cli_app_uuid, app.app_id)
+            authorize_application(uuid.UUID(ONEFUZZ_CLI_APP), app.app_id)
 
         self.results["client_id"] = app.app_id
         self.results["client_secret"] = password
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -328,7 +328,9 @@ class Client:
 
         (password_id, password) = self.create_password(app.object_id)
 
-        cli_app = list(client.applications.list(filter="appId eq '%s'" % ONEFUZZ_CLI_APP))
+        cli_app = list(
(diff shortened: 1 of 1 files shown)


bmc-msft:
Running integration tests

src/deployment/deploy.py
```diff
@@ -328,11 +328,11 @@ def setup_rbac(self) -> None:
...
 
-        cli_app = client.applications.list(filter="appId eq '%s'" % ONEFUZZ_CLI_APP)
-
-        onefuzz_cli_app_uuid = uuid.UUID(ONEFUZZ_CLI_APP)
+        cli_app = list(
+            client.applications.list(filter="appId eq '%s'" % ONEFUZZ_CLI_APP)
+        )
 
-        if not cli_app:
+        if len(cli_app) == 0:
```
ranweiler, line 335: Can you describe what the bug was? Is is that `client.applications.list()` does not return a `list`, but some kind of generator or other iterable?

bmc-msft: almost every azure-sdk API returns iterators, not lists.

bmc-msft (approved):

src/deployment/deploy.py
ranweiler: Makes sense. So the `list()` call above is the only thing that _needed_ to be in this PR. I'm fine with the more explicit `len()` check here, it is just equivalent to `not cli_app` for a `list`, so I was wondering if there was anything else going on.

chkeita: yes it was returning an empty iterator  when the no application was found.
the previous test was always evaluating to false

PR #489, opened 2021-01-29
Continually try updating scaleset configs until successful

bmc-msft:
As is, if a scaleset is "failed" for some reason (such as Azure VM allocation failed), the service does not attempt to push updated VM configs, which can cause our time-limited SAS tokens in the VM config to timeout.

This PR makes 2 changes:
1. attempts to perform any update as long as the scaleset isn't in the "Updating" state
2. moves the config update attempts from once per day configured once per day, but runs every 90 seconds until it's successful.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/pools.py
+++ b/src/api-service/__app__/onefuzzlib/pools.py
@@ -1146,21 +1146,26 @@ class Scaleset(BASE_SCALESET, ORMMixin):
             self.nodes.append(node_state)
 
     def update_configs(self) -> None:
-        if self.state != ScalesetState.running:
-            logging.debug(
-                "scaleset not running, not updating configs: %s", self.scaleset_id
-            )
-            return
+        if not self.needs_config_update:
+            logging.debug("config update not needed: %s", self.scaleset_id)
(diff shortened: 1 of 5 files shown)


src/api-service/__app__/onefuzzlib/pools.py
```diff
@@ -1146,21 +1146,26 @@ def update_nodes(self) -> None:
...
+
+        logging.info("updating scaleset configs: %s", self.scaleset_id)
 
         pool = Pool.get_by_name(self.pool_name)
         if isinstance(pool, Error):
+            logging.error(
+                "unable to find pool during config update: %s - %s",
+                self.scaleset_id,
+                pool,
+            )
```
chkeita, line 1160: i think needs_config_update  should be to set false here too

bmc-msft: In this case, we're marking the scaleset as failed, which moves it into the failed state.  If we ever have a scaleset without a pool, it's an _all-stop_ level event, as something is seriously broken.

chkeita (approved):

PR #490, opened 2021-01-30
address clippy issues in agent

bmc-msft:
Addresses clippy issues in the agent & re-enables clippy verification during CI


bmc-msft:
Note, this leaves commented out code relating to the supervisor in src/local/common.  This is intentional, I have a non-pushed PR that adds a local supervisor task that uses this code.

src/agent/debugger/src/target.rs
```diff
@@ -616,18 +618,15 @@ impl Target {
     pub(crate) fn handle_single_step(&mut self, step_state: StepState) -> Result<()> {
         self.single_step.remove(&self.current_thread_handle);
 
-        match step_state {
-            StepState::Breakpoint { pc } => {
-                write_instruction_byte(self.process_handle, pc, 0xcc)?;
+        if let StepState::Breakpoint { pc } = step_state {
```
lzybkr, line 621: Please revert this change - it will cause an annoying merge conflict for my PR.

src/agent/input-tester/src/tester.rs
```diff
@@ -420,11 +420,11 @@ impl Tester {
 }
 
 fn exception_from_throw(exception: &Exception) -> bool {
-    match exception.description {
+    matches!(
```
lzybkr, line 423: I think this is more confusing that the match statement.

src/agent/input-tester/src/test_result/mod.rs
```diff
@@ -343,17 +343,11 @@ impl ExitStatus {
     }
 
     pub fn is_normal_exit(&self) -> bool {
-        match self {
-            ExitStatus::Code(_) => true,
-            _ => false,
-        }
+        matches!(self, ExitStatus::Code(_))
```
lzybkr, line 346: I'm not sure I like this change.

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -12,7 +12,7 @@ use anyhow::{Error, Result};
 use std::{collections::HashMap, path::Path, time::Duration};
 use tempfile::tempdir;
 
-const DEFAULT_TIMEOUT_SECS: u64 = 5;
+const DEFAULT_TIMEOUT_SECS: Duration = Duration::from_secs(5);
```
lzybkr, line 15: You can drop `_SECS` since it is a Duration now.

lzybkr (changes requested):

src/agent/debugger/src/target.rs
bmc-msft: reverted

src/agent/input-tester/src/test_result/mod.rs
ranweiler: I'm fine with this type, where we are matching against one variant.

src/agent/input-tester/src/tester.rs
ranweiler: I agree; when matching against several cases, I suggest using a full `match` expr.

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -26,17 +26,17 @@ pub const CHECK_FUZZER_HELP: &str = "check_fuzzer_help";
 pub const TARGET_EXE: &str = "target_exe";
 pub const TARGET_ENV: &str = "target_env";
 pub const TARGET_OPTIONS: &str = "target_options";
-pub const SUPERVISOR_EXE: &str = "supervisor_exe";
-pub const SUPERVISOR_ENV: &str = "supervisor_env";
-pub const SUPERVISOR_OPTIONS: &str = "supervisor_options";
+// pub const SUPERVISOR_EXE: &str = "supervisor_exe";
```
ranweiler, line 29: Assuming this is unused but intentionally kept for later?

ranweiler (approved):

src/agent/onefuzz-agent/src/local/common.rs
bmc-msft: Yes.

src/agent/input-tester/src/test_result/mod.rs
bmc-msft: leaving this one, since it's a single use variant

src/agent/input-tester/src/tester.rs
bmc-msft: reverted this one.

lzybkr (approved):

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -47,26 +47,45 @@ impl<'a> Tester<'a> {
...
-            check_retry_count,
+            timeout: DEFAULT_TIMEOUT,
+            check_asan_log: false,
+            check_asan_stderr: false,
+            check_debugger: true,
+            check_retry_count: 0,
         }
     }
 
+    pub fn timeout(&mut self, value: u64) -> &mut Self {
```
chkeita, line 64: apologies for the late review but was this fixing a clippy issue or just a style change ?

ranweiler: @chkeita, my guess is that we were hitting the [`too_many_arguments`](https://rust-lang.github.io/rust-clippy/master/#too_many_arguments) lint, and @bmc-msft just used it a cue to try to improve the overall API.

chkeita: Fair enough, I would like to propose that we keep the type immutable in this case.
Something like 

``` rust
pub fn timeout(self, value: u64) -> Self {
    Self{
        timeout : Duration::from_secs(value),
        ..self
    }
}
```

My reasoning for avoiding mutability is to make it easier to reason about the state of variables because the state can  only be defined at initialization.

bmc-msft: @ranweiler is correct here. 

I modeled the code after our existing mechanism that uses the builder paradigm, Expand (the variable expansion functionality).  I'm open to switching to the immutable version, but we should probably update both to follow the same paradigm.

chkeita: created #500 with the proposed changed

PR #493, opened 2021-02-01
add telemetry stats utility

bmc-msft:


src/utils/telemetry-stats/src/main.rs
```diff
@@ -0,0 +1,147 @@
+#[macro_use]
+extern crate serde;
+
+use anyhow::Result;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::io::{self, BufRead, BufReader};
+use chrono::{DateTime, Utc, Duration};
+//use std::time::Duration;
```
ranweiler, line 9: Commented code.

src/utils/telemetry-stats/src/main.rs
```diff
@@ -0,0 +1,147 @@
+#[macro_use]
```
ranweiler, line 1: Should we add a mod-level doc comment noting the expected API version / schema?

src/utils/telemetry-stats/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+struct Entry {
+    event: Vec<Event>,
+    context: Context,
+}
+
+fn read_file(filename: &str) -> Result<Vec<Entry>> {
+    let mut results = vec![];
+    let file = File::open(filename)?;
+    let reader = BufReader::new(file);
+    for (_, line) in reader.lines().enumerate() {
```
ranweiler, line 54: No need to `enumerate()` if we're just binding the index to `_`, right?

ranweiler (approved):

PR #502, opened 2021-02-04
Update Proxy heartbeat & logging

### LINKED ISSUE #498: Proxy VM is not sending heartbeats in certain circumstances
## Information

- Onefuzz version: 2.1.0
- OS: Windows (CLI)

This happens in our production instance which you may be familiar with c***rod05, but it does not seem to happen on our less busy or newer instances. 

1. Run the 'debug' command 
$ onefuzz.exe debug node rdp <guid>

2. Connect as you would if you wanted to inspect some onefuzz logs on the node

3. Browse the web, watch TV, or just leave the RDP session open

After 10 to 15 minutes you will be disconnected.

4. You can recreate the debug session using the same command in Step 1, and will see the node in the same state it was before you disconnected.

Upon further investigation, the instance activity log shows the App Function (System-Assigned Managed Identity) has deleted the Proxy VM and that App Insights did not log a heartbeat for 10 minutes before deleting the Proxy VM. So partly, the behavior was expected. That is why we think this issue has something to do with the heartbeats. 

### Expected result

We expect the Proxy VM to send heartbeats and then for the App Function to process them so it does not delete the Proxy VM.

 

AB#39994700

bmc-msft:
This PR adds:
* application insights logging to the VM proxy
* moved heartbeat updates to be more consistent
 
Fixes #498


src/proxy-manager/src/main.rs
```diff
@@ -2,39 +2,47 @@
...
         info!("checking updates");
-        if proxy_config.update().await? {
-            last_updated = std::time::Instant::now();
-        } else if last_updated + MINIMUM_NOTIFY_INTERVAL < std::time::Instant::now() {
+        proxy_config.update().await?;
+
+        if last_updated + MINIMUM_NOTIFY_INTERVAL < Instant::now() {
             proxy_config.notify().await?;
-            last_updated = std::time::Instant::now();
+            last_updated = Instant::now();
```
ranweiler, line 37: Should we rename this var to `last_notified`?

ranweiler (approved):

PR #503, opened 2021-02-04
update pools before nodes or scalesets

bmc-msft:
Handling pool updates prior to nodes or scalesets speeds up the scaleset shutdowns and autoscaling operations.


src/api-service/__app__/timer_workers/__init__.py
```diff
@@ -28,6 +28,14 @@ def process_scaleset(scaleset: Scaleset) -> None:
 
 
 def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
+    pools = Pool.search()
```
ranweiler, line 31: Can we add a comment here noting that order matters wrt perf?

ranweiler: (and noting _why_, in particular)

ranweiler (approved):

PR #508, opened 2021-02-04
local_run refactoring

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/storage-queue/src/local_queue.rs
```diff
@@ -0,0 +1,147 @@
...
+        })
+    }
+
+    pub async fn enqueue(&self, data: impl Serialize) -> Result<()> {
+        let sender = self
+            .sender
+            .lock()
+            .map_err(|_| anyhow::anyhow!("unable to acquire lock"))?;
+        let mut buffer = Vec::new();
+        serde_xml_rs::to_writer(&mut buffer, &data)
```
bmc-msft, line 129: Can you speak to why you chose XML as the local queue serialization?

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -57,6 +63,10 @@ impl SyncedDir {
     }
 
     pub async fn init(&self) -> Result<()> {
+        if let Some(path) = self.url.as_file_path() {
+            fs::create_dir_all(path).await?;
```
bmc-msft, line 67: I'm confused by this.

Don't we `fs::create_dir_all(&self.path)` later on?  What context does the url need creating?

src/agent/storage-queue/src/local_queue.rs
chkeita: I chose xml to be compatible with azure. The body of message in the queue is [in xml](https://docs.microsoft.com/en-us/rest/api/storageservices/get-messages#response-body)

src/agent/onefuzz-supervisor/src/work.rs
```diff
@@ -113,45 +115,42 @@ impl IWorkQueue for WorkQueue {
...
     pub work_set: WorkSet,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct Receipt(pub storage_queue::Receipt);
-
 pub struct WorkQueue {
     queue: QueueClient,
-    registration: Registration,
+    _registration: Registration,
```
bmc-msft, line 133: why is this _registration?

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -66,41 +58,12 @@ pub enum CrashTestResult {
...
-            if !exists(&path).await? {
-                let data = serde_json::to_vec(&report)?;
-                fs::write(path, data).await?;
-                Ok(true)
-            } else {
-                Ok(false)
-            }
-        }
-    }
+    container.upload(dest_name, report).await
```
bmc-msft, line 66: 👍

src/agent/onefuzz-agent/src/tasks/fuzz/supervisor.rs
```diff
@@ -31,7 +31,7 @@ use tokio::{
     sync::Notify,
 };
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize)]
```
bmc-msft, line 34: Why did you remove the `Default`?  This was useful, considering how below you had to specify a suite of values are None.

src/agent/onefuzz-agent/src/tasks/fuzz/generator.rs
```diff
@@ -234,23 +233,23 @@ mod tests {
...
             crashes: SyncedDir {
-                path: crashes.to_path_buf(),
-                url: None,
+                path: tempfile::tempdir().unwrap().path().into(),
+                url: BlobContainerUrl::parse(Url::from_directory_path(crashes).unwrap())?,
             },
             tools: Some(SyncedDir {
-                path: radamsa_dir.to_path_buf(),
-                url: None,
+                path: tempfile::tempdir().unwrap().path().into(),
```
bmc-msft, line 251: Does this work?  It's my understanding that as soon as `tempdir()` goes out of scope, the underlying paths get deleted.

src/agent/onefuzz-agent/src/tasks/fuzz/supervisor.rs
chkeita: I had to remove it because the SyncDir No longer have an optional url. and Url does not implement Default. 
Unfortunately the `..Default::default()` syntax does not work even if you already specified the fields that require a value.

src/agent/onefuzz/src/syncdir.rs
chkeita: it is way to ensure the remote directory exist. it also simplify creation because the directory get created as needed. 
I will rename the variable to remote path to avoid the confusion with path below.

src/agent/onefuzz-supervisor/src/work.rs
chkeita: That was the result of some experimentation, will rename

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -174,3 +198,59 @@ pub fn build_common_config(args: &ArgMatches<'_>) -> Result<CommonConfig> {
...
+
+        Ok(DirectoryMonitorQueue {
+            directory_path,
+            queue_client,
+            handle,
+        })
+    }
+}
+
+pub async fn wait_for_dir(path: impl AsRef<Path>) -> Result<()> {
```
bmc-msft, line 246: can the `backoff` crate be used here?

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -174,3 +198,59 @@ pub fn build_common_config(args: &ArgMatches<'_>) -> Result<CommonConfig> {
...
+            storage_queue::local_queue::ChannelQueueClient::new()?,
+        );
+        let queue = queue_client.clone();
+        let handle: tokio::task::JoinHandle<std::result::Result<(), anyhow::Error>> =
+            tokio::spawn(async move {
+                let mut monitor = DirectoryMonitor::new(directory_path_clone.clone());
+                monitor.start()?;
+                loop {
+                    match monitor.poll_file() {
+                        Poll::Ready(Some(crash)) => {
```
bmc-msft, line 225: perhaps rename this from crash to filename?

src/agent/storage-queue/src/local_queue.rs
chkeita: never mind I thought i was looking at azure queue.
This was probably the result of a copy paste from the azure queue. I can convert to json instead

src/agent/onefuzz/src/fs.rs
```diff
@@ -170,6 +172,104 @@ impl OwnedDir {
     }
 }
 
+#[cfg(target_os = "linux")]
+pub async fn _sync(
```
chkeita, line 176: self reminder to test it on linux

bmc-msft:
relative paths don't appear to work.

Tested via 
`RUST_LOG=debug onefuzz-agent local libfuzzer --crashes_dir ./crashes --inputs_dir ./inputs --target_exe ./fuzz.exe --unique_reports_dir ./unique_reports`

This fails with the error `Error: invalid file path`.  Adding a bunch of anyhow context messages localy, found this is the case for Url::from_file_path that has the issue.

bmc-msft:
Running our basic libfuzzer sample, I get lots of crashes & inputs, but no crash reports.

`onefuzz-agent local libfuzzer --crashes_dir /tmp/wut/crashes --inputs_dir /tmp/wut/inputs --target_exe /tmp/wut/fuzz.exe --unique_reports_dir /tmp/wut/unique_reports`

Let it run for a while, no reports ever generated.

bmc-msft:
Is the an intent to automatically clean up the job/task directories at some point in the future?  I've run the same command a few times, and now I have many huge subdirectories in `00000000-0000-0000-0000-000000000000/RANDOM_GUID`

src/agent/onefuzz-agent/src/local/libfuzzer.rs
```diff
@@ -3,46 +3,101 @@
...
+                ..common
+            },
+        )?;
+        let analysis_task = spawn(async move { run_analysis(analysis_config).await });
+
+        task_handles.push(analysis_task);
+        task_handles.push(analysis_input_monitor.handle);
+    }
+
+    try_join_all(task_handles).await?;
```
ranweiler, line 99: I think this needs something equivalent to:
```suggestion
    let results = try_join_all(task_handles).await?;
    for result in results {
        result?;
    }
```
(there's definitely nicer ways to write that)

Otherwise, I think we only propagate the `JoinError` from a task that panics, not the `Ok(Err(_))` from a task that completed successfully, but whose return value was an `Err(_)`.

src/agent/onefuzz-agent/src/local/libfuzzer.rs
```diff
@@ -3,46 +3,101 @@
...
+        task_handles.push(crash_report_input_monitor.handle);
+    }
 
     if args.is_present(COVERAGE_DIR) {
-        let coverage_config = build_coverage_config(args, true)?;
-        let coverage = CoverageTask::new(coverage_config);
-        let coverage_task = spawn(async move { coverage.local_run().await });
-
-        let result = tokio::try_join!(fuzz_task, report_task, coverage_task)?;
-        result.0?;
```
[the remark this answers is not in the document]
chkeita: This still does not stop the all the task as soon as one fails. It basically waits for all the task to finish and return an Vec<Result<()>>. [see here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b9f317bcf27c62c249b2d607f42f4fad).
The solution i found was to unwrap() the result inside the spawn.

src/agent/storage-queue/src/lib.rs
```diff
@@ -1,200 +1,138 @@
...
-            .error_for_status()
-            .context("storage queue enqueue failed with error")?;
-        Ok(())
+impl<'de> Deserialize<'de> for QueueClient {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Url::deserialize(deserializer)
+            .map(|u| QueueClient::new(u).expect("Unable to create queue client"))
```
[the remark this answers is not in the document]
chkeita: i could not find a way to translate the anyhow error to a D::Error. any suggestions ?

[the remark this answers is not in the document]
chkeita: found a way

src/agent/onefuzz-agent/src/local/libfuzzer.rs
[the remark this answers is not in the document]
ranweiler: Ah right! Yeah, we actually need a `select()` so we can check the inner (non-`JoinHandle`) `Result`.

src/agent/onefuzz-agent/src/local/libfuzzer.rs
ranweiler: (as noted elsewhere, we really need a `select()` if we want to short-circuit on the inner values)

src/agent/storage-queue/src/local_queue.rs
chkeita: i reverted this back to xml because of the serialisation behavior of simple string. 
the json serializer puts quotes around the string which breaks the file monitoring currently

src/agent/onefuzz-agent/src/local/cmd.rs
```diff
@@ -5,17 +5,20 @@ use anyhow::Result;
...
+    libfuzzer, libfuzzer_coverage, libfuzzer_crash_report, libfuzzer_fuzz, libfuzzer_merge,
+    radamsa,
 };
 
 const RADAMSA: &str = "radamsa";
 const LIBFUZZER: &str = "libfuzzer";
 const LIBFUZZER_FUZZ: &str = "libfuzzer-fuzz";
 const LIBFUZZER_CRASH_REPORT: &str = "libfuzzer-crash-report";
 const LIBFUZZER_COVERAGE: &str = "libfuzzer-coverage";
+const LIBFUZZER_MERGE: &str = "libfuzzer_merge";
```
bmc-msft, line 18: Use the `-` instead of `_` to match others.

```suggestion
const LIBFUZZER_MERGE: &str = "libfuzzer-merge";
```

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -2,10 +2,20 @@ use crate::tasks::config::CommonConfig;
...
+use reqwest::Url;
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    time::Duration,
+};
 use uuid::Uuid;
 
+use backoff::{future::retry, Error as BackoffError, ExponentialBackoff};
+use path_absolutize::*;
```
bmc-msft, line 16:
```suggestion
use path_absolutize::Absolutize;
```

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -142,17 +130,56 @@ pub fn add_common_config(app: App<'static, 'static>) -> App<'static, 'static> {
...
+    task_id: Uuid,
+    args: &ArgMatches<'_>,
+) -> Result<Vec<SyncedDir>> {
+    let current_dir = std::env::current_dir()?;
+    let dirs: Result<Vec<SyncedDir>> = value_t!(args, name, PathBuf)?
+        .iter()
+        .enumerate()
+        .map(|(index, remote_path)| {
+            let path = PathBuf::from(remote_path);
+            let remote_path = path.absolutize()?;
```
bmc-msft, line 150: the `absolutize` doesn't do what I would expect when it comes to reports.

I ran the following:
`./target/debug/onefuzz-agent local libfuzzer --crashes_dir /tmp/local-test/crashes  --inputs_dir /tmp/local-test/inputs --unique_reports_dir /tmp/local-test/unique   --target_exe ./test.exe`

In the crash reports, this set the input blob to: 
```
  "input_blob": {
    "account": null,
    "container": null,
    "name": "tmp/local-test/crashes/crash-dc57cfd90a06ff5647a255718a2684209851a6bc"
  }
```

I would expect this to be `/tmp/local-tests/crashes/crash-...`.

Additionally, I would expect if I said `--crashes_dir crashes`, the name to be `crashes/crash-...`.  Instead, it ends up being the full path to $CWD/crashes minus the leading slash.

I think we can probably merge this PR and address this with a future PR.

src/agent/onefuzz/src/blob/url.rs
```diff
@@ -83,36 +106,54 @@ pub struct BlobContainerUrl {
 impl BlobContainerUrl {
     pub fn new(url: Url) -> Result<Self> {
         if !possible_blob_container_url(&url) {
-            bail!("Invalid container URL: {}", url);
+            bail!("Invalid container URL 1: {}", url);
```
bmc-msft, line 109:
```suggestion
            bail!("Invalid container URL: {}", url);
```

bmc-msft (approved):

src/agent/onefuzz-agent/src/local/common.rs
chkeita: as discussed this will be addressed in a separate PR

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -20,35 +27,35 @@ pub enum SyncOperation {
...
-        }
-
-        let dir = &self.path;
-        let url = self.url.as_ref().unwrap().url();
-        let url = url.as_ref();
-        debug!("syncing {:?} {}", operation, dir.display());
-        match operation {
-            SyncOperation::Push => az_copy::sync(dir, url, delete_dst).await,
-            SyncOperation::Pull => az_copy::sync(url, dir, delete_dst).await,
+        // Adding a trailing '/' to specify a copy of the dir content on linux
```
ranweiler, line 38: Wow, I didn't even know we could do this in this way.

If this is necessary, maybe it should be the responsibility of the `az_copy` module?

src/agent/onefuzz/src/fs.rs
```diff
@@ -170,6 +172,104 @@ impl OwnedDir {
...
+
+    let output = cmd
+        .spawn()
+        .context("robocopy failed to start")?
+        .wait_with_output()
+        .await
+        .context("robocopy failed to run")?;
+
+    if let Some(code) = output.status.code() {
+        // any value >= 8 indicates that there was at least one failure during the copy operation
```
ranweiler, line 245: 👍

src/agent/onefuzz/src/fs.rs
```diff
@@ -170,6 +172,104 @@ impl OwnedDir {
...
+            stdout,
+            stderr
+        );
+    }
+
+    Ok(())
+}
+
+#[cfg(target_os = "windows")]
+pub async fn _sync(
```
ranweiler, line 216: WDYT about naming this (and above) `sync_impl()`?

src/agent/onefuzz/src/fs.rs
```diff
@@ -170,6 +172,104 @@ impl OwnedDir {
...
+
+    if !output.status.success() {
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        anyhow::bail!(
+            "sync failed src:{:?} dst:{:?} stdout:{:?} stderr:{:?}",
+            src.as_ref(),
+            dst.as_ref(),
+            stdout,
+            stderr
```
ranweiler, line 208: Huh, did `rustfmt` not add a trailing comma?

src/agent/onefuzz/src/blob/url.rs
```diff
@@ -1,24 +1,28 @@
...
+use std::{fmt, path::PathBuf};
 
 use anyhow::Result;
 use reqwest::Url;
 use serde::{de, Serialize, Serializer};
 
 #[derive(Clone, Eq, PartialEq)]
-pub struct BlobUrl {
-    url: Url,
+pub enum BlobUrl {
```
ranweiler, line 11: The goal of this abstraction was to be a `Url`, but one that we know to follow the default Azure blob storage pattern, which guarantees that we can extract a container and account name.

I would have liked if we kept it as-is, but then added a _new_ `enum` that can be either/or, as in:
```rust
pub enum Blob {
    Remote(BlobUrl),
    Local(PathBuf),
}
```

src/agent/onefuzz/src/blob/url.rs
```diff
@@ -28,47 +32,66 @@ impl BlobUrl {
...
+        match self {
+            Self::LocalFile(path) => {
+                Url::from_file_path(path).expect("Could not convert path to url")
+            }
+            Self::AzureBlob(url) => url.clone(),
+        }
     }
 
-    pub fn account(&self) -> String {
-        // Ctor checks that domain has at least one subdomain.
```
ranweiler, line 35: Whatever happens, we shouldn't remove comments like this that justify `unwrap()` operations.

src/agent/onefuzz-supervisor/src/work.rs
```diff
@@ -171,38 +170,44 @@ impl WorkQueue {
...
-            if is_auth_error(err) {
-                self.renew().await.context("unable to renew registration")?;
-                self.queue
-                    .delete(receipt)
-                    .await
-                    .context("unable to claim work from queue")?;
+        Ok(Some(msg))
+    }
+
+    pub async fn claim(&mut self, message: Message) -> Result<WorkSet> {
```
ranweiler, line 189: Why have `claim()` return a work set? If the caller needs it, they can do a partial move or `clone()` it.

ranweiler: In my mind, the argument to this method should just be the metadata required to claim the message (i.e. finalize the queue pop), and that + the message payload could be in their own struct.

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -101,18 +107,66 @@ impl SyncedDir {
...
+
+            while let Some(item) = monitor.next().await {
+                event!(event.clone(); EventData::Path = item.display().to_string());
 
-        while let Some(item) = monitor.next().await {
-            event!(event.clone(); EventData::Path = item.display().to_string());
-            if let Some(uploader) = &mut uploader {
                 if let Err(err) = uploader.upload(item.clone()).await {
-                    bail!(
+                    warn!(
```
bmc-msft, line 169: Why did you change this from a failure condition to warn?  I'm concerned this will swallow errors.

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -101,18 +107,66 @@ impl SyncedDir {
...
+            fs::create_dir_all(&path).await?;
+
+            while let Some(item) = monitor.next().await {
+                event!(event.clone(); EventData::Path = item.display().to_string());
+                let file_name = item
+                    .file_name()
+                    .ok_or_else(|| anyhow!("invalid file path"))?;
+                let destination = path.join(file_name);
+                if let Err(err) = fs::copy(&item, &destination).await {
+                    warn!(
```
bmc-msft, line 156: if fs::copy fails, we should fail hard, otherwise we'll swallow errors in practice

src/agent/onefuzz/src/fs.rs
```diff
@@ -170,6 +172,104 @@ impl OwnedDir {
...
+}
+
+#[cfg(target_os = "windows")]
+pub async fn _sync(
+    src: impl AsRef<OsStr>,
+    dst: impl AsRef<OsStr>,
+    delete_dst: bool,
+    recursive: bool,
+) -> Result<()> {
+    let mut cmd = Command::new("robocopy");
```
bmc-msft, line 222: Probably not for this PR, but we probably want to do early-checking the platform specific tools (`robocoby` or `rsync`) from here, and `cdb` or `gdb` for the crash reporting tasks during initialization.

src/agent/onefuzz/src/syncdir.rs
chkeita: This was because when monitoring lib fuzzer crashes i saw an issue where the monitor would pick up what seemed to be a temporary file and the file would disappear before the copy was invoked which would cause an error

src/agent/onefuzz/src/syncdir.rs
chkeita: see my other comment about above. This change was to have the same behavior for local and managed mode.
My thought was that we would not want to stop monitoring the whole directory if one upload fails

src/agent/onefuzz/src/fs.rs
chkeita: creates #669 to track this

src/agent/onefuzz/src/blob/url.rs
chkeita: I agree however this will be a big change as it might involve doing the same thing to the blobContainer abstraction
This will be addressed in a different PR. I created #670 to keep track of this

src/agent/onefuzz/src/fs.rs
chkeita: 👍

src/agent/onefuzz/src/syncdir.rs
bmc-msft: We handled the race condition for syncing inputs from libFuzzer in the remote mode in #204 .

As a compromise, can we attempt the move, if it fails, check to see if the source still exists.  If the source still exists, then bail?

src/agent/onefuzz/src/syncdir.rs
bmc-msft: upload uses the reqwest-retry logic to try to recover.  My thought is if we fail even with automatic recovery, we should hard-fail the task.

src/agent/onefuzz/src/syncdir.rs
chkeita: #204 is addressing a different problem. Here is what the error looks like 
```
Error: Couldn't upload file.  source:"/mnt/c/work/onefuzz/src/agent/00000000-0000-0000-0000-000000000000/c00fed1e-cced-46d3-9ccd-1f5d645122a5/inputs_dir/.2f654f29901d09cd755b50314cdbabecb25b43d7.5O4S1O" destination:"/mnt/c/work/onefuzz-samples/examples/simple-libfuzzer_linux/seeds/.2f654f29901d09cd755b50314cdbabecb25b43d7.5O4S1O" err:No such file or directory (os error 2)
``` 
I added a check to see if the file still exists as you suggested

src/agent/onefuzz/src/syncdir.rs
chkeita: added a file existence check to decide between warning and bailing

src/agent/onefuzz/src/fs.rs
chkeita: yup

src/agent/onefuzz/src/syncdir.rs
chkeita: moved the logic into the SyncPath type

src/agent/onefuzz-supervisor/src/work.rs
chkeita: when claiming the message we call delete on the inner queue_message which takes ownership of the message. 
So the Message argument is not usable after the call to this. Returning the work_set allows us to avoid a clone. The caller still has the choice of not using the return value.

ranweiler: This comment is _not_ a PR blocker, but I just want to clarify my suggestion. Feel free to resolve once read.

Taking ownership of the inner `queue_message` field is not a problem (unless there is something new to the situation I missed!). Until this PR, we already did what I am suggesting (with the previous version of the field, which held a `Receipt`). We apparently had a `work_set.clone()` in one of the branches, but it is actually redundant.

So if we maintained / switch back to a similar API, it would look like:
```rust
pub struct Metadata(Option<QueueMessage>);

pub struct Message {
    pub metadata: Metadata,  // Before, this was always a `Receipt`.
    pub work_set: WorkSet,
}

// Assume the caller splits off `Metadata` from the `Message.metadata` field.
// Now `claim()` only takes ownership of what it actually needs, 
// and doesn't even need to "know" the message payload type.
pub async fn claim(&mut self, metadata: Metadata) -> Result<()>;
```
Then we can handle the `Message`/`claim()` like we did before:
```rust
// Move one field out of the struct.
// This is fine as long as we only try to use other (un-moved) fields later on, 
// and not the entire struct. Note we don't _need_ this temp variable.
let work_set = message.work_set;

// Use the other, un-moved field. We can now access `work_set` as needed, since it is in scope.
match self.work_queue.claim(message.metadata) {
    // Handle the different cases.
    Ok(..) => {
        do_something_with(work_set);
    }
    Err(..) => {
        print_warning_about(work_set);
    }
}
```

chkeita:
testing in progress

src/agent/onefuzz-agent/src/local/libfuzzer.rs
```diff
@@ -3,46 +3,102 @@
 
 use crate::{
     local::{
-        common::COVERAGE_DIR,
+        common::{
```
chkeita, line 6: testing in progress

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -43,6 +43,7 @@ pub async fn send_retry_reqwest<F: Fn() -> Result<reqwest::RequestBuilder> + Sen
             Ok(response)
         }
     };
+
```
bmc-msft, line 46:
```suggestion

```

PR #512, opened 2021-02-05
update nodes and scalesets in a consistent order

bmc-msft:
With autoscaling, it is important that scalesets and nodes are updated
in a consistent order to prevent update thrashing.


src/api-service/__app__/timer_workers/__init__.py
```diff
@@ -42,13 +43,17 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
             autoscale_pool(pool)
 
     Node.mark_outdated_nodes()
+    # ensure nodes are updated in a consistent order, which is needed for
+    # autoscaling of pools
```
ranweiler, line 47: Could you elaborate on this just a bit? I take this to mean "consistent across timer-triggered invocations", but it would be helpful to explain why, and how it interacts with auto-scaling in particular.

ranweiler (approved):

PR #514, opened 2021-02-05
use the primitive types in more places

bmc-msft:
In refactoring the autoscaling code, I noticed that the primitive types that provides consistent data validation for Containers, Pool Names, and Regions were not used everywhere.  This PR corrects that issue.

This more broadly uses at-deserialization-time validation, rather than relying on our ORM's identification that `str` is an unchecked and therefore only used as a post-query filter.

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -170,23 +170,23 @@ class Files(Endpoint):
     endpoint = "files"
 
     @cached(ttl=ONE_HOUR_IN_SECONDS)
-    def _get_client(self, container: str) -> ContainerWrapper:
+    def _get_client(self, container: primitives.Container) -> ContainerWrapper:
         sas = self.onefuzz.containers.get(container).sas_url
         return ContainerWrapper(sas)
 
-    def list(self, container: str) -> models.Files:
+    def list(self, container: primitives.Container) -> models.Files:
         """ Get a list of files in a container """
         self.logger.debug("listing files in container: %s", container)
         client = self._get_client(container)
         return models.Files(files=client.list_blobs())
 
-    def delete(self, container: str, filename: str) -> None:
+    def delete(self, container: primitives.Container, filename: str) -> None:
         """ delete a file from a container """
         self.logger.debug("deleting in container: %s:%s", container, filename)
         client = self._get_client(container)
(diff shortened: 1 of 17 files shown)


bmc-msft:
For review:


Type implementations:
https://github.com/microsoft/onefuzz/blob/e3dfcb8b95d6d0bc8e5f3c8b667f62b7bfd90040/src/pytypes/onefuzztypes/primitives.py#L15-L33

Validator implementations:
https://github.com/microsoft/onefuzz/blob/e3dfcb8b95d6d0bc8e5f3c8b667f62b7bfd90040/src/pytypes/onefuzztypes/validators.py#L6-L25

Checked string types:
https://github.com/microsoft/onefuzz/blob/e3dfcb8b95d6d0bc8e5f3c8b667f62b7bfd90040/src/api-service/__app__/onefuzzlib/orm.py#L67

Filter building code:
https://github.com/microsoft/onefuzz/blob/e3dfcb8b95d6d0bc8e5f3c8b667f62b7bfd90040/src/api-service/__app__/onefuzzlib/orm.py#L146-L167

src/cli/onefuzz/api.py
```diff
@@ -170,31 +170,34 @@ class Files(Endpoint):
     endpoint = "files"
 
     @cached(ttl=ONE_HOUR_IN_SECONDS)
-    def _get_client(self, container: str) -> ContainerWrapper:
+    def _get_client(self, container: primitives.Container) -> ContainerWrapper:
```
ranweiler, line 173: Style suggestion, maybe just import `Container` into this module so we don't have to keep qualifying it.

ranweiler: I guess my suggestion includes the primitives we use below, too.

ranweiler (approved):

chkeita (approved):

src/cli/onefuzz/api.py
bmc-msft: We previously did this, but changed to avoid conflict with the class names for the API endpoints (such as `Containers`)

bmc-msft: Leaving for now, we can revisit that in the future, but it would be a larger impact.

PR #518, opened 2021-02-08
add missing scaleset nodes

bmc-msft:
Scalesets can have nodes that never check in (such as broken OS setup scripts).

This will add nodes that Azure knows about but have not checked in such that the `dead node` detection will eventually reimage the node.

NOTE: If node setup takes longer than NODE_EXPIRATION_TIME (1 hour), this will cause the nodes to continuously get reimaged.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/workers/nodes.py
+++ b/src/api-service/__app__/onefuzzlib/workers/nodes.py
@@ -49,6 +49,7 @@ class Node(BASE_NODE, ORMMixin):
         machine_id: UUID,
         scaleset_id: Optional[UUID],
         version: str,
+        new: bool = False,
     ) -> "Node":
         node = cls(
             pool_name=pool_name,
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -49,21 +49,25 @@ def create(
...
-        node.save()
-        send_event(
-            EventNodeCreated(
-                machine_id=node.machine_id,
-                scaleset_id=node.scaleset_id,
-                pool_name=node.pool_name,
+        # `save` only returns None when `new` is True and an object already
+        # exists in Azure Tables. If we're in that case, don't send an event.
+        result = node.save(new=new)
+        if result is None:
```
ranweiler, line 63: Per the comment, did you want:
```suggestion
        if result is not None:
```

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -300,6 +301,35 @@ def cleanup_nodes(self) -> bool:
                 )
                 node.delete()
 
+        # Scalesets can have nodes that never check in (such as broken OS setup
```
ranweiler, line 304: 👍 helpful explanation, thanks.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -300,6 +301,35 @@ def cleanup_nodes(self) -> bool:
...
+            if machine_id in node_machine_ids:
+                continue
+
+            logging.info(
+                "scaleset - adding missing azure node: %s:%s",
+                self.scaleset_id,
+                machine_id,
+            )
+
+            # Note, using `new=True` makes it such that if a node already has
```
ranweiler, line 323: 👍

src/api-service/__app__/onefuzzlib/workers/nodes.py
bmc-msft: The comment was worded oddly which lead to the misinterpretation.   Then I clicked the wrong button in github, which accepted this suggestion.  I reverted the suggestion and updated the comment.

ranweiler (approved):

PR #528, opened 2021-02-09
check libfuzzer `-help=1` prior to starting the heartbeat

bmc-msft:
Currently, the a bad libfuzzer (such as missing a prereq dll) transitions from `setting_up`, to `running`, then to `failure`.

This PR, in combination with #530, will make `--wait_for_running` actually useful in determining if the libfuzzer target is meaningfully launched.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/config.rs
+++ b/src/agent/onefuzz-agent/src/tasks/config.rs
@@ -162,7 +162,7 @@ impl Config {
         match self {
             Config::LibFuzzerFuzz(config) => {
                 fuzz::libfuzzer_fuzz::LibFuzzerFuzzTask::new(config)?
-                    .check_run()
+                    .managed_run()
                     .await
             }
             Config::LibFuzzerReport(config) => {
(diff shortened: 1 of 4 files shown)


chkeita (approved):

ranweiler (approved):

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -80,6 +80,11 @@ impl LibFuzzerFuzzTask {
         }
     }
 
+    pub async fn check_run(&self) -> Result<()> {
```
ranweiler, line 83: I think `managed_run()` would be more consistent with the coverage task.

bmc-msft, line 83:
```suggestion
    pub async fn managed_run(&self) -> Result<()> {
```

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -162,7 +162,7 @@ impl Config {
         match self {
             Config::LibFuzzerFuzz(config) => {
                 fuzz::libfuzzer_fuzz::LibFuzzerFuzzTask::new(config)?
-                    .run()
+                    .check_run()
```
bmc-msft, line 165:
```suggestion
                    .managed_run()
```

PR #530, opened 2021-02-09
Fix wait_for_running for templates

bmc-msft:
This fixes the checks from task & job state checking.

Additionally, this PR waits until we've seen at least one heartbeat before saying the task is fully `running`.

### DIFF
--- a/src/cli/onefuzz/templates/__init__.py
+++ b/src/cli/onefuzz/templates/__init__.py
@@ -233,7 +233,7 @@ class JobHelper:
         }
         self.wait_for_running = wait_for_running
 
-    def check_current_job(self) -> List[Task]:
+    def get_running_tasks_checked(self) -> List[Task]:
         self.job = self.onefuzz.jobs.get(self.job.job_id)
         if self.job.state in JobState.shutting_down():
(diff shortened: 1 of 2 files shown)


ranweiler (approved):

src/cli/onefuzz/templates/__init__.py
```diff
@@ -233,30 +233,38 @@ def wait_on(
         }
         self.wait_for_running = wait_for_running
 
-    def check_current_job(self) -> Job:
-        job = self.onefuzz.jobs.get(self.job.job_id)
-        if job.state in ["stopped", "stopping"]:
+    def check_current_job(self) -> List[Task]:
```
chkeita, line 236: i think this should be renamed to reflect the fact that we are getting the tasks
I propose 
`get_running_tasks_checked` or `check_running_job_tasks`

chkeita (approved):

src/cli/onefuzz/templates/__init__.py
bmc-msft, line 236:
```suggestion
    def get_running_tasks_checked(self) -> List[Task]:
```

PR #535, opened 2021-02-10
libfuzzer-dotnet integration

bmc-msft:


docs/how-to/fuzzing-dotnet-with-libfuzzer.md
```diff
@@ -0,0 +1,222 @@
+# Fuzzing .Net using libfuzzer-dotnet
+
+Fuzzing .Net using libFuzzer makes use of three Open Source capabilities.
```
chkeita, line 3: Is this only for dotnet core library ? if that is the case we should probably say it this doc

docs/how-to/fuzzing-dotnet-with-libfuzzer.md
```diff
@@ -0,0 +1,222 @@
...
+  ```
+  
+  For our example [problems](../../src/integration-tests/libfuzzer-dotnet/problems/) library, our callback for `Fuzzer.LibFuzzer.Run` is straight forwards.  `Func` already takes a `ReadOnlySpan<byte>`.  If your functions takes strings, this would be the place to convert the span of bytes to strings.
+  [wrapper/program.cs](../../src/integration-tests/libfuzzer-dotnet/wrapper/program.cs)
+  ```C#
+  using SharpFuzz;
+  namespace Wrapper {
+    public class Program {
+      public static void Main(string[] args) {
+        Fuzzer.LibFuzzer.Run(stream => { Problems.Problems.Func(stream); });
```
chkeita, line 62:
```suggestion 
        Fuzzer.LibFuzzer.Run(Problems.Problems.Func);
```

chkeita (approved):

docs/how-to/fuzzing-dotnet-with-libfuzzer.md
bmc-msft: I would prefer to leave this as is, primarily to give a clear point to do data conversion from the span of bytes into something else.

docs/how-to/fuzzing-dotnet-with-libfuzzer.md
```diff
@@ -0,0 +1,222 @@
+# Fuzzing .Net using libfuzzer-dotnet
+
```
chkeita, line 2:
``` suggestion
##Supported version
We currently support net45 framework or any version that support least netstandard1.6
Please refer to [this page](https://dotnet.microsoft.com/platform/dotnet-standard) check if your framework version is supported 
```

PR #545, opened 2021-02-12
set windows linker to rust-lld.exe

bmc-msft:


ranweiler (approved):
Could you just document (PR message is fine) the speedup you observed in switching, and any differences in artifact size?

PR #561, opened 2021-02-18
Update command variable expansion

bmc-msft:
* Documents `crashes_account` and `crashes_container`
* Adds `reports_dir` and support for `unique_reports`, `reports`, and `no_repro` containers to the generic analysis task
* Adds `microsoft_telemetry_key` and `instance_telemetry_key` to generic supervisor, generator, and analysis tasks


bmc-msft:
NOTE: the rename was driven because each time I wanted to use `telemetry_key`, I needed to refer back to the `onefuzz-telemetry` library.  This significantly reduces the burden to figure out which key does what.

However, this would be specifically good to get careful verification (both in code, as well as checking our telemetry recording) before this gets merged.

bmc-msft:
To clarify, the intent:

* `instrumentation_key` should now be called `instance_telemetry_key`
* `telemetry_key` should now be called `microsoft_telemetry_key`

src/agent/onefuzz-supervisor/src/config.rs
```diff
@@ -103,17 +107,19 @@ impl StaticConfig {
             None
         };
 
-        let instrumentation_key = if let Ok(key) = std::env::var("ONEFUZZ_INSTRUMENTATION_KEY") {
-            Some(Uuid::parse_str(&key)?)
-        } else {
-            None
-        };
+        let instance_telemetry_key =
```
bmc-msft, line 110: note: this is currently unexercised functionality intended to enable the `docker container` scenario.

chkeita (approved):

bmc-msft:
From integration tests:

```
ERROR:integration:task shutdown early: linux-libfuzzer-rust - libfuzzer_fuzz
ERROR:integration:task shutdown early: linux-libfuzzer-rust - libfuzzer_crash_report
ERROR:integration:task shutdown early: linux-libfuzzer-rust - libfuzzer_coverage
```

src/agent/onefuzz-agent/src/tasks/analysis/generic.rs
```diff
@@ -46,21 +53,59 @@ pub async fn spawn(config: Config) -> Result<()> {
     config.analysis.init().await?;
     config.tools.init_pull().await?;
 
+    // setup reports
```
ranweiler, line 56: This comment doesn't really give me any particular idea of what is supposed to happen below. Could you expand it a bit, maybe addressing our expectation for the corner case of "no syncdirs configured"?

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -26,11 +26,13 @@ pub struct CommonConfig {
 
     pub instance_id: Uuid,
 
-    pub instrumentation_key: Option<Uuid>,
-
     pub heartbeat_queue: Option<Url>,
 
-    pub telemetry_key: Option<Uuid>,
+    #[serde(alias = "instrumentation_key")]
```
ranweiler, line 31: For posterity, please document the reason for these aliases, and when we can remove them.

src/agent/onefuzz-supervisor/src/config.rs
```diff
@@ -25,9 +25,11 @@ pub struct StaticConfig {
 
     pub onefuzz_url: Url,
 
-    pub instrumentation_key: Option<Uuid>,
+    #[serde(alias = "instrumentation_key")]
```
ranweiler, line 28: Apply comment re serde aliases.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -261,10 +261,10 @@ mod global {
     }
 }
 
-pub fn set_appinsights_clients(ikey: Option<Uuid>, tkey: Option<Uuid>) {
-    let instance_client = ikey.map(|k| TelemetryClient::new(k.to_string()));
-    let shared_client = tkey.map(|k| TelemetryClient::new(k.to_string()));
-    global::set_clients(instance_client, shared_client);
+pub fn set_appinsights_clients(instance_key: Option<Uuid>, microsoft_key: Option<Uuid>) {
```
ranweiler, line 264: While we're here, maybe we should add newtypes/wrapper types for `Uuid` named `MicrosoftTelemetryKey` and `InstanceTelemetryKey` so we can't ever accidentally swap keys at call sites.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -318,7 +318,7 @@ pub fn set_property(entry: EventData) {
     let (key, value) = entry.as_values();
 
     if entry.can_share() {
```
ranweiler, line 320: Should we rename this property/predicate to include the name `microsoft`?

docs/command-replacements.md
```diff
@@ -18,6 +18,11 @@ The following values are replaced with the specific values at runtime.
 * `{setup_dir}` : Path to the setup directory
 * `{job_id}`: UUID that indicates the Job ID
 * `{task_id}`: UUID that indicates the Task ID
+* `{reports_dir}`: Path to the directory to write crash reports
+* `{crashes_account}`: Azure Storage Account name for the `crashes` container
+* `{crashes_container}`: Container name for the `crashes` container
+* `{microsoft_telemetry_key}`: Application Insights key used for [non-attributable telemetry](telemetry.md) collection.
```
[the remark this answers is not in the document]
bmc-msft: non-attributable is the phrase our existing (approved) documentation includes, but the other parts are good.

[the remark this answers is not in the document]
bmc-msft:
```suggestion
* `{microsoft_telemetry_key}`: Application Insights key used for collecting [non-attributable telemetry](telemetry.md) to improve OneFuzz.
```

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -6,10 +6,13 @@ use uuid::Uuid;
 
 pub use appinsights::telemetry::SeverityLevel::{Critical, Error, Information, Verbose, Warning};
 
+pub type MicrosoftTelemetryKey = Uuid;
```
ranweiler, line 9: Using a `type` alias is good documentation, but still lets us use the wrong key in the wrong place.

To get typeck to help us, we'd need to define types like:
```rust
// We could make the inner field `pub`, or be safer and add accessors/`From` impls (at the cost of boilerplate).
pub struct MicrosoftTelemetryKey(Uuid);
pub struct InstanceTelemetryKey(Uuid);
```
And then ideally update the config definitions to deserialize each UUID directly as the correct type. This is made easy in a struct context via the serde [`#[serde(transparent)]`](https://serde.rs/container-attrs.html#transparent) attribute.

ranweiler (approved):
Updates look good overall, but please note the type aliasing gotcha that I called out.

src/agent/onefuzz-telemetry/src/lib.rs
bmc-msft: added, please take a look

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -1,13 +1,36 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(transparent)]
+pub struct MicrosoftTelemetryKey(Uuid);
+impl MicrosoftTelemetryKey {
+    pub fn to_string(&self) -> String {
+        self.0.to_string()
+    }
+
+    pub fn new(value: Uuid) -> Self {
+        Self { 0: value }
```
ranweiler, line 16: Whoa, I didn't even know you could write it this way for tuple structs. TIL.

I'd suggest just returning `Self(value)` here and below, but this is purely stylistic.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -1,13 +1,36 @@
...
 
 pub use appinsights::telemetry::SeverityLevel::{Critical, Error, Information, Verbose, Warning};
 
-pub type MicrosoftTelemetryKey = Uuid;
-pub type InstanceTelemetryKey = Uuid;
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(transparent)]
+pub struct MicrosoftTelemetryKey(Uuid);
+impl MicrosoftTelemetryKey {
+    pub fn to_string(&self) -> String {
```
ranweiler: Consider adding actual impls for [`std::string::ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html), or even better, impl [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html), which could be useful generally, and triggers a [blanket impl](https://doc.rust-lang.org/std/string/trait.ToString.html#impl-ToString-4) for `ToString`.

ranweiler (approved):

Issue #562, opened 2021-02-18
Add support for multi-tenant authentication


andrew-slutsky:
Add support for multi-tenant authentication

PR #563, opened 2021-02-18
Added support for multi-tenant authentication

andrew-slutsky:

Added support for multi-tenant authentication and tested both with and without the deploy.py option. I created pools, scalesets, and submitted basic libfuzzer template jobs and verified a unique crash report was where I expected it to me. 

## PR Checklist
* [X] Applies to work item: #562
* [x] Tests passed
* [No] Requires documentation to be updated

## Info on Pull Request

The end-to-end changes needed to have onefuzz deployed with multi-tenant authentication.

## Validation Steps Performed

I ran deploy.py with and without the new argument "--multi_tenant_domain pmecloud.onmicrosoft.com" and verified the functionality by submitting a template libfuzzer job and finding a unique report in the expected crash report container.

E.g.
$ ./onefuzz-cli-2.5.0.exe containers files list oft-unique-reports-3543e721e2f856759fab8fd09b5582ed
{
    "files": [
        "482338715289d865cf40deec880abf777e21a8afcccfe1a24070c0f07d2b0975.json"
    ]
}


src/deployment/registration.py
```diff
@@ -504,6 +504,21 @@ def assign_scaleset_role(onefuzz_instance_name: str, scaleset_name: str) -> None
...
+        query_microsoft_graph(
+            method="PATCH",
+            resource="applications/%s" % objectId,
+            body={"signInAudience": "AzureADMultipleOrgs"},
+        )
+    except adal.AdalError:
+        raise Exception(
+            "error setting signInAudience in ad application %s: %s"
+            % (objectId, adal.AdalError)
+        )
```
chkeita, line 519: In this case can you print out the command to make the assignment see `assign_scaleset_role_manually`

anslutsk: Yeah, I noticed you did that. Why did you do that? If the "try" block fails then asking the user to run the command again right after this operation failed, isn't going to change the output? 

Regardless, I'm happy to add a similar line: "execute the following query in the azure portal bash shell: "
Are you suggesting I put the print line and replace the 'raise Exception' call?

src/agent/onefuzz-supervisor/src/auth.rs
```diff
@@ -87,26 +87,46 @@ pub struct ClientCredentials {
...
             resource,
+            multi_tenant_domain,
             tenant,
         }
     }
 
     pub async fn access_token(&self) -> Result<AccessToken> {
+        let (authority, resource) = if let Some(domain) = &self.multi_tenant_domain {
+            let url = Url::parse(&self.resource.clone())?;
+            let host = url.host_str().unwrap();
```
ranweiler, line 116: We haven't validated that we have a `host_str()` here, so this `unwrap()` may panic, which will abort the program uninformatively. Instead, we could return early with an `Result::Err` and error message.

The `anyhow::bail!()` macro lets us do this explicitly. Example:
```suggestion
            let host = if let Some(host) = url.host_str() {
                host
            } else {
                anyhow::bail!("resource URL does not have a host string: {}", url);
            }
```
Note that the type of the `else` block is unit, and does not match the first block. Since it is an early return, this is okay.

Alternately, you can use the [`Option::ok_or_else()` method](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) to translate `Option::Some(x)` to `Result::Ok(x)` and the `None` case to a `Result::Err(..)` variant of your choosing (using a closure, to avoid unnecessary computation). This would look like:
```suggestion
            let host = url
                .host_str()
                .ok_or_else(|| anyhow::format_err!("resource URL does not have a host string: {}", url)))?;
```
Note that we use `anyhow::format_err!()` to create a `Result::Err` value (with no early return), and the use of `?` to express "unwrap or return early at this site".

src/agent/onefuzz-supervisor/src/auth.rs
```diff
@@ -87,26 +87,46 @@ pub struct ClientCredentials {
...
+            let instance: Vec<&str> = host.split('.').collect();
+            (
```
ranweiler, line 118: Good use of tuples as an alternative to mutation.

src/deployment/registration.py
chkeita: We had an issue recently on linux machines that could not execute the registration code. We added this as a mitigation strategy to allow people to complete the registration in that case. 
Yes the suggestion is to replace the raise excepetion unless this registration is required for the rest of the script to continue.

src/deployment/deploy.py
```diff
@@ -371,12 +396,31 @@ def deploy_template(self) -> None:
...
+
         params = {
+            "app_func_audience": {"value": app_func_audience},
             "name": {"value": self.application_name},
             "owner": {"value": self.owner},
             "clientId": {"value": self.results["client_id"]},
             "clientSecret": {"value": self.results["client_secret"]},
+            "app_func_issuer": {"value": app_func_issuer},
             "signedExpiry": {"value": expiry},
+            "multi_tenant_domain": multi_tenant_domain,
```
chkeita, line 423:
``` Suggestion
            "multi_tenant_domain": {"value": multi_tenant_domain},
```

chkeita, line 423: seems like this key should not be added when the multi_tenant_domain is not defined. 
With the current code the value will be an empty string.
i propose you don't set the value when there are no multitenant domain and conditionally add the key like this 
``` Python
{
            "multi_tenant_domain" if multi_tenant_domain: multi_tenant_domain
}
```

same goes for `multi_tenant_domain`

src/agent/onefuzz-supervisor/src/auth.rs
```diff
@@ -147,20 +167,32 @@ impl From<ClientAccessTokenBody> for AccessToken {
...
+            multi_tenant_domain,
+        }
     }
 
     fn url(&self) -> Url {
         let mut url = Url::parse(MANAGED_IDENTITY_URL).unwrap();
-        url.query_pairs_mut()
-            .append_pair("resource", &self.resource);
+        let resource = if let Some(domain) = &self.multi_tenant_domain {
+            let uri = Url::parse(&self.resource).unwrap();
```
gdhuper, line 187: Just a suggestion, as pointed out before, this `unwrap` may panic if it's an invalid url. Since the return type is not `Result<>`,  an alternative could be using [expect](https://learning-rust.github.io/docs/e4.unwrap_and_expect.html#expect), which will still panic but allows you to pass a string message.

ranweiler: That's a better alternative, but still panics, which we want to avoid.

I suggest doing some validation earlier. For example, you could validate and precompute `multi_tenant_domain` in the ctor (`ManagedIdentityCredentials::new()`), which would then return a `Result`.

@anslutsk, using some stand-in values, could you share what we expect `resource` and a valid `managed_identity_url` to look like?

ranweiler (changes requested):
Please see comments about avoiding panics, and questions about the expected form of unvalidated inputs. The most useful fix might be to introduce a new validating wrapper type for multi-tenant domain URLs.

src/deployment/registration.py
anslutsk: Thanks for pointing this out. I was also using the wrong exception. So, my updated changes will require the user to manually make the change if the azure graph call fails, otherwise deployment will not continue until the setting is enabled. This will provide better support for Linux users, where the azure graph call fails.

src/deployment/deploy.py
```diff
@@ -323,6 +344,9 @@ def setup_rbac(self) -> None:
                     app.object_id, ApplicationUpdateParameters(app_roles=app_roles)
                 )
 
+            if self.multi_tenant_domain and app.sign_in_audience == "AzureADMyOrg":
+                assign_multi_tenant_auth(app.object_id)
```
chkeita, line 348: you should do assignment outside of the if clause so an exisiting deployment can be upgrades to multi tenant

anslutsk: Ahhhh, yes. I did not do testing back and forth between the settings on a deployment. I have just completed this. It works great now, even including the failure of the graph API.

anslutsk: I suppose this can be done once outside the clause but there is some consistency in the logic since the corresponding action sets the URI here inside the 'if' segment:

            params = ApplicationCreateParameters(
                display_name=self.application_name,
                identifier_uris=[url],
                reply_urls=[url + "/.auth/login/aad/callback"],
                optional_claims=OptionalClaims(id_token=[], access_token=[]),
                required_resource_access=[
                    RequiredResourceAccess(
                        resource_access=[
                            ResourceAccess(id=USER_IMPERSONATION, type="Scope")
                        ],
                        resource_app_id="00000002-0000-0000-c000-000000000000",
                    )
                ],
                app_roles=app_roles,
                
So, for this reason I think we should keep it in both segments even though it appears redundant.

src/agent/onefuzz-supervisor/src/auth.rs
anslutsk: I put in some temporary println statements to provide in context here.

In the case where OneFuzz is deployed in single tenant mode, the config.json file contains an entry for "multi_tenant_config" which has value "null" and the self.resource String value is "https://anslutsk-26.azurewebsites.net" and thus, the pattern "if let Some(domain)" is not triggered, so the else clause is run, which executes the line: self.resource.clone().

In the case where OneFuzz is deployed in multi tenant mode, the config.json file contains the entry "multi_tenant_config" with the value "mspmecloud.onmicrosoft.com" and the self.resource String value is "https://anslutsk-26.azurewebsites.net" and so the pattern "if let Some(domain)" is triggered, so the corresponding code block is executed and the value of self.instance[0] is "anslutsk-26"

anslutsk: When there was an invalid domain in the "multi_tenant_config" json config item, for example the String value "**....**" the program did not crash and we get the expected error during while trying to obtain a Bearer token from IMS:

[2021-02-24T07:12:19Z ERROR onefuzz_telemetry] error running supervisor agent: 400 Bad Request: {"error":"invalid_resource","error_description":"AADSTS500011: The resource principal named https://**....**/anslutsk-multi-26 was not found in the tenant named 975f013f-7f24-47e8-a7d3-abc4752bf346. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You might have sent your authentication request to the wrong tenant.\r\nTrace ID: 0eec3338-bfa4-49f1-a7aa-7d13f9842200\r\nCorrelation ID: 02d18935-c309-4c71-8757-7a9770a98805\r\nTimestamp: 2021-02-24 07:12:19Z","error_codes":[500011],"timestamp":"2021-02-24 07:12:19Z","trace_id":"0eec3338-bfa4-49f1-a7aa-7d13f9842200","correlation_id":"02d18935-c309-4c71-8757-7a9770a98805","error_uri":"https://westus2.login.microsoft.com/error?code=500011"}

In addition:

When there was an invalid domain in the "multi_tenant_config" json config item, for example the String value "**asdfasdfasdf**" the program did not crash and we get the expected error during while trying to obtain a Bearer token from IMS:

[2021-02-24T07:16:33Z ERROR onefuzz_telemetry] error running supervisor agent: 400 Bad Request: {"error":"invalid_resource","error_description":"AADSTS500011: The resource principal named https://**asdfasdfasdf**/anslutsk-multi-26 was not found in the tenant named 975f013f-7f24-47e8-a7d3-abc4752bf346. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You might have sent your authentication request to the wrong tenant.\r\nTrace ID: b0c8b83c-e552-4f2f-9f77-d280785d2600\r\nCorrelation ID: 1982fe67-c50a-4e0a-ac63-186134839aae\r\nTimestamp: 2021-02-24 07:16:33Z","error_codes":[500011],"timestamp":"2021-02-24 07:16:33Z","trace_id":"b0c8b83c-e552-4f2f-9f77-d280785d2600","correlation_id":"1982fe67-c50a-4e0a-ac63-186134839aae","error_uri":"https://westus2.login.microsoft.com/error?code=500011"}

I do not see the value in adding additional error handling. Am I missing something?

anslutsk: The value assigned to self.resource is taken from the config item "onefuzz_url" and if the value is invalid, the program crashes like this:

[2021-02-24T15:54:35Z ERROR onefuzz_telemetry] error loading supervisor agent config: invalid value: string "asdf", expected description() is deprecated; use Display at line 1 column 50

If the "onefuzz_url" is set to "null" the following error is displayed:

[2021-02-24T16:03:23Z ERROR onefuzz_telemetry] error loading supervisor agent config: invalid type: null, expected a string representing an URL at line 1 column 48
Error: invalid type: null, expected a string representing an URL at line 1 column 48

anslutsk: I don't think it's necessary to change this section of the code, but that doesn't mean that I don't support making a change! If either of you would like to suggest a specific change then I'd be happy to test it. It would need to be pretty specific though since I don't share either of your view points.

ranweiler: Thanks for sharing examples of expected/valid inputs, and your investigation into the current failure modes.

> I do not see the point in adding additional error handling. Am I missing something?

We still want to change what we're doing here, because this de facto current unreachability is very non-local, and depends on the behavior of the larger system. But if we ever accidentally introduce a bug elsewhere that makes the `unwrap()` reachable, we want to fail in a way that will be easier to understand and fix, and get [reported here](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-supervisor/src/main.rs#L109-L113).

So, the constraint here is that we must not `unwrap()`. We could make `ManagedIdentityCredentials::url()` return a `Result<Url>`, but that would be a bit silly, since we'd then have to change all use sites, and a non-`Ok()` result is actually an error that we can detect much earlier (in the constructor).

Instead, I suggest changing `ManagedIdentityCredentials::new()` to have the signature `new(Url, Option<Url>) -> Result<Self>`, taking care to generate the `resource` string _without_ a trailing `/` in the non-multi-tenant case.

anslutsk (commented):
I'm confused

anslutsk:
testing123

anslutsk:
testing123123

src/agent/onefuzz-supervisor/src/auth.rs
anslutsk: test123

anslutsk (commented):
test

bmc-msft:
OBE: #746

PR #565, opened 2021-02-18
stop jobs that do not start within 30 days

bmc-msft:
If a job does not start within 30 days, stop the job and mark all of the tasks as `failed`.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/jobs.py
+++ b/src/api-service/__app__/onefuzzlib/jobs.py
@@ -7,13 +7,17 @@ import logging
 from datetime import datetime, timedelta
 from typing import List, Optional, Tuple
 
-from onefuzztypes.enums import JobState, TaskState
+from onefuzztypes.enums import ErrorCode, JobState, TaskState
 from onefuzztypes.events import EventJobCreated, EventJobStopped
+from onefuzztypes.models import Error
 from onefuzztypes.models import Job as BASE_JOB
 
 from .events import send_event
 from .orm import MappingIntStrAny, ORMMixin, QueryFilter
 from .tasks.main import Task
+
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/jobs.py
```diff
@@ -36,6 +40,34 @@ def search_expired(cls) -> List["Job"]:
...
+            (datetime.utcnow() - JOB_NEVER_STARTED_DURATION).isoformat()
+        )
+        for job in cls.search(
+            query={
+                "state": [JobState.enabled],
+            },
+            raw_unchecked_filter=time_filter,
+        ):
+            # if the job has started, leave it be
+            if job.end_time is not None:
```
chkeita, line 55: can this be added to the db query?

chkeita (approved):

src/api-service/__app__/onefuzzlib/jobs.py
bmc-msft: I got a failure when I tried in a fresh instance, as the field won't be created yet.

chkeita: I am ok to leave it as is but there is trick to check fields that might not exist we use it to [find outdated nodes](https://github.com/microsoft/onefuzz/blob/657af9722ceae3acc2e027f4181b3c62e96555fd/src/api-service/__app__/onefuzzlib/workers/nodes.py#L113)

bmc-msft: oh neat.  I'll investigate that.

bmc-msft: Updated following that example.

chkeita (approved):

PR #567, opened 2021-02-18
allow nodes with multiple tasks to continue on task stop

bmc-msft:
As is, when multiple tasks are running on a single node, if any one of them stops, the node gets reimaged.

This changes the behavior such that when a node with multiple tasks has one task stop, the other tasks will continue.


bmc-msft:
Note, this should get specific testing from the service/node interaction, as this is the first use of the stop_task node command

ranweiler (approved):

PR #568, opened 2021-02-18
clarify scaleset logging

bmc-msft:


src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -217,7 +224,10 @@ def setup(self) -> None:
                 self.set_failed(pool)
                 return
 
-            logging.info("creating scaleset: %s", self.scaleset_id)
+            logging.info(
+                SCALESET_LOG_PREFIX + "creating scaleset scalset_id:%s",
```
ranweiler, line 228:
```suggestion
                SCALESET_LOG_PREFIX + "creating scaleset. scaleset_id:%s",
```

ranweiler: You change convention in these messages wrt initial capitalization, and the boundary between the context message and the key/value data. Consider doing a pass and making it the same throughout the touched code.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -150,7 +151,7 @@ def set_failed(self, error: Error) -> None:
     def init(self) -> None:
         from .pools import Pool
 
-        logging.info("scaleset init: %s", self.scaleset_id)
+        logging.info(SCALESET_LOG_PREFIX + "init. scaleset_id:%s", self.scaleset_id)
```
ranweiler, line 154: Is there a reason to concat using `+` vs. going all-in on interpolation? Example of what I mean:
```suggestion
        logging.info("%s init. scaleset_id:%s", SCALESET_LOG_PREFIX, self.scaleset_id)
```
(removing the trailing space from the prefix constant)

ranweiler (approved):
In the future, it might be nice to have a generic mechanism for context-scoped logging.

bmc-msft:
> In the future, it might be nice to have a generic mechanism for context-scoped logging.

Agreed.

Eventually, it would be great to move onto `opentelemetry` for similar reasons we're wanting to do that with Rust.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
bmc-msft: We use this method in a few places already, as such thought it would be the least impact.

bmc-msft: My preference would be to use a context scoped logging impl, something on top of opentelemetry.  Once it's available in public preview in Azure, we should move to it.

PR #570, opened 2021-02-18
Add symbol cache and filtering

### LINKED ISSUE #285: Cache module disassembly for generic coverage recording
In a `generic_coverage` task, we will definitely disassemble basic blocks for the same modules, over and over again, between runs. This may happen within runs, too, as an edge case, but it is guaranteed between.

Add support for using a local cache so we only ever disassemble on the first run.

ranweiler:
- Add caching to symbol table-driven module disassembly on Linux.
- Add configurable regex-based filtering for coverage collection, by module and module-scoped symbol name.

Block coverage recording can be manually tested using the `block_coverage` example in the `coverage` crate. See `./block_coverage -h` for expected args.

The filter file is optional. The file format is JSON like this:
```json
{
    "modules": {
        "allow": [
            "<module-path-regex-1>",
            "<module-path-regex-2>",
        ]
    },
    "symbols": {
        "<module-path-regex-1>": {
            "allow": [
                "<symbol-name-regex-1>",
                "<symbol-name-regex-2>",
            ]
        },
        "<module-path-regex-2>": {
            "deny": [
                "<symbol-name-regex-3>",
                "<symbol-name-regex-4>",
            ]
        }
    }
}
```

Closes #285.


src/agent/coverage/src/region.rs
```diff
@@ -0,0 +1,128 @@
...
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::ops::Range;
+
+use serde::{Deserialize, Serialize};
+
+/// A non-overlapping set of regions of program data.
+#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(transparent)]
+pub struct RegionIndex<R> {
```
ranweiler, line 13: The impl of this type deserves unit tests. I will add those when I update it to support overlapping and duplicate regions.

ranweiler:
Hmm, maybe a tagged repr would be better for the filter rules. Example of such an alternative:

```json
{
    "modules": {
        "type": "allow",
        "matches": [
            "<module-path-regex-1>",
            "<module-path-regex-2>"
        ]
    },
    "symbols": {
        "<module-path-regex-1>": {
            "type": "allow",
            "matches": [
                "<symbol-name-regex-1>",
                "<symbol-name-regex-2>"
            ]
        },
        "<module-path-regex-1>": {
            "type": "deny",
            "matches": [
                "<symbol-name-regex-3>",
                "<symbol-name-regex-4>"
            ]
        }
    }
}
```
The existing repr is a little weird because we only permit either an `allow` xor `deny` rule, so each rule serializes as a filter object with exactly one key.

src/agent/coverage/src/block/linux.rs
```diff
@@ -245,13 +251,17 @@ impl Images {
 #[derive(Clone, Debug, PartialEq)]
 pub struct ModuleImage {
     map: MemoryMap,
+    path: ModulePath,
```
bmc-msft, line 254: Why do you wrap `PathBuf` rather than just using `PathBuf` directly?

ranweiler: Its ctor validates some additional conditions on a `PathBuf`, which lets us treat otherwise-partial and fallible accessors as total. It also documents the provenance of the path, and lets us keep our resulting assumptions in one place.

src/agent/coverage/src/block/linux.rs
```diff
@@ -130,60 +150,46 @@ impl Recorder {
...
 
     fn on_module_load(&mut self, tracee: &mut Tracee, image: &ModuleImage) -> Result<()> {
-        log::info!("module load: {}", image.path().display());
-
-        let blocks = find_module_blocks(image.path())?;
+        log::info!("module load: {}", image.path());
 
-        log::debug!("found {} blocks", blocks.len());
+        // Fetch disassembled module info via cache.
+        let info = self.modules.fetch(image.path())?;
```
bmc-msft, line 156: Rather than handling this option via is_none+bail, followed by unwrap later on, use `ok_or_else` to convert the missing fetch to a failure.

```suggestion
        let info = self.modules.fetch(image.path())?.ok_or_else(|| format_err("unable to fetch info for module: {}", image.path()))?;
```

src/agent/coverage/src/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+        where
+            A: SeqAccess<'d>,
+        {
+            let mut patterns = Vec::<String>::new();
+
+            while let Some(p) = seq.next_element()? {
+                patterns.push(p);
+            }
+
+            let regexes = RegexSet::new(patterns).unwrap();
```
bmc-msft, line 139: If I'm reading this correctly, this aborts if the user provides an invalid regex.  Instead of abort, the error condition should be raised.

src/agent/coverage/examples/block_coverage.rs
```diff
@@ -28,42 +29,70 @@ fn main() -> Result<()> {
...
-            };
-
-            let marker = if b.count == 0 { " " } else { "x" };
-
-            log::debug!("  [{}] {}+{:x}", marker, name, b.offset);
+            }
+
+            let marker = if block.count == 0 { " " } else { "x" };
+
+            let module = recorder.modules.cached.get(module_path).unwrap();
```
bmc-msft, line 81: It would be useful to use `expect()` here to explain what the failure is in the case of an abort.

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
+
+    pub fn path_lossy(&self) -> String {
+        self.path.to_string_lossy().into_owned()
+    }
+
+    pub fn name(&self) -> &OsStr {
+        // Unwrap checked in ctor.
```
bmc-msft, line 59:
```suggestion
        // Unwrap checked in constructor.
```

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+pub struct SymbolFilter {
+    /// Pre-compiled regex set for fast matching.
+    regexes: RegexSet,
+
+    /// Maps module name regexes to filters for symbol names.
+    filters: BTreeMap<usize, Filter>,
+}
+
+impl SymbolFilter {
+    pub fn new(mut spec: SymbolFilterSpec) -> Result<Self> {
```
bmc-msft, line 270: The construct of using `spec.filters.keys()` followed by the values later on by string/pattern is difficult to reason about.

This would be simplified using [RegexSet::patterns](https://docs.rs/regex/1.1.0/regex/struct.RegexSet.html), rather than saving a copy of the string representations of the regex.

src/agent/coverage/src/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+#[serde(transparent)]
+pub struct Allow {
+    #[serde(with = "self::regex_set")]
+    regexes: RegexSet,
+}
+
+impl Allow {
+    /// Build a filter that allows all names.
+    pub fn all() -> Self {
+        let regexes = RegexSet::new(&[".*"]).unwrap();
```
bmc-msft, line 41:
```suggestion
        Self::new(&[".*"]).unwrap()
```

src/agent/coverage/src/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+        Self { regexes }
```
bmc-msft, line 42:
```suggestion
```

src/agent/coverage/src/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+pub struct Deny {
+    #[serde(with = "self::regex_set")]
+    regexes: RegexSet,
+}
+
+impl Deny {
+    /// Build a filter that allows all names.
+    pub fn none() -> Self {
+        let empty: &[&str] = &[];
+        let regexes = RegexSet::new(empty).unwrap();
```
bmc-msft, line 77:
```suggestion
        Self::new(empty).unwrap()
```

src/agent/coverage/src/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+        Self { regexes }
```
bmc-msft, line 78:
```suggestion
```

bmc-msft (approved):
see re style comments.  Otherwise, LGTM

src/agent/coverage/src/block/linux.rs
```diff
@@ -96,7 +111,11 @@ impl Recorder {
     fn on_breakpoint(&mut self, tracee: &mut Tracee) -> Result<()> {
         let mut regs = tracee.registers()?;
 
-        log::trace!("hit breakpoint: {:x} (~{})", regs.rip - 1, tracee.pid);
+        log::trace!(
+            "hit breakpoint: pc = {:x}, pid = {}",
+            regs.rip - 1,
```
lzybkr, line 116: I suggest using the local variable (computed below, so move up) instead.

src/agent/coverage/src/cache.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn fetch(&mut self, path: &ModulePath) -> Result<Option<&ModuleInfo>> {
+        if !self.cached.contains_key(path) {
+            self.insert(path)?;
+        }
+
+        Ok(self.cached.get(path))
+    }
+
+    pub fn insert(&mut self, path: &ModulePath) -> Result<()> {
+        let entry = ModuleInfo::new_elf(path)?;
```
lzybkr, line 32: Should this function be Linux only, or will you have an equivalent Windows api so just this line needs to be conditionally compiled?

src/agent/coverage/src/cache.rs
```diff
@@ -0,0 +1,56 @@
...
+pub struct ModuleInfo {
+    /// Index of the module segments and symbol metadata.
+    pub module: ModuleIndex,
+
+    /// Set of image offsets of basic blocks.
+    pub blocks: BTreeSet<u64>,
+}
+
+impl ModuleInfo {
+    pub fn new_elf(path: &ModulePath) -> Result<Self> {
```
lzybkr, line 48: Maybe add: `#[cfg(target_os = "linux")]`

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        if path.file_name().is_none() {
+            bail!("module path has no file name");
+        }
+
+        Ok(Self { path })
+    }
+
+    pub fn existing(path: impl AsRef<Path>) -> Result<Self> {
+        let path = path.as_ref().canonicalize()?;
```
lzybkr, line 46: Canonicalize does weird things on Windows - consider using [dunce::canonicalize](https://docs.rs/dunce/1.0.1/dunce/fn.canonicalize.html).

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ModuleIndex {
+    pub path: ModulePath,
+    pub base_va: u64,
+    pub symbols: SymbolIndex,
+}
+
+impl ModuleIndex {
+    pub fn parse_elf(path: ModulePath, data: &[u8]) -> Result<Self> {
```
lzybkr, line 100: Maybe add `#[cfg(target_os = "linux")]`.

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub image_offset: u64,
+
+    /// Total size in bytes of the symbol definition.
+    pub size: u64,
+}
+
+impl Symbol {
+    pub fn contains_file_offset(&self, offset: u64) -> bool {
+        let lo = self.file_offset;
+        let hi = lo + self.size;
```
lzybkr, line 232: Should this be a checked add?

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+        (lo..hi).contains(&offset)
+    }
+
+    pub fn contains_image_offset(&self, offset: u64) -> bool {
+        let lo = self.image_offset;
+        let hi = lo + self.size;
```
lzybkr, line 238: Should this be a checked add?

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+pub struct SymbolFilterSpec {
+    filters: HashMap<String, Filter>,
+}
+
+#[derive(Clone, Debug)]
+pub struct SymbolFilter {
+    /// Pre-compiled regex set for fast matching.
+    regexes: RegexSet,
+
+    /// Maps module name regexes to filters for symbol names.
```
lzybkr, line 265: Can you expand this comment to better explain the key is an index into the `regexes` field?

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    pub fn is_allowed(&self, module: &ModulePath, name: &str) -> bool {
+        let rules = self.regexes.matches(&module.path_lossy());
+
+        // Check if there is some symbol rule for the module path.
+        //
+        // If many rules would apply, the first rule is used.
+        if let Some(idx) = rules.iter().next() {
+            // Guaranteed by ctor.
```
lzybkr, line 292: s/ctor/constructor/

src/agent/coverage/src/disasm.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(Self { module, data })
+    }
+
+    /// Find block entry points for every symbol in the module.
+    pub fn find_blocks(&self) -> BTreeSet<u64> {
+        let mut blocks = BTreeSet::new();
+
+        for symbol in self.module.symbols.iter() {
+            let sym_blocks = self.find_symbol_blocks(symbol);
+            blocks.extend(sym_blocks);
```
lzybkr, line 27: Maybe `find_symbol_blocks` should modify an existing BTreeSet instead of constructing a new BTreeSet.

src/agent/coverage/src/disasm.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        blocks
+    }
+
+    /// Find all entry points for blocks contained within the region of `symbol`.
+    pub fn find_symbol_blocks(&self, symbol: &Symbol) -> BTreeSet<u64> {
+        // Slice the symbol's instruction data from the module file data.
+        let data = {
+            let lo = symbol.file_offset as usize;
+            let hi = lo + (symbol.size as usize);
```
lzybkr, line 38: Should this be a checked add?

src/agent/coverage/src/disasm.rs
```diff
@@ -0,0 +1,110 @@
...
+            &self.data[lo..hi]
+        };
+
+        // Initialize a decoder for the current symbol.
+        let mut decoder = Decoder::new(64, data, DecoderOptions::NONE);
```
lzybkr, line 43: Presumably we check the bitness of the target elsewhere?

src/agent/coverage/src/disasm.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        // Compute the VA of the symbol, assuming preferred module base VA.
+        let va = self.module.base_va + symbol.image_offset;
```
lzybkr, line 46: Do we need a checked add?

lzybkr (approved):

src/agent/coverage/src/block/linux.rs
```diff
@@ -130,60 +150,46 @@ impl Recorder {
...
 
-        if new {
-            for b in blocks {
-                let va = image.offset_to_va(b.offset);
-                log::trace!("set breakpoint: {:x} (~{})", va, tracee.pid);
-                self.breakpoints.set(tracee, va)?;
+        // Collect blocks allowed by the symbol filter.
+        let mut allowed_blocks = vec![];
+
+        for symbol in info.module.symbols.iter() {
```
chkeita, line 167: i believe you can achieve the same without allocating the vector first with flat_map

``` rust
let allowed_blocks =
      info.module.symbols.iter()
      .filter(|symbol| {
          self.symbol_filter
              .is_allowed(&info.module1.path, &symbol.name)
      })
      .flat_map(|symbol| info.blocks.range(symbol.range()))
      .collect();
```

src/agent/coverage/src/code.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        Ok(Self { regexes, filters })
+    }
+
+    pub fn is_allowed(&self, module: &ModulePath, name: &str) -> bool {
+        let rules = self.regexes.matches(&module.path_lossy());
+
+        // Check if there is some symbol rule for the module path.
+        //
+        // If many rules would apply, the first rule is used.
```
chkeita, line 290: Just wondering if there is any concern about impact of the ordering of the rules
It seems that this function could give a different result depending on which rule came first.

chkeita (approved):

bmc-msft (changes requested):
Marking this as request-changes so we don't accidentally merge as is, given the number of comments.  @ranweiler , can you resolve the comments (either close and or address them)?

src/agent/coverage/examples/block_coverage.rs
```diff
@@ -28,42 +29,70 @@ fn main() -> Result<()> {
...
 
-    let argv: Vec<_> = env::args().skip(1).collect();
-    if argv.is_empty() {
-        anyhow::bail!("empty target argv");
-    }
-    let mut cmd = Command::new(&argv[0]);
-    if argv.len() > 1 {
-        cmd.args(&argv[1..]);
-    }
+    let opt = Opt::from_args();
```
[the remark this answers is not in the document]
ranweiler: Yep, I'll be cleaning up and unifying the platforms in a follow-up PR.

src/agent/coverage/src/block.rs
```diff
@@ -1,35 +1,72 @@
...
+pub struct CommandBlockCov {
+    modules: BTreeMap<ModulePath, ModuleCov>,
+}
+
+impl CommandBlockCov {
+    /// Return early with `false` if the module did not need to be inserted.
+    ///
+    /// Else initializes the block coverage map for the module.
+    pub fn insert(&mut self, path: &ModulePath, offsets: impl Iterator<Item = u64>) -> bool {
+        if self.modules.contains_key(path) {
```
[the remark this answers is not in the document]
ranweiler: I didn't want to `clone()` the `path` argument, which is required by the signature of `entry()`. But I think most calls to this method will be for uninitialized modules, which means we'd `clone()` anyway. The rewrite is more readable, too.

src/agent/coverage/src/cache.rs
ranweiler: We'll conditionally compile in some form (follow-up PR).

src/agent/coverage/src/code.rs
ranweiler: If the add were to overflow, the `Symbol` info itself is nonsense. This could happen with a malformed binary file, so we should really validate earlier.

src/agent/coverage/src/code.rs
ranweiler: Will address by validating when we construct the `Symbol`.

src/agent/coverage/src/code.rs
ranweiler: Good call.

src/agent/coverage/src/code.rs
ranweiler: It is definitely order-dependent. I consider it reasonable because the order is exposed to and fully controlled by the user.

My thinking was:
1. We're letting users provide multiple regexes to match symbols and modules (vs. _exact_ names).
2. We cannot statically know if the regexes will overlap.
3. By (1) and (2), we need to be able to handle conflicts / multiple matches.
4. We want the conflict handling to be predictable by the user, and debuggable.

Deferring to the user-provided "definition order" seemed like it'd be easy to understand and document. I'm open to other ideas!

src/agent/coverage/src/disasm.rs
ranweiler: I like that. We could always wrap a mutating method in one that returns a `BTreeSet`, if we wanted.

src/agent/coverage/src/disasm.rs
ranweiler: My other comments about earlier validation apply here. I'll add `Range`-returning methods to `Symbol` to encapsulate this.

src/agent/coverage/src/disasm.rs
ranweiler: As exposed, this function could actually panic, if an invalid `Symbol` was passed. Fixing that, too.

src/agent/coverage/src/disasm.rs
ranweiler: We don't right now, as this is x64-only on Linux. Will open an issue.

ranweiler: Captured in #606.

src/agent/coverage/src/block/linux.rs
ranweiler: True, though I find this style harder to read, so I think I'm going to leave as-is.

src/agent/coverage/src/disasm.rs
ranweiler: Yes, that'd be good to do here, since we can't locally validate that in the `Symbol` ctor.

src/agent/coverage/src/filter.rs
ranweiler: Good catch! That `unwrap()` wasn't meant to hang around.

src/agent/coverage/src/code.rs
ranweiler: Sorry, I'm not really sure what you're suggesting here. I'll add some comments and we can revisit.

ranweiler: Ah! Actually, there's some implementation-determined runtime behavior in the module matching that we really don't want. 

I'm going to revisit this to be array-based (on the user/JSON side), and I think that will make the implementation easier to read and understand.

src/agent/coverage/src/code.rs
ranweiler: Actually, the outer _module_ matching is not fixed and user-controlled, it is controlled by the implementation of `HashMap`. I'm going to revisit this to make my above comments true at the module level, too.

src/agent/coverage/src/code.rs
ranweiler: (I plan on addressing this in a follow-up PR).

src/agent/coverage/src/code.rs
ranweiler: Will address in a follow-up PR.

bmc-msft (approved):

lzybkr (approved):

PR #582, opened 2021-02-19
handle error condition when creating a container that is being deleted

bmc-msft:
When users try to create a container immediately after deleting it, Azure will fail saying the deletion is in-progress.

catching ResourceExistsError during create handles this error.


bmc-msft:
The exception we get back looks like:
```
Exception while executing function: Functions.containers Result: Failure
Exception: ResourceExistsError: The specified container is being deleted. Try operation later.
RequestId:5770196c-b01e-0036-0717-079780000000
Time:2021-02-19T23:35:23.2776166Z
ErrorCode:ContainerBeingDeleted
Error:None
```

ranweiler (approved):

PR #585, opened 2021-02-22
handle delayed AAD resources in deployments

bmc-msft:
This handles exceptions generated during deployment when newly created AAD resources are not yet available for use by the ARM deployment.

An example error message:

```
Message: {'error': {'code': 'PrincipalNotFound', 'message': 'Principal HEXGUID does not exist in the directory GUID.'}}
```


src/deployment/deploy.py
```diff
@@ -384,16 +384,36 @@ def deploy_template(self) -> None:
...
+            try:
+                result = client.deployments.create_or_update(
+                    self.resource_group, gen_guid(), deployment
+                ).result()
+                if result.properties.provisioning_state != "Succeeded":
+                    logger.error(
+                        "error deploying: %s",
+                        json.dumps(result.as_dict(), indent=4, sort_keys=True),
+                    )
+                    sys.exit(1)
```
ranweiler, line 402: Is there a reason we `exit(1)` here, but throw an exception in other error paths? Is the discrepancy just a refactoring artifact?

src/deployment/registration.py
```diff
@@ -235,7 +235,10 @@ def add_application_password(app_object_id: UUID) -> Tuple[str, str]:
             return add_application_password_impl(app_object_id)
         except GraphQueryError as err:
             error = err
-            logging.warning("unable to create app password: %s", err.message)
+            if "Request_ResourceNotFound" in repr(err):
```
ranweiler, line 238: Can we cite something for the origin of this search substring? Or at least comment on where it is empirically derived?

Same applies to `PrincipalNotFound`, above.

ranweiler (approved):

src/deployment/deploy.py
bmc-msft: This is a refactoring artifact.

bmc-msft: This is a refactoring artifact.

chkeita (approved):

src/deployment/registration.py
bmc-msft: I added some refs to how this is similar to az-cli

Issue #542, opened 2021-02-11
LibFuzzer coverage task failed to synchronize coverage to container


jagunter:
## Information

- Onefuzz version: 2.4.1
- OS: MicrosoftWindowsDesktop:Windows-10:20h2-pro:latest

## Provide detailed reproduction steps (if any)

Job was submitted with the following arguments:

```
onefuzz template libfuzzer basic --setup_dir <setup> --target_exe <target> --inputs <inputs> --duration 72 --notification_config <config> --reboot_after_setup -v <project> <name> <build> <pool>
```

1. Submit artifacts of job id `40eee076-df8b-4c4f-b6cd-4970625fde1b` from the internal instance. 
2. Wait ~20 minutes
3. Verify coverage files are not synchronized to container
4. Connect to host directly via SSH or RDP and confirm coverage files have been generated locally

### Expected result

Coverage files would appear in container

### Actual result

Coverage files do not appear in container

bmc-msft:
@jagunter we fixed an issue that we believe addresses this issue, #542.

PR #588, opened 2021-02-23
Actively tail worker stdio from supervisor agent

ranweiler:
In the supervisor agent, incrementally read from the running worker agent's redirected stderr and stdout, instead of waiting until it exits.

The worker agent's stderr and stdout are piped to the supervisor when tasks are run. The supervisor's `WorkerRunner` does _not_ use `wait_with_output()`, which handles this (at the cost of blocking). Instead, it makes repeated calls to to `try_wait()` on timer-based state transitions, and does not try to read the pipes until the worker exits. But when one of the child's pipes is full, the child can block forever waiting on a `write(2)`, such as in a `log` facade implementation.

This bug has not been caught because we control the child worker agent, and until recently, it mostly only wrote to these streams using `env_logger` at its default log level. But recent work: (1) set more-verbose `INFO` level default logging, (2) logged stderr/stdout lines of child processes of _the worker_, and (3) some user targets logged very verbosely for debugging. This surfaced the underlying issue.

### DIFF
--- a/src/agent/onefuzz-supervisor/src/worker.rs
+++ b/src/agent/onefuzz-supervisor/src/worker.rs
@@ -251,8 +251,14 @@ struct RedirectedChild {
 
 impl RedirectedChild {
     pub fn new(mut child: Child) -> Result<Self> {
-        let stderr = child.stderr.take().ok_or_else(|| format_err!("onefuzz-agent stderr not piped"))?;
-        let stdout = child.stdout.take().ok_or_else(|| format_err!("onefuzz-agent stdout not piped"))?;
+        let stderr = child
+            .stderr
+            .take()
+            .ok_or_else(|| format_err!("onefuzz-agent stderr not piped"))?;
+        let stdout = child
+            .stdout
+            .take()
+            .ok_or_else(|| format_err!("onefuzz-agent stdout not piped"))?;
         let streams = Some(StreamReaderThreads::new(stderr, stdout));
         Ok(Self { child, streams })
     }
@@ -305,8 +311,16 @@ impl StreamReaderThreads {
     }
 
     pub fn join(self) -> Result<CapturedStreams> {
-        let stderr = self.stderr.join().map_err(|_| format_err!("stderr tail thread panicked"))?.to_string_lossy();
-        let stdout = self.stdout.join().map_err(|_| format_err!("stdout tail thread panicked"))?.to_string_lossy();
+        let stderr = self
+            .stderr
+            .join()
+            .map_err(|_| format_err!("stderr tail thread panicked"))?
+            .to_string_lossy();
+        let stdout = self
+            .stdout
+            .join()
+            .map_err(|_| format_err!("stdout tail thread panicked"))?
+            .to_string_lossy();
 
         Ok(CapturedStreams { stderr, stdout })
     }
@@ -317,7 +331,9 @@ impl IWorkerChild for RedirectedChild {
         let output = if let Some(exit_status) = self.child.try_wait()? {
             let exit_status = exit_status.into();
             let streams = self.streams.take();
-            let streams = streams.ok_or_else(|| format_err!("onefuzz-agent streams not captured"))?.join()?;
+            let streams = streams
+                .ok_or_else(|| format_err!("onefuzz-agent streams not captured"))?
+                .join()?;
 
             Some(Output {
                 exit_status,
@@ -388,7 +404,6 @@ impl std::io::Write for CircularBuffer {
         Ok(())
     }
 }
-
 
 #[cfg(test)]
 pub mod double;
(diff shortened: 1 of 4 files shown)


ranweiler:
This is probably the root cause for #542.

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -228,8 +228,6 @@ impl IWorkerRunner for WorkerRunner {
         cmd.arg("managed");
         cmd.arg("config.json");
         cmd.arg(setup_dir);
-        cmd.stderr(Stdio::piped());
-        cmd.stdout(Stdio::piped());
```
bmc-msft, line 232: This isn't a viable fix.

We capture stdout & stderr from the agent to report to the user why something went wrong.  The stdout & stderr is reported back to the service on task failure.  Right now, the service only saves the last 4k of stdout and 4k of stderr.

bmc-msft: An *extremely* naive impl:

```rust
async fn last_4k_string(stream: impl AsyncRead + Unpin) -> Result<String> {
    let mut result = String::new();

    let mut stream = BufReader::new(stream);
    loop {
        let mut buf = vec![];
        let bytes_read = stream
            .read_until(b'\n', &mut buf)
            .await
            .map_err(|e| format_err!("unable to read stream {}", e))?;
        if bytes_read == 0 {
            break;
        }
        let line = String::from_utf8_lossy(&buf).to_string();
        result.push_str(&line);
        result.truncate(4096);
    }
    Ok(result)
}


pub async fn monitor_process_last_4k(mut process: Child) -> Result<Output> {
    let stdout = process
        .stdout
        .take()
        .ok_or_else(|| format_err!("stdout not captured"))?;
    let stderr = process
        .stderr
        .take()
        .ok_or_else(|| format_err!("stderr not captured"))?;

    let stdout_monitor = last_4k_string(stdout);
    let stderr_monitor = last_4k_string(stderr);

    // even if stdout/stderr read fails, wait on the others.
    let (stdout_result, stderr_result, process_result) =
        futures::join!(stdout_monitor, stderr_monitor, process);
    let exit_status = process_result?.into();

    let stdout = stdout_result.unwrap_or_else(|x| format!("unable to read stdout: {}", x));
    let stderr = stderr_result.unwrap_or_else(|x| format!("unable to read stderr: {}", x));

    Ok(Output {
        stdout,
        stderr,
        exit_status,
    })
}
```

ranweiler: Ah, right, last night I just wasn't seeing that we were still doing that!

In that case I agree, something like a 4kb tail is the least viable solution. Will update this PR accordingly.

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -228,25 +232,97 @@ impl IWorkerRunner for WorkerRunner {
...
+struct RedirectedChild {
+    /// The child process.
+    child: Child,
+
+    /// Worker threads which continuously read from the redirected streams.
+    streams: Option<StreamReaderThreads>,
+}
 
-        Ok(child)
+impl RedirectedChild {
```
bmc-msft, line 252: This would be *really* good to write a unit test around, something akin to:
`perl -e 'print STDOUT "a"x4090 . "b"x4090; print STDERR "c"x4090 . "d"x4090'`

ranweiler: Agreed, this and the circular buffer both need unit tests (and I need to end-to-end test it still).

src/agent/onefuzz-supervisor/src/worker/tests.rs
```diff
@@ -226,3 +226,36 @@ async fn test_worker_done() {
     assert!(matches!(worker, Worker::Done(..)));
     assert_eq!(events, vec![]);
 }
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_redirected_child() {
```
ranweiler, line 232: Should add tests for some error cases (and Windows).

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -231,24 +235,118 @@ impl IWorkerRunner for WorkerRunner {
...
+    streams: Option<StreamReaderThreads>,
+}
+
+impl RedirectedChild {
+    pub fn spawn(mut cmd: Command) -> Result<Self> {
+        // Make sure we capture the child's output streams.
+        cmd.stderr(Stdio::piped());
+        cmd.stdout(Stdio::piped());
+
+        let mut child = cmd.spawn().context("onefuzz-agent failed to start")?;
```
bmc-msft, line 257: I imagine the redirect could will be useful in other places eventually.

If/when https://github.com/rust-lang/rust/issues/44434 gets merged, we should move this to

```
let mut child = cmd.spawn().with_context(|| format!("child process failed to start: {:?}", cmd.get_program()))?;
```

bmc-msft (approved):

PR #592, opened 2021-02-24
Update `onefuzz-agent` unit tests

ranweiler:
- Update the `onefuzz-agent` unit tests and re-enable in CI
- Disable `reqwest-retry` integration test that depends on a local server

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/generic/input_poller.rs
+++ b/src/agent/onefuzz-agent/src/tasks/generic/input_poller.rs
@@ -27,6 +27,26 @@ pub enum State<M> {
     Processed(M),
 }
 
+impl<M: PartialEq> PartialEq for State<M> {
+    fn eq(&self, other: &State<M>) -> bool {
+        use State::*;
+
+        match (self, other) {
+            (Ready, Ready) => true,
+            (Polled(l), Polled(r)) =>
+                l == r,
+            (Parsed(l0, l1), Parsed(r0, r1)) =>
+                l0 == r0 && l1 == r1,
+            (Downloaded(l0, l1, l2, l3), Downloaded(r0, r1, r2, r3)) =>
+                l0 == r0 && l1 == r1 && l2 == r2 && l3.path() == r3.path(),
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz-agent/src/tasks/generic/input_poller.rs
```diff
@@ -27,6 +27,26 @@ pub enum State<M> {
...
+        use State::*;
+
+        match (self, other) {
+            (Ready, Ready) => true,
+            (Polled(l), Polled(r)) =>
+                l == r,
+            (Parsed(l0, l1), Parsed(r0, r1)) =>
+                l0 == r0 && l1 == r1,
+            (Downloaded(l0, l1, l2, l3), Downloaded(r0, r1, r2, r3)) =>
+                l0 == r0 && l1 == r1 && l2 == r2 && l3.path() == r3.path(),
```
ranweiler, line 41: The last clause is why we can't just derive `PartialEq`: we ust compare the two `tempfile::TempDir` values via their projection to `path()`, since `TempDir: !PartialEq`.

In a future refactor we can wrap `TempDir` in a newtype that impls `PartialEq`, or upstream the same.

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -169,19 +169,20 @@ impl SendRetry for reqwest::RequestBuilder {
     }
 }
 
-#[cfg(test)]
-mod test {
-    use super::*;
+// TODO: convert to feature-gated integration test.
```
bmc-msft, line 172: Instead of commenting these out, can we use `#[ignore]`

ranweiler, line 172: I think we need to make this an integration test, and we should document the expected behavior of the test endpoint.

ranweiler: Ah, good call!

bmc-msft (approved):

chkeita (approved):

PR #593, opened 2021-02-24
Use non-deprecated atomic method

ranweiler:
As of `rustc` 1.50.0, [`AtomicUsize::compare_and_swap()`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.compare_and_swap) is deprecated.

The appropriate migration is described [here](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#migrating-to-compare_exchange-and-compare_exchange_weak).

### DIFF
--- a/src/agent/onefuzz-telemetry/src/lib.rs
+++ b/src/agent/onefuzz-telemetry/src/lib.rs
@@ -200,7 +200,7 @@ mod global {
     pub fn set_clients(instance: Option<TelemetryClient>, shared: Option<TelemetryClient>) {
         use Ordering::SeqCst;
 
-        let last_state = STATE.compare_and_swap(UNSET, SETTING, SeqCst);
+        let last_state = STATE.compare_exchange(UNSET, SETTING, SeqCst, SeqCst);
 
         if last_state == SETTING {
             panic!("race while setting telemetry client");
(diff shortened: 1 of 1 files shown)


chkeita (approved):

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -200,7 +200,7 @@ mod global {
     pub fn set_clients(instance: Option<TelemetryClient>, shared: Option<TelemetryClient>) {
         use Ordering::SeqCst;
 
-        let last_state = STATE.compare_and_swap(UNSET, SETTING, SeqCst);
+        let last_state = STATE.compare_exchange(UNSET, SETTING, SeqCst, SeqCst);
```
bmc-msft, line 203: This isn't a drop in replacement, the return types are different.

bmc-msft (changes requested):

src/agent/onefuzz-telemetry/src/lib.rs
ranweiler: Speculative `git push` (-;

bmc-msft:
Integration tests in queue

bmc-msft (approved):

PR #597, opened 2021-02-25
fix bool queries

bmc-msft:
This addresses broken queries used for identifying outdated nodes.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/orm.py
+++ b/src/api-service/__app__/onefuzzlib/orm.py
@@ -148,8 +148,8 @@ def build_filters(
             for x in values:
                 if not isinstance(x, bool):
                     raise TypeError("unexpected type")
-                parts.append("%s eq %s" % (field_name, str(x).lower))
-        if isinstance(values[0], int):
+                parts.append("%s eq %s" % (field_name, str(x).lower()))
+        elif isinstance(values[0], int):
             parts = []
             for x in values:
                 if not isinstance(x, int):


src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -148,8 +148,8 @@ def build_filters(
             for x in values:
                 if not isinstance(x, bool):
                     raise TypeError("unexpected type")
-                parts.append("%s eq %s" % (field_name, str(x).lower))
-        if isinstance(values[0], int):
+                parts.append("%s eq %s" % (field_name, str(x).lower()))
```
ranweiler, line 151: This is where the actual bug was.

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -148,8 +148,8 @@ def build_filters(
...
+        elif isinstance(values[0], int):
```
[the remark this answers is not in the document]
bmc-msft: Unfortunately, bool is a subclass of int.  This was generating queries that were `field == 0`, where we needed `field eq false`

```
Python 3.8.5 (default, Jan 27 2021, 15:41:15)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = False
>>> isinstance(a, int)
True
>>>
```

src/api-service/__app__/onefuzzlib/orm.py
bmc-msft: This is a two-part bug.

chkeita (approved):

PR #600, opened 2021-02-26
add non-x86_64 architecture libfuzzer target support using qemu-user

bmc-msft:
This provides support to run non-x86_64 based targets using user mode QEMU.  

This is done by wrapping the target_exe in a shell script that executes the target underneath qemu-(ARCH) as part of the task template.

Given these limitations, I've added warnings sent to the user in the CLI to indicate this is a preview feature and may change in the future.

Limitations:
* Does not support live-repro.   This is actually simpler, rather than launching it in our own gdbserver, we can just use qemu's built in GDB.
* Does not support coverage extraction.  We would need to replace the wrapper script with functionality in the agent that does the wrapping, move to using QEMU's gdb server support, and then fix up our gdb scripts to use `target remote`.


bmc-msft:
Note crash reporting is extremely slow.  Our default timeout of 5 seconds is far too short.  As such, this sets the timeout for significantly longer.

docs/how-to/fuzzing-other-architectures-on-azure.md
```diff
@@ -0,0 +1,151 @@
...
+Source capabilities:
+
+2. [libFuzzer](https://www.llvm.org/docs/LibFuzzer.html), which performs the fuzzing
+3. [qemu-user](https://qemu.readthedocs.io/en/latest/user/main.html), which provides user-space emulation.
+
+TL/DR: check out our [libfuzzer-aarch64-crosscompile example](../../src/integration-tests/libfuzzer-aarch64-crosscompile/)
+
+## Issues using qemu_user based LibFuzzer in OneFuzz
+
+There are a few notable limitations using qemu-user based libFuzzer targets on OneFuzz
```
ranweiler, line 13:
```suggestion
There are a few notable limitations when using `qemu-user` based libFuzzer targets in OneFuzz.
```

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -488,3 +494,197 @@ def dotnet(
...
+        libfuzzer tasks, wrapped via qemu-user (PREVIEW FEATURE)
+        """
+
+        self.logger.warning(
+            "qemu_user jobs are a preview feature and may change in the future"
+        )
+
+        pool = self.onefuzz.pools.get(pool_name)
+        if pool.os != OS.linux:
+            raise Exception("libfuzzer qemu-user jobs are only compatable on linux")
```
ranweiler, line 537:
```suggestion
            raise Exception("libfuzzer qemu-user jobs are only compatible with with Linux")
```

src/integration-tests/integration-test.py
```diff
@@ -42,6 +42,7 @@
 class TemplateType(Enum):
     libfuzzer = "libfuzzer"
     libfuzzer_dotnet = "libfuzzer_dotnet"
+    libfuzzer_qemu = "libfuzzer_qemu"
```
ranweiler, line 45: Would it be more consistent to use a more explicit variant name like:
```suggestion
    libfuzzer_qemu_user = "libfuzzer_qemu_user"
```

src/integration-tests/integration-test.py
```diff
@@ -407,6 +429,10 @@ def launch_repro(self) -> None:
         has_cdb = bool(which("cdb.exe"))
         has_gdb = bool(which("gdb"))
         for job_id in self.successful_jobs:
+            if not TARGETS[self.target_jobs[job_id]].test_repro:
```
ranweiler, line 432: I'm assuming this lookup is infallible in context.

src/integration-tests/libfuzzer-aarch64-crosscompile/Makefile
```diff
@@ -0,0 +1,28 @@
...
+fuzz.exe: fuzz-libs fuzz.c
+	aarch64-linux-gnu-g++ -pthread -lasan -o fuzz.exe fuzz-libs/*.o fuzz.c -fsanitize=address -fsanitize-coverage=trace-pc
+
+check: fuzz.exe
+	ASAN_OPTIONS=:detect_leaks=0 qemu-aarch64 -L /usr/aarch64-linux-gnu ./fuzz.exe -runs=1
+
+fuzz: check
+	ASAN_OPTIONS=:detect_leaks=0 qemu-aarch64 -L /usr/aarch64-linux-gnu ./fuzz.exe
+
+.PHONY: clean
```
ranweiler, line 25:
```suggestion
.PHONY: check clean fuzz
```

src/integration-tests/libfuzzer-aarch64-crosscompile/Makefile
```diff
@@ -0,0 +1,28 @@
+all: check
```
ranweiler, line 1: Needs copyright notice.

ranweiler (approved):

src/integration-tests/integration-test.py
bmc-msft: yep.

src/cli/onefuzz/templates/libfuzzer.py
bmc-msft:
```suggestion
            raise Exception("libfuzzer qemu-user jobs are only compatible with Linux")
```

src/integration-tests/integration-test.py
```diff
@@ -84,6 +86,15 @@ class Integration(BaseModel):
         use_setup=True,
         wait_for_files=[ContainerType.inputs, ContainerType.crashes],
     ),
+    "linux-libfuzzer-aarch64-crosscompile": Integration(
+        template=TemplateType.libfuzzer_qemu,
```
bmc-msft, line 90:
```suggestion
        template=TemplateType.libfuzzer_qemu_user,
```

src/integration-tests/integration-test.py
```diff
@@ -237,6 +248,17 @@ def launch(self, path: str) -> None:
                     duration=1,
                     vm_count=1,
                 )
+            elif config.template == TemplateType.libfuzzer_qemu:
```
bmc-msft, line 251:
```suggestion
            elif config.template == TemplateType.libfuzzer_qemu_user:
```

src/integration-tests/libfuzzer-aarch64-crosscompile/Makefile
bmc-msft, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

```

bmc-msft:
```suggestion
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

all: check
```

bmc-msft:
On windows, using --target_exe ./foo sends the target_exe to `./foo-wrapper.sh`.  Due to a service issue, this is an invalid blob store path which causes a failure in the blob store library in the service.

The CLI shouldn't pass ./ here.  (and a follow-on, the service could attempt to catch ./ in the path to catch this as an error, rather than a 500 error when the blob store rejects it)

bmc-msft:
Follow-up PRs addressed the service issue.

PR #613, opened 2021-03-01
require target_exe to be a canonicalized relative path

bmc-msft:
This prevents querying the blob store for paths such as "./foo" instead of "foo", which fails.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/tasks/config.py
+++ b/src/api-service/__app__/onefuzzlib/tasks/config.py
@@ -104,6 +104,10 @@ def check_target_exe(config: TaskConfig, definition: TaskDefinition) -> None:
 
         return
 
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/tasks/config.py
```diff
@@ -102,6 +104,12 @@ def check_target_exe(config: TaskConfig, definition: TaskDefinition) -> None:
 
         return
 
+    if (
```
ranweiler, line 107: Maybe just add a comment reminding ourselves why we're doing this: we get relative paths as input, but blob storage doesn't have a notion of path. It deals in structureless string _names_, and so has no upstream notion of normalizing/traversal when comparing the equality of blob names.

ranweiler (approved):

PR #615, opened 2021-03-01
allow unnecessary_wraps in coverage

bmc-msft:
Addresses a warning from `clippy` with a raised priority warning.


bmc-msft:
OBE by #616

PR #616, opened 2021-03-01
Rewrite redundant `Result` wraps

ranweiler:
Remove redundant `Result` wrapping flagged by the linter.

### DIFF
--- a/src/agent/coverage/src/block/windows.rs
+++ b/src/agent/coverage/src/block/windows.rs
@@ -94,7 +94,7 @@ impl BlockCoverageHandler {
         dbg.quit_debugging();
     }
 
-    fn try_on_create_process(&mut self, dbg: &mut Debugger, module: &Module) -> Result<()> {
+    fn handle_on_create_process(&mut self, dbg: &mut Debugger, module: &Module) -> Result<()> {
         dbg.target().sym_initialize()?;
 
         log::info!(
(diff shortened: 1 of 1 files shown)


src/agent/coverage/src/block/windows.rs
```diff
@@ -139,38 +137,32 @@ impl BlockCoverageHandler {
...
     fn on_breakpoint(&mut self, dbg: &mut Debugger, bp: BreakpointId) {
-        if self.try_on_breakpoint(dbg, bp).is_err() {
+        if self.handle_on_breakpoint(dbg, bp).is_err() {
             self.stop(dbg);
         }
     }
 
     fn on_poll(&mut self, dbg: &mut Debugger) {
-        if self.try_on_poll(dbg).is_err() {
-            self.stop(dbg);
```
bmc-msft, line 173: This no longer calls self.stop on failure.  Is that intended?

ranweiler: Yes, because that was unreachable. The original pattern was only mean to be a within-code reminder that if one of non-trait implementation methods were to fail, we should call `quit_debugging()`. The `on_poll()` method just exists to do exactly this _on timeout_ waiting for a debug event.

chkeita (approved):

bmc-msft (approved):

PR #623, opened 2021-03-02
fix agent `retry` on connection level failures

### LINKED ISSUE #263: Log in agent retries for debugging
When making agent HTTP requests with retries and backoff, log what's happening to provide insight and debuggability.

bmc-msft:
In debugging the connection retry issues, I dug into this more.  

Apparently, some of hyper's connection errors are not mapped to std::io::Error, rendering the existing downcast impl ineffective.

As such, this PR makes the following updates:
1. Any request that fails for what `reqwest` calls a `connection` error is considered transient.
2. Updates the retry notify code to use our `warn` macro such that the events show up in application insights.
3. Updates the unit test to demonstrate that failures by trying to connect to `http://localhost:81/`, which shouldn't be listening on any system.
4. Adds a simple unit test to verify with send_retry_default, connections to https://www.microsoft.com work

Fixes #263

### DIFF
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -186,7 +186,8 @@ mod test {
         reqwest::Client::new()
             .get("https://www.microsoft.com")
             .send_retry_default()
-            .await?.error_for_status()?;
+            .await?
+            .error_for_status()?;
 
         Ok(())
     }
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -43,9 +43,7 @@ fn to_backoff_response(
 ) -> Result<Response, backoff::Error<anyhow::Error>> {
     match result {
         Err(error) => {
-            if error.is_connect() {
-                Err(backoff::Error::Transient(anyhow::Error::from(error)))
-            } else if is_transient_socket_error(&error) {
+            if error.is_connect() || is_transient_socket_error(&error) {
                 Err(backoff::Error::Transient(anyhow::Error::from(error)))
             } else {
                 Err(backoff::Error::Permanent(anyhow::Error::from(error)))
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -182,6 +182,16 @@ mod test {
     use super::*;
 
     #[tokio::test]
+    async fn retry_should_pass() -> Result<()> {
+        reqwest::Client::new()
+            .get("https://www.microsoft.com")
+            .send_retry_default()
+            .await?.error_for_status()?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
     async fn retry_should_fail() -> Result<()> {
         let resp = reqwest::Client::new()
             .get("http://localhost:81/test.txt")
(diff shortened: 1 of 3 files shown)


src/agent/reqwest-retry/src/lib.rs
```diff
@@ -173,17 +181,34 @@ impl SendRetry for reqwest::RequestBuilder {
...
 
-    // TODO: convert to feature-gated integration test.
-    #[ignore]
     #[tokio::test]
-    async fn empty_stack() -> Result<()> {
-        let resp = reqwest::Client::new()
-            .get("http://localhost:5000/api/testGet")
+    async fn retry_should_pass() -> Result<()> {
+        reqwest::Client::new()
+            .get("https://www.microsoft.com")
```
ranweiler, line 187:
```suggestion
            .get("https://www.contoso.com")
```

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -173,17 +181,34 @@ impl SendRetry for reqwest::RequestBuilder {
...
-            .await?;
-        println!("{:?}", resp);
+            .await?.error_for_status()?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn retry_should_fail() -> Result<()> {
+        let resp = reqwest::Client::new()
```
ranweiler, line 197:
```suggestion
        let invalid_url = "http://localhost:81/test.txt";
        let resp = reqwest::Client::new()
```

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -173,17 +181,34 @@ impl SendRetry for reqwest::RequestBuilder {
...
+            .get("http://localhost:81/test.txt")
```
ranweiler, line 198:
```suggestion
            .get(invalid_url)
```
The above are to make the expectation / setup a bit more self-documenting.

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -173,17 +181,34 @@ impl SendRetry for reqwest::RequestBuilder {
...
+                3i32,
+                to_backoff_response,
+            )
+            .await;
+
+        assert!(resp.is_err());
+        if let Err(err) = &resp {
+            let as_text = format!("{}", err);
+            assert!(as_text.contains("Maximum number of attempts reached for this request"));
+        }
```
ranweiler, line 211: Instead of the preceding `assert!()`, maybe add an `else` branch here, which panics.

ranweiler (approved):

src/agent/reqwest-retry/src/lib.rs
bmc-msft: Disagree with this.  While the URL exists, it redirects to www.microsoft.com.  This makes it clear in the test that it's making a request to a microsoft property, where contoso is not.

chkeita (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -107,7 +115,7 @@ pub async fn send_retry_reqwest<
                 max_elapsed_time: Some(max_elapsed_time),
                 ..ExponentialBackoff::default()
             },
-            |err, _| println!("Transient error: {}", err),
+            |err, _| warn!("transient http error: {}", err),
```
chkeita, line 118: can we add the url in the message here it might be helpful when debugging.

bmc-msft: Unfortunately, the URL here isn't easily accessible.  At this context, we have a Builder, which we'd have to TryClone to get access into the URL.

I agree that it would be valuable.  Perhaps a follow-up can refactor this to make it accessible?

PR #628, opened 2021-03-02
Simplify module/symbol filter definition format, impl

ranweiler:
Update the filter rule format and implementation to be simpler and user-predictable. In particular, we remove an accidental dependence of rule application on hash map iteration order.

Example of the new rule format (with explanation inline):
```js
[
    // Track any .exe, but only functions whose symbols match some regex (ignoring the rest).
    {
        "module": ".exe$",
        "include": [
            "^main$",
            "test_input$",
            "read_all$"
        ]
    },
    // Track `libabc.so`, `libabc.so.1`, &c, excluding only those symbols whose
    // name contains the string `asan` or starts with `_private` (tracking the rest).
    {
        "module": "libabc.so",
        "exclude": [
            "asan",
            "^_private.*"
        ]
    },
    // Track the entire module `libxyz.so`, and all its symbols.
    // By default, this would be redundant, except for the next rule.
    {
        "module": "libxyz.so",
        "include": true
    },
    // If a module does not match any of the above, do not track it.
    // This means we won't even try to parse backing file.
    // This effectively inverts our default behavior.
    {
        "module": ".*",
        "include": false
    }
]
```
Note that our JSON parsing does not actually accept comment syntax.


src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
+struct ModuleRuleDef {
+    pub module: String,
 
-    /// Maps module names (via regexes) to filters for the symbol names provided
-    /// by the matched module(s).
+    #[serde(flatten)]
+    pub rule: RuleDef,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
ranweiler, line 316: Add comment explaining how this type exists to be a serde target and support a slightly more complex (and intuitive) user-facing file format.

bmc-msft:
`include` and `exclude` can be either a bool or a list of regexes?

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
+#[serde(untagged)]
+enum RuleDef {
+    Include { include: bool },
+    Exclude { exclude: bool },
+    Filter(Box<Filter>),
```
ranweiler, line 321: This is boxed per a clippy lint: the `Filter` struct is on the order of 500 bytes, which forces any value of type `RuleDef` to be that size... event a variant that is otherwise just a `bool`!

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
+}
+
+#[derive(Clone, Debug)]
```
ranweiler, line 324: Add a comment that explains that this is essentially a normalized lowering of the serde-implementing `Def` type, to reduce the complexity of the actual filtering logic.

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
+    /// distinction at this level lets us avoid parsing modules that we want to
+    /// ignore.
+    IncludeModule(bool),
+
+    /// The entire module should be tracked and parsed, with a filter applied to
+    /// its symbols.
+    FilterSymbols(Box<Filter>),
+}
+
+#[derive(Clone, Debug)]
```
ranweiler, line 339: Document.

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
 
-        Ok(Self { regexes, filters })
-    }
+        let regexes = RegexSet::new(&modules)?;
 
-    pub fn is_allowed(&self, module: &ModulePath, name: &str) -> bool {
-        let rules = self.regexes.matches(&module.path_lossy());
+        Ok(Self { regexes, rules })
+    }
+    pub fn includes_module(&self, module: &ModulePath) -> bool {
```
ranweiler, line 366: Add unit tests.

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
-        // If many rules would apply, the first rule is used.
-        if let Some(idx) = rules.iter().next() {
-            // Guaranteed by constructor.
-            let filter = self.filters.get(&idx).unwrap();
-
-            filter.is_allowed(name)
-        } else {
-            // If no module-level rule exists, allow by default.
-            true
+    pub fn includes_symbol(&self, module: &ModulePath, symbol: impl AsRef<str>) -> bool {
```
ranweiler, line 385: Add unit tests.

src/agent/coverage/src/code/tests.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#[test]
+fn test_module_filter_def_include_filter() {
+    let text = r#"{ "module": "abc.exe", "include": [] }"#;
+    let def: ModuleRuleDef = serde_json::from_str(text).unwrap();
+
+    assert_eq!(def.module, "abc.exe");
+
+    if let RuleDef::Filter(filter) = def.rule {
```
ranweiler, line 43: We can't just use `matches!` here, since `filter` has type `Box<Filter>`, and `box_patterns` is unstable.

src/agent/coverage/src/code/tests.rs
```diff
@@ -0,0 +1,80 @@
...
+        assert!(matches!(*filter, Filter::Include(_)));
+    } else {
+        panic!("expected a `Filter` rule");
+    }
+
+    let text = r#"{ "module": "abc.exe", "include": [ "^parse_data$" ] }"#;
```
bmc-msft, line 49: It's conceptually complex to have `include` and `exclude` either a list of regexes or a Boolean.  Can this be simplified?

src/agent/coverage/src/code/tests.rs
```diff
@@ -0,0 +1,80 @@
...
+    let def: ModuleRuleDef = serde_json::from_str(text).unwrap();
+
+    assert_eq!(def.module, "abc.exe");
+    assert!(matches!(def.rule, RuleDef::Exclude { exclude: true }));
+
+    let text = r#"{ "module": "abc.exe", "exclude": false }"#;
+    let def: ModuleRuleDef = serde_json::from_str(text).unwrap();
+
+    assert_eq!(def.module, "abc.exe");
+    assert!(matches!(def.rule, RuleDef::Exclude { exclude: false }));
```
bmc-msft, line 33: How does `include: true` and `exclude: false` differ?  Or `exclude: true` and `include: false`?

ranweiler:
> `include` and `exclude` can be either a bool or a list of regexes?

Correct. It isn't hard to change that, if we want.

src/agent/coverage/src/code.rs
```diff
@@ -302,83 +301,113 @@ impl Region for Symbol {
...
+#[derive(Clone, Debug)]
+enum Rule {
+    /// Asserts that the entire sould be be tracked (and its symbols included),
+    /// or ignored, and its symbols excluded.
     ///
-    /// The integer key to this map is an index into the `regexes` field, and so
-    /// identifies a (module-matching) regex in that set.
-    filters: BTreeMap<usize, Filter>,
+    /// The implied symbol tracking behavior could be encoded by a filter, but a
+    /// distinction at this level lets us avoid parsing modules that we want to
```
ranweiler, line 330: @bmc-msft, see this comment for one motivation behind allowing `"exclude": true`. I also find that to be more direct (from a user perspective) than asking them to encode that behavior with the right regex-based rule.

We could say "use regex lists, only", at the expense of some static information. The cost there isn't super high. Worst case it means we'd attempt to parse every module that is linked or loaded by the target. But we only do that once per task (due to our cache), so it is definitely amortized.

src/agent/coverage/src/code/tests.rs
ranweiler: They have identical semantics.

src/agent/coverage/src/code/tests.rs
ranweiler: See other comment for motivation / discussion.

ranweiler:
Add new doc that explains the coverage tracking rule format, with worked examples.

docs/how-to/coverage-filtering.md
```diff
@@ -0,0 +1,117 @@
...
+]
+```
+
+The filter list an array of objects that describe per-module rules.
+
+A `<module-include>` looks like:
+```
+{
+  "module": <regex>,
+  "include": <bool | Array<regex>>
```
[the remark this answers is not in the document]
bmc-msft: Since we're discussing the json schema, following the json schema names is probably useful.

https://json-schema.org/understanding-json-schema/

docs/how-to/coverage-filtering.md
```diff
@@ -0,0 +1,117 @@
+# Coverage Filtering
```
bmc-msft, line 1: This should probably include a comment that this is for an upcoming feature, as this document will be part of the `main` before the uninstrumented target coverage tasks are available.

PR #629, opened 2021-03-03
add info to help clarify success after warnings

bmc-msft:
During a deployment this morning, AAD failed the first attempt to create an application password and upon retry succeeded.

The error message given by AAD was scary:

```
WARNING:root:unable to create app password: request did not succeed: HTTP 403 - {
  "error": {
    "code": "Authorization_RequestDenied",
    "message": "Insufficient privileges to complete the operation.",
    "innerError": {
      "date": "2021-03-03T15:39:35",
      "request-id": "GUID",
      "client-request-id": "GUID"
    }
  }
}
```

This error gave me the impression that the AAD password creation failed, yet the deployment continued.

This PR adds a comment in the success case to clarify when the password creation succeeded.

Now, the above case will look like this:
```
WARNING:root:unable to create app password: request did not succeed: ...
INFO:root:app password created
```


src/deployment/registration.py
```diff
@@ -232,7 +232,9 @@ def add_application_password(app_object_id: UUID) -> Tuple[str, str]:
     while count < tries:
         count += 1
```
ranweiler, line 233: Could we add a log message to this loop along the lines of "retrying app password creation", so if you get the (now-nicer) warning message, it also tells you that it is retrying before you get the success messsage?

ranweiler (approved):

src/deployment/registration.py
bmc-msft: added.

PR #650, opened 2021-03-09
mark dependant tasks failed upon failure

### LINKED ISSUE #644: Tasks should be marked failed upon failure of a prereq task
Lets assume I create a fuzzing task and a crash reporting task.  If the fuzzing task fails, we should stop the crash reporting task, marking it as failed because it's prereq failed.

bmc-msft:
Fix #644


chkeita (approved):

bmc-msft:
@chkeita I changed the logic slightly.  Please re-review.

chkeita (approved):

PR #654, opened 2021-03-09
enable long-running integration tests

bmc-msft:
This refactors the way integration tests are run.

Previously, this would keep a record of everything it launched, then check based on that record.

Now, it launches jobs similar to before, but during `check_jobs` and `check_repros`, it's entirely based on the data coming back from 1f.

This allows us to perform tests either via:

* run the whole thing like today: `python integration-test.py test  /tmp/tests/`
* run in chunks (like if we were doing a long run integration test)
     * start the tests and walk away: `python integration-test.py launch /tmp/tests/`
     * check the status of any of the jobs: `python integration-test.py check_jobs <TEST_ID>`
     * verify `repro` works for any of the jobs with crash reports: `python integration-test.py check_repros <TEST_ID>`
     * cleanup all of the allocated resources: `python integration-test.py cleanup <TEST_ID>`

### DIFF
--- a/src/integration-tests/integration-test.py
+++ b/src/integration-tests/integration-test.py
@@ -7,13 +7,13 @@
 """ Launch multiple templates using samples to verify Onefuzz works end-to-end """
 
 # NOTE:
-# 1. This script uses pre-built fuzzing samples from the onefuzz-samples project.
-#    https://github.com/microsoft/onefuzz-samples/releases/latest
+# 1. This script uses an unpacked version of the `integration-test-results`
+#    from the CI pipeline.
 #
-# 2. This script will create new pools & managed scalesets during the testing by
-#    default.  To use pre-existing pools, specify `--user_pools os=pool_name`
+#    Check out https://github.com/microsoft/onefuzz/actions/workflows/
+#       ci.yml?query=branch%3Amain+is%3Asuccess
 #
-# 3. For each stage, this script launches everything for the stage in batch, then
+# 2. For each stage, this script launches everything for the stage in batch, then
 #    checks on each of the created items for the stage.  This batch processing
 #    allows testing multiple components concurrently.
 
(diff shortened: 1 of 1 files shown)


src/integration-tests/integration-test.py
```diff
@@ -313,279 +272,315 @@ def check_task(self, task_id: UUID, scalesets: List[Scaleset]) -> Optional[str]:
...
-                        del self.tasks[task_id]
+            if job_tasks:
+                scalesets = self.of.scalesets.list()
+                for job_id in job_tasks:
+                    finished_tasks: Set[UUID] = set()
+                    for task in job_tasks[job_id]:
+                        if job_id not in jobs:
+                            continue
+                        task_result = self.check_task(jobs[job_id], task, scalesets)
+                        if task_result is None:
```
chkeita, line 367: I imagine that some task might never run because of a bug and we will need to know which ones.

I suggest we keep track the list of the pending task here and either log them or return them as part of the result. Also it would be helpful to have an option that fail the check if we have any pending tasks.

chkeita (approved):

src/integration-tests/integration-test.py
bmc-msft: Addressed

src/integration-tests/integration-test.py
```diff
@@ -291,301 +267,345 @@ def launch(self, path: str) -> None:
...
-                    result = self.of.repro.connect(
-                        repro.vm_id,
-                        delete_after_use=True,
-                        debug_command=commands[repro.os][0],
+            commands: Dict[OS, Tuple[str, str]] = {
+                OS.windows: ("r rip", r"^rip=[a-f0-9]{16}"),
+                OS.linux: ("info reg rip", r"^rip\s+0x[a-f0-9]+\s+0x[a-f0-9]+"),
+            }
+
+            for (job, repro) in list(repros.values()):
```
ranweiler, line 519: Is the `list()` here (and below) redundant, since we're turning an iterable into a list, which we then iterate over?

ranweiler (approved):

src/integration-tests/integration-test.py
bmc-msft: This allows us to modify `repros` within the loop, which is otherwise not allowed.

PR #658, opened 2021-03-10
Fix incorrect offset in stack reports

lzybkr:
The wrong PC was from the context record and incorrectly adjusted by 1 which worked well enough to see the correct function name and line numbers, but was the wrong offset.

The correct PC is passed in a STACKFRAME_EX struct.

### DIFF
--- a/src/agent/debugger/src/dbghelp.rs
+++ b/src/agent/debugger/src/dbghelp.rs
@@ -514,7 +514,7 @@ impl DebugHelpGuard {
         }
     }
 
-    pub fn stackwalk_ex<F: FnMut(&FrameContext, &STACKFRAME_EX) -> bool>(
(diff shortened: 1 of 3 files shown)


bmc-msft (approved):

ranweiler:
I assume this changes the OneFuzz crash reports / hashes, due to the changed PC?

lzybkr:
Yes, it could, but I don't think the hashing is used for libfuzzer tasks. Note that my other PR definitely changes the stack hashes as we use the default Rust generated hash impl and I changed the struct to include the offset.

ranweiler:
@bmc-msft: I verified that this PR _does_ change the function-relative offsets of stack frames with no source info (e.g. from `kernel32`, `ntdll`), and thus changes the `call_stack` and `call_stack_sha256` fields of generic crash reports.

lzybkr:
Note that a Windows update would just as easily change the hash for such stacks.

PR #667, opened 2021-03-11
fix racy tests for expected failures in web requests

bmc-msft:
Thanks primarily due to lookups on `localhost` but also impacted by host-based firewalls silently dropping packets, this test could be racy.

This addresses those issues.

### DIFF
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -205,9 +205,7 @@ mod test {
 
         if let Err(err) = &resp {
             let as_text = format!("{}", err);
-            assert!(
-                as_text.contains("Maximum number of attempts reached for this request")
-            );
+            assert!(as_text.contains(MAX_RETRY_ERROR_MESSAGE));
         } else {
             anyhow::bail!("response to {} was expected to fail", invalid_url);
         }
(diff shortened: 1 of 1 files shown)


ranweiler (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -192,23 +192,20 @@ mod test {
...
             )
             .await;
 
         if let Err(err) = &resp {
             let as_text = format!("{}", err);
-            assert!(
-                as_text.contains("Maximum number of attempts reached for this request")
-                    || as_text.contains("os error 10061")
-            );
+            assert!(as_text.contains(MAX_RETRY_ERROR_MESSAGE));
```
ranweiler, line 208: I'm assuming this reversion was you root-causing the test failure and removing a recently-added band-aid.

lzybkr: 10061 was the error I was seeing locally which is WSAECONNREFUSED. I guessed (perhaps incorrectly) that retry wouldn't happen after this error.

ranweiler: That's legit, just making sure that the churn here is intentional!

bmc-msft: The underlying WSAECONNREFUSED which @lzybkr was hitting is handled by the retry code.  

However, we were hitting "maximum duration for testing" instead of "tried too many times".  This PR adjusts the test such that we always hit the former.

PR #671, opened 2021-03-12
address issues from by latest regex update

bmc-msft:
This PR addresses an issue exposed when updating to the latest `regex` version.

Prior to this PR, the latest regex would give the following error:
```
error[E0597]: `re` does not live long enough
   --> onefuzz/src/libfuzzer.rs:214:15
    |
214 |         match re.captures_iter(&output_text).next() {
    |               ^^----------------------------
    |               |
    |               borrowed value does not live long enough
    |               a temporary with access to the borrow is created here ...
...
228 |     }
    |     -
    |     |
    |     `re` dropped here while still borrowed
    |     ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `regex::CaptureMatches<'_, '_>`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
214 |         let x = match re.captures_iter(&output_text).next() {
215 |             Some(captures) => {
216 |                 let added_files_count = captures.get(1).unwrap().as_str().parse::<i32>().unwrap();
217 |                 let added_feature_count = captures.get(2).unwrap().as_str().parse::<i32>().unwrap();
218 |                 Ok(LibFuzzerMergeOutput {
219 |                     added_files_count,
  ...

```

### DIFF
--- a/src/agent/onefuzz/src/libfuzzer.rs
+++ b/src/agent/onefuzz/src/libfuzzer.rs
@@ -211,7 +211,8 @@ impl<'a> LibFuzzer<'a> {
         let output_text = String::from_utf8_lossy(&output.stderr);
         let pat = r"MERGE-OUTER: (\d+) new files with (\d+) new features added";
         let re = regex::Regex::new(pat).unwrap();
-        let result = match re.captures_iter(&output_text).next() {
+        let captures = re.captures_iter(&output_text).next();
+        match captures {
             Some(captures) => {
                 let added_files_count = captures.get(1).unwrap().as_str().parse::<i32>().unwrap();
                 let added_feature_count = captures.get(2).unwrap().as_str().parse::<i32>().unwrap();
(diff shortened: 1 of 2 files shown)


src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -211,7 +211,7 @@ impl<'a> LibFuzzer<'a> {
         let output_text = String::from_utf8_lossy(&output.stderr);
         let pat = r"MERGE-OUTER: (\d+) new files with (\d+) new features added";
         let re = regex::Regex::new(pat).unwrap();
-        match re.captures_iter(&output_text).next() {
+        let result = match re.captures_iter(&output_text).next() {
```
ranweiler, line 214: It might address the compiler complaint a bit more directly if we were to do:
```suggestion
        let captures = re.captures_iter(&output_text).next();
        match captures {
```

ranweiler (approved):

PR #674, opened 2021-03-15
retry any failed request regardless of status

bmc-msft:
Simplify the retry logic to always retry any failed request up to 5 times.  The initial delay is now 5 seconds (plus jitter).

Additionally, upon failure we now indicate how many times we failed as well as the timing for each of the request.

### DIFF
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -1,108 +1,46 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use anyhow::Result;
+use anyhow::{format_err, Result};
 use async_trait::async_trait;
 use backoff::{self, future::FutureOperation, ExponentialBackoff};
 use onefuzz_telemetry::warn;
-use reqwest::{Response, StatusCode};
+use reqwest::Response;
 use std::{
-    sync::atomic::{AtomicI32, Ordering},
+    sync::atomic::{AtomicUsize, Ordering},
     time::Duration,
 };
 
-use std::error::Error as StdError;
-use std::io::ErrorKind;
-
-const DEFAULT_RETRY_PERIOD: Duration = Duration::from_secs(2);
-const MAX_ELAPSED_TIME: Duration = Duration::from_secs(30);
-const MAX_RETRY_ATTEMPTS: i32 = 5;
-const MAX_RETRY_ERROR_MESSAGE: &str = "Maximum number of attempts reached for this request";
-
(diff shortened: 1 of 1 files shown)


src/agent/reqwest-retry/src/lib.rs
```diff
@@ -195,17 +105,12 @@ mod test {
...
-                2i32,
-                to_backoff_response,
-            )
+            .send_retry(Duration::from_millis(1), 3)
             .await;
 
         if let Err(err) = &resp {
             let as_text = format!("{}", err);
-            assert!(as_text.contains(MAX_RETRY_ERROR_MESSAGE));
+            assert!(as_text.contains("request failed after"));
```
chkeita, line 113: I know this was the existing logic but maybe we need to introduce a RetryError for this case.
not a blocking comment.

chkeita (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -1,173 +1,83 @@
...
-use anyhow::Result;
+use anyhow::{format_err, Result};
 use async_trait::async_trait;
 use backoff::{self, future::FutureOperation, ExponentialBackoff};
 use onefuzz_telemetry::warn;
-use reqwest::{Response, StatusCode};
+use reqwest::Response;
 use std::{
-    sync::atomic::{AtomicI32, Ordering},
+    sync::atomic::{AtomicUsize, Ordering},
```
ranweiler, line 10: 👍

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -1,173 +1,83 @@
...
     let result = op
         .retry_notify(
             ExponentialBackoff {
                 current_interval: retry_period,
                 initial_interval: retry_period,
-                max_elapsed_time: Some(max_elapsed_time),
                 ..ExponentialBackoff::default()
             },
-            |err, _| warn!("transient http error: {}", err),
+            |err, dur| warn!("request attempt failed at {:?}: {}", dur, err),
```
ranweiler, line 53:
```suggestion
            |err, dur| warn!("request attempt failed after {:?}: {}", dur, err),
```

ranweiler (approved):

src/agent/reqwest-retry/src/lib.rs
chkeita: The siggestion here is to have send_retry return a Result<Response, SendRetryError> and move the type of failure inside the SendRetryError. That way we don't have to do string contains

bmc-msft: That's reasonable, but let's do that as a follow-on PR.

PR #683, opened 2021-03-17
simplify config update notification

bmc-msft:


ranweiler (approved):

src/proxy-manager/src/config.rs
```diff
@@ -154,18 +154,14 @@ impl Config {
         Ok(())
     }
 
-    pub async fn update(&mut self) -> Result<bool> {
+    pub async fn update(&mut self) -> Result<()> {
         if self.fetch().await? {
             info!("config updated");
             self.save().await?;
+            proxy::update(&self.data).await?;
+            self.notify().await?;
```
ranweiler, line 162: Before, it looks like we were trying to only `notify()` if the config had changed. Why was that check unnecessary?

bmc-msft: `fetch` tells us if the data is new or not already.

fetch:
https://github.com/microsoft/onefuzz/blob/6e60a8cf10dff8c5c2a48caa0ce49d9dc095a54d/src/proxy-manager/src/config.rs#L128-L141

PR #689, opened 2021-03-18
add LoadLibrary-based libfuzzer integration test

bmc-msft:


src/integration-tests/integration-test.py
```diff
@@ -178,6 +178,18 @@ class Integration(BaseModel):
...
     ),
+    "windows-libfuzzer-load-library": Integration(
+        template=TemplateType.libfuzzer,
+        os=OS.windows,
+        target_exe="fuzz.exe",
+        inputs="seeds",
+        wait_for_files={
+            ContainerType.inputs: 2,
+            ContainerType.unique_reports: 1,
+            ContainerType.coverage: 1,
```
ranweiler, line 189: We wait for files to appear, but currently have no ability to download/make assertions about them, right?

ranweiler (approved):

src/integration-tests/integration-test.py
bmc-msft: Yep.

PR #696, opened 2021-03-18
include the full failure when we give up from retrying

bmc-msft:
Previously,  after 4 attempts, we'd raise the error `request failed after {} attempts` without the underlying error context from reqwest.  This moves to raising the underlying error, but with context that it failed after N attempts.


ranweiler (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -109,8 +113,8 @@ mod test {
             .await;
 
         if let Err(err) = &resp {
-            let as_text = format!("{}", err);
-            assert!(as_text.contains("request failed after"));
+            let as_text = format!("{:?}", err);
+            assert!(as_text.contains("request attempt 4 failed"));
```
chkeita, line 117: this should match the max number of attempts (3) specified in the call line 112

bmc-msft: In practice, this count is actually the number of retries.  1 initial, 3 retries.  so, 4 attempts.

chkeita (approved):

PR #699, opened 2021-03-19
Update Windows generic coverage recording

ranweiler:
- Reimplement Windows generic coverage using new coverage format
- Remove old format
- Update and unify examples


ranweiler:
Will be adding filtering in follow-up PR.

src/agent/input-tester/src/crash_detector.rs
```diff
@@ -160,48 +152,10 @@ impl<'a> CrashDetectorEventHandler<'a> {
             stderr_buffer: vec![],
             debugger_output: String::new(),
             exceptions: vec![],
-            id_to_block: None,
```
ranweiler, line 163: Splitting up the new `Recorder` from its debugger event handler, and will splice it into this struct.

lzybkr: Beyond what we discussed, this will also be useful for other fuzzers like radamsa.

src/agent/input-tester/src/crash_detector.rs
```diff
@@ -313,26 +270,43 @@ impl<'a> DebugEventHandler for CrashDetectorEventHandler<'a> {
...
 /// the process or any children processes.
-pub fn test_process(
+pub fn test_process<'a>(
     app_path: impl AsRef<OsStr>,
     args: &[impl AsRef<OsStr>],
     env: &HashMap<String, String>,
     max_duration: Duration,
     ignore_first_chance_exceptions: bool,
-    coverage_map: Option<&mut AppCoverageBlocks>,
+    cache: Option<&'a mut ModuleCache>,
```
ranweiler, line 309: Assumes we retain just this module disassembly cache in-memory across runs, or reload it from disk.

src/agent/coverage/src/block/windows.rs
```diff
@@ -7,162 +7,242 @@ use std::time::{Duration, Instant};
...
         }
     }
 
     fn on_poll(&mut self, dbg: &mut Debugger) {
-        self.handle_on_poll(dbg);
+        self.on_poll(dbg);
     }
 }
+
+#[derive(Clone, Debug, Default)]
```
ranweiler, line 205: Add doc comment.

src/agent/coverage/src/block/windows.rs
```diff
@@ -7,162 +7,257 @@ use std::time::{Duration, Instant};
...
 
-        log::info!(
-            "exe loaded: {}, {} bytes",
-            module.path().display(),
-            module.image_size(),
-        );
+    // Note: this could also be a reference to enable reuse across runs, to
+    // support implicit calculation of total coverage for a corpus. For now,
+    // assume callers will merge this into a separate struct when needed.
+    coverage: CommandBlockCov,
```
lzybkr, line 84: It seems reasonable for `Recorder` to **not** do any merging.

I do wonder though what mechanism we would use to skip collecting coverage on already hit blocks. This would speed up collecting total coverage for a corpus and would also speed up fuzzing iterations when using radamsa.

ranweiler: Good point. I was thinking we'll want the `ModuleCache` itself to be append-only. However, we could still add a second optional reference to an existing (total) `CommandBlockCov`, and when setting breakpoints, we could skip registration if a hit already occurs in the existing cov. I think this may warrant separate profiling, however.

src/agent/coverage/src/block/windows.rs
```diff
@@ -7,162 +7,257 @@ use std::time::{Duration, Instant};
...
+                    log::trace!("{:>16x}: {}+{:x}", pc, name, offset);
+                }
             }
 
-            self.coverage.report_block_hit(m, b);
+            self.coverage
+                .increment(breakpoint.module, breakpoint.offset)?;
         } else {
-            log::error!("no block for breakpoint");
+            log::error!("hit breakpoint without data: {}", id.0);
```
lzybkr, line 153: You'll probably want to log RIP here as well.

src/agent/coverage/src/code.rs
```diff
@@ -195,6 +195,20 @@ impl ModuleIndex {
             symbols,
         })
     }
+
+    #[cfg(target_os = "windows")]
+    pub fn index_pe(path: ModulePath, pe: &goblin::pe::PE) -> Self {
```
lzybkr, line 200: With no impl or comments, it's hard to tell why there is a `parse_elf` for Linux but no `index_elf`, and a `index_pe` for Windows bug no `parse_pe`.

src/agent/input-tester/src/tester.rs
```diff
@@ -78,7 +71,6 @@ pub struct Tester {
     ignore_first_chance_exceptions: bool,
     appverif_controller: Option<AppVerifierController>,
     bugs_found_dir: PathBuf,
-    coverage_map: Option<AppCoverageBlocks>,
```
lzybkr, line 81: Is the plan to add a `Option<CommandgBlockCov>` here in a future PR?

lzybkr (approved):

src/agent/coverage/src/code.rs
ranweiler: Will clean up the naming and document.

src/agent/input-tester/src/tester.rs
ranweiler: Yeah, but I can just add that right now.

ranweiler: Ah right, I take it back, it makes more sense to do this in a follow-up. Before we were loading it from a directory in the ctor, but there's more to think about now.

ranweiler: https://github.com/microsoft/onefuzz/issues/712

ranweiler:
Integration tests queued.

PR #700, opened 2021-03-19
add log checking to refactored integration check

bmc-msft:
In practice, Application Insights can take up to 3 minutes before something sent to it is available via KQL.

This PR logs a start and stop marker such that the integration tests only search for logs during the integration tests. This reduces the complexity when using the integration tests during the development process.

Note: this migrated the new functionality from #356 into the latest integration test tools.


src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -214,9 +214,6 @@ def on_worker_event_done(machine_id: UUID, event: WorkerDoneEvent) -> Result[Non
             node.debug_keep_node = True
             node.save()
     else:
-        logging.error(
```
bmc-msft, line 217: note, this is now handled by `task.mark_failed`.  Since we don't record errors on tasks already marked as failures, this moves the error longing to model our state handling.

src/api-service/__app__/onefuzzlib/azure/vm.py
```diff
@@ -136,7 +136,7 @@ def get_extension(vm_name: str, extension_name: str) -> Optional[Any]:
             resource_group, vm_name, extension_name
         )
     except (ResourceNotFoundError, CloudError) as err:
-        logging.error("extension does not exist %s", err)
+        logging.info("extension does not exist %s", err)
```
bmc-msft, line 139: this is common place when we're deploying VMs.  shouldn't be an error.

src/api-service/__app__/onefuzzlib/notifications/main.py
```diff
@@ -108,9 +108,12 @@ def get_queue_tasks() -> Sequence[Tuple[Task, Sequence[str]]]:
 
 
 def new_files(container: Container, filename: str) -> None:
-    report = get_report_or_regression(container, filename)
-
     notifications = get_notifications(container)
+
+    report = get_report_or_regression(
```
bmc-msft, line 113: since we're going to try to inspect every new file as a potential report, this sets up the parsing (seen later in this PR) to only log errors if we actually expect the files to be a report, rather than just optimistically being a report.

src/api-service/__app__/onefuzzlib/tasks/main.py
```diff
@@ -205,6 +205,8 @@ def mark_failed(
             )
             return
 
+        logging.error("task failed %s:%s - %s", self.job_id, self.task_id, error)
```
bmc-msft, line 208: See previous comment about `mark_failed`.

src/integration-tests/integration-test.py
```diff
@@ -686,6 +692,88 @@ def cleanup(self) -> None:
...
+        wait(self.check_log_end_marker, frequency=5.0)
+        self.logger.info("application insights log flushed")
+
+        logs = self.of.debug.logs.keyword("error", limit=100000, timespan="PT3H")
+
+        seen_errors = False
+        seen_stop = False
+        for entry in logs:
+            entry_as_str = json.dumps(entry, sort_keys=True)
+            if not seen_stop:
```
chkeita, line 751: it looks like you are iterating over the logs in reverse order. Maybe add a comment to that effect.

src/integration-tests/integration-test.py
```diff
@@ -686,6 +692,88 @@ def cleanup(self) -> None:
...
+                if self.stop_log_marker in entry_as_str:
+                    seen_stop = True
+                continue
+
+            if self.start_log_marker in entry_as_str:
+                break
```
chkeita, line 757: it should be faster to just compare the markers to entry.get("message") directly instead going through the entire json

chkeita (approved):

PR #704, opened 2021-03-20
automatically retry supervisor requests

### LINKED ISSUE #677: supervisor should automatically retry it's API requests
The supervisor's coordinator, which polls the Azure Functions instance will only retry hitting the API endpoint if the status code is `UNAUTHORIZED`.  This means that supervisors that poll the service during an upgrade or intermittent downtime will bail.

These should be updated to use the `reqwest-retry` library.

Current source:
https://github.com/microsoft/onefuzz/blob/7ebdeac537218bd6094c556e45710be5f695de5e/src/agent/onefuzz-supervisor/src/coordinator.rs#L236-L254

chkeita:

closes #677


src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -233,7 +234,13 @@ impl Coordinator {
         request_type: RequestType<'a>,
     ) -> Result<Response> {
         let request = self.build_request(request_type.clone())?;
-        let mut response = self.client.execute(request).await?;
```
bmc-msft, line 236: since self.client isn't used anymore, it should probably get removed all together.

chkeita: it is still used to create the RequestBuilder see poll_commands_request on line 267

bmc-msft: ah, yeah I missed that.

bmc-msft (dismissed):

bmc-msft:
During two different integration test runs, every task is failing, though I think the issue isn't related to this change.  

As such, I've marked this as do-not-merge while I debug it.

bmc-msft (approved):

PR #713, opened 2021-03-22
run until end of LLVMFuzzerTestOneInput for collecting coverage

bmc-msft:
As is, we ran the libfuzzer target until `NtTerminateProcess`.  For the DLLs that got auto-unloaded, this doesn't record the coverage.

Instead, this moves to breaking on return from `LLVMFuzzerTestOneInput`, which allows gathering coverage for the DLLs that are auto-unloading on the targets in question.


src/agent/script/win64/libfuzzer-coverage/DumpCounters.js
```diff
@@ -98,8 +98,12 @@ function dumpCounters(results_dir, should_disable_sympath) {
     // Disable FCE from sanitizers.
     execute("sxd av");
 
-    // Run to exit break in `ntdll!NtTerminateProcess`.
-    execute("g");
+    // run until LLVMFuzzerTestOneInput
```
ranweiler, line 101:
```suggestion
    // Run until `LLVMFuzzerTestOneInput()`.
    // This makes us unlikely to have unloaded any modules that the user dynamically loaded,
    // and so we will still be able to dump the coverage tables for those modules.
```

ranweiler: I believe this newly narrows the applicability of this script to libFuzzer targets only, which wasn't the case previously (or at least originally). In practice, that's just fine.

ranweiler (approved):

PR #714, opened 2021-03-23
make asan_log optional in CrashReport

bmc-msft:


bmc-msft:
@ranweiler, I was mistaken.  asan_log was already an Option.

chkeita (approved):

PR #716, opened 2021-03-23
work around AAD service principal race condition

bmc-msft:
This works around an issue in AAD service principal creation.  The race condition in AAD shows up as:

```
  File "deploy.py", line 926, in <module>
    main()
  File "deploy.py", line 920, in main
    state[1](client)
  File "deploy.py", line 303, in setup_rbac
    client.service_principals.create(service_principal_params)
  File "/tmp/tmpp2x7ybfg/deploy-venv/lib/python3.8/site-packages/azure/graphrbac/operations/service_principals_operations.py", line 87, in create
    raise models.GraphErrorException(self._deserialize, response)
azure.graphrbac.models.graph_error_py3.GraphErrorException: When using this permission, the backing application of the service principal being created must in the local tenant
```

The azure-cli has the same issue, as seen in https://github.com/Azure/azure-cli/issues/14767


src/deployment/deploy.py
```diff
@@ -300,7 +300,22 @@ def setup_rbac(self) -> None:
...
+                if (
+                    "service principal being created must in the local tenant"
+                    not in str(err)
+                ):
+                    raise err
+                logging.warning(
+                    "creating service principal failed with an error that occurs "
+                    "due to AAD race conditions.  retrying after delay"
+                )
+                time.sleep(60)
```
ranweiler, line 317: It looks like we're only retrying once, after a 1 min sleep.
1. Would it make more sense to retry N times with backoff?
2. If we fail with with the suspicious string in the `err` text, it would be good to emphasize that it may still be a resource creation race, but that we intentionally gave up after retrying.

ranweiler (approved):

PR #723, opened 2021-03-23
add symbol and module names to StackFrame

bmc-msft:
This exposes the module_info and symbol name from debugger in the StackFrame.  This enables the stack minimization function work on function names.


bmc-msft:
Current output:
```json
{
  "crash_report": {
    "input_sha256": "8abb9b1b07f17b5442c5d71b38eaf599ddd37fa38a3062c709a7e68412e57922",
    "executable": "C:\\Users\\bcaswell\\projects\\test-artifacts\\windows-trivial-crash\\fuzz.exe",
    "crash_type": "WriteAccessViolation",
    "crash_site": "fuzz!strncpy minkernel\\crts\\ucrt\\src\\appcrt\\string\\amd64\\strncpy.asm:46",
    "call_stack": [
      "fuzz!strncpy minkernel\\crts\\ucrt\\src\\appcrt\\string\\amd64\\strncpy.asm:46",
      "fuzz!check D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:21",
      "fuzz!from_file D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:67",
      "<corrupt frame(s)>"
    ],
    "call_stack_sha256": "8c0026819742a73973b587fd9e18545b1fd2770ebd65b9141f24bcc6c4e47b9c",
    "minimized_stack": [
      "fuzz!strncpy minkernel\\crts\\ucrt\\src\\appcrt\\string\\amd64\\strncpy.asm:46",
      "fuzz!check D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:21",
      "fuzz!from_file D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:67",
      "<corrupt frame(s)>"
    ],
    "minimized_stack_sha256": "8c0026819742a73973b587fd9e18545b1fd2770ebd65b9141f24bcc6c4e47b9c",
    "minimized_stack_function_names": [
      "fuzz!strncpy",
      "fuzz!check",
      "fuzz!from_file"
    ],
    "minimized_stack_function_names_sha256": "58cd174f108dcd9cf568209a818dcae72d4fb7b0b85037598e26c214f3c44fce",
    "task_id": "00000000-0000-0000-0000-000000000000",
    "job_id": "00000000-0000-0000-0000-000000000000"
  }
}
```

New output:
```json
{
  "crash_report": {
    "input_sha256": "8abb9b1b07f17b5442c5d71b38eaf599ddd37fa38a3062c709a7e68412e57922",
    "executable": "C:\\Users\\USERNAME\\projects\\test-artifacts\\windows-trivial-crash\\fuzz.exe",
    "crash_type": "ReadAccessViolation",
    "crash_site": "fuzz!from_file D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:70",
    "call_stack": [
      "fuzz!from_file D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:70",
      "<corrupt frame(s)>"
    ],
    "call_stack_sha256": "361ea42294b00fd685652fb327a66231907d8f9703155b436eb2401c43d14a8c",
    "minimized_stack": [
      "fuzz!from_file D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:70",
      "<corrupt frame(s)>"
    ],
    "minimized_stack_sha256": "361ea42294b00fd685652fb327a66231907d8f9703155b436eb2401c43d14a8c",
    "minimized_stack_function_names": [
      "from_file"
    ],
    "minimized_stack_function_names_sha256": "710dc924cf18883634a64c005c2b88f0199c041c10fa3fa360f62ca3a83de2da",
    "task_id": "00000000-0000-0000-0000-000000000000",
    "job_id": "00000000-0000-0000-0000-000000000000"
  }
}

File: src/agent/debugger/src/stack.rs [hunk 1]
```diff
@@ -86,13 +85,25 @@ pub enum DebugStackFrame {
...
+        module_name: Option<String>,
     },
     CorruptFrame,
 }
 
 impl DebugStackFrame {
-    pub fn new(function: String, location: DebugFunctionLocation) -> DebugStackFrame {
-        DebugStackFrame::Frame { function, location }
+    pub fn new(
+        function: String,
```
lzybkr, line 96: This is a little confusing now, maybe it'd be better to remove the `function` field.

That said, I'm not sure it's a good idea to allow frames with no module or symbol as you have no position to render.

Obviously some locations won't have a module (e.g. JIT compiled code) and some locations won't have symbols (again, JIT compiled code).

I do think it's useful to call out an unknown module. The current rendering uses the VA though, and that's less useful for hashing as it is likely not stable. Better would be to convert to some sort of RVA, perhaps based on the memory region's base address (`win_util::memory::get_memory_info` can return the base address of the region for a given address.)

ranweiler: Before this PR, it looks like we set `function` to the _module name_ when we can't find the symbol info: https://github.com/microsoft/onefuzz/blob/main/src/agent/debugger/src/stack.rs#L209-L212

So if we're going to change the `DebugStackFrame::Frame` variant, it might be a good opportunity to clean that up, too. We should always have a module (even if just `UNKNOWN_MODULE`) and relative offset, and that's part of what makes the current definition even harder to follow, I think. The optionality is sort of backwards.

ranweiler: > Better would be to convert to some sort of RVA, perhaps based [...]

@lzybkr, this is exactly what you impl'd [here](https://github.com/microsoft/onefuzz/blob/main/src/agent/debugger/src/stack.rs#L225-L227), right?

lzybkr: Yep, forgot about that, though maybe that's why I added `get_memory_info` in the first place.

src/agent/debugger/src/stack.rs
```diff
@@ -84,15 +83,24 @@ impl Display for DebugFunctionLocation {
 #[derive(Clone, Debug, Hash, PartialEq)]
 pub enum DebugStackFrame {
     Frame {
-        function: String,
+        module_name: String,
         location: DebugFunctionLocation,
```
ranweiler, line 87: This is basically what I had in mind.

The only latently weird thing is the `location` field, which has type `DebugFunctionLocation`. It is now misnamed (when `symbol` is `None`), and we are overloading it semantically.

WDYT about a struct variant like:
```rust
Frame {
    module_name: String,
    module_offset: u64,
    symbol: Option<Symbol>,
    symbol_location: Option<DebugFunctionLocation)>,
}
```
We'd maybe want to group the latter two into a single optional value, if they should always be present/absent together.

bmc-msft: I'm fine with that, though if we're promoting module_offset to Frame, the only entity in DebugFrameLocation is file_info, which could get promoted and renamed.  

Any thoughts on this @lzybkr?

bmc-msft: updated to incorporate @ranweiler's suggestions.

src/agent/debugger/src/stack.rs
```diff
@@ -120,18 +81,19 @@ impl Display for DebugStackFrame {
...
                 } else {
                     write!(formatter, "{}", module_name)?;
                 }
-                if let Some(file_info) = &location.file_info {
-                    write!(formatter, " {}", file_info)
+                if let Some(symbol_location) = &symbol_location {
+                    write!(formatter, " {}", symbol_location)
                 } else {
-                    write!(formatter, "+0x{:x}", location.displacement)
+                    write!(formatter, "+0x{:x}", module_offset)
```
ranweiler, line 96: This is looking good, I think we just lost one thing.

Before this PR, we overloaded the `DebugFunctionLocation.displacement` field, so it would either mean "offset, relative to the image base of the module" or "offset from the base of the _symbol's_ mapped image base". With this most recent change, it always means "module-relative offset" (unless I misread something). 

Suppose:
- `mymodule` is mapped to `0x1000`
- `mymodule!myfunc` is mapped to `0x1200`
- The PC is `0x1234`

When we don't have source line info, we want to see any of the following, in order of decreasing preference:
1. `mymodule!myfunc+0x34`
2.  `mymodule+0x234`  (if no symbols)
3. `<UnknownModule>+0x234`  (if our debugger APIs fail to give us the module name for some reason)
4. `<corrupt stack frame(s)>` (if we can't even use the above fallback where try to guess a module base)

As written, I think in case (1), we'd emit `mymodule!myfunc+0x234`. We really want to grab, save, and use the [`SymInfo.displacement`](https://github.com/microsoft/onefuzz/blob/main/src/agent/debugger/src/stack.rs#L197) when we can get it (under the hood, from [`SymFromInlineContextW()`](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfrominlinecontextw)).

ranweiler (changes requested):

src/agent/debugger/src/stack.rs
bmc-msft: addressed

ranweiler (approved):

bmc-msft:
Note, with the above changes, the output now looks like:
```json
{
  "crash_report": {
    "input_sha256": "24c3bb3be2e50c36320b9e737f81438e0ec34835ce39def8bec9921b64ac17e8",
    "executable": "C:\\Users\\USERNAME\\projects\\test-artifacts\\windows-trivial-crash\\fuzz.exe",
    "crash_type": "WriteAccessViolation",
    "crash_site": "fuzz!strncpy+0x30 minkernel\\crts\\ucrt\\src\\appcrt\\string\\amd64\\strncpy.asm:46",
    "call_stack": [
      "fuzz!strncpy+0x30 minkernel\\crts\\ucrt\\src\\appcrt\\string\\amd64\\strncpy.asm:46",
      "fuzz!check+0x1d D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:21",
      "fuzz!from_file+0xa7 D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:67",
      "<corrupt frame(s)>"
    ],
    "call_stack_sha256": "da4ee7a170c47a7d34d92bda06d030e84060120396f26e8f938538e2dae9ceba",
    "minimized_stack": [
      "fuzz!check+0x1d D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:21",
      "fuzz!from_file+0xa7 D:\\a\\onefuzz\\onefuzz\\src\\integration-tests\\trivial-crash\\fuzz.c:67",
      "<corrupt frame(s)>"
    ],
    "minimized_stack_sha256": "9d706a2aadca7cb39d81b01a0bab178c17792c83baa2a085baa2bfe167e5711e",
    "minimized_stack_function_names": [
      "check",
      "from_file"
    ],
    "minimized_stack_function_names_sha256": "a6577cb7f0b98e9eb955baedc8ed6d2192852f1aa2acb562985c853748a9f664",
    "task_id": "00000000-0000-0000-0000-000000000000",
    "job_id": "00000000-0000-0000-0000-000000000000"
  }
}
```

lzybkr (approved):

PR #730, opened 2021-03-25
updates related to cargo clippy 1.51.0

bmc-msft:


src/agent/onefuzz-agent/src/tasks/stats/common.rs
```diff
@@ -8,6 +8,9 @@ use onefuzz_telemetry::{track_event, Event::runtime_stats};
 use serde::Deserialize;
 pub const STATS_DELAY: std::time::Duration = std::time::Duration::from_secs(30);
 
+// TODO - remove unkonwn_lints once GitHub build agents are at 1.51.0 or later
+#[allow(unknown_lints)]
+#[allow(clippy::upper_case_acronyms)]
 #[derive(Debug, Deserialize, Clone)]
 pub enum StatsFormat {
     AFL,
```
ranweiler, line 16: FWIW, I do think we should just change this one to `Afl`, assuming it isn't too intrusive.

ranweiler (approved):

PR #736, opened 2021-03-26
address telemetry log message locality

bmc-msft:
Prior to this fix, all logging would show up as:

```
[2021-03-26T19:53:31Z INFO  onefuzz_telemetry] Starting libFuzzer crash report task
```

This moves the actual calls to `log::log!()` to the macros, which puts it in the source file in question.

Now the events look like:
```
[2021-03-26T21:20:14Z INFO  onefuzz_agent::tasks::report::libfuzzer_report] Starting libFuzzer crash report task
```

Most importantly, this allows us to use env_logger's ability to filter on a per module basis in a meaningful fashion.  As an example, this allows:
```
RUST_LOG="info,onefuzz::syncdir=warn,onefuzz_agent::tasks::fuzz::libfuzzer=warn"
```

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/stats/common.rs
+++ b/src/agent/onefuzz-agent/src/tasks/stats/common.rs
@@ -21,7 +21,7 @@ pub async fn monitor_stats(path: Option<String>, format: Option<StatsFormat>) ->
                     StatsFormat::AFL => afl::read_stats(&path).await,
                 };
                 if let Ok(stats) = stats {
-                    event_entries!(runtime_stats; stats);
+                    log_events!(runtime_stats; stats);
                 }
                 delay_with_jitter(STATS_DELAY).await;
             }
--- a/src/agent/onefuzz-agent/src/tasks/stats/common.rs
+++ b/src/agent/onefuzz-agent/src/tasks/stats/common.rs
@@ -4,7 +4,7 @@
 use super::afl;
 use anyhow::{Error, Result};
 use onefuzz::jitter::delay_with_jitter;
-use onefuzz_telemetry::{track_event, Event::runtime_stats};
+use onefuzz_telemetry::Event::runtime_stats;
 use serde::Deserialize;
 pub const STATS_DELAY: std::time::Duration = std::time::Duration::from_secs(30);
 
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz-agent/src/tasks/stats/common.rs
```diff
@@ -21,7 +21,7 @@ pub async fn monitor_stats(path: Option<String>, format: Option<StatsFormat>) ->
                     StatsFormat::AFL => afl::read_stats(&path).await,
                 };
                 if let Ok(stats) = stats {
-                    track_event(runtime_stats, stats);
+                    event_entries!(runtime_stats; stats);
```
bmc-msft, line 24: previously, we did not log the AFL stats.  it was only sent to Application Insights.

ranweiler: That's strange. It seems like it should have, since `track_event()` called the now-removed `local_log_event()`.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -492,23 +492,22 @@ pub fn set_property(entry: EventData) {
...
-fn local_log_event(event: &Event, properties: &[EventData]) {
-    let as_values = properties
+pub fn format_events(events: &[EventData]) -> String {
+    events
         .iter()
         .map(|x| x.as_values())
         .map(|(x, y)| format!("{}:{}", x, y))
         .collect::<Vec<String>>()
-        .join(" ");
-    log::log!(log::Level::Info, "{} {}", event.as_str(), as_values);
```
ranweiler, line 502: So this line was the core problem, right? `log::log!()` got expanded in `onefuzz-telemetry`, and not in call sites in the consumer crate. As a result, it'd pick up the wrong module name.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -540,18 +538,24 @@ pub fn to_log_level(level: &appinsights::telemetry::SeverityLevel) -> log::Level
...
 pub fn log_message(level: appinsights::telemetry::SeverityLevel, msg: String) {
-    let log_level = to_log_level(&level);
-    log::log!(log_level, "{}", msg);
     if let Some(client) = client(ClientType::Instance) {
         client.track_trace(msg, level);
     }
 }
 
+#[macro_export]
+macro_rules! event_entries {
```
ranweiler, line 548: WDYT about naming this "`log_events!()`"?

ranweiler (approved):

PR #738, opened 2021-03-27
cleanup local job_id directory upon clean exit

bmc-msft:


src/agent/onefuzz-agent/src/local/libfuzzer_fuzz.rs
```diff
@@ -51,7 +51,9 @@ pub fn build_fuzz_config(args: &clap::ArgMatches<'_>, common: CommonConfig) -> R
 
 pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
     let common = build_common_config(args, true)?;
+    register_cleanup(common.job_id)?;
     let config = build_fuzz_config(args, common)?;
+
```
bmc-msft, line 56:
```suggestion
```

src/agent/onefuzz-agent/src/local/radamsa.rs
```diff
@@ -18,7 +18,9 @@ use uuid::Uuid;
 
 pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
     let common = build_common_config(args, true)?;
+    register_cleanup(common.job_id)?;
     let fuzz_config = build_fuzz_config(args, common.clone())?;
+
```
bmc-msft, line 23:
```suggestion
```

src/agent/onefuzz-agent/src/local/generic_analysis.rs
```diff
@@ -61,6 +62,7 @@ pub fn build_analysis_config(
 
 pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
     let common = build_common_config(args, true)?;
+    register_cleanup(common.job_id)?;
```
chkeita, line 65: can we make the behavior configurable with a paramter. I used 'keep_job_dir' in my local run ui PR

chkeita (approved):

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -162,6 +167,30 @@ pub fn get_synced_dirs(
     Ok(dirs?)
 }
 
+fn recursive_remove(path: &Path) -> Result<()> {
```
ranweiler, line 170: It seems like this is recreating `remove_dir_all()`, while also calling it. What's the motivation here?

AFAICT, the differences are that we (1) include paths in some of the error messages, when recursively deleting, (2) we allow each recursive remove call to possibly fail one time due to some sort of race.

bmc-msft: The issue I was experiencing was due to Windows recursion.  I've updated the impl to just use the [remove_dir_all](https://docs.rs/remove_dir_all/0.7.0/remove_dir_all/) crate which handles the issues in a more coherent fashion.

ranweiler (approved):

PR #739, opened 2021-03-27
change the scale of timeouts to seconds instead of minutes

bmc-msft:


src/agent/onefuzz-agent/src/local/cmd.rs
```diff
@@ -49,8 +49,8 @@ pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
         }
     };
 
-    if let Some(minutes) = running_duration {
-        if let Ok(run) = timeout(Duration::from_secs(minutes * 60), run).await {
+    if let Some(seconds) = running_duration {
```
chkeita, line 52: Can you update the description of the timeout paramter to specify that the unit is now second

chkeita (approved):

PR #743, opened 2021-03-27
setup the coverage task to ignore the INPUTS_DIR when used with libFuzzer

chkeita:
The current local libfuzzer job was setting up the coverage task to read from the `INPUTS_DIR` and the input queue by default. 
This was causing the coverage task to start running on all the inputs at the beginning of the fuzzing session potentially consuming a lot of the resource in the process. 
This PR rmoves the `INPUTS_DIR` from the coverage task. it can still be specified with the `READONLY_INPUTS` parameter


bmc-msft:
If I have an existing input file when I start the local libFuzzer job, under this change, does the existing input get analyzed for coverage?

chkeita:
> If I have an existing input file when I start the local libFuzzer job, under this change, does the existing input get analyzed for coverage?

No the existing inputs will not get picked up unless the the dir is specified in the `READONLY_INPUTS` paramter

bmc-msft:
> > If I have an existing input file when I start the local libFuzzer job, under this change, does the existing input get analyzed for coverage?
> 
> No the existing inputs will not get picked up unless the the dir is specified in the `READONLY_INPUTS` paramter

bmc-msft:
This is a pretty significant difference to how our cloud-based libFuzzer job works and not what I would expect as a user.  I expect adding `--coverage_dir` to gather coverage for all of the inputs for my target, not just the ones found from the start of the new job.

With this change, if a user provides a  libFuzzer target where their inputs include all possible paths, this change would have the coverage report will say 0% covered rather than 100% covered.

chkeita:
My intent was to make the behavior of the coverage task explicit so we can control it when using the local task. I could change the behavior to take the input file and add an additional paramter to ignore it if we want

PR #750, opened 2021-03-29
Yield on each enqueue and pop on the local queue

chkeita:


src/agent/storage-queue/src/local_queue.rs
```diff
@@ -113,19 +113,28 @@ pub struct ChannelQueueClient {
     sender: Arc<Mutex<UnboundedSender<Vec<u8>>>>,
     receiver: Arc<Mutex<UnboundedReceiver<Vec<u8>>>>,
     pub url: reqwest::Url,
+    low_resource: bool,
 }
 
 impl ChannelQueueClient {
     pub fn new() -> Result<Self> {
         let (sender, receiver) = unbounded_channel();
+        let cpus = num_cpus::get() as u64;
```
bmc-msft, line 122: We shouldn't need to cast this to u64.
```suggestion
        let cpus = num_cpus::get();
```

bmc-msft (approved):

PR #752, opened 2021-03-30
verify libfuzzer targets at the start of the task

bmc-msft:
Changes in this PR:
1. In addition to the existing `-help=1` verification, this adds checking up to 5 inputs from each available seeds directory.
2. If no seeds directories are provided for the task, then this checks that an empty file parses.
3. Adds a helper that unifies libFuzzer argument builders, which ensure all libFuzzer related tasks use the same behavior for argument buliding.
4. Changes libfuzzer_fuzz to pull seeds prior to starting the fuzzing.  Previously, we started fuzzing and pulled the seeds as we went.  This allowed targets with huge input corpus containers to start fuzzing immediately at the cost of potentially duplicating input seed discovery.  As we want to verify some number of seeds parse as expected, this undoes that optimization.


src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -170,7 +170,9 @@ pub fn get_synced_dirs(
 fn register_cleanup(job_id: Uuid) -> Result<()> {
     let path = std::env::current_dir()?.join(job_id.to_string());
     atexit::register(move || {
-        remove_dir_all(&path).expect("cleanup failed");
+        if std::fs::metadata(&path).is_ok() {
+            remove_dir_all(&path).expect("cleanup failed");
```
chkeita, line 174: maybe check that the path still exist before throwing here

chkeita (approved):

PR #754, opened 2021-03-30
remove supervisor-is-done upon test completion

bmc-msft:


src/agent/onefuzz-supervisor/src/agent/tests.rs
```diff
@@ -221,4 +221,8 @@ async fn test_emitted_state_failed_setup() {
     let coordinator: &CoordinatorDouble = agent.coordinator.downcast_ref().unwrap();
     let events = &coordinator.events;
     assert_eq!(events, &expected_events);
+
+    tokio::fs::remove_file(crate::done::done_path().unwrap())
```
ranweiler, line 225: This is kind of a band-aid over the real problem, which is that the agent impl is writing to the filesystem in a unit test where IO is otherwise abstracted.

I can understand not wanting to go down that rabbit hole ATM, but we should add a comment for whoever ends up fixing this later.

PR #756, opened 2021-03-30
Add command block coverage merging

ranweiler:
- Add coverage merging for module and command-level block coverage, tests
- Improve an error message


src/agent/coverage/src/block.rs
```diff
@@ -40,12 +40,10 @@ impl CommandBlockCov {
 
     pub fn increment(&mut self, path: &ModulePath, offset: u64) -> Result<()> {
         if let Some(module) = self.modules.get_mut(path) {
```
ranweiler, line 42: Now, when we _merge_ command coverage, we now add unknown modules to the overall coverage. Here, we drop increment attempts for unknown modules. The reason this changed is that although right now we expect to precompute possible coverage locations, we'd like this format to support coverage discovery.

We should probably create a new (empty) `ModuleCov` for `path` as-needed.

src/agent/coverage/src/block.rs
```diff
@@ -69,6 +74,22 @@ impl ModuleCov {
         let blocks = offsets.map(|o| (o, BlockCov::new(o))).collect();
         Self { blocks }
     }
+
+    pub fn increment(&mut self, offset: u64) {
+        if let Some(block) = self.blocks.get_mut(&offset) {
```
ranweiler, line 79: Again, we should probably create an entry on demand.

src/agent/coverage/src/block.rs
```diff
@@ -95,3 +118,151 @@ impl BlockCov {
...
+        cov
+    }
+
+    // Builds a vec of `(count, offset)` tuples from a `ModuleCov`.
+    fn to_vec(cov: &ModuleCov) -> Vec<(u64, u32)> {
+        cov.blocks.iter().map(|(o, b)| (*o, b.count)).collect()
+    }
+
+    #[test]
+    fn test_to_from_vec_roundtrip() {
```
lzybkr, line 144: Based on the test name and code, it looks like it only tests test code, but I think it's meant to test `increment` so maybe a different function name would be better?

ranweiler: Oops, nope, this is just a vestigial self-check test.

lzybkr (approved):

PR #765, opened 2021-04-01
Use custom serialization for block coverage

ranweiler:
Update command block coverage serialization to support a more compact JSON format. See tests for examples.

### DIFF
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -101,7 +101,6 @@ impl ModuleCov {
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct BlockCov {
     /// Offset of the block, relative to the module base load address.
-    #[serde(with = "hex")]
     pub offset: u64,
 
     /// Number of times a block was seen to be executed, relative to some input
@@ -172,46 +171,6 @@ mod array {
             }
 
             Ok(map)
-        }
-    }
-}
-
-mod hex {
-    use std::fmt;
-
-    use serde::de::{self, Deserializer, Visitor};
-    use serde::ser::Serializer;
-
-    pub fn serialize<S>(data: &u64, ser: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        let s = format!("0x{:x}", data);
-        ser.serialize_str(&s)
-    }
-
-    pub fn deserialize<'d, D>(de: D) -> Result<u64, D::Error>
-    where
-        D: Deserializer<'d>,
-    {
-        de.deserialize_str(HexVisitor)
-    }
-
-    struct HexVisitor;
-
-    impl<'d> Visitor<'d> for HexVisitor {
-        type Value = u64;
-
-        fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
-            write!(f, "lowercase hex string with `0x` prefix")
-        }
-
-        fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
-        where
-            E: de::Error,
-        {
-            let h = s.strip_prefix("0x").unwrap_or(s);
-            u64::from_str_radix(h, 16).map_err(de::Error::custom)
         }
     }
 }
@@ -381,13 +340,13 @@ mod tests {
     #[test]
     fn test_block_cov_serde() {
         let block = BlockCov {
-            offset: 0x123,
+            offset: 123,
             count: 456,
         };
 
         let ser = serde_json::to_string(&block).unwrap();
 
-        let text = r#"{"offset":"0x123","count":456}"#;
+        let text = r#"{"offset":123,"count":456}"#;
         assert_eq!(ser, text);
 
         let de: BlockCov = serde_json::from_str(&ser).unwrap();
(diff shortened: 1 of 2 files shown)


src/agent/coverage/src/block.rs
```diff
@@ -16,7 +16,8 @@ use crate::code::ModulePath;
 /// Block coverage for a command invocation.
 ///
 /// Organized by module.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
```
ranweiler, line 19: Note that this also impls the `serde` traits for this type, which is new.

src/agent/coverage/src/block.rs
```diff
@@ -280,4 +377,58 @@ mod tests {
...
+            cov.increment(&some_dll, 0x30);
+            cov
+        };
+
+        let ser = serde_json::to_string(&cov).unwrap();
+
+        #[cfg(target_os = "linux")]
+        let text = r#"{"/lib/some.dll":[{"offset":"0x2","count":0},{"offset":"0x30","count":1},{"offset":"0x400","count":0}],"/main.exe":[{"offset":"0x1","count":1},{"offset":"0x20","count":0},{"offset":"0x300","count":1}]}"#;
+
+        #[cfg(target_os = "windows")]
```
ranweiler, line 426: This is bad, as is the conditional compilation in the existing tests. This has to do with how `ModulePath` validates absolute paths in an OS-dependent way. Will address separately.

src/agent/coverage/src/block.rs
```diff
@@ -119,6 +123,99 @@ impl BlockCov {
...
+    use std::fmt;
+
+    use serde::de::{self, Deserializer, Visitor};
+    use serde::ser::Serializer;
+
+    pub fn serialize<S>(data: &u64, ser: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let s = format!("0x{:x}", data);
```
bmc-msft, line 189: This isn't supported by python's json parser.  We shouldn't use a serialization format that requires custom parsers. 

Example:

```
>>> json.loads('{"a": 3}')
{'a': 3}
>>> json.loads('{"a": 0x3}')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.8/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 8 (char 7)```

Comment (bmc-msft) [reply src/agent/coverage/src/block.rs:189]: A string hex value is specifically suboptimal, as you can't immediately use it.

Comment (ranweiler) [reply src/agent/coverage/src/block.rs:189]: It doesn't get serialized as a prefix numeric literal, it gets serialized as a string (in double quotes).

Comment (bmc-msft) [reply src/agent/coverage/src/block.rs:189]: Understood.

That's what I meant by a string hex value can't be immediately used.  It means we have to write a custom parser to use the data.

"0x20" isn't a value that users can immediately use.  They have to parse the string as hex, then use that.

Comment (ranweiler) [reply src/agent/coverage/src/block.rs:189]: Ah. Yeah, that's true if we want to optimize for doing arithmetic.

There are two problems I'm trying to solve here:
1. JSON numbers _can_ be lossy (e.g. if they pass through a JS interpreter)
2. We are talking about memory locations and offsets. I assume users will typically want to _display_ those addrs/offsets, rather than do arithmetic on them. In shell one-liner, it'd be annoying to format a decimal number as hex. So, I'm also optimizing for a different use case.

Comment (ranweiler) [reply src/agent/coverage/src/block.rs:189]: To be clear on my ambient goal: make it trivial to get modoffset (with hex offsets) from the JSON format.

We shouldn't ever run into `Number.MAX_SAFE_INTEGER` in practice with these _offsets_. Arguably we should require that they be `u32` values in the first place. I'd feel more strongly about it if we intended to represent arbitrary (including possibly invalid) 64-bit addresses.

Comment (ranweiler) [reply src/agent/coverage/src/block.rs:189]: Per discussion offline: for this format, we'll switch to serializing the offsets as `Number` values, since we control their generation, and they are _really_ `u32` values. Will change and add comments to that effect.

File: src/agent/coverage/src/block.rs [hunk 4]
```diff
@@ -280,4 +343,61 @@ mod tests {
...
+            cov.increment(&main_exe, 300);
+            cov.insert(&some_dll, vec![2, 30, 400].into_iter());
+            cov.increment(&some_dll, 30);
+            cov
+        };
+
+        let ser = serde_json::to_string(&cov).unwrap();
+
+        #[cfg(target_os = "linux")]
+        let text = r#"{"/lib/some.dll":[{"offset":2,"count":0},{"offset":30,"count":1},{"offset":400,"count":0}],"/main.exe":[{"offset":1,"count":1},{"offset":20,"count":0},{"offset":300,"count":1}]}"#;
```
lzybkr, line 393: Serializing full property names feels unnecessarily verbose for a format that is essentially meant only for machines.

I think we'll see on the order of thousands of such instances, so the impact is neither trivial nor major, so I don't have a strong opinion on using something shorter.

src/agent/coverage/src/block.rs
```diff
@@ -98,6 +101,13 @@ impl ModuleCov {
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct BlockCov {
     /// Offset of the block, relative to the module base load address.
+    //
+    // These offsets are represented as `u64` values, but since they are offsets
+    // within an assumed well-formed module, they should never exceed a `u32`.
+    // We thus assume that they can be losslessly serialized to an `f64`.
+    //
+    // If we need to handle malformed binaries or arbitrary addresses, then this
```
lzybkr, line 109: Maybe add an assertion in the constructor?

lzybkr (approved):

src/agent/coverage/src/block.rs
ranweiler: I'll be addressing this separately in a follow-up PR, with explicit failure if the assumption is violated.

bmc-msft (approved):

src/agent/coverage/src/block.rs
ranweiler: Yes, it could definitely be made even more terse. That said, there should only be one of these files per VM in the coverage task, and we will eventually support other output formats if needed. So I'll leave this for now.

PR #766, opened 2021-04-02
fix checking status codes for failure

bmc-msft:
This fixes retrying on bad status codes.

Note, this adds support to specify specific error codes as "successful".  This is important for the If-None-Match conditional upload case (blob uploading).  The response we get back is 409 (Conflict).

Previously, we would "fail fast" the 409.  However, what we want is "409 is basically success here" and every other call 409 should be a failure.

### DIFF
--- a/src/agent/reqwest-retry/src/lib.rs
+++ b/src/agent/reqwest-retry/src/lib.rs
@@ -19,12 +19,20 @@ pub async fn send_retry_reqwest_default<
 >(
     build_request: F,
 ) -> Result<Response> {
-    send_retry_reqwest(build_request, [], DEFAULT_RETRY_PERIOD, MAX_RETRY_ATTEMPTS).await
+    send_retry_reqwest(
+        build_request,
+        [],
+        [],
+        DEFAULT_RETRY_PERIOD,
+        MAX_RETRY_ATTEMPTS,
+    )
+    .await
 }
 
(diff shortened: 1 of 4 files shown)


chkeita (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -19,12 +19,20 @@ pub async fn send_retry_reqwest_default<
...
+        DEFAULT_RETRY_PERIOD,
+        MAX_RETRY_ATTEMPTS,
+    )
+    .await
 }
 
 pub async fn send_retry_reqwest<F: Fn() -> Result<reqwest::RequestBuilder> + Send + Sync>(
     build_request: F,
     fail_fast_status: impl AsRef<[StatusCode]>,
+    extra_success_status: impl AsRef<[StatusCode]>,
```
chkeita, line 35: instead of adding more and more paramter for each expcetion why not jus bring back the error_mapper that we [had before](https://github.com/microsoft/onefuzz/pull/623/files#diff-f3dc2a81b9ec9e427ffda01c77947d7fc06990acf0e3b2d5d460c22e8d874d74R92) 
This will make the behavior easier to configure

bmc-msft: the `error_mapper` method was error prone, though I understand your concern. 

Now, if you want to do something beyond the default, you can provide a function that allows for arbitrary configuration of retry based on status codes.

chkeita (approved):

PR #769, opened 2021-04-05
rename telemetry key names at the service level

bmc-msft:
In #766, the telemetry keys were renamed to reduce confusion.  However, this broke the telemetry keys within the agent due to a known issue using flattened json with aliases (See https://github.com/serde-rs/serde/issues/1976).

This PR removes the aliases and fixes the agent level telemetry & logging.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/proxy.py
+++ b/src/api-service/__app__/onefuzzlib/proxy.py
@@ -222,8 +222,8 @@ class Proxy(ORMMixin):
             ),
             forwards=forwards,
             region=self.region,
-            instrumentation_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
-            telemetry_key=os.environ.get("ONEFUZZ_TELEMETRY"),
+            instance_telemetry_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
+            microsoft_telemetry_key=os.environ.get("ONEFUZZ_TELEMETRY"),
             instance_id=get_instance_id(),
         )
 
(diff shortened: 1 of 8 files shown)


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -222,8 +222,8 @@ def save_proxy_config(self) -> None:
             ),
             forwards=forwards,
             region=self.region,
-            instrumentation_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
-            telemetry_key=os.environ.get("ONEFUZZ_TELEMETRY"),
+            instance_telemetry_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
+            microsoft_telemetry_key=os.environ.get("ONEFUZZ_TELEMETRY"),
```
ranweiler, line 226: In the future, we should consider renaming this var to include the string `MICROSOFT`.

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -222,8 +222,8 @@ def save_proxy_config(self) -> None:
             ),
             forwards=forwards,
             region=self.region,
-            instrumentation_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
-            telemetry_key=os.environ.get("ONEFUZZ_TELEMETRY"),
+            instance_telemetry_key=os.environ.get("APPINSIGHTS_INSTRUMENTATIONKEY"),
```
ranweiler, line 225: Right now, whenever we grab these key from the environment, we need to remember and preserve the mapping `instance <-> instrumentation`, `microsoft <-> telemetry`.

I suggest renaming the env vars elsewhere, but in the meantime, maybe we should define wrappers to `os.environ.get`, e.g.:
```python
def get_microsoft_telemetry_key() -> str:
    return os.environ.get("ONEFUZZ_TELEMETRY")
```
We could also just exports constants.

ranweiler (approved):

PR #774, opened 2021-04-06
allow deployment to non-default subscriptions

bmc-msft:
As is, 1f deployments deploy to the current az-cli profile subscription.  It would be useful to be able to deploy to a user specified subscription rather than the current one.

### DIFF
--- a/src/deployment/registration.py
+++ b/src/deployment/registration.py
@@ -80,9 +80,10 @@ def query_microsoft_graph(
         )
 
 
-def get_graph_client() -> GraphRbacManagementClient:
+def get_graph_client(subscription_id: str) -> GraphRbacManagementClient:
     client: GraphRbacManagementClient = get_client_from_cli_profile(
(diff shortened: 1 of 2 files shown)


src/deployment/registration.py
```diff
@@ -221,7 +224,9 @@ def create_application_registration(
     return registered_app
 
 
-def add_application_password(app_object_id: UUID) -> Tuple[str, str]:
+def add_application_password(
+    app_object_id: UUID, subscription_id: str
```
ranweiler, line 228: Can the `subscription_id` be narrowed to a `UUID`, too?

ranweiler (approved):

src/deployment/registration.py
bmc-msft: Not without covering the code with casts back and forth.  All of the underlying Azure APIs expect it as a string.

PR #779, opened 2021-04-08
Use dbghelp APIs to find PDB for PE when disassembling

### LINKED ISSUE #777: PDB paths are not searched by PE module processor
When we process PE modules for disassembly in `coverage::pe::process_modules()`, we _only_ try to resolve the PDB using the CV `filename` attr.

Instead, we should ideally search in the same way dbghelp API would, as seen [here](https://docs.microsoft.com/en-us/windows/win32/debug/symbol-paths#example-search-path). In particular, we must make sure that we include the directory containing the PE.

ranweiler:
Closes #777.


src/agent/debugger/src/dbghelp.rs
```diff
@@ -670,6 +674,94 @@ impl DebugHelpGuard {
...
+        // We zero-initialize `found_file_data`, and assume that `SymFindFileInPathW`
+        // only succeeds if it wrote a NUL-terminated wide string.
+        let found_file =
+            unsafe { win_util::string::os_string_from_wide_ptr(found_file_data.as_ptr()) };
+
+        Ok(found_file.into())
+    }
+
+    pub fn sym_get_search_path(&self, process_handle: HANDLE) -> Result<OsString> {
+        let mut search_path_data = Vec::<u16>::with_capacity(MAX_PATH * 8);
```
ranweiler, line 734: This was arbitrary; any ideas for a more motivated choice?

Will move this value to a constant and statically ensure it fits in a `u32`.

src/agent/debugger/src/dbghelp.rs
```diff
@@ -670,6 +674,94 @@ impl DebugHelpGuard {
...
+        &self,
+        process_handle: HANDLE,
+        file_name: impl AsRef<Path>,
+        pdb_signature: u32,
+        pdb_age: u32,
+    ) -> Result<PathBuf> {
+        let file_name = win_util::string::to_wstring(file_name);
+
+        // Must be at least `MAX_PATH` characters in length.
+        let mut found_file_data = Box::new([0u16; MAX_PATH]);
```
ranweiler, line 694: Eh, might as well just use a `Vec` to allocate this, like we do below.

src/agent/coverage/src/pe.rs
```diff
@@ -289,6 +280,78 @@ pub fn process_module(
     anyhow::bail!("PE missing codeview pdb debug info")
 }
 
+fn find_pdb_path(pe_path: &Path, cv: &CodeviewPDB70DebugInfo) -> Result<PathBuf> {
+    let cv_filename = CStr::from_bytes_with_nul(cv.filename)?.to_str()?;
```
ranweiler, line 284: Add `Context` calls throughout.

src/agent/coverage/src/pe.rs
```diff
@@ -239,21 +240,11 @@ pub fn process_module(
         codeview_pdb70_debug_info: Some(cv),
     }) = pe.debug_data
     {
-        let pdb_path = CStr::from_bytes_with_nul(cv.filename)?.to_str()?;
-        let pdb_file = File::open(pdb_path)?;
+        let pdb_path = find_pdb_path(pe_path.as_ref(), &cv)?;
```
ranweiler, line 243: Add `Context` calls throughout, as well as more logging.

src/agent/coverage/src/pe.rs
```diff
@@ -289,6 +280,78 @@ pub fn process_module(
     anyhow::bail!("PE missing codeview pdb debug info")
 }
 
+fn find_pdb_path(pe_path: &Path, cv: &CodeviewPDB70DebugInfo) -> Result<PathBuf> {
+    let cv_filename = CStr::from_bytes_with_nul(cv.filename)?.to_str()?;
+
+    // This field is named `filename`, but it may be an absolute path.
+    // The callee `sym_find_pdb_file_in_path()` handles either.
+    let cv_filename = Path::new(cv_filename);
```
ranweiler, line 288: We should try this path directly first, before invoking dbghelp.

src/agent/coverage/src/pe.rs
```diff
@@ -289,6 +280,78 @@ pub fn process_module(
...
+    // The docs warn that we must not intialize the dbghelp symbol handler
+    // with the current process handle when debugging another process. When
+    // we invoke this function, we will often be doing exactly this.
+    //
+    // Assume this is only relevant when making queries about some debuggee's
+    // address space, which we will not do here.
+    let handle = win_util::process::current_process_handle();
+
+    let dbghelp = debugger::dbghelp::lock()?;
+    dbghelp.sym_initialize(handle)?;
```
lzybkr, line 301: Is this necessary? I can see this if we want to build the cache when not debugging, but I think this is only called in a debugging context where symbols are already loaded using the target process handle.

Note that the value passed can be anything, it need not be a process handle, and in fact, it would probably be better to use something else because the warning in the docs getting unexpected results.

src/agent/coverage/src/pe.rs
```diff
@@ -289,6 +280,78 @@ pub fn process_module(
...
+
+    // Enable signature and age checking.
+    let options = dbghelp.sym_get_options();
+    dbghelp.sym_set_options(options | SYMOPT_EXACT_SYMBOLS);
+
+    let mut search_path = dbghelp.sym_get_search_path(handle)?;
+
+    log::debug!("initial search path = {:?}", search_path);
+
+    // Try to add the directory of the PE to the PDB search path.
```
lzybkr, line 312: Is this really necessary? I'm thinking onefuzz should be setting the symbol path for other reasons, e.g. the input tester needs the correct symbol path to produce good call stacks, and also if the user remotes in to debug a crash, the symbol path should be set correctly for them already.

For the "local" scenario where we run coverage on the same machine the target is built on, the full path in the PE should be all we need.

src/agent/debugger/src/dbghelp.rs
```diff
@@ -670,6 +674,94 @@ impl DebugHelpGuard {
...
+        // to `NULL`, the default search path is used.
+        let search_path = std::ptr::null_mut();
+
+        // See: https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfindfileinpathw#remarks
+        let id = pdb_signature as *mut _;
+        let two = pdb_age;
+        let three = 0;
+
+        // SSRVOPT_DWORD
+        let flags = 0x0002;
```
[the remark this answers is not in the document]
ranweiler: Those constants are actually upstream now (thanks to you!), so I removed those and added a new one.

src/agent/coverage/src/pe.rs
ranweiler: >  I'm thinking onefuzz should be setting the symbol path for other reasons

Agreed. The goal here was to make this code path more robust in various contexts. This is inessential and we could remove it.

> [...] the full path in the PE should be all we need.

Maybe. I found that this was painfully brittle in practice. In local fuzzing, I would actually expect that the user typically would NOT build in the same directory the target would be invoked from. But again, OneFuzz should really be setting `_NT_SYMBOL_PATH` to include the setup dir, which will handle almost all cases. However, the instant a user nests a target module in the setup dir, it will then break coverage.

The original inspiration was actually item (3) [here](https://docs.microsoft.com/en-us/windows/win32/debug/symbol-paths#specifying-search-paths), which appears to actually be _false_, and in need of correction. The actual behavior is described by the docs for the [`UserSearchPath` parameter](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitializew#parameters) of `SymInitializeW`. That said, IMO, the incorrectly-described behavior is more intuitive.

src/agent/coverage/src/pe.rs
ranweiler: So, at one point I actually threaded an optional `handle` parameter through so we could reuse an existing handle-addressed symbol resolver context, and only fell back on this. But, it ended up touching a lot of places, so I shelved it, thinking that it'd be better to revisit more holistically.

After we address everything else, I'll push a commit for you to take a peek at.

ranweiler: Take a look at: https://github.com/microsoft/onefuzz/pull/779/commits/98670cfbfbcf88b672f008796dae5ddae7db885c

As discussed, ideally we'll eventually migrate the handle parameter to something like a `DbghelpContext`.

lzybkr (approved):

PR #786, opened 2021-04-09
combine execs/sec and iteration counts from multiple workers

bmc-msft:
As is, the execs/sec and iteration counts require significant work from the user to understand how much work their fuzzer has done and how many tests per second they are seeing.  This tries to simplify that data from a given runner's perspective.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
+++ b/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
@@ -349,7 +349,7 @@ impl TotalStats {
         if let Some(current) = self.worker_stats.get(&worker_data.worker_id) {
             // if it's the same run, only add the differences
             if current.run_id == worker_data.run_id {
-                self.count += worker_data.count - current.count;
+                self.count += worker_data.count.saturating_sub(current.count);
             } else {
                 self.count += worker_data.count;
             }
--- a/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
+++ b/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
@@ -22,7 +22,7 @@ use tokio::{
     io::{AsyncBufReadExt, BufReader},
     sync::mpsc,
     task,
-    time::{self, Duration},
+    time::{delay_for, Duration},
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


bmc-msft:
Note: while this isn't specifically a breaking change, it will change how stats are reported in a meaningful fashion.

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -370,35 +392,98 @@ impl Timer {
...
-        std::iter::repeat(None).take(workers).collect();
+    let mut total = TotalStats::default();
+
+    // report all zeros to start
+    total.report();
 
     let timer = Timer::new(RUNTIME_STATS_PERIOD);
 
     loop {
         tokio::select! {
```
chkeita, line 410: Not a blocking comment but i noticed that this pattern seems a bit slow when dequeuing a lot of messages. 
i have been using try_recv instead for polling at a regular interval
``` Rust
match stats_channel.try_recv() {
        Ok(stats) => {
            heartbeat_client.alive();
            total.update(stats);
            total.report();
        }
        Err(mpsc::error::TryRecvError::Empty) => {
           total.report();
           delay_for(RUNTIME_STATS_PERIOD).await,
        }
        Err(mpsc::error::TryRecvError::Closed) => break,
    }
```
I don't think this would b a big deal for this PR though

chkeita (approved):

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -323,26 +323,48 @@ async fn report_fuzzer_sys_info(worker_id: usize, run_id: Uuid, fuzzer_pid: u32)
...
+    count: u64,
+    execs_sec: f64,
+}
+
+impl TotalStats {
+    fn update(&mut self, worker_data: RuntimeStats) {
+        if let Some(current) = self.worker_stats.get(&worker_data.worker_id) {
+            // if it's the same run, only add the differences
+            if current.run_id == worker_data.run_id {
+                self.count += worker_data.count - current.count;
```
ranweiler, line 352: Just to be cautious:
```suggestion
                self.count += worker_data.count.saturating_sub(current.count);
```

ranweiler (approved):

PR #790, opened 2021-04-10
Refactor debugger breakpoint handling

lzybkr:
There is a bunch of refactoring in the PR but also a couple of important fixes.

Breakpoints are now cleared after `ContinueDebugEvent` returns and restored just before calling `ContinueDebugEvent` again. This ensures that debugger clients can read process memory and not see the breakpoints.

Additionally, the original code byte for a breakpoint is reread to correctly handle the rare case of self modifying code.

Note that with this change, debug event processing is now significantly slower than before this PR. For infrequent events it might not be a big deal, but single stepping performance is much worse.

I may look at adding a way to simulate process reads to account for our breakpoints to improve performance, but I think this refactoring is worth taking as is because it is more correct and also I think cleans up breakpoints in general.


src/agent/debugger/src/target.rs
```diff
@@ -272,24 +207,159 @@ impl Target {
...
+        Ok(())
+    }
+
+    pub fn initial_wow64_bp(&mut self) {
+        self.saw_initial_wow64_bp = true;
+    }
+
+    fn try_resolve_all_unresolved_breakpoints(&mut self) {
+        // borrowck - take ownership from self so we call `try_resolve_unresolved_breakpoint`.
+        let mut unresolved_breakpoints = std::mem::take(&mut self.unresolved_breakpoints);
```
ranweiler, line 230: Noting that we're allocating a new (empty) vec each time we do this, and could avoid that if we used `Option::take()`. But we only do this once per breakpoint, so it is no doubt lost in the perf noise.

src/agent/debugger/src/target.rs
```diff
@@ -272,24 +207,159 @@ impl Target {
...
+                                    debug!(
+                                        "unknown symbol {}!{}",
+                                        module.name().display(),
+                                        sym_info.function()
+                                    );
+                                }
+                            }
+                        }
+                        Err(e) => {
+                            error!("Can't set symbolic breakpoints: {}", e);
```
ranweiler, line 280:
```suggestion
                            error!("Can't set symbolic breakpoints: {:?}", e);
```
To include causes (and someday, backtraces when available).

src/agent/debugger/src/target.rs
```diff
@@ -272,24 +207,159 @@ impl Target {
...
+        let dbghelp = dbghelp::lock().expect("can't lock dbghelp to find module");
+        dbghelp.get_module_base(self.process_handle, address)
     }
 
-    pub fn breakpoint_set_at_addr(&self, address: u64) -> bool {
-        self.breakpoints.contains_key(&address)
+    fn module_from_address(&mut self, address: u64) -> &mut Module {
+        let module_base = self
+            .module_base_from_address(address)
+            .unwrap_or(unsafe { NonZeroU64::new_unchecked(module::UNKNOWN_MODULE_BASE_ADDRESS) });
```
ranweiler, line 341: How about avoiding `unsafe` and doing:
```suggestion
            .unwrap_or(NonZeroU64::new(module::UNKNOWN_MODULE_BASE_ADDRESS).unwrap());
```
Alternately, it looks like `new_unchecked()` is a `const fn`, so maybe we could just make the exported constant a `NonZeroU64`.

src/agent/debugger/src/target.rs
```diff
@@ -300,30 +370,43 @@ impl Target {
...
             }
 
-            self.sym_initialized = true;
+            return Err(e);
+        }
+
+        for (_, module) in self.modules.iter_mut() {
+            if let Err(e) = module.sym_load_module(self.process_handle) {
+                error!(
+                    "Error loading symbols for module {}: {}",
```
ranweiler, line 405:
```suggestion
                    "Error loading symbols for module {}: {:?}",
```

src/agent/debugger/src/target.rs
```diff
@@ -504,12 +468,17 @@ impl Target {
...
         // To avoid these possible races, when resuming, we only let a single thread go **if**
         // we're single stepping any threads.
-        //
-        // First, if we last stepped because of hitting a breakpoint, we restore the breakpoint
-        // so that whichever thread is resumed, it can't miss the breakpoint.
-        if let Some(CodeByteToUpdate { address, byte }) = self.code_byte_to_update.take() {
-            trace!("Updating breakpoint at 0x{:x}", address);
-            write_instruction_byte(self.process_handle, address, byte)?;
+
+        // We now re-enable the breakpoint so that the next time we step, the breakpoint
```
ranweiler, line 472: 👍

src/agent/debugger/src/breakpoint.rs
```diff
@@ -0,0 +1,298 @@
...
+pub(crate) enum ExtraInfo {
+    Rva(RvaExtraInfo),
+    UnresolvedSymbol(SymbolExtraInfo),
+    Resolved(AddressExtraInfo),
+}
+
+pub struct Breakpoint {
+    id: BreakpointId,
+    kind: BreakpointType,
+    disabled: u32,
```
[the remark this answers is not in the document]
ranweiler: I see now. Some docs here would be useful.

src/agent/debugger/src/target.rs
lzybkr: We have too many `unwrap` calls already and I started replacing them with `expect` or adding proper error handling.

Unfortunately `const fn` doesn't alleviate the need to use `unsafe`, and you can't use `unwrap` or `expect` on a `const`.

src/agent/debugger/src/target.rs
lzybkr: Thanks for the tip, I've been using nightly anyway and it is handy getting the backtraces.

src/agent/debugger/src/target.rs
lzybkr: An empty Vec does not allocate, and using an option requires additional checks when pushing, so I'm going to leave this as is.

src/agent/debugger/src/breakpoint.rs
```diff
@@ -0,0 +1,326 @@
...
+        )?;
+        Ok(())
+    }
+}
+
+fn write_instruction_byte(process_handle: HANDLE, ip: u64, b: u8) -> Result<()> {
+    let orig_byte = [b; 1];
+    let remote_address = ip as _;
+    process::write_memory_slice(process_handle, remote_address, &orig_byte)?;
+    process::flush_instruction_cache(process_handle, remote_address, orig_byte.len())?;
```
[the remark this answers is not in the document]
lzybkr: I'm guessing it doesn't matter as we won't hit this code too often for a collection of breakpoints, and we are/will use this function for non-collections, so leaving as is.

src/agent/debugger/src/breakpoint.rs
```diff
@@ -0,0 +1,326 @@
...
+    breakpoints: fnv::FnvHashMap<u64, Breakpoint>,
+    min_breakpoint_addr: u64,
+    max_breakpoint_addr: u64,
+}
+
+impl BreakpointCollection {
+    pub fn new() -> Self {
+        BreakpointCollection {
+            breakpoints: fnv::FnvHashMap::default(),
+            min_breakpoint_addr: u64::MAX,
```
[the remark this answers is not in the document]
lzybkr: I'm going to punt on this as it does look more painful than care for right now.

Note that I'm switching to a `BTreeMap` and eventually one could replace these fields with `first_key` and `last_key` and then one would naturally have an `Option` to deal with.

src/agent/debugger/src/target.rs
ranweiler: Ah, right! This is def better, then.

src/agent/debugger/src/target.rs
ranweiler: Ah, darn. Maybe someday soon: https://github.com/rust-lang/rust/issues/51999#issuecomment-818925156

src/agent/debugger/src/target.rs
```diff
@@ -567,8 +565,26 @@ impl Target {
...
+        // We don't remove breakpoints when processing an event, so it's possible that the
+        // memory we read contains **our** breakpoints instead of the original code.
+        let remote_address = remote_address as u64;
+        let module = self.module_from_address(remote_address);
+        let range = remote_address..=(remote_address + buf.len() as u64);
+
+        let u8_buf = unsafe {
+            std::slice::from_raw_parts_mut(buf.as_mut_ptr() as *mut u8, std::mem::size_of_val(buf))
+        };
+        for (address, breakpoint) in module.breakpoints_for_range(range) {
```
ranweiler, line 580: 👍

src/agent/debugger/src/target.rs
```diff
@@ -567,8 +565,26 @@ impl Target {
         Ok(current_context.get_flags())
     }
 
-    pub fn read_memory(&self, remote_address: LPCVOID, buf: &mut [impl Copy]) -> Result<()> {
-        process::read_memory_array(self.process_handle, remote_address, buf)
+    pub fn read_memory(&mut self, remote_address: LPCVOID, buf: &mut [impl Copy]) -> Result<()> {
```
ranweiler, line 568: WDYT about making this method itself `unsafe`?

src/agent/debugger/src/target.rs
```diff
@@ -567,8 +565,26 @@ impl Target {
...
+        process::read_memory_array(self.process_handle, remote_address, buf)?;
+
+        // We don't remove breakpoints when processing an event, so it's possible that the
+        // memory we read contains **our** breakpoints instead of the original code.
+        let remote_address = remote_address as u64;
+        let module = self.module_from_address(remote_address);
+        let range = remote_address..=(remote_address + buf.len() as u64);
```
ranweiler, line 575: Do we want a _non_-inclusive upper bound? I think this can panic if we have a breakpoint set at `remote_address + buf.len()`. If `buf` is already a `&[u8]`, then its max index will be `buf.len() - 1`, but in the loop, we would index into it at `(remote_address + buf.len()) - remote_address == buf.len()`.

lzybkr: Yep, good catch.

src/agent/debugger/src/target.rs
lzybkr: I'm not sure that would add value as it's unclear how far up the callstack to stop adding `unsafe` and I think this function is easily reviewed for safety concerns.

ranweiler: My concern is that, among other things, the current API would let us create an invalid reference in safe Rust, for example, if `buf` had type `&mut [&[T]]`.

I think `unsafe` for this method is the right place to have it: it's on the caller to ensure that they are safely transmuting the underlying remote bytes to a valid Rust type.

lzybkr: Opened https://github.com/microsoft/onefuzz/issues/815 to track.

ranweiler (approved):

PR #792, opened 2021-04-12
invert expect_crash_on_failure config option defaults

bmc-msft:
This matches the existing job template behavior


ranweiler (approved):

src/agent/onefuzz-agent/src/local/libfuzzer_fuzz.rs
```diff
@@ -94,9 +94,9 @@ pub fn build_shared_args() -> Vec<Arg<'static, 'static>> {
         Arg::with_name(CHECK_FUZZER_HELP)
             .takes_value(false)
             .long(CHECK_FUZZER_HELP),
-        Arg::with_name(DISABLE_EXPECT_CRASH_ON_FAILURE)
+        Arg::with_name(EXPECT_CRASH_ON_FAILURE)
             .takes_value(false)
```
ranweiler, line 98: Our now naming now agrees with the non-local `libfuzzer-fuzz` task, but there the `expect_crash_on_failure` defaults defaults to `true`. Do we want to do that here, too, to invert the name, but keep the semantics?

ranweiler: Ref: https://github.com/microsoft/onefuzz/blob/458a142232a288950793d52426403d9ef5cebe13/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs#L64-L65

bmc-msft: good point.  I added an additional helper for the default value to be false.

src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -61,7 +65,7 @@ pub struct Config {
     #[serde(default = "default_bool_true")]
     pub check_fuzzer_help: bool,
 
-    #[serde(default = "default_bool_true")]
+    #[serde(default = "default_bool_false")]
```
ranweiler, line 68: If you like, this could be:
```suggestion
    #[serde(default)]
```

ranweiler (approved):

Issue #758, opened 2021-03-31
regression_reported event should include the task config


bmc-msft:
See #757

PR #793, opened 2021-04-12
add TaskConfig to crash_reported and regression_reported events

### LINKED ISSUE #757: crash_reported event should include the task config
Right now the crash_reported event includes job id and task id, it needs the task config too

bmc-msft:
resolves #757 and #758


src/api-service/__app__/onefuzzlib/notifications/main.py
```diff
@@ -97,6 +97,32 @@ def get_notifications(container: Container) -> List[Notification]:
...
+            return Task.get(
+                report.crash_test_result.crash_report.job_id,
+                report.crash_test_result.crash_report.task_id,
+            )
+        if report.crash_test_result.no_repro:
+            return Task.get(
+                report.crash_test_result.no_repro.job_id,
+                report.crash_test_result.no_repro.task_id,
+            )
+    elif report.original_crash_test_result:
```
ranweiler, line 112: Since `crash_test_result` is a required field, but `original_` is `Optional`, won't this be unreachable?

bmc-msft: Good catch.

src/api-service/__app__/onefuzzlib/notifications/main.py
```diff
@@ -97,6 +97,34 @@ def get_notifications(container: Container) -> List[Notification]:
...
+        return Task.get(
+            report.crash_test_result.no_repro.job_id,
+            report.crash_test_result.no_repro.task_id,
+        )
+
+    # original_crash_test_result is not required, nor are report & no_repro.
+    if report.original_crash_test_result:
+        if report.original_crash_test_result.crash_report:
+            return Task.get(
+                report.original_crash_test_result.crash_report.job_id,
```
ranweiler, line 117: One thing I don't follow here: I would expect this to be the `job_id`/`task_id` for the task that _originally_ found crash that has now regressed. That means here we'd be returning a totally different (past) task, compared to the other cases. Am I mistaken?

bmc-msft: By using the current `job_id` and `task_id`, we provide the metadata related to the task that added the regression.  This includes the setup containers and tags related to the regression.

ranweiler: My expectation: we want to use the _current_ job ID/task ID.

My reading of the code in this branch: in this branch, but not the other branches we are extracting and returning the _old_ job ID/task ID, which is not what we want.

bmc-msft: I believe I understand your concern.

In the case the new regression report is missing both a crash report and a norepo, which shouldn't happen but isn't protected against, this will fall back to identifying the previous task.

I think it's a toss up, but I leaned toward providing the old data rather than no data, given the context.

ranweiler (approved):

PR #794, opened 2021-04-12
Allow the local run to work without copying the synced dirs

chkeita:
Added an option to local run to not create a job directory


src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -6,7 +6,6 @@ use clap::{App, Arg, ArgMatches};
 use onefuzz::jitter::delay_with_jitter;
 use onefuzz::{blob::url::BlobContainerUrl, monitor::DirectoryMonitor, syncdir::SyncedDir};
 use path_absolutize::Absolutize;
-use remove_dir_all::remove_dir_all;
```
bmc-msft, line 9: This crate can be removed now, since this was the only place it was used.

bmc-msft (approved):

PR #796, opened 2021-04-12
add Timestamp to multiple models

bmc-msft:
Expose the Azure storage table's "Timestamp" for the models where Timestamp should be user-accessible and remove the Timestamp field from models that did not sign up for it.

The behavior where Timestamp is only set by Azure Storage is kept.


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -42,6 +42,7 @@
 # This isn't intended to ever be shared to the client, hence not being in
 # onefuzztypes
 class Proxy(ORMMixin):
+    Timestamp: Optional[datetime.datetime]
```
ranweiler, line 45: Is there a straightforward way to avoid using a different casing convention in the Python code?

ranweiler (approved):

ranweiler (approved):

PR #798, opened 2021-04-13
add basic context to every retried request

bmc-msft:
In addition, this PR reduces the retry (but not failure) log to `debug`.


src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -255,12 +253,18 @@ impl Coordinator {
 
             // And try one more time.
             let request = self.get_request_builder(request_type);
-            response = request.send_retry_default().await?;
+            response = request
+                .send_retry_default()
+                .await
+                .context("Coordinator.send_with_auth_retry after refreshing access token")?;
```
ranweiler, line 259: We renamed this method, so we need to grep for stale context strings.

ranweiler (approved):

PR #800, opened 2021-04-13
Upgrade to tokio 1.5.0

chkeita:
Upgrade to tokio 1.5.0
replaced tokio::sync::mpsc:: with [flume ](https://crates.io/crates/flume) because try_recv was removed

### DIFF
--- a/src/agent/onefuzz-agent/src/local/generic_analysis.rs
+++ b/src/agent/onefuzz-agent/src/local/generic_analysis.rs
@@ -15,13 +15,14 @@ use crate::{
 };
 use anyhow::Result;
 use clap::{App, Arg, SubCommand};
+use flume::Sender;
 use storage_queue::QueueClient;
 
 pub fn build_analysis_config(
     args: &clap::ArgMatches<'_>,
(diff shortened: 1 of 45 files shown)


bmc-msft (approved):

src/agent/onefuzz-agent/src/local/generic_analysis.rs
```diff
@@ -16,13 +16,12 @@ use crate::{
...
 use clap::{App, Arg, SubCommand};
 use storage_queue::QueueClient;
-use tokio::sync::mpsc::UnboundedSender;
 
 pub fn build_analysis_config(
     args: &clap::ArgMatches<'_>,
     input_queue: Option<QueueClient>,
     common: CommonConfig,
-    event_sender: Option<UnboundedSender<UiEvent>>,
+    event_sender: Option<flume::Sender<UiEvent>>,
```
ranweiler, line 24: Only suggestion: in all the places where we use `flume::Sender`, just `use flume::Sender` at the top of the file so we don't have to qualify the name.

ranweiler (approved):

PR #810, opened 2021-04-15
Simplify handling of OS-sensitive test cases

ranweiler:
- Add `module!` test helper to let us always define tests paths using POSIX-style paths
- Use `serde_json::json!` macro for maintainable definitions of expected test results

### DIFF
--- a/src/agent/coverage/src/block.rs
+++ b/src/agent/coverage/src/block.rs
@@ -256,26 +256,24 @@ mod tests {
 
     // Given a POSIX-style path as a string, construct a valid absolute path for
     // the target OS and return it as a checked `ModulePath`.
-    macro_rules! module {
-        ($posix_path: expr) => {{
-            let mut p = std::path::PathBuf::default();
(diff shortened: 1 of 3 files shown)


src/agent/coverage/src/block.rs
```diff
@@ -251,37 +254,45 @@ mod tests {
         );
     }
 
-    fn module_path(path: &str) -> ModulePath {
-        ModulePath::new(std::path::PathBuf::from(path)).unwrap()
+    // Given a POSIX-style path as a string, construct a valid absolute path for
+    // the target OS and return it as a checked `ModulePath`.
+    macro_rules! module {
```
chkeita, line 259: is there an advantage of using a module instead of a function ?

ranweiler: Not anymore (other than the use site standing out a bit more). I was originally going to make it accept an input spec in a different format, then ended up going with this simpler / more function-friendly impl.

bmc-msft (approved):

chkeita (approved):

PR #811, opened 2021-04-15
Use `u32` to represent module offsets

ranweiler:
In #765, we introduced an assumption that all module-relative instruction offsets are representable by a `u32`. Move to an explicit `u32` repr for offsets, and add validation to check for overflow at conversion sites.

### DIFF
--- a/src/agent/coverage/src/block/linux.rs
+++ b/src/agent/coverage/src/block/linux.rs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 use std::collections::BTreeMap;
+use std::convert::TryInto;
 use std::ffi::OsStr;
 use std::process::Command;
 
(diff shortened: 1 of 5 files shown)


src/agent/coverage/src/block/linux.rs
```diff
@@ -307,12 +316,16 @@ impl ModuleImage {
         (self.map.address.0)..(self.map.address.1)
     }
 
-    pub fn va_to_offset(&self, va: u64) -> u64 {
-        va - self.base()
+    pub fn va_to_offset(&self, va: u64) -> Result<u32> {
+        if let Some(offset) = va.checked_sub(self.base()) {
+            Ok(offset.try_into()?)
```
lzybkr, line 321: Add context here?

lzybkr (approved):

PR #812, opened 2021-04-15
Add computation of block coverage statistics

ranweiler:


src/agent/coverage/src/block.rs
```diff
@@ -57,12 +57,39 @@ impl CommandBlockCov {
         self.modules.iter()
     }
 
+    /// Total count of covered blocks across all modules.
+    pub fn covered(&self) -> u64 {
+        self.modules.values().map(|m| m.covered()).sum()
+    }
+
+    /// Total count of known blocks across all modules.
+    pub fn features(&self) -> u64 {
```
lzybkr, line 66: I think `known_blocks` would be a better name.

src/agent/coverage/src/block.rs
```diff
@@ -78,6 +105,36 @@ impl ModuleCov {
         Self { blocks }
     }
 
+    /// Total count of blocks that have been reached (have a nonzero count).
+    pub fn covered(&self) -> u64 {
+        self.blocks
+            .values()
+            .map(|b| u64::min(1, b.count as u64))
```
lzybkr, line 112: It took me a second to understand. I wonder if this would be clearer:
```rust
    .filter(|b| b.count > 0)
    .count()
```

src/agent/coverage/src/block.rs
```diff
@@ -78,6 +105,36 @@ impl ModuleCov {
...
+            .sum()
+    }
+
+    /// Total count of known blocks.
+    pub fn features(&self) -> u64 {
+        self.blocks.len() as u64
+    }
+
+    /// Total count of blocks covered by `self` but not `other`.
```
lzybkr, line 121: This is implied by your comment but maybe worth noting anyway - a difference of `0` does not imply identical coverage.

lzybkr (approved):

src/agent/coverage/src/block.rs
ranweiler: Will switch. This more-abstracted name was chosen for consistency with OneFuzz, but I agree that `known_blocks` would be clearer (and this doesn't impact user-facing serialization, so we are free to diverge).

PR #833, opened 2021-04-26
move to Instant.elapsed() instead of time::Sleep timer

bmc-msft:
In current tokio, time::sleep().elapsed does not update unless the Sleep
is polled.  as such, the execute_pending_commands never fires.  This
replaces the sleep().elapsed with Instant.elapsed().

### DIFF
--- a/src/agent/onefuzz-supervisor/src/agent.rs
+++ b/src/agent/onefuzz-supervisor/src/agent.rs
@@ -13,6 +13,8 @@ use crate::setup::*;
 use crate::work::IWorkQueue;
 use crate::worker::IWorkerRunner;
 
+const PENDING_COMMANDS_DELAY: time::Duration = time::Duration::from_secs(10);
+
 pub struct Agent {
     coordinator: Box<dyn ICoordinator>,
     reboot: Box<dyn IReboot>,
@@ -50,7 +52,7 @@ impl Agent {
     }
 
     pub async fn run(&mut self) -> Result<()> {
-        let mut delay = command_delay();
+        let mut instant = time::Instant::now();
 
         // Tell the service that the agent has started.
         //
@@ -68,9 +70,9 @@ impl Agent {
 
         loop {
             self.heartbeat.alive();
-            if delay.is_elapsed() {
+            if instant.elapsed() >= PENDING_COMMANDS_DELAY {
                 self.execute_pending_commands().await?;
-                delay = command_delay();
+                instant = time::Instant::now();
             }
 
             let done = self.update().await?;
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-supervisor/src/agent.rs
```diff
@@ -68,9 +70,9 @@ impl Agent {
 
         loop {
             self.heartbeat.alive();
-            if delay.is_elapsed() {
+            if instant.elapsed() >= PENDING_COMMANDS_DELAY {
```
ranweiler, line 73: Maybe we should update this to `select!` in a loop, against two [`Interval`](https://docs.rs/tokio/1.5.0/tokio/time/struct.Interval.html) futures that explicitly control how often we attempt to execute pending commands or update the state machine. Something like:

```rust
let cmd_interval = time::interval(PENDING_COMMANDS_INTERVAL);
let update_interval = time::interval(UPDATE_INTERVAL);

loop {
    self.heartbeat.alive();

    tokio::select! {
        _ = cmd_interval.tick() => {
            self.execute_pending_commands().await?;            
        }
        _ = update_interval.tick() => {
            if self.update().await? {
                debug!("agent done, exiting loop");
                break;                                
            }
        }
    }
}
```

ranweiler: We'd probably want to update the `busy()` state impl too, so it checks if an `Interval` has ticked before polling the remote queue.

ranweiler: Also, we'd want to replace [this `sleep()`](https://github.com/microsoft/onefuzz/blob/cf3d9049406f46c8a798be38fe883d43103677a1/src/agent/onefuzz-supervisor/src/agent.rs#L187) with an `Interval`. The logic we'd want for the `ready` state is: "every `UPDATE_INTERVAL`, give the inner state handler an _opportunity_ to check for available work. The state handler will then consult its own timer, so there is at least `WORK_POLL_INTERVAL` seconds between calls to the remote work queue."

bmc-msft: It's underlying issue is made complex based on how the state machine is implemented. 

* Free needs to sleep because it's response is dependent on the server.
* Busy shouldn't sleep, but instead await on the output of the workers.
* All of the other states should proceed immediately without delays.

In my opinion, the state machine's delay logic belongs in each of the individual states and instead of a loop where we manage the transitions, I think we should move to a model akin to the [A Fist Full of States](https://deislabs.io/posts/a-fistful-of-states/) from Deis Labs.

bmc-msft: Either way, that's a large amount of work and should be a follow-on PR.

ranweiler (approved):

PR #834, opened 2021-04-26
when processing node updates, don't wait on the node in cases it should be stopped

bmc-msft:
In situations when the node should be done, mark it as done without
waiting for the node to respond to the Done command.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/workers/nodes.py
+++ b/src/api-service/__app__/onefuzzlib/workers/nodes.py
@@ -217,7 +217,7 @@ class Node(BASE_NODE, ORMMixin):
             "node: stopping busy node with all tasks complete: %s",
             self.machine_id,
         )
-        self.stop()
+        self.stop(done=True)
         return True
 
     def mark_tasks_stopped_early(self, error: Optional[Error] = None) -> None:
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -267,15 +267,15 @@ def can_process_new_work(self) -> bool:
                 "can_schedule is set to be deleted.  machine_id:%s",
                 self.machine_id,
             )
-            self.stop()
+            self.stop(done=True)
```
chkeita, line 270: Could we be introducing a racing condition here ? 
The node could have picked up a new item from the work queue before receiving the done command.
In that case we could be loosing some work.
same for the `reimage_requested` case below

bmc-msft: I don't think so.

`can_process_new_work` is called from the supervisor's endpoint verifying that work from the queue can be run.  In the places we're updating (checking `node.is_outdated`, `node.delete_requested`, and `node.reimage_requested`) are node specific responses.  All worksets checked for this node would fail until the node is reimaged.

chkeita (approved):

PR #836, opened 2021-04-26
Add node messages to node get

bmc-msft:
This exposes the node commands that have yet to be processed by the node.

Example use case:  The SDK can now ask "has this node installed my SSH key"

### DIFF
--- a/src/pytypes/onefuzztypes/models.py
+++ b/src/pytypes/onefuzztypes/models.py
@@ -557,6 +557,29 @@ class NodeHeartbeatEntry(BaseModel):
     data: List[Dict[str, HeartbeatType]]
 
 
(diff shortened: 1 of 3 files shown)


src/pytypes/onefuzztypes/models.py
```diff
@@ -557,13 +557,37 @@ class NodeHeartbeatEntry(BaseModel):
     data: List[Dict[str, HeartbeatType]]
 
 
+class StopNodeCommand(BaseModel):
```
bmc-msft, line 560: Note, these are just moved from later in models.py to enable us to not have to call `Node.update_forward_refs()`

chkeita (approved):

PR #838, opened 2021-04-28
Ignore inline frames when getting current frame

lzybkr:


src/agent/debugger/src/stack.rs
```diff
@@ -228,39 +228,45 @@ pub fn get_stack(
...
-                    &dbghlp,
-                    &module_info,
-                    process_handle,
-                    program_counter,
-                    frame.InlineFrameContext,
-                )
+    dbghlp.stackwalk_ex(
+        process_handle,
+        thread_handle,
+        true, /* visit inline frames */
```
lzybkr, line 234: This is the only change here - the rest is `rustfmt`.

ranweiler (approved):

PR #843, opened 2021-05-02
Move PDB functions into own module

ranweiler:
This does not implement any logic changes, but reorganizes code for easier re-use.

PR #842, opened 2021-05-01
Update `pdb` to 0.7

ranweiler:
Picks up some small bugfixes and panic removals.

PR #844, opened 2021-05-02
Try to use Sancov tables when recording Windows coverage

ranweiler:
Includes #842, #843, merge after.

- Add Sancov table discovery for Windows modules with debug info
- When recording coverage, try to recover coverage blocks from Sancov tables
- Not tested against x86-32 targets, but should be close

### DIFF
--- a/src/agent/coverage/src/block/pe_provider.rs
+++ b/src/agent/coverage/src/block/pe_provider.rs
@@ -29,27 +29,38 @@ where
         Self { data, pe, pdb }
     }
 
+    /// Try to provide basic block offsets using available Sancov table symbols.
+    ///
+    /// If PC tables are available and definitely well-formed, use those
+    /// directly. Otherwise, look for an inline counter or bool flag array, then
+    /// disassemble all functions to reverse the instrumentation sites.
     pub fn provide(&mut self) -> Result<BTreeSet<u32>> {
         let mut visitor = SancovDelimiterVisitor::new(self.pdb.address_map()?);
 
         let global_symbols = self.pdb.global_symbols()?;
         let mut iter = global_symbols.iter();
 
+        // Search symbols which delimit Sancov tables.
         while let Some(symbol) = iter.next()? {
             if let Ok(SymbolData::Data(data)) = symbol.parse() {
                 visitor.visit_data_symbol(&data)?;
             }
         }
 
+        // If we found a non-empty PC table, try to parse it.
         if let Some(pcs_table) = visitor.sancov_pcs_table() {
-            // Discovering and parsing the PC table can be error-prone, if we even have it. Recover
-            // it if we can, with some strict assumptions. If we can't, fall back on reversing the
-            // inline counter accesses.
+            // Discovering and parsing the PC table can be error-prone, if we even have it. Mine it
+            // for PCs if we can, with some strict assumptions. If we can't, fall back on reversing
+            // the inline table accesses.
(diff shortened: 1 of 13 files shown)


src/agent/coverage/src/block/pe_provider.rs
```diff
@@ -0,0 +1,620 @@
...
+
+use std::collections::BTreeSet;
+use std::convert::TryInto;
+
+use anyhow::{format_err, Result};
+use goblin::pe::PE;
+use pdb::{AddressMap, DataSymbol, FallibleIterator, Rva, ProcedureSymbol, Source, SymbolData, PDB};
+
+/// Basic block offset provider for uninstrumented PE modules.
+pub struct PeBasicBlockProvider {}
```
ranweiler, line 12: In a future PR, we will revisit and move the PE `process_` code into this type.

src/agent/coverage/src/block/pe_provider.rs
```diff
@@ -0,0 +1,620 @@
...
+        )?;
+
+        let debug_info = self.pdb.debug_information()?;
+        let mut modules = debug_info.modules()?;
+
+        while let Some(module) = modules.next()? {
+            if let Some(module_info) = self.pdb.module_info(&module)? {
+                let mut symbols = module_info.symbols()?;
+                while let Some(symbol) = symbols.next()? {
+
```
ranweiler, line 76: Stray newline.

src/agent/coverage/src/block/pe_provider.rs
```diff
@@ -0,0 +1,623 @@
...
+            access_offsets,
+            address_map,
+            data,
+            pe,
+            table,
+        })
+    }
+
+    pub fn visit_procedure_symbol(&mut self, proc: &ProcedureSymbol) -> Result<()> {
+        use iced_x86::{Decoder, DecoderOptions, Instruction, Mnemonic, OpKind};
```
bmc-msft, line 507: Can this be imported at the top please?

bmc-msft:
This PR needs #847 merged to fix the build.

src/agent/coverage/src/block/pe_provider.rs
```diff
@@ -0,0 +1,623 @@
...
+        Ok(pcs)
+    }
+}
+
+/// Searches a PDB for data symbols that delimit various Sancov tables.
+#[derive(Default)]
+pub struct SancovDelimiterVisitor<'am> {
+    address_map: AddressMap<'am>,
+
+    llvm_bools_start: Option<u32>,
```
bmc-msft, line 147: Since these always should come in pairs, I propose `llvm_bools: Option<u32, u32>` (with similar updates elsewhere where both start & stop are defined & needed.

It took re-reviewing this a few times to realize that the macro define_table_getter was handling the case where one symbol was set and not the other with `self.$start?` and `self.$stop?`.

ranweiler: > should come in pairs

They _should_, but we only get to see them one at a time, with an undefined order of symbol visitation, so we can't use the type you suggest.

> It took re-viewing a few times

That's fair, the macro (and much else) needs doc commenting. FYI, I wrote the various getters _without_ a macro at first, but it really was more error-prone and less clear in the end.

ranweiler: Added many comments.

bmc-msft (approved):

bmc-msft (approved):

PR #855, opened 2021-05-05
Factor out Sancov analysis

ranweiler:
- Factor out OS-independent Sancov analysis logic
- Remove some unwraps and unchecked casts
- In the inline access scanner, use a VA for the disassembler PC and explicitly compute offsets


bmc-msft (approved):
Note, I reviewed #858 first, though the same comments apply.

PR #858, opened 2021-05-06
Explicitly pad start delimiters on Windows

ranweiler:
Sancov table start delimiters are padded on Windows, but not Linux. Expose this in the `SancovDelimiters` interface and update the PE block provider.

Based on #855, merge after.

### DIFF
--- a/src/agent/coverage/src/sancov.rs
+++ b/src/agent/coverage/src/sancov.rs
@@ -7,6 +7,12 @@ use std::convert::TryInto;
 use anyhow::{format_err, Result};
 use iced_x86::{Decoder, DecoderOptions, Instruction, Mnemonic, OpKind};
 
+/// Size of padding inserted (on Window) between `__start_` delimiter symbols
+/// and the first entry of the delimited table's array.
+///
+/// To find the true start offset of the table, add this to the symbol value.
+const DELIMITER_START_PADDING: u32 = 8;
+
 #[derive(Default)]
 pub struct SancovDelimiters {
     llvm_bools_start: Option<u32>,
@@ -280,12 +286,6 @@ pub enum Delimiter {
     MsvcPreviewCountersStart,
     MsvcPreviewCountersStop,
 }
-
-/// Size of padding inserted (on Window) between `__start_` delimiter symbols
-/// and the first entry of the delimited table's array.
-///
-/// To find the true start offset of the table, add this to the symbol value.
-const DELIMITER_START_PADDING: u32 = 8;
 
 impl std::str::FromStr for Delimiter {
     type Err = anyhow::Error;
--- a/src/agent/coverage/src/sancov.rs
+++ b/src/agent/coverage/src/sancov.rs
@@ -87,7 +87,6 @@ impl SancovDelimiters {
         }
 
         if let Some(table) = self.msvc_preview_counters_table(pad) {
-
             return Some(table);
         }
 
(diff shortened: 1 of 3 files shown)


src/agent/coverage/src/sancov.rs
```diff
@@ -0,0 +1,429 @@
...
+
+    define_table_getter!(
+        name = msvc_preview_counters_table,
+        start = msvc_preview_counters_start,
+        stop = msvc_preview_counters_stop,
+        ty = SancovTableTy::Counters,
+    );
+
+    pub fn insert(&mut self, delimiter: Delimiter, offset: u32) {
+        use Delimiter::*;
```
bmc-msft, line 170: Importing from an enum defined in this file is a unusual pattern that I don't think we do anywhere else.  I recognize this is copied from a previous commit, but seeing it here I found it odd.

src/agent/coverage/src/sancov.rs
```diff
@@ -0,0 +1,429 @@
...
+    MsvcPcsStop,
+    MsvcPreviewCountersStart,
+    MsvcPreviewCountersStop,
+}
+
+/// Size of padding inserted (on Window) between `__start_` delimiter symbols
+/// and the first entry of the delimited table's array.
+///
+/// To find the true start offset of the table, add this to the symbol value.
+const DELIMITER_START_PADDING: u32 = 8;
```
bmc-msft, line 288: Can we move this earlier in the file?  I thought you missed it in the reorg when reviewing define_table_getter

src/agent/coverage/src/sancov.rs
```diff
@@ -67,58 +66,54 @@ macro_rules! define_table_getter {
...
         }
 
-        if let Some(table) = self.msvc_counters_table() {
+        if let Some(table) = self.msvc_counters_table(pad) {
             return Some(table);
         }
 
-        if let Some(table) = self.msvc_preview_counters_table() {
+        if let Some(table) = self.msvc_preview_counters_table(pad) {
+
```
bmc-msft, line 90:
```suggestion
```

src/agent/coverage/src/sancov.rs
```diff
@@ -67,58 +66,54 @@ macro_rules! define_table_getter {
...
         // No MSVC tables found. Check for LLVM-emitted tables.
 
-        if let Some(table) = self.llvm_bools_table() {
+        if let Some(table) = self.llvm_bools_table(pad) {
             return Some(table);
         }
 
-        if let Some(table) = self.llvm_counters_table() {
+        if let Some(table) = self.llvm_counters_table(pad) {
+
```
bmc-msft, line 101:
```suggestion
```

src/agent/coverage/src/sancov.rs
```diff
@@ -67,58 +66,54 @@ macro_rules! define_table_getter {
...
+    pub fn pcs_table(&self, pad: bool) -> Option<SancovTable> {
         // Check for MSVC tables first.
-        if let Some(table) = self.msvc_pcs_table() {
+        if let Some(table) = self.msvc_pcs_table(pad) {
             return Some(table);
         }
 
-        if let Some(table) = self.llvm_pcs_table() {
+        if let Some(table) = self.llvm_pcs_table(pad) {
+
```
bmc-msft, line 116:
```suggestion
```

src/agent/coverage/src/sancov.rs
ranweiler: Changed. There are a small number of other places where we do it, but it's not super high value here.

bmc-msft (approved):

PR #859, opened 2021-05-06
explicitly ignore azcopy temporary files from directory monitoring

### LINKED ISSUE #856: Files of the format ".azDownload-guid-input" are being uploaded to inputs container
## Information

- Onefuzz version: 2.14.0
- OS: Canonical:UbuntuServer:18.04-LTS:latest

1. Submit job with generic supervisor task
2. `onefuzz containers files list <inputs container>`

### Expected result

Only input files generated by the fuzzer are uploaded to the inputs container

### Actual result

Files such as

```
.azDownload-12e6ac53-bf7c-864c-4254-b16917146cb6-<input-file>
```

are also uploaded to the inputs container

In the worst cases this creates a feedback loop, resulting in files like

```
.azDownload-a4f3c77f-1d1b-1b4f-49e3-206532bf66e2-.azDownload-33927fcb-efe8-214c-4fed-2d300b04f527-.azDownload-147eab04-03d9-e44d-5e3f-3bc5fc02bf01-<input-file>
```

In these cases eventually the fuzzing task will fail due to an azcopy error.

bmc-msft:
Fixes #856

### DIFF
--- a/src/agent/onefuzz/src/syncdir.rs
+++ b/src/agent/onefuzz/src/syncdir.rs
@@ -203,7 +203,15 @@ impl SyncedDir {
                 let file_name = item
                     .file_name()
                     .ok_or_else(|| anyhow!("invalid file path"))?;
-                if ignore_dotfiles && file_name.to_string_lossy().starts_with('.') {
+                let file_name_str = file_name.to_string_lossy();
+
+                // explicitly ignore azcopy temporary files
+                // https://github.com/Azure/azure-storage-azcopy/blob/main/ste/xfer-remoteToLocal-file.go#L35
+                if file_name_str.starts_with(".azDownload-") {
+                    continue;
+                }
+
+                if ignore_dotfiles && file_name_str.starts_with('.') {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz/src/syncdir.rs
```diff
@@ -203,7 +203,15 @@ impl SyncedDir {
                 let file_name = item
                     .file_name()
                     .ok_or_else(|| anyhow!("invalid file path"))?;
-                if ignore_dotfiles && file_name.to_string_lossy().starts_with('.') {
+                let file_name_str = file_name.to_string_lossy();
+
+                // explicitly ignore azcopy temporary files
+                // https://github.com/Azure/azure-storage-azcopy/blob/main/ste/xfer-remoteToLocal-file.go#L35
```
ranweiler, line 209: 👍

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -231,11 +239,19 @@ impl SyncedDir {
...
+                if file_name_str.starts_with(".azDownload-") {
                     continue;
                 }
-                event!(event.clone(); EventData::Path = item.display().to_string());
 
+                if ignore_dotfiles && file_name_str.starts_with('.') {
```
ranweiler, line 250: It looks like we've been exactly duplicating logic in each branch that should always be the same. Since it is becoming a bit more complex now, consider introducing a helper function like `should_ignore_file(path, ignore_dotfiles) -> bool` that will tell us whether to `continue` based on filename.

src/agent/onefuzz/src/syncdir.rs
ranweiler: I'm assuming there's currently no way to control the temporary path that `azcopy` uses.

ranweiler (approved):

src/agent/onefuzz/src/syncdir.rs
ranweiler: Yeah, looks like not: https://github.com/Azure/azure-storage-azcopy/blob/067b2fc3f79aca43a378ee1a6761268c7d1e4f38/ste/xfer-remoteToLocal-file.go#L471-L477

bmc-msft: It's for a good reason.  Generally speaking, rename of files in the same directory are atomic, where move is not.   It's significantly more complicated (and sometimes not possible) to ensure users get the benefit of downloading to a temp file then renaming if you have to use an alternate directory.

PR #864, opened 2021-05-06
Improve precision of inline table access scanning 

ranweiler:
- Detect Sancov table accesses via static displacements, not just PC-relative
- Explicitly skip instructions with no memory operand
- Skip instructions whose effective address relies on (non-PC) register values

### DIFF
--- a/src/agent/coverage/src/sancov.rs
+++ b/src/agent/coverage/src/sancov.rs
@@ -5,7 +5,7 @@ use std::collections::BTreeSet;
 use std::convert::TryInto;
 
 use anyhow::{format_err, Result};
-use iced_x86::{Decoder, DecoderOptions, Instruction, Mnemonic, OpKind};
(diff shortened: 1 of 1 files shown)


src/agent/coverage/src/sancov.rs
```diff
@@ -337,9 +337,30 @@ impl SancovInlineAccessScanner {
         decoder.set_ip(va);
 
         let mut inst = Instruction::default();
+
         while decoder.can_decode() {
             decoder.decode_out(&mut inst);
 
+            // If no memory operand, there is no table access.
```
ranweiler, line 344: In a future cleanup pass, we'll factor out all this detection logic into a standalone function or (configurable) struct.

ranweiler:
Build failed due to`cargo audit` check, addressed in #865.

bmc-msft (approved):

PR #868, opened 2021-05-07
Try to use Sancov tables when analyzing ELF modules

### LINKED ISSUE #846: Try to use Sancov tables for Linux coverage
Attempt the same Sancov table recovery that we do for PE + PDB modules, but for ELF modules. Reuse the `SancovTable` types and logic as much as possible.

ranweiler:
Closes #846.

### DIFF
--- a/src/agent/coverage/src/elf.rs
+++ b/src/agent/coverage/src/elf.rs
@@ -1,0 +1,280 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::collections::BTreeSet;
+use std::convert::{TryFrom, TryInto};
+
+use anyhow::{format_err, Result};
+use goblin::elf::{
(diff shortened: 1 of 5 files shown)


src/agent/coverage/src/elf.rs
```diff
@@ -0,0 +1,280 @@
...
+        let file_offset = self.ctx.va_to_file_offset(va, None)?;
+        let file_range = file_offset..(file_offset + pcs_table.size);
+
+        let table_data = self
+            .ctx
+            .data
+            .get(file_range)
+            .ok_or_else(|| format_err!("Sancov table data out of file range"))?;
+
+        // Assumes x86-64, `sizeof(uintptr_t) == 8`.
```
ranweiler, line 183: We'll address this in a future PR by introducing a PC table parser abstraction, which we will use for both PE and ELF modules.

PR #839, opened 2021-04-29
Refactoring proxy lifetime to only shutdown when proxy is out-of-date.

nharper285:
## Summary of the Pull Request

_What is this about?_
We'd like to refactor the proxy lifecycle to only delete when the proxy is out-of-date - i.e. when the proxy is older than 7 days or a mismatched version. I've changed two files, proxy.py and timer_daily\init.py to check for the version and timestamp before stopping a live proxy. 

## PR Checklist
* [ ] Applies to work item: #xxx
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.
* [ ] Tests added/passed
* [ ] Requires documentation to be updated
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Info on Pull Request

_What does this include?_
Changes to two files: 
proxy.py: 
- get_or_create() edited to check if timestamp is >7 days.
- Created is_outdated() to check version and timestamp for out-of-date proxy. 
timer_daily/init.py
- Proxy check now includes is_outdated() before determining if a proxy should be shutdown. 

## Validation Steps Performed
Deploying test instance to determine if proxy lives past a single day.

PR #883, opened 2021-05-13
force ASCII generation of markdown document generation

bmc-msft:
This uplifts functionality from #839 in a simplified fashion.

### DIFF
--- a/src/pytypes/extra/generate-docs.py
+++ b/src/pytypes/extra/generate-docs.py
@@ -310,7 +310,6 @@ def main() -> None:
         2, "Full Event Schema", message.schema_json(indent=4, sort_keys=True), "json"
     )
 
-    filename = sys.argv[1]
     with open(filename, "w", newline="\n", encoding="ascii") as handle:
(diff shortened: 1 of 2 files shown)


bmc-msft:
cc: @nharper285

src/pytypes/extra/generate-docs.py
```diff
@@ -257,42 +264,54 @@ def main() -> None:
...
+        result += typed(
+            4, "Schema", example.schema_json(indent=4, sort_keys=True), "json"
+        )
+
+    result += typed(
+        2, "Full Event Schema", message.schema_json(indent=4, sort_keys=True), "json"
+    )
 
-    typed(2, "Full Event Schema", message.schema_json(indent=4, sort_keys=True), "json")
+    with open(filename, "w", newline="\n", encoding="ascii") as handle:
```
ranweiler, line 313: I suggest we use `utf8` instead of `ascii`.

ranweiler (approved):

PR #888, opened 2021-05-13
Updating default VM_SKU on request. 

nharper285:

_What is this about?_
A developer requested we update the default CLI value for VM_SKU to Standard_D4s_v3, as this SKU can more efficiently use 4 cores to run standard jobs - esp. libfuzzer jobs. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Changed files:
- modified:   src/cli/onefuzz/api.py
- modified:   src/cli/onefuzz/templates/ossfuzz.py
Updating default value to new Standard. 

## Validation Steps Performed

Integration tests.

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -1276,7 +1276,7 @@ class Scaleset(Endpoint):
         size: int,
         *,
         image: Optional[str] = None,
-        vm_sku: Optional[str] = "Standard_D2s_v3",
+        vm_sku: Optional[str] = "Standard_D4s_v3",
         region: Optional[primitives.Region] = None,
         spot_instances: bool = False,
         ephemeral_os_disks: bool = False,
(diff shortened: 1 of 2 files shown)


bmc-msft:
This change would increase CPU costs for fuzzing workloads by 2x, based on information from the public Azure Pricing Calculator.  We should be hesitant to make changes that impact the cost to all users of OneFuzz.

Could your developer's needs be served by setting up the scalesets in the pools for your instance to use this VM SKU?

nharper285:
> This change would increase CPU costs for fuzzing workloads by 2x, based on information from the public Azure Pricing Calculator. We should be hesitant to make changes that impact the cost to all users of OneFuzz.
> 
> Could your developer's needs be served by setting up the scalesets in the pools for your instance to use this VM SKU?

PR #893, opened 2021-05-14
handle issue from azure-mgmt-resource 17.0.0 upgrade

bmc-msft:
In azure-mgmt-resource 17.0.0, the resources.get_by_id().properties is now a string, rather than a dict.

In practice, it's not a json string (uses single quotes rather than double quotes).  This converts the quotes & uses json to decode it.

For example, in practice, the value is: "{'tenantId': 'GUID', 'principalId': 'GUID', 'clientId': 'GUID'}

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/creds.py
+++ b/src/api-service/__app__/onefuzzlib/azure/creds.py
@@ -3,6 +3,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+import json
 import os
 from typing import Any, List
 from uuid import UUID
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -130,7 +131,19 @@ def get_scaleset_principal_id() -> UUID:
         credential=get_identity(), subscription_id=get_subscription()
     )
     uid = client.resources.get_by_id(get_scaleset_identity_resource_path(), api_version)
-    return UUID(uid.properties["principalId"])
+
+    # workaround issue from azure-mgmt-resource, where properties is now a str
+    # instead of an obj.
+    # https://github.com/Azure/azure-sdk-for-python/pull/18686/files
```
ranweiler, line 137: Was this meant to link to something more specific, or is the `/files` stray?

bmc-msft:
For reference: https://github.com/Azure/azure-sdk-for-python/issues/18763

ranweiler (approved):

src/api-service/__app__/onefuzzlib/azure/creds.py
bmc-msft: It was more of a "look at the massive # of `object` that are now `str` in this PR"

PR #894, opened 2021-05-14
Use assign User.Read from Microsoft graph instead of AAD Graph

### LINKED ISSUE #877: The deployment script should configure the application registion to use Microsoft graph instead of Azure ad graph
Azure Ad graph will [soon be deprecated](https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-planning-checklist#:~:text=Azure%20AD%20Graph%20API%20is,technical%20support%20or%20security%20updates.). So we need to migrate the configuration of our application registration to Microsoft Graph.
The deployment script adds the `user_impersonation` rights [from Azure Ad graph](https://github.com/microsoft/onefuzz/blob/54d49a973e4007b3db2ff0d134d368c648183524/src/deployment/deploy.py#L314) (00000002-0000-0000-c000-000000000000 is the [guid of Azure Ad Graph](https://www.shawntabrizi.com/aad/common-microsoft-resources-azure-active-directory/))

chkeita:
Replace the permission in the application registration. 
closes #877


src/deployment/deploy.py
```diff
@@ -72,7 +72,8 @@
     update_pool_registration,
 )
 
-USER_IMPERSONATION = "311a71cc-e848-46a1-bdf8-97ff7156d8e6"
+USER_READ_PERMISSION = "e1fe6dd8-ba31-4d61-89e7-88639da4683d"
```
ranweiler, line 75: For each of these constants, are there docs we can link to?

ranweiler (approved):

src/deployment/deploy.py
chkeita: could not find an external doc with all the guids for the permissions. So I added a comment explaining how I got those values

bmc-msft: Can we link to this document, which demonstrates it?

https://docs.microsoft.com/en-us/azure/aks/azure-ad-integration-cli#create-azure-ad-server-component

PR #872, opened 2021-05-10
Single tenant auth, issuer url fix

muhammadmeisam:
## Info on Pull Request
The deploy.py is only able to set the issuer url to the redmond tenant in the case of single tenant auth. For multi tenant, since the the tenant_domain is provided specifically, it works fine. In the case of single tenant, if the instance is not in the redmond tenant, the cli is unable to authenticate itself.

_What does this include?_

## Validation Steps Performed
A test instance was deployed in the PME instance and it was verified that the cli could authenticate with the instance.

PR #898, opened 2021-05-17
fix deployment using client_id

bmc-msft:
Fix issue in client creation from #872

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -438,7 +438,7 @@ class Client:
                     % onefuzz_client.config.tenant_id
                 )
             self.cli_config = {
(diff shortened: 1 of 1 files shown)


bmc-msft:
In the aforementioned PR, `Application.client_id` isn't available.

It causes the exception:
```
  File "deploy.py", line 1082, in <module>
    main()
  File "deploy.py", line 1076, in main
    state[1](client)
  File "deploy.py", line 441, in setup_rbac
    "client_id": onefuzz_cli_app.client_id,
AttributeError: 'Application' object has no attribute 'client_id'
```

chkeita (approved):

PR #904, opened 2021-05-20
simplify long-life proxy cleanup

bmc-msft:
After merging #839 , I hit a race condition where I created proxy session while the proxy was being shut down.

This PR addresses this race condition by
1. unifies all of the proxy state actuation into timer_proxy rather than two distinct timer functions.
2. sets outdated at the end of the timer_proxy function, and checks during the next iteration (30s later).

Additionally, this works addresses another issue I had not seen during the review of #839 but noticed upon handling the above issue.  Currently, an up-to-date proxy is always maintained for any region a proxy has been created.  If the worker VMs all move from eastus to eastus2, we don't need to continue to manage proxies in eastus.

This PR also includes EventProxyStateUpdated, which adds in debugging this aforementioned issue.

This PR also changes the VM names for proxies to be `proxy-<base58(proxy_id)>` to handle the automatic disk name generation limits.  Otherwise, disk names are truncated from the VM names and never get deleted.

### DIFF
--- a/docs/webhook_events.md
+++ b/docs/webhook_events.md
@@ -37,6 +37,7 @@ Each event will be submitted via HTTP POST to the user provided URL.
 * [proxy_created](#proxy_created)
 * [proxy_deleted](#proxy_deleted)
 * [proxy_failed](#proxy_failed)
+* [proxy_state_updated](#proxy_state_updated)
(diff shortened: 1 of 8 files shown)


bmc-msft:
CC: @nharper285

PR #907, opened 2021-05-21
encode proxy name as base58 to allow full deletion of resources

bmc-msft:
As is, the proxy name gets truncated when turned into the disk name.
This prevents the resource cleanup from completion.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/proxy.py
+++ b/src/api-service/__app__/onefuzzlib/proxy.py
@@ -65,7 +65,7 @@ class Proxy(ORMMixin):
 
     def get_vm(self) -> VM:
         vm = VM(
-            name="proxy-%s" % base58.b58encode(self.proxy_id).decode(),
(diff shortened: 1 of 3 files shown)


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -64,7 +65,7 @@ def key_fields(cls) -> Tuple[str, Optional[str]]:
 
     def get_vm(self) -> VM:
         vm = VM(
-            name="proxy-%s-%s" % (self.region, self.proxy_id),
+            name="proxy-%s" % base58.b58encode(self.proxy_id.bytes).decode(),
```
ranweiler, line 68: Could you document the motivation for this, and the max length you observed?

ranweiler (approved):

src/api-service/__app__/onefuzzlib/proxy.py
bmc-msft: added validation with comments.

PR #909, opened 2021-05-21
Fix multi proxy race condition

bmc-msft:
Refactored PR of #904 for easier review.  Once #908 is reviewed & merged, this will be easier to review.

### DIFF
--- a/src/api-service/__app__/timer_proxy/__init__.py
+++ b/src/api-service/__app__/timer_proxy/__init__.py
@@ -49,7 +49,7 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
     scalesets = Scaleset.search()
     regions = set(x.region for x in scalesets)
     for region in regions:
-        if not any(not x.outdated for x in proxies if x.region == region):
+        if all(x.outdated for x in proxies if x.region == region):
             Proxy.get_or_create(region)
 
     events = get_events()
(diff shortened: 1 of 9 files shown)


bmc-msft:
To compare just the changes for this PR without the changes in #908, use this link:

https://github.com/bmc-msft/onefuzz/compare/proxy-state-update-event...fix-multi-proxy-race-condition

src/api-service/__app__/timer_proxy/__init__.py
```diff
@@ -34,6 +41,17 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
...
 
+        if proxy.is_outdated():
+            proxy.outdated = True
+            proxy.save()
+
+    # make sure there is a proxy for every currently active region
+    scalesets = Scaleset.search()
+    regions = set(x.region for x in scalesets)
+    for region in regions:
+        if not any(not x.outdated for x in proxies if x.region == region):
```
ranweiler, line 52:
```suggestion
        if all(x.outdated for x in proxies if x.region == region):
```

src/api-service/__app__/timer_proxy/__init__.py
```diff
@@ -34,6 +41,17 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
             )
             process_state_updates(proxy)
 
+        if proxy.is_outdated():
+            proxy.outdated = True
```
ranweiler, line 45: In the future, should `outdated` really be a `state`, instead of an additional field? So we'd have a `ProxyState` that would be an extension of `VmState`.

ranweiler (approved):

PR #912, opened 2021-05-21
Implement fast coverage-specific breakpoints

ranweiler:
Thanks, @lzybkr!

### DIFF
--- a/src/agent/coverage/src/pdb.rs
+++ b/src/agent/coverage/src/pdb.rs
@@ -41,7 +41,7 @@ pub fn find_pdb_path(pe_path: &Path, pe: &PE, target_handle: Option<HANDLE>) -> 
     let cv_filename = Path::new(cv_filename);
 
     // If the PE-specified PDB file exists on disk, use that.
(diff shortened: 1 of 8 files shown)


src/agent/coverage/src/pdb.rs
```diff
@@ -41,7 +41,7 @@ pub fn find_pdb_path(pe_path: &Path, pe: &PE, target_handle: Option<HANDLE>) ->
     let cv_filename = Path::new(cv_filename);
 
     // If the PE-specified PDB file exists on disk, use that.
-    if std::fs::metadata(&cv_filename)?.is_file() {
+    if std::fs::metadata(&cv_filename).map_or(false, |md| md.is_file()) {
```
ranweiler, line 44: Let's split this off as a separate bugfix.

ranweiler: A slight variant of this is part of #913.

ranweiler:
The code in question has been rewritten, so closing this out.

PR #928, opened 2021-05-25
only set the SSH key permissions on Windows once

bmc-msft:
On Windows, removing permissions that have already been removed fails.  As such, this needs to happen once and only once.  

NOTE: SSH keys are added as node commands, which are processed serially.

### DIFF
--- a/src/agent/onefuzz-supervisor/src/commands.rs
+++ b/src/agent/onefuzz-supervisor/src/commands.rs
@@ -10,7 +10,7 @@ use tokio::{fs, io::AsyncWriteExt, process::Command};
 use std::{env, path::PathBuf};
 
 #[cfg(target_os = "windows")]
-use tokio::sync::OnceCell;
+use tokio::sync::{OnceCell, SetError};
 
 #[cfg(target_os = "linux")]
 use users::{get_user_by_name, os::unix::UserExt};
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -43,69 +53,72 @@ pub async fn add_ssh_key(key_info: SshKeyInfo) -> Result<()> {
...
-        .wait_with_output()
-        .await
-        .context("Powershell Get-ACL | Set-ACL failed to run")?;
-    if !result.status.success() {
-        bail!(
-            "set authorized_keys ({}) permissions failed: {:?}",
-            admin_auth_keys_path.display(),
-            result
-        );
+    if SET_PERMISSION_ONCE.set(()).is_ok() {
```
ranweiler, line 56: I suggest we match on the result of `set()`, and only fall through on `Err(SetErr::AlreadyInitialized)`.

ranweiler: In the current impl, is it correct that a caller could attempt this once, succeed with _some_ `icacls` invocations, then fail, and then call this again with a partially-initialized state?

ranweiler: I'm wondering if we really need either multiple `OnceCell`'s or an atomics + `lazy_static` state machine.

bmc-msft: OnceCell is already thread-safe.  I'm not sure why we need atomics on top of this.

Beyond that, we have significant other issues if we handle executing node commands concurrently, so we should not see this executed in anything but a serial fashion.

If any node command fails, we hard fail the supervisor.  As such, we should only ever enter this block once and only once.

ranweiler: > I'm not sure why we need atomics on top of this.

Not suggesting atomics are needed for `OnceCell`. Only mentioned them wrt the idea of using a `lazy_static` state machine that required mutation to drive forward (which I'm not necessarily claiming we should do!).

>  so we should not see this executed in anything but a serial fashion.

I understand. My suggestion is that also actually checking this doesn't take too much work (just the `match`, vs. conditional check), so it seems worthwhile. Your call.

> If any node command fails, we hard fail the supervisor.

True, currently. Again, your call, just thinking about future-proofing.

ranweiler (approved):

src/agent/onefuzz-supervisor/src/commands.rs
bmc-msft: Added the other match statements

PR #929, opened 2021-05-26
Add timeout when recording Linux coverage

### LINKED ISSUE #923: Enable configurable coverage timeout on Linux
At a minimum, the coverage recorder needs to return a PID or PID wrapper that can be used to kill the child process.

ranweiler:
- Simplify the Linux block coverage `Recorder` API 
- Add a (mandatory) timeout to recording runs

Closes #923.

### DIFF
--- a/src/agent/coverage/src/block/linux.rs
+++ b/src/agent/coverage/src/block/linux.rs
@@ -28,7 +28,6 @@ pub struct Recorder<'c> {
     filter: CmdFilter,
     images: Option<Images>,
     tracer: Ptracer,
-    tracee: Tracee,
 }
 
 impl<'c> Recorder<'c> {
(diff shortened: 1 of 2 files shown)


src/agent/coverage/src/block/linux.rs
```diff
@@ -408,3 +404,52 @@ fn continue_to_init_execve(tracer: &mut Ptracer) -> Result<Tracee> {
 
     anyhow::bail!("did not see initial execve() in tracee while recording coverage");
 }
+
+const MAX_POLL_PERIOD: Duration = Duration::from_millis(100);
```
ranweiler, line 408: This should probably be much longer (our Windows equivalent is 1 sec).

bmc-msft (approved):

PR #932, opened 2021-05-26
fix issue deleting proxy from storage tables

bmc-msft:
This issue was introduced in 2.18.0.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/proxy.py
+++ b/src/api-service/__app__/onefuzzlib/proxy.py
@@ -165,8 +165,7 @@ class Proxy(ORMMixin):
             self.stopped()
 
     def stopped(self) -> None:
-        if self.state != VmState.stopped:
-            self.set_state(VmState.stopped)
+        self.set_state(VmState.stopped)
         logging.info(PROXY_LOG_PREFIX + "removing proxy: %s", self.region)
         send_event(EventProxyDeleted(region=self.region, proxy_id=self.proxy_id))
         self.delete()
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -165,7 +165,10 @@ def stopping(self) -> None:
             self.stopped()
 
     def stopped(self) -> None:
+        if self.state != VmState.stopped:
```
ranweiler, line 168: Should `set_state()` be responsible for this "set to `<state>` if not `<state>`" logic? Do we ever _not_ want this?

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -165,7 +165,10 @@ def stopping(self) -> None:
...
+            self.set_state(VmState.stopped)
         logging.info(PROXY_LOG_PREFIX + "removing proxy: %s", self.region)
+        send_event(EventProxyDeleted(region=self.region, proxy_id=self.proxy_id))
```
ranweiler, line 171: Should we only send this _after_ successfully calling `delete()`?

src/api-service/__app__/onefuzzlib/proxy.py
bmc-msft: This is already done in set_state, so I removed the additional check here.

src/api-service/__app__/onefuzzlib/proxy.py
bmc-msft: I think it's a toss-up.  I was trying to remove the conceptual use-after-delete.

ranweiler: That's legit. This and the other (related) comments about events / states are more discussion points. We can revisit the big picture going forward; I think the only thing that matters is that we have a well defined point in the business logic at which we expect to fire a given event!

ranweiler (approved):

PR #935, opened 2021-05-26
mark tasks that are stopped that never started with an error

### LINKED ISSUE #934: explain why some tasks don't get started in a job
Let's assume we have a pool with only a single VM.

When we launch a libfuzzer basic template job for one hour, the fuzzing task will launch and run for one hour.  Once any task in the job starts executing, the job timer begins.  After one hour, the job is stopped.  However, in this hour, the coverage and crash reporting tasks stay in the "scheduled" state until the job is eventually stopped because a VM never tried to start that task.

In this limited VM count example, the user is shown the task "stopped" successfully.  While this is true, it would be more useful to say "this task never launched".

bmc-msft:
Fixes #934

### DIFF
--- a/src/api-service/__app__/timer_tasks/__init__.py
+++ b/src/api-service/__app__/timer_tasks/__init__.py
@@ -19,7 +19,7 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
     expired_tasks = Task.search_expired()
     for task in expired_tasks:
         logging.info("stopping expired task: %s", task.job_id)
-        task.stopping()
+        task.mark_stopping()
 
     expired_jobs = Job.search_expired()
     for job in expired_jobs:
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/timer_tasks/__init__.py
```diff
@@ -19,7 +19,7 @@ def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa:
     expired_tasks = Task.search_expired()
     for task in expired_tasks:
         logging.info("stopping expired task: %s", task.job_id)
-        task.stopping()
+        task.mark_stopping()
```
ranweiler, line 22: How does this change relate to the other? Is it required?

bmc-msft: It is required, such that instead of calling the state handler directly, it does all of the other related stuff.

State handling is something we could use another refactor on, but that's out of scope for fixing this issue.

ranweiler (approved):

PR #936, opened 2021-05-27
prevent object id collision in hide_secrets

bmc-msft:
This fixes an issue related to object id reuse that can occur making the object identification cache fail.  Instead, this simplifies `hide_secrets` to always recurse and use `setattr` to always set the value based on the recursion.

Note, the object id reuse issue was seen in the `events.filter_event_recurse` development and this was the fix for the id reuse there.

Python documentation states:

> id(object):
>     Return the “identity” of an object. This is an integer (or long integer)
>     which is guaranteed to be unique and constant for this object during its
>     lifetime. Two objects with non-overlapping lifetimes may have the same
>     id() value.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/orm.py
+++ b/src/api-service/__app__/onefuzzlib/orm.py
@@ -15,7 +15,6 @@ from typing import (
     List,
     Mapping,
     Optional,
-    Set,
     Tuple,
     Type,
     TypeVar,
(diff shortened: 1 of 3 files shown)


bmc-msft:
@ranweiler this should fix the build issues you're seeing.

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -223,6 +222,33 @@ def raw(
...
+
+def hide_secrets(data: B, hider: Callable[[SecretData], SecretData]) -> B:
+    for field in data.__fields__:
+        field_data = getattr(data, field)
+
+        if isinstance(field_data, SecretData):
+            field_data = hider(field_data)
+        elif isinstance(field_data, list):
+            if len(field_data) > 0:
+                if not isinstance(field_data[0], BaseModel):
```
ranweiler, line 236: Is this meant to check that `field_data : [BaseModel]`, or something else?

If the former, I'm not sure why checking the first element is enough (i.e. why we can assume that `field_data` is a homogenous list, if it is a list).

ranweiler: Ah, is it because we have typed `BaseModel` fields, and we assume our field-typing regime enforces homogenous lists?

If so, that seems better, but still not quite enough (couldn't they be `Union[]` values?). Again, I might just be missing something.

ranweiler: Oh, you just _moved_ this code. I guess it'd still be good to confirm 😄 .

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -281,38 +307,8 @@ def get_keys(self) -> Tuple[KEY, KEY]:
...
         return (partition_key, row_key)
 
-    @classmethod
-    def hide_secrets(
-        cls,
-        model: BaseModel,
-        hider: Callable[["SecretData"], None],
-        visited: Set[int] = set(),
-    ) -> None:
-        if id(model) in visited:
```
ranweiler, line 291: So _this_ is where the action is. Great catch!

ranweiler (approved):

src/api-service/__app__/onefuzzlib/orm.py
bmc-msft: The assumption was if the first field was a BaseModel, all the fields would be.  However, I agree that's a naïve approach.  

Addressed.

chkeita (approved):

PR #830, opened 2021-04-23
Crash repro as task

bmc-msft:

PR #937, opened 2021-05-27
add an optional initial delay for heartbeats

bmc-msft:
One of the difficulties in crash repro as task is a race condition where we the client tries to connect before the cdb is running.

This makes it such that we can use the heartbeat to identify if the task has started before connecting in.

NOTE:  In this PR, it's always set to None.  See #830 for it's actual usage.  However, I split out the PR for easier review.

### DIFF
--- a/src/agent/onefuzz/src/heartbeat.rs
+++ b/src/agent/onefuzz/src/heartbeat.rs
@@ -84,7 +84,6 @@ where
         TContext: Send + Sync + 'static,
     {
         let heartbeat_period = heartbeat_period.unwrap_or(DEFAULT_HEARTBEAT_PERIOD);
-        let initial_delay = initial_delay.unwrap_or(DEFAULT_HEARTBEAT_PERIOD);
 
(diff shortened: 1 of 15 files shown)


ranweiler (approved):

src/agent/onefuzz/src/heartbeat.rs
```diff
@@ -93,7 +95,7 @@ where
 
         let flush_context = context.clone();
         let heartbeat_process = task::spawn(async move {
-            random_delay(heartbeat_period).await;
+            random_delay(initial_delay).await;
```
chkeita, line 98: should the initial delay still be random not that it is explicitly specified ? I think a jitter here is more appropriate here

bmc-msft: Normally, yes.  However, this is a user-initiated task, not something like launching 3000 tasks from CICD.  We shouldn't see a thundering herd here.

chkeita (approved):

PR #939, opened 2021-05-28
work around issue with discriminated typed unions

bmc-msft:
We're experiencing a bug where Unions of sub-models are getting downcast, which causes a loss of information.  

As an example, EventScalesetCreated was getting downcast to EventScalesetDeleted.  I have not figured out why, nor can I replicate it locally to minimize the bug send upstream, but I was able to reliably replicate it on the service.

While working through this issue, I noticed that deserialization of SignalR events was frequently wrong, leaving things like tasks as "init" in `status top`.

Both of these issues are Unions of models with a type field, so it's likely these are related.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/events.py
+++ b/src/api-service/__app__/onefuzzlib/events.py
@@ -81,7 +81,7 @@ def send_event(event: Event) -> None:
         instance_name=get_instance_name(),
     )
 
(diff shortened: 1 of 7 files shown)


chkeita (approved):

bmc-msft:
Apparently, this is well known.

https://github.com/samuelcolvin/pydantic/issues/2079

Issue #943, opened 2021-06-01
command to delete job files


bmc-msft:
It would be useful to add command to delete all of the contextual containers used by a job.

PR #949, opened 2021-06-01
add `onefuzz jobs containers delete JOB_ID`

bmc-msft:
Addresses #943

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -1017,14 +1017,10 @@ class JobContainers(Endpoint):
             self.logger.info("not removing: %s", container_name)
 
         for container_name in to_delete:
-            try:
-                self.onefuzz.containers.get(container_name)
-                self.onefuzz.containers.delete(container_name)
+            if self.onefuzz.containers.delete(container_name).result:
                 self.logger.info("removed container: %s", container_name)
-            except Exception as err:
-                self.logger.error(
-                    "error removing container %s: %s", container_name, err
-                )
+            else:
+                self.logger.info("container already removed: %s", container_name)
 
 
 class JobTasks(Endpoint):
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -970,6 +970,62 @@ class JobContainers(Endpoint):
             results[container] = self.onefuzz.containers.files.list(container).files
         return results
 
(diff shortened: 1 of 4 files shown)


src/cli/onefuzz/api.py
```diff
@@ -970,6 +970,58 @@ def list(
...
+                        build=job.config.build,
+                        platform=task.os,
+                    )
+                    == container.name
+                ):
+                    to_delete.add(container.name)
+                else:
+                    to_keep.add(container.name)
+
+        for container_name in to_keep:
```
ranweiler, line 1011: In the absence of unit testing, it'd be nice if we were using non-intersecting sets / a structure that partitioned the containers.

I don't think you need to change it in this PR (your comment below helps a lot), but since this is about deletion, I'm inclined to be extra cautious.

src/cli/onefuzz/api.py
```diff
@@ -970,6 +970,58 @@ def list(
             results[container] = self.onefuzz.containers.files.list(container).files
         return results
 
+    def delete(self, job_id: UUID_EXPANSION, *, only_job_specific: bool = True) -> None:
```
ranweiler, line 973: Could we make this support a dry-run mode, without too much pain?

src/cli/onefuzz/api.py
```diff
@@ -1495,6 +1547,39 @@ def namespaced_guid(
             identifiers.append(platform)
         return uuid.uuid5(ONEFUZZ_GUID_NAMESPACE, ":".join(identifiers))
 
+    def build_container_name(
```
ranweiler, line 1550: 👍

ranweiler (approved):

src/cli/onefuzz/api.py
bmc-msft: added

src/cli/onefuzz/api.py
bmc-msft: added

src/cli/onefuzz/api.py
```diff
@@ -1005,19 +1012,15 @@ def delete(self, job_id: UUID_EXPANSION, *, only_job_specific: bool = True) -> N
                     == container.name
                 ):
                     to_delete.add(container.name)
-                else:
-                    to_keep.add(container.name)
 
+        to_keep = containers - to_delete
```
ranweiler, line 1016: 👍

ranweiler (approved):
Thanks, new changes make everything even more obvious.

PR #951, opened 2021-06-02
update deploy contrib dependencies

bmc-msft:
Update deploy-onefuzz-via-azure-devops 3rd party dependencies, primarily to update urllib3 which addresses CVE-2021-33503.

This fixes an issue when parsing authority of HTTP requests.  In the scope of this tool, the requests that are parsed come from GitHub, which is trusted in this context.


bmc-msft:
dup of #950

PR #953, opened 2021-06-02
stabilize `onefuzz jobs containers download`

bmc-msft:


src/cli/onefuzz/api.py
```diff
@@ -970,6 +971,33 @@ def list(
...
+        if not tasks:
+            raise Exception("no tasks with job_id:%s" % job_id)
+
+        for task in tasks:
+            for container in task.config.containers:
+                info = self.onefuzz.containers.get(container.name)
+                name = os.path.join(container.type.name, container.name)
+                to_download[name] = info.sas_url
+
+        if output is None:
```
ranweiler, line 988: Is it correct that this PR just makes `output` default to the current directory?

If so, I suggest just editing the title or description to match the commit.

ranweiler (approved):

src/cli/onefuzz/api.py
bmc-msft: no, this moves the implementation from `onefuzz debug job download_files` to jobs containers download.

In (probably only internally) discussions, we've stated the debug subset as not really stable.  This is an attempt to address one of the commands that is continuously useful.

ranweiler: Ah! That's right.

Issue #958, opened 2021-06-02
Create a job from a previous job using OneFuzz CLI


jagunter:
It would be nice if there was a way to easily rerun a job from the CLI. This could be useful in a couple of scenarios:

1. Extending a fuzzing run. If ran a fuzzer for 1 week and then want to run it for an additional week, it would be nice to be able to run the fuzzer again without needing to specify the exact arguments.
2. Fixing a fuzzer bug. If a fuzzing task fails due to an error in the fuzzer (not a crash in the target), it could be useful to rerun it after updating the fuzzer in the container.
3. Reproducing a bug. If a fuzzing task fails, it could be useful to rerun in some scenarios, e.g. testing after a OneFuzz upgrade.

ranweiler:
I like the idea of making this as easy as possible. It is natural to want to "run this job again, but with just one thing changed". I think we need to tease apart the different use cases and compare them to the existing data model.

> 1. Extending a fuzzing run. If ran a fuzzer for 1 week and then want to run it for an additional week, it would be nice to be able to run the fuzzer again without needing to specify the exact arguments.

This should definitely be easy. There is stuff to think about with container existence and identity, but it seems reasonable to have something that fails fast if any dependent resource arguments (storage containers, pools) changed or are missing. We could solve this by both (1) always allowing argument _overrides_, and (2) having a notion of "invalidated" arguments that _must_ be overridden (treated as missing, and thus must be re-specified) to re-run the job (which should get a new ID).

> 2. Fixing a fuzzer bug. If a fuzzing task fails due to an error in the fuzzer (not a crash in the target), it could be useful to rerun it after updating the fuzzer in the container.

Let's think about what happens if we create a job from a template. Our setup container will be named `oft-setup-$x`, where `$x = hash(project, name, build, platform)`. If we change the actual fuzzer code, we should definitely change `build`, which implies the name of the container should change (at least in a template-based setting). This again could be solved by the argument invalidation idea, but now there is _dependency_: if you invalidate `build`, then the `setup` container should be invalidated, and should be re-specified. A complication: containers are really per-task, and in the general (non-template-based) case of job creation, there could be several `setup` containers. Should we only support re-running for template-based jobs?

> 3. Reproducing a bug. <...>

Agreed, like (1), this should be easy.

[...]

PR #960, opened 2021-06-03
add `onefuzz debug job rerun JOB_ID` command

bmc-msft:
This allows the specification of an alternate pool or duration when rerunning an existing job.


src/cli/onefuzz/api.py
```diff
@@ -1129,6 +1129,70 @@ def list(
...
+                        added.add(task_id)
+                else:
+                    task = self.onefuzz.tasks.create_with_config(config)
+                    self.logger.info(
+                        "created task: %s - %s",
+                        task.task_id,
+                        task.config.task.type.name,
+                    )
+                    new_task_ids[task_id] = task.task_id
+                    added.add(task_id)
```
chkeita, line 1187: At this point we  re created the current task but some of its pre requisite are not created yet. 
if we put it in the added list here we will not be revisiting it to update the pre requisite  with the new ids

chkeita (approved):

ranweiler (commented):
I suggest we discuss the larger design bit more before merging this. I've added some initial thoughts in the [issue comments](https://github.com/microsoft/onefuzz/issues/958#issuecomment-854975092).

ranweiler (approved):
This looks good to me; my only suggestion is that we ship this under the `debug` subcommand namespace so we can iterate on it before committing it to our stable API.

PR #966, opened 2021-06-07
migrate to msgraph

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/deployment/registration.py
+++ b/src/deployment/registration.py
@@ -12,6 +12,8 @@ from enum import Enum
 from typing import Any, Callable, Dict, List, NamedTuple, Optional, Tuple, TypeVar
 from uuid import UUID, uuid4
 
+# https://github.com/microsoftgraph/msgraph-sdk-python-core
+
 import requests
 from azure.cli.core.azclierror import AuthenticationError
 from azure.common.client_factory import get_client_from_cli_profile
@@ -141,39 +143,29 @@ def register_application(
     subscription_id: str,
 ) -> ApplicationInfo:
     logger.info("retrieving the application registration %s" % registration_name)
-    client = get_graph_client(subscription_id)
-    apps: List[Application] = list(
-        client.applications.list(filter="displayName eq '%s'" % registration_name)
-    )
 
-    if len(apps) == 0:
+    app = get_application(display_name=registration_name)
+
+    if not app:
         logger.info("No existing registration found. creating a new one")
         app = create_application_registration(
             onefuzz_instance_name, registration_name, approle, subscription_id
         )
     else:
-        app = apps[0]
         logger.info(
             "Found existing application objectId '%s' - appid '%s'"
-            % (app.object_id, app.app_id)
+            % (app.object_id, app["appId"] )
         )
 
-    onefuzz_apps: List[Application] = list(
-        client.applications.list(filter="displayName eq '%s'" % onefuzz_instance_name)
-    )
+    onefuzz_app = get_application(display_name=onefuzz_instance_name)
 
-    if len(onefuzz_apps) == 0:
+    if not(onefuzz_app):
         raise Exception("onefuzz app not found")
 
-    onefuzz_app = onefuzz_apps[0]
-    pre_authorized_applications = (
-        onefuzz_app.pre_authorized_applications
-        if onefuzz_app.pre_authorized_applications is not None
-        else []
-    )
+    pre_authorized_applications = onefuzz_app["apiApplication"]["preAuthorizedApplications"]
 
-    if app.app_id not in [app.app_id for app in pre_authorized_applications]:
-        authorize_application(UUID(app.app_id), UUID(onefuzz_app.app_id))
+    if app["appId"] not in [app["appId"] for app in pre_authorized_applications]:
+        authorize_application(UUID(app["appId"]), UUID(onefuzz_app["appId"]))
 
     password = create_application_credential(registration_name, subscription_id)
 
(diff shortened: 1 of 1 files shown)


chkeita:
Postponing the merge of this until linux machines are allowed to connect to Azure active directory.

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -99,18 +99,46 @@ def get_regions() -> List[Region]:
     return sorted([Region(x.name) for x in locations])
 
 
-@cached
-def get_graph_client() -> GraphRbacManagementClient:
-    return GraphRbacManagementClient(get_msi(), get_subscription())
+def query_microsoft_graph(
```
chkeita, line 102: Testing in progress
- [x] deploying from a linux machine 
- [ ] running integration tests

src/deployment/registration.py
```diff
@@ -147,140 +128,135 @@ def register_application(
     subscription_id: str,
 ) -> ApplicationInfo:
     logger.info("retrieving the application registration %s" % registration_name)
-    client = get_graph_client(subscription_id)
-    apps: List[Application] = list(
-        client.applications.list(filter="displayName eq '%s'" % registration_name)
+
+    app = get_application(
+        display_name=registration_name, subscription_id=subscription_id
```
ranweiler, line 133: Moving the search into `get_application()` makes things much more readable! We might want a `find_` or `query_` function that returns a list (see other comment), but this is a good move overall.

src/deployment/registration.py
```diff
@@ -147,140 +128,135 @@ def register_application(
...
-    registered_app: Application = client.applications.create(params)
+    }
 
-    logger.info("creating service principal")
-    service_principal_params = ServicePrincipalCreateParameters(
-        account_enabled=True,
-        app_role_assignment_required=False,
-        service_principal_type="Application",
-        app_id=registered_app.app_id,
+    registered_app: Dict = query_microsoft_graph(
```
ranweiler, line 221: Hmm, worst case, can `query_microsoft_graph()` legitimately return a `Dict`?

src/deployment/deploy.py
```diff
@@ -287,78 +286,99 @@ def setup_rbac(self) -> None:  # noqa: C901
...
-            ),
-            AppRole(
-                allowed_member_types=["Application"],
-                display_name=OnefuzzAppRole.ManagedNode.value,
-                id=str(uuid.uuid4()),
-                is_enabled=True,
-                description="Allow access from a lab machine.",
-                value=OnefuzzAppRole.ManagedNode.value,
-            ),
+            {
```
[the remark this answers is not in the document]
chkeita: Nope for now we are stuck with direct call to msbraph

src/deployment/deploy.py
```diff
@@ -287,78 +286,99 @@ def setup_rbac(self) -> None:  # noqa: C901
...
+                        }
+                    ]
+                },
+                "web": {
+                    "implicitGrantSettings": {
+                        "enableAccessTokenIssuance": False,
+                        "enableIdTokenIssuance": True,
+                    },
+                    "redirectUris": [
+                        f"{self.get_instance_url()}/.auth/login/aad/callback"
```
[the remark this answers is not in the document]
chkeita: i moved `get_identifier_url()` not because of the string interpolation but because the value depends on the configuration of the deployment. The redirect uri only depends on the instance_url

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -99,18 +103,68 @@ def get_regions() -> List[Region]:
...
+    params: Optional[Dict] = None,
+    body: Optional[Dict] = None,
+) -> List[Dict]:
+    result = query_microsoft_graph(
+        method,
+        resource,
+        params,
+        body,
+    )
+    if result["value"]:
```
ranweiler, line 156: 1. We don't truly _know_ that the key `"value"` must occur in `result`.
2. Even if does,  if `result["value"] == []`, then the conditional expr will evaluate `False`. IOW, as-written, I don't think it is possible for this query to return the empty list, even though that may be a valid response payload.
3. Ideally we should use an `isinstance` check before casting.

One solution could be to rewrite this as something like:
```suggestion
    value = result.get("value")
    if isinstance(value, list):
```
Here, the `isinstance` is also basically doing a `None`-check for us.

src/deployment/registration.py
```diff
@@ -10,78 +10,94 @@
...
-
-    response.status_code
-
     if 200 <= response.status_code < 300:
-        try:
-            return response.json()
-        except ValueError:
-            return None
+        if response.content and response.content.strip():
+            return cast(Dict, response.json())
```
ranweiler, line 57: Since a JSON document can also be an array of values, let's add a final `isinstance` check here, too.

So something like:
```suggestion
            json = response.json()
            if isinstance(json, dict):
                return cast(Dict, json)
            else:
                # return error
```

ranweiler: (if we do the `isinstance` checks, I think we don't need to cast anymore, too)

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -99,18 +103,77 @@ def get_regions() -> List[Region]:
...
-    client = get_graph_client()
-    return bool(
-        client.groups.is_member_of(
-            CheckGroupMembershipParameters(group_id=group_id, member_id=member_id)
-        ).value
+    body = {"groupIds": [group_id]}
+    response = query_microsoft_graph_list(
+        method="POST", resource=f"users/{member_id}/checkMemberGroups", body=body
     )
+    return group_id in response
```
ranweiler, line 176: One last question: is this _ever_ true? It seems like it can't be, based on the types.

- `query_microsoft_graph_list()` returns `List[Dict]`
- `group_id` has type `str`

I'm surprised it even typechecks, so I may just be misreading something?

chkeita: The expected result is a [list of strings](https://docs.microsoft.com/en-us/graph/api/user-checkmembergroups?view=graph-rest-1.0&tabs=http#response-1)

I guess the type annotation is wrong in this particular case but the logic is correct

ranweiler (commented):
Can we explicitly test (if we haven't already) and document (in the PR text) the new expected upgrade behavior if the `$ONEFUZZ_AAD_GROUP_ID` variable is set?

PR #976, opened 2021-06-09
fix service authentication refresh errors

bmc-msft:


src/agent/reqwest-retry/src/lib.rs
```diff
@@ -172,6 +172,19 @@ impl SendRetry for reqwest::RequestBuilder {
...
 
+pub fn is_auth_failure(response: &Result<Response>) -> bool {
+    if let Err(error) = response {
+        if let Some(error) = error.downcast_ref::<reqwest::Error>() {
+            if let Some(status) = error.status() {
+                if status == StatusCode::UNAUTHORIZED {
+                    return true;
+                }
+            }
+        }
```
chkeita, line 183: his only check the case where the request was converted to an error with a call to `error_for_status` most likely.
We should also check the response status code without the downcast too.

```suggestion
        if let Some(error) = error.downcast_ref::<reqwest::Error>() {
            if let Some(status) = error.status() {
                if status == StatusCode::UNAUTHORIZED {
                    return true;
                }
            }
        } else {
            return response.status() == StatusCode::UNAUTHORIZED;
        }
```

bmc-msft: For calls from reqwest-retry, always done within the context of error_for_status.

However, I see the value of adding it.  Note, it needs to be done up a level.  I've pushed an update to add this functionality.

chkeita (approved):

src/agent/reqwest-retry/src/lib.rs
```diff
@@ -172,6 +172,22 @@ impl SendRetry for reqwest::RequestBuilder {
     }
 }
 
+pub fn is_auth_failure(response: &Result<Response>) -> bool {
+    match response {
```
ranweiler, line 176:
```suggestion
    // Check both cases to support `error_for_status()`.
    match response {
```

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -310,9 +312,11 @@ impl Coordinator {
             response = request
                 .send_retry_default()
                 .await
-                .context("Coordinator.send after refreshing access token")?;
+                .context("Coordinator.send after refreshing access token");
         };
 
+        let response = response?;
```
ranweiler, line 318: Might be good to have something like:
```suggestion
        let response = response.context("non-status error after ensuring valid access token")?;
```

ranweiler (approved):

PR #978, opened 2021-06-09
Retry coverage recording

### LINKED ISSUE #975: Only fail coverage task for truly fatal tracing errors
When recording coverage, a tracee may occasionally die or exit unexpectedly, for reasons outside of our control. We should permit this to some extent, probably retrying, and failing the OneFuzz task only after some threshold has been reached.

ranweiler:
Closes #975.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -13,7 +13,7 @@ use coverage::cache::ModuleCache;
 use coverage::code::{CmdFilter, CmdFilterDef};
 use onefuzz::expand::Expand;
 use onefuzz::syncdir::SyncedDir;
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -188,6 +189,28 @@ impl<'a> TaskContext<'a> {
     }
 
     pub async fn record_input(&mut self, input: &Path) -> Result<()> {
+        let attempts = MAX_COVERAGE_RECORDING_ATTEMPTS;
+
+        for _ in 0..attempts {
+            if let Err(err) = self.try_record_input(input).await {
```
bmc-msft, line 195: Can we keep the result here, and just add context stating it failed after N attempts as this functions Result::Err?

bmc-msft (approved):

PR #984, opened 2021-06-11
Check scaleset size for missing nodes

bmc-msft:


src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -449,6 +449,29 @@ def _resize_shrink(self, to_remove: int) -> None:
         for node in nodes:
             node.send_stop_if_free()
 
+    def check_scaleset_size(self) -> None:
```
ranweiler, line 452: Can we add a doc comment summarizing the responsibilities of this function?

Also, I find the name a bit confusing (`check` makes me think of a read-only predicate or guard function, but this does a lot of mutating). How about something like `sync_scaleset_size()`?

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -449,6 +449,29 @@ def _resize_shrink(self, to_remove: int) -> None:
...
+        if self.state != ScalesetState.running:
+            return
+
+        size = get_vmss_size(self.scaleset_id)
+        if size is None:
+            logging.info(
+                SCALESET_LOG_PREFIX + "scaleset is unavailable. scaleset_id:%s",
+                self.scaleset_id,
+            )
+            self.set_shutdown(now=True)
```
ranweiler, line 462: If the scaleset is unavailable, why does it make sense to `set_shutdown()`?

ranweiler (approved):

src/api-service/__app__/onefuzzlib/workers/scalesets.py
bmc-msft: The scaleset shutdown code handles cleaning up all of the resources for the scaleset.  This includes deleting nodes, stopping tasks on said nodes, deleting scaleset specific storage queues, etc.

ranweiler: So is this just an optimization? If we don't call this here, what happens?

bmc-msft: The "Azure says this scaleset doesn't exist" error handling is to move to shutting it down.  I copied the comment from the other places we handle this same error condition.

Issue #981, opened 2021-06-10
github issue notification in templates not working


bmc-msft:
There is an issue using the GitHub notification integration from job templates.

This command:
```
onefuzz template libfuzzer A B C linux --notification_config @~/onefuzz/contrib/onefuzz-job-github-actions/github-issues.json
```
results in this error:
```
ERROR:cli:command failed: request did not succeed: HTTP 400 - {"code": 450, "errors": ["config.base_url: field required", "config.auth_token: field required", "config.project: field required", "config.type: field required", "config.unique_fields: field required", "config.ado_fields: field required", "config.on_duplicate.increment: field required", "config.on_duplicate.set_state: field required", "config.on_duplicate.ado_fields: field required", "config.url: field required", "config.auth.url: field required"]}
```

Note, these are the configuration errors for Azure Devops notifications, not GitHub.  Both should be usable here.

PR #985, opened 2021-06-11
handle serialization of secrets sent from the CLI

bmc-msft:
This normalizes the SecretData serialization from the client to address #981.

When serializing objects sent to the service with secrets, we would turn it into a SecretData

We use SecretData to convert this:
`{"auth": {"user": "A", "personal_access_token": "B"}}`
to this:
`"auth": { "secret": { "url": "https://KEYVAULT-URL" }}`

Currently, in the case we have a SecretData we've not yet saved, the serialized form looks like this:
`{"auth": { "secret": {"user": "A", "personal_access_token": "B"}}}`

This PR simplifies the client side serialization to this:
`{"auth": {"user": "A", "personal_access_token": "B"}}`

### DIFF
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -379,7 +379,6 @@ def serialize(data: Any) -> Any:
     if data is None:
         return data
     if isinstance(data, SecretData) and not isinstance(data.secret, SecretAddress):
-        print(type(data.secret))
         return serialize(data.secret)
     if isinstance(data, BaseModel):
         return {serialize(a): serialize(b) for (a, b) in data.dict().items()}
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -30,6 +30,7 @@ from uuid import UUID
 import msal
 import requests
 from azure.storage.blob import ContainerClient
+from onefuzztypes.models import SecretAddress, SecretData
 from pydantic import BaseModel, Field
 from tenacity import Future as tenacity_future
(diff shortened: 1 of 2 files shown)


src/cli/onefuzz/backend.py
```diff
@@ -377,6 +378,8 @@ def container_file_path(container_url: str, blob_name: str) -> str:
 def serialize(data: Any) -> Any:
     if data is None:
         return data
+    if isinstance(data, SecretData) and not isinstance(data.secret, SecretAddress):
```
ranweiler, line 381: Can we add some unit tests, to automate your example in the PR message?

ranweiler (approved):
This issue suggests `SecretData` should be more misuse-resistant. Is the deeper fix to also distinguish (at the type level) between saved and unsaved `SecretData`? It seems risky that the difference between serialization as plaintext and a gated reference is just whether or not a model has been saved.

src/cli/onefuzz/backend.py
bmc-msft: > This issue suggests `SecretData` should be more misuse-resistant. Is the deeper fix to also distinguish (at the type level) between saved and unsaved `SecretData`? It seems risky that the difference between serialization as plaintext and a gated reference is just whether or not a model has been saved.

I agree with the concern that `SecretData` needs work.  However, I am concern that will be a significantly larger issue.

Note, this is serialization used in the CLI backend, not SecretData in general.

I added a unittest that validates the functionality.

PR #990, opened 2021-06-14
fix GithubIssueTemplate deserialization 

chkeita:
Revert change made in #985 
fixed  GithubIssueTemplate deserialization


chkeita:
> ADOTemplate and TeamsTemplate are likely impacted by this issue as well for the same reasons. Is it possible to make this a fix within the SecretData implementation rather than one that needs to be replicated for each implementation?

I added two tests for `ADOTemplate` and `TeamsTemplate`. They were no affected by the issue .

I could not put the fix in `SecretData` because it is a generic type and we need to have access to the Generic argument at runtime to do the deserialization.

src/api-service/tests/test_secrets.py
```diff
@@ -91,3 +93,43 @@ def test_read_secret(self) -> None:
...
+            c = NotificationCreate.parse_obj(b)
+            d = c.json()
+            e = json.loads(d)
+            NotificationCreate.parse_obj(e)
+
+    def test_roundtrip_team_issue(self) -> None:
+        a = """
+            {
+                "config" : {"url": "http://test"},
+                "container": "bmc"
```
bmc-msft, line 114:
```suggestion
                "container": "testing"
```

src/api-service/tests/test_secrets.py
```diff
@@ -91,3 +93,43 @@ def test_read_secret(self) -> None:
...
+        NotificationCreate.parse_obj(e)
+
+    def test_roundtrip_ado(self) -> None:
+        current_path = pathlib.Path(__file__).parent.absolute()
+        with open(
+            f"{current_path}"
+            + "/../../../contrib/onefuzz-job-azure-devops-pipeline/ado-work-items.json"  # noqa
+        ) as json_file:
+            b = json.load(json_file)
+            b["container"] = "bmc"
```
bmc-msft, line 131:
```suggestion
            b["container"] = "testing"
```

src/api-service/tests/test_secrets.py
```diff
@@ -91,3 +93,43 @@ def test_read_secret(self) -> None:
...
             self.fail(f"Invalid config type {type(notification.config)}")
+
+    def test_roundtrip_github_issue(self) -> None:
+        current_path = pathlib.Path(__file__).parent.absolute()
+        with open(
+            f"{current_path}"
+            + "/../../../contrib/onefuzz-job-github-actions/github-issues.json"
+        ) as json_file:
+            b = json.load(json_file)
+            b["container"] = "bmc"
```
bmc-msft, line 104:
```suggestion
            b["container"] = "testing"
```

bmc-msft (approved):

PR #992, opened 2021-06-15
updates from upstream libclusterfuzz

bmc-msft:


ranweiler (approved):

ranweiler (commented):
In the PR or commit message, could you just note what upstream commit you are pulling these from?

bmc-msft:
This merges the results of running the update tool.   The tool always pull the latest from `main`.

https://github.com/microsoft/onefuzz/tree/main/src/agent/libclusterfuzz/third-party

ranweiler:
> This merges the results of running the update tool. The tool always pull the latest from main.

Understood, my suggestion is that it may be handy to have an explicit history of what commit `clusterfuzz` was at whenever we last updated.

In this case, it looks like: https://github.com/google/clusterfuzz/commit/6f27c1d64167838d39162fb82afa3427141d30fb

PR #995, opened 2021-06-16
add context to errors in syncdir

bmc-msft:


src/agent/onefuzz/src/blob/url.rs
```diff
@@ -151,7 +151,8 @@ impl BlobContainerUrl {
     pub fn url(&self) -> Result<Url> {
         match self {
             Self::BlobContainer(url) => Ok(url.clone()),
-            Self::Path(p) => Ok(Url::from_file_path(p).map_err(|_| anyhow!("invalid path"))?),
+            Self::Path(p) => Ok(Url::from_file_path(p)
```
ranweiler, line 154: WDYT about changing the existing `map_err()` calls to uses of `Context`?

ranweiler (approved):

src/agent/onefuzz/src/blob/url.rs
bmc-msft: `Anyhow::Context` doesn't support the results from `Url::from_file_path`, currently a `Result<Url, ()>`.

PR #996, opened 2021-06-16
fix SyncedDir.remote_url issue with relative paths

bmc-msft:
Url::from_file_path does not support creating `Url` with a relative
path.  This change converts relative paths to canonicalized absolute
paths based on CWD.


chkeita (approved):

ranweiler (approved):
Could we add some unit tests to exercise this?

ranweiler (approved):

PR #1000, opened 2021-06-16
Explicitly track debuggee thread state to detect thread exit/debug event races

ranweiler:
- Define an enum to track the debugger's current understanding of debuggee thread state
- Update our private suspend/resume methods to update and return the current state
- Detect thread exit/debug event races in suspend/resume calls

The win32 `SuspendThread()`/`ResumeThread()` calls increment/decrement a per-thread suspend counter. When the counter is 0, the thread is runnable, else it is suspended.

The win32 functions return the _previous_ suspend count, and `-1` on error. The full set of error descriptions is not part of the public API, but can definitely be caused by a lack of [thread access rights](https://docs.microsoft.com/en-us/windows/win32/procthread/thread-security-and-access-rights), or an exited thread.

We always have the `THREAD_SUSPEND_RESUME` access right for any `ThreadInfo` thread, because we observe it as a debugger of the thread's process. If we get an "access denied" error, then this means that the thread has exited during our current debugger loop operation, but our debugger has not _yet_ been delivered the relevant debug event.

This PR does not _fix_ our handling of racy thread exits, but updates our modeling of debuggee threads to _enable_ robust handling of threads when performing multi-threaded single-stepping.


src/agent/debugger/src/target.rs
```diff
@@ -31,55 +31,88 @@ pub(crate) enum StepState {
     SingleStep,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum ThreadState {
+    Running,
```
ranweiler, line 36: This is misleading; rename to `Runnable`.

stishkin (approved):

bmc-msft (approved):

ranweiler:
Manually testing.

ranweiler:
Manually tested coverage recording and SARIF generation.

ranweiler:
Manual testing complete.

PR #941, opened 2021-05-29
libfuzzer fuzzing task perf improvements

chkeita:
- resuse the regex to parse the output of libfuzzer
- added a cancellation notification to report_fuzzer_sys_info.
   ~~The code seems to be actively waiting this function and consuming some cpu time~~
   The notification allows us to reduce the time waiting for the fuzzing loop to terminate. 

## Summary of the Pull Request

_What is this about?_

## PR Checklist
* [ ] Applies to work item: #xxx
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.
* [ ] Tests added/passed
* [ ] Requires documentation to be updated
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

_How does someone test & validate?_

PR #1002, opened 2021-06-18
add cooloff period for rapidly exiting libFuzzer targets

bmc-msft:
For rapidly exiting libFuzzer targets, the agent can get into a state where it never has a chance for the handles to get reaped.  This eventually causes an OOM error in the system.

This enforces that within a given worker, at least 5 seconds have occurred between each execution of the target.

Note, we effectively had a sleep before, but the performance improvements in #941 removed the effective sleep.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
+++ b/src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
@@ -24,7 +24,7 @@ use tokio::{
     select,
     sync::{mpsc, Notify},
     task,
-    time::{sleep, Duration},
+    time::{sleep, Duration, Instant},
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-agent/src/tasks/fuzz/libfuzzer_fuzz.rs
```diff
@@ -181,6 +184,12 @@ impl LibFuzzerFuzzTask {
                         )
                     })?;
             }
+
+            // if libFuzzer is exiting rapidly, give some breathing room to allow the
+            // handles to be reaped.
+            if instant.elapsed() < COOLOFF_PERIOD {
+                sleep(COOLOFF_PERIOD).await;
+            }
```
chkeita, line 192: The sleep should be for the remaining time instead of the whole cooloff period
```Suggestion
            let remaining_time = instant.elapsed() - COOLOFF_PERIOD; 
            if  remaining_time < 0 {
                sleep(remaining_time).await;
            }
```

bmc-msft: Duration will panic if you try to create a negative duration.  I implemented the idea though.

chkeita (approved):

PR #1005, opened 2021-06-18
remove nodes from db upon reimage

bmc-msft:
The flag `Node.reimage_queued` is intended to stop nodes from reimaging repeatedly.  

In #970, in order to work around Azure API failures, this flag was cycled if the node was already set to cleanup.  Unfortunately, reimaging can sometimes take a significant amount of time, causing this change to get nodes multiple times.

Instead of using `reimage_queued` as a flag, this PR deletes the node from the storage table upon reimage.  When the node registers OR the next time through `Scaleset.cleanup_nodes`, the Node will be recreated automatically, whichever comes first.


src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -551,6 +545,9 @@ def delete_nodes(self, nodes: List[Node]) -> None:
             machine_ids,
         )
         delete_vmss_nodes(self.scaleset_id, machine_ids)
+        for node in nodes:
+            if node.machine_id in machine_ids:
+                node.delete()
```
ranweiler, line 550: Can we just iterate over `machine_ids` directly? By construction, its IDs are a subset of those in `nodes`, right? If there's a reason to iterate over `nodes` instead, can you document it?

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -551,6 +545,9 @@ def delete_nodes(self, nodes: List[Node]) -> None:
             machine_ids,
         )
         delete_vmss_nodes(self.scaleset_id, machine_ids)
```
ranweiler, line 547: So this begins an async deletion operation, and internally gets access to an `LROPoller`. We ignore that, and return `Option[Error]` In practice, it looks like we always return `None`.

Does the initial state of the poller contain data we should be examining? If deletion fails right away, how do we detect it?

ranweiler: (or if deletion fails later, when do we detect it?)

ranweiler (commented):
IIUC, this moves us from a more explicit model (where we flag reimage-queued nodes) to an implicit one, where we rely on some expected future behavior to cause the right end result. I find the correctness of this very hard to reason about.

Is there another variation we could use? Would it be better to add _more_ state to `reimage_request`, e.g. some form of counter, or a new state like `reimage_finishing`?

src/api-service/__app__/onefuzzlib/workers/scalesets.py
bmc-msft: No?  In order to delete the `Node` object, we need the `Node` instance.

`nodes` is of type `Node`.  machine_ids is `List[UUID]`.

We could iterate the other way around, but it would be less efficient.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
bmc-msft: To answer your question about the LROPoller, the initial state of the poller isn't immediately useful.  The request itself fails if it's immediately bad.  I've already started investigating using the LROPoller.from_continuation_token to reconstruct the poller, but that's a relatively large change.

With regarding recovery from Azure failures, it will take some time, but we'll eventually recover.

* For reimaging, `cleanup_nodes` will recreate it, the heartbeat will eventually be too long expired, and we'll reimage it.
* For deletion, `cleanup_nodes` will recreate it and we'll reimage it like above, but now `sync_scaleset_size` will eventually trigger deletion of a node.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
ranweiler: Ah, right right. Mental typechecker failure.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
ranweiler: Makes sense, and the deeper fix to the eventing complexity is out of scope here.

ranweiler (approved):

PR #1008, opened 2021-06-21
Fix issue when deserializing GithubIssueTemplate with hidden secret

chkeita:
Fixed issue when deserializing GithubIssueTemplate with hidden secret
Updated unit tests to check the expected data type for SecretData


src/api-service/tests/test_secrets.py
```diff
@@ -100,12 +100,40 @@ def test_roundtrip_github_issue(self) -> None:
...
+            notification_dict = json.load(json_file)
+            notification_dict["container"] = "testing"
+            notification1 = NotificationCreate.parse_obj(notification_dict)
+            self.assertIsInstance(
+                notification1.config, GithubIssueTemplate, "invalid config type"
+            )
+            notification1_config = cast(GithubIssueTemplate, notification1.config)
+            self.assertIsInstance(
+                notification1_config.auth.secret, GithubAuth, "Invalid secret type"
+            )
```
bmc-msft, line 112: I am concerned that `cast` will hide errors from mypy in the future.

Using `assertIsInstance` makes the unit test output better and is the better choice in most cases, but using a raw `assert` here enables `mypy` to verify the data structures for us without `cast`.
 
```suggestion
            notification1 = NotificationCreate.parse_obj(notification_dict)
            assert isinstance(notification1.config, GithubIssueTemplate)
            self.assertIsInstance(
                notification1.config.auth.secret, GithubAuth, "Invalid secret type"
            )
```

bmc-msft (changes requested):
see feedback regarding `cast`

bmc-msft:
Has this change been verified to create github issues end-to-end?

chkeita:
> Has this change been verified to create github issues end-to-end?

yes i was able to successfully create github issues with these changes

bmc-msft (approved):

PR #1009, opened 2021-06-22
update third-party Rust dependencies

bmc-msft:


bmc-msft:
Note: 

`goblin::pe::utils::find_offset` now takes a new param `opts`, which is a `ParseOptions`.  The default value makes `find_offset` work as it currently does.

Ref: https://docs.rs/goblin/0.4.1/src/goblin/pe/utils.rs.html#62-92
Ref: https://docs.rs/goblin/0.4.1/goblin/pe/options/struct.ParseOptions.html#structfield.resolve_rva

ranweiler (approved):

PR #1010, opened 2021-06-22
instance wide configuration

bmc-msft:
TODO:
* [x] add setting initial set of admins during deployment


src/api-service/__app__/instance_config/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+from ..onefuzzlib.events import get_events
+from ..onefuzzlib.request import not_ok, ok, parse_request
+from ..onefuzzlib.user_credentials import parse_jwt_token
+
+
+def get(req: func.HttpRequest) -> func.HttpResponse:
+    return ok(InstanceConfig.fetch())
+
+
+def post(req: func.HttpRequest) -> func.HttpResponse:
```
ranweiler, line 22: I suggest we refactor this to do all request parsing up front, and only then invoke a separate function that acts on the results. We could then unit test some different scenarios.

src/api-service/__app__/instance_config/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    if isinstance(request, Error):
+        return not_ok(request, context="instance_config_update")
+
+    config = InstanceConfig.fetch()
+
+    user_info = parse_jwt_token(req)
+    if isinstance(user_info, Error):
+        return not_ok(user_info, context="missing user info")
+
+    if config.admins:
```
ranweiler, line 33: IIUC, we permit an initial deployment scenario where there are _no_ explicit admins, and thus every authenticated CLI user is a de facto admin (the status quo). But now, can't any one of these users submit an instance config update setting (only) themselves as admin?

I'd like to see a much stricter separation between the two scenarios. Ideally, we should only allow the explicitly-allowed admin case, and require that it be seeded by default and/or out of band.

ranweiler (commented):
Still have some open questions on what we're implementing here. I think it will be important to add some detailed documentation on the intended implementation when we merge this.

src/pytypes/onefuzztypes/models.py
```diff
@@ -856,4 +856,9 @@ class Task(BaseModel):
     user_info: Optional[UserInfo]
 
 
+class InstanceConfig(BaseModel):
+    admins: Optional[List[UUID]] = None
```
ranweiler, line 860: Could we add comments documenting the meaning of these fields? The most important is what it means to be an admin, but also whether pool `modification` encompasses creation and deletion.

src/cli/onefuzz/api.py
```diff
@@ -1557,6 +1557,20 @@ def list(self) -> responses.ProxyList:
         return self._req_model("GET", responses.ProxyList, data=requests.ProxyGet())
 
 
+class InstanceConfigCmd(Endpoint):
```
ranweiler, line 1560: What does the error message look like when this command fails?

src/api-service/__app__/pool/__init__.py
```diff
@@ -78,6 +78,15 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if isinstance(request, Error):
         return not_ok(request, context="PoolCreate")
 
+    if not can_modify_pools(req):
```
ranweiler, line 81: Can we pull this check out into `main()`? This and `call_if_user` chould be part of a common `call_if_authorized` function.

src/api-service/__app__/onefuzzlib/notifications/ado.py
```diff
@@ -232,13 +233,12 @@ def notify_ado(
...
         ado = ADO(container, filename, config, report)
         ado.process()
-    except AzureDevOpsAuthenticationError as err:
-        fail_task(report, err)
-    except AzureDevOpsClientError as err:
-        fail_task(report, err)
-    except AzureDevOpsServiceError as err:
-        fail_task(report, err)
-    except AzureDevOpsClientRequestError as err:
+    except (
```
ranweiler, line 236: 👍

src/api-service/__app__/onefuzzlib/notifications/ado.py
```diff
@@ -201,6 +201,7 @@ def process(self) -> None:
         for work_item in self.existing_work_items():
             self.update_existing(work_item)
             seen = True
+            break
```
ranweiler, line 204: Is this a stray change from other WIP?

Doesn't this `break` prevent this function from updating more than one work item?

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,38 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    if config.admins is None:
+        return True
+
+    return user_info.object_id in config.admins
+
+
+def can_modify_pools(req: func.HttpRequest) -> bool:
+    user_info = parse_jwt_token(req)
+    if not isinstance(user_info, UserInfo):
+        False
```
ranweiler, line 64: Should be a `return False`, and that's why you need the `assert` below?

ranweiler: (apply above)

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,38 @@ def is_agent(token_data: UserInfo) -> bool:
     return False
 
 
+def can_modify_config(req: func.HttpRequest, config: InstanceConfig) -> bool:
```
ranweiler, line 47: It seems like this and `can_modify_pools` should just accept a `UserInfo` instead off an `HttpRequest`.

src/api-service/__app__/instance_config/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+    config = InstanceConfig.fetch()
+
+    if not can_modify_config(req, config):
+        return not_ok(
+            Error(code=ErrorCode.INVALID_PERMISSION, errors=["unauthorized"]),
+            context="instance_config_update",
+        )
+
+    for field in config.__fields__:
```
ranweiler, line 34: Can we document (and ideally, unit test) this update logic? Perhaps as a method on the `InstanceConfig`?

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,38 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    # make mypy happy
+    assert isinstance(user_info, UserInfo)
+
+    if config.admins is None:
+        return True
+
+    return user_info.object_id in config.admins
+
+
+def can_modify_pools(req: func.HttpRequest) -> bool:
```
chkeita, line 61: can we be specific here and distinguish between "no admins specified" and "The user is not an authorized admin". This will allows us give a more meaningful message when the access fail.
maybe if we return a `Union[bool, AuthtenticationError]`

src/api-service/__app__/onefuzzlib/config.py
```diff
@@ -0,0 +1,34 @@
...
+from onefuzztypes.events import EventInstanceConfigUpdated
+from onefuzztypes.models import InstanceConfig as BASE_CONFIG
+from pydantic import Field
+
+from .azure.creds import get_instance_name
+from .events import send_event
+from .orm import ORMMixin
+
+
+class InstanceConfig(BASE_CONFIG, ORMMixin):
```
chkeita, line 17: I think the PR is missing the initilization of the admins in the deployment script. Unless i missed that logic somewhere else

src/api-service/__app__/onefuzzlib/notifications/ado.py
bmc-msft: This wasn't meant to be included in this PR.  backing out.

src/api-service/__app__/onefuzzlib/notifications/ado.py
bmc-msft: This wasn't meant to be included in this PR.  backing out.

src/api-service/__app__/onefuzzlib/config.py
bmc-msft: You didn't miss it.  It's listed in the PR's todo list.

src/cli/onefuzz/api.py
bmc-msft: The same way all errors look at the CLI.

We generate a 400 error.  The JSON is printed.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
bmc-msft: The 4 callers do not have a `UserInfo`, but they do have a `HttpRequest`.  Instead of having all 4 callers call parse_jwt_token, do it once here.

src/api-service/__app__/pool/__init__.py
bmc-msft: While I think that level of reorg should happen, IMO, that should be done as part of the larger RBAC feature we've discussed.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
bmc-msft: addressed.

src/api-service/__app__/onefuzzlib/config.py
bmc-msft: added

src/api-service/__app__/pool/__init__.py
ranweiler: That seems sensible.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
ranweiler: I see your point about factoring out the JWT inspection. I do think it is being factored into the wrong place; it should go up the stack, closer to the HTTP/auth layer.

That said, if you want to keep this for practical reasons, then that's fine. My suggestion here is still to change the signature, and require the callers to invoke `parse_jwt_token()`, even though it is currently repetitive. I think it will make it easer to do the right thing in later work. Your call.

src/api-service/__app__/instance_config/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+from onefuzztypes.requests import InstanceConfigUpdate
+
+from ..onefuzzlib.config import InstanceConfig
+from ..onefuzzlib.endpoint_authorization import call_if_user, can_modify_config
+from ..onefuzzlib.events import get_events
+from ..onefuzzlib.request import not_ok, ok, parse_request
+
+
+def get(req: func.HttpRequest) -> func.HttpResponse:
+    return ok(InstanceConfig.fetch())
```
ranweiler, line 18: Worth explicitly considering: should non-admins be able to query the list of admins? As implemented, they can, right?

src/pytypes/onefuzztypes/models.py
```diff
@@ -856,4 +856,28 @@ class Task(BaseModel):
...
+            if field == "admins" and self.admins is None:
+                continue
+
+            if hasattr(self, field):
+                setattr(self, field, getattr(config, field))
+
+    @validator("admins", allow_reuse=True)
+    def check_admins(cls, value: Optional[List[UUID]]) -> Optional[List[UUID]]:
+        if value is not None and len(value) == 0:
+            raise ValueError("admins must be None or at least UUID")
```
ranweiler, line 879:
```suggestion
            raise ValueError("admins must be None or contain at least one UUID")
```

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,35 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    user_info = parse_jwt_token(req)
+    if isinstance(user_info, Error):
+        return user_info
+
+    config = InstanceConfig.fetch()
+    if config.allow_pool_management:
+        return None
+
+    if config.admins is None:
+        return Error(code=ErrorCode.UNAUTHORIZED, errors=["pool modification disabled"])
```
ranweiler, line 68: If no `admins` are set, then any user should be able to manage pools, right?

So this should be:
```suggestion
        return None
```
We should probably hoist this above the `allow_pool_management` check, too, since that field is only sensible for a config with at least one admin, right?

src/pytypes/onefuzztypes/models.py
```diff
@@ -856,4 +856,28 @@ class Task(BaseModel):
     user_info: Optional[UserInfo]
 
 
+class InstanceConfig(BaseModel):
+    # initial set of admins can only be set during deployment.
+    # if admins are set, only admins can update instance configs.
+    admins: Optional[List[UUID]] = None
+
+    # if set, only admins can manage pools or scalesets
+    allow_pool_management: bool = Field(default=True)
```
ranweiler, line 865: I like the name you landed on.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,35 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    if not isinstance(user_info, UserInfo):
+        return False
+
+    if config.admins is None:
+        return True
+
+    return user_info.object_id in config.admins
+
+
+def check_can_manage_pools(req: func.HttpRequest) -> Optional[Error]:
```
ranweiler, line 58: I think we need to unit test this and `can_modify()`.

src/api-service/__app__/instance_config/__init__.py
bmc-msft: It was done here on purpose.  I find it exceedingly frustrating when I can't find out who owns a thing I am using.

ranweiler: For sure (though "admin" here doesn't imply "owner").

I think this is fine, I just want us to be explicit about our model: we assume all authenticated users (admin or not) can enumerate other users. We don't offer APIs for this purpose, but APIs we provide can be used to do this. Before this PR, it was already possible on some retention horizon, via log queries. Now, it continues to be possible, with this other channel (which discloses admins only). None of this violates any of our existing confidentiality assumptions.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
bmc-msft: My intent was to soft-disable managing pools.

ranweiler: Ah, by "soft-disable" do you mean in a way analogous to admin-overridable GitHub branch protections?

I.e., you want to support this scenario:
> I have a OneFuzz instance with no admin allow-list, so **all** users have admin rights. However, as a footgun avoidance _convention_, I want to disable pool management by default. This convention expects users to temporarily enable pool management, make a pool change, then disable it again.

bmc-msft: yes.

chkeita (approved):

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
ranweiler: I think it would be helpful to add a comment about that intent.

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,46 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    user_info = parse_jwt_token(req)
+    if isinstance(user_info, Error):
+        return user_info
+
+    # When there are no admins in the `admins` list, all users are considered
```
ranweiler, line 63: This is excellent, thank you!

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,46 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    # admins.  However, allow_pool_modification is still useful to protect from
```
ranweiler, line 64:
```suggestion
    # admins.  However, `allow_pool_management` is still useful to protect from
```

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,46 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    # mistakes.
+    #
+    # To make changes while still protecting against accidental changes to
+    # pools, do the follwoing:
```
ranweiler, line 68:
```suggestion
    # pools, do the following:
```

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,46 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    #
+    # 1. set allow_pool_management to True
```
ranweiler, line 70:
```suggestion
    # 1. set `allow_pool_management` to `True`
```

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
```diff
@@ -43,6 +44,46 @@ def is_agent(token_data: UserInfo) -> bool:
...
+    # 2. make the change
+    # 3. set allow_pool_management to False
```
ranweiler, line 72:
```suggestion
    # 3. set `allow_pool_management` to `False`
```

src/api-service/__app__/onefuzzlib/endpoint_authorization.py
ranweiler: @bmc-msft, to be more specific: I suggest some tests that check the relationship between the `InstanceConfig` and `allow_pool_management`.

One way we could do this is have these top-level functions extract the user from the `HttpRequeset`, and then pass it to instance methods on `InstanceConfig`. These could be predicates that return `bool` based on whether or not a user is an admin generally, and if `allow_pool_management` is set, and should be easy to test in isolation.

ranweiler (approved):

PR #1011, opened 2021-06-22
fix debug report

chkeita:
The debug report created by the command `onefuzz debug  notification job <job id>`  is causing a crash in the regression task 
```
error running task: libfuzzer regression

Caused by:
    0: handling crash reports
    1: unable to parse crash report: fake-crash-sample.json
    2: unable to parse report: task_unique_reports_2/fake-crash-sample.json - "{\"input_url\": null, \"input_blob\": {\"account\": \"fuzz27ee6imdmr5gy\", \"container\": \"oft-crashes-cecbd958a1f257688f9768edaaf6c94d\", \"name\": \"fake-crash-sample\"}, \"executable\": \"fuzz.exe\", \"crash_type\": \"fake crash report\", \"crash_site\": \"fake crash site\", \"call_stack\": [\"#0 fake\", \"#1 call\", \"#2 stack\"], \"call_stack_sha256\": \"0000000000000000000000000000000000000000000000000000000000000000\", \"input_sha256\": \"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\", \"asan_log\": \"fake asan log\", \"task_id\": \"b1107de0-c3cb-43ff-ab68-5accc579f4d4\", \"job_id\": \"afa45e3e-9a75-4a47-8d59-ef3154599fc7\", \"scariness_score\": null, \"scariness_description\": null, \"minimized_stack\": null, \"minimized_stack_sha256\": null, \"minimized_stack_function_names\": null, \"minimized_stack_function_names_sha256\": null}"
```

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
+++ b/src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
@@ -32,15 +32,12 @@ pub struct CrashReport {
     pub call_stack: Vec<String>,
     pub call_stack_sha256: String,
 
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    #[serde(deserialize_with = "deserialize_null_default")]
-    pub minimized_stack: Vec<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub minimized_stack: Option<Vec<String>>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub minimized_stack_sha256: Option<String>,
-
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    #[serde(deserialize_with = "deserialize_null_default")]
-    pub minimized_stack_function_names: Vec<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub minimized_stack_function_names: Option<Vec<String>>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub minimized_stack_function_names_sha256: Option<String>,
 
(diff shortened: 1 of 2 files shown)


bmc-msft:
I think this should be fixed in the agent, not the debug command.

This means we have existing reports that could break the regression task.

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -33,11 +33,13 @@ pub struct CrashReport {
     pub call_stack_sha256: String,
 
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    #[serde(deserialize_with = "deserialize_null_default")]
```
bmc-msft, line 36: Instead of a custom `deserialize_with`, these could be optional which matches the python side:

https://github.com/microsoft/onefuzz/blob/97cba8eda1caf25c0ed47bc11be13dcccc75451c/src/pytypes/onefuzztypes/models.py#L256-L259

chkeita: I implemented the changes to match the python side and other configs.
but I personally think that having `Option<Vec<string>>` is redundant since since `None` and `Some(vec![])` can be used to represent the absence of a list.

src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -11,7 +11,7 @@ use onefuzz_telemetry::{
     },
     EventData,
 };
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
```
bmc-msft, line 14:
```suggestion
use serde::{Deserialize, Serialize};
```

bmc-msft (approved):

PR #1012, opened 2021-06-23
update azure-keyvault-secrets to 4.3.0

bmc-msft:
Note, this also removes azure-keyvault and azure-keyvault-keys which are
not used at this time.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/secrets.py
+++ b/src/api-service/__app__/onefuzzlib/secrets.py
@@ -5,7 +5,7 @@
 
 
 import os
-from typing import Tuple, Type, TypeVar, cast
+from typing import Tuple, Type, TypeVar
 from urllib.parse import urlparse
 from uuid import uuid4
 
@@ -38,7 +38,7 @@ def save_to_keyvault(secret_data: SecretData) -> SecretData:
 def get_secret_string_value(self: SecretData[str]) -> str:
     if isinstance(self.secret, SecretAddress):
         secret = get_secret(self.secret.url)
-        return cast(str, secret.value)
+        return secret.value
     else:
         return self.secret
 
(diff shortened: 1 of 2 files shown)


bmc-msft:
Note, the upstream typing signature for `begin_delete_secret` appears to be wrong.

https://github.com/Azure/azure-sdk-for-python/issues/19411

However, as this does not depend on the result, this PR removes the call to `wait`.

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -79,4 +79,4 @@ def get_secret_obj(secret_url: str, model: Type[A]) -> A:
 def delete_secret(secret_url: str) -> None:
     (vault_url, secret_name) = parse_secret_url(secret_url)
     keyvault_client = get_keyvault_client(vault_url)
-    keyvault_client.begin_delete_secret(secret_name).wait()
```
ranweiler, line 82: Doesn't this change the semantics of this function? Are we just removing the `wait()` for the sake of the bad type signature?

ranweiler: If we really don't care that we blocked until deletion completed, I'd suggest renaming this method and documenting that it only starts deletion, and does not allow callers to determine when deletion has actually completed.

The main thing that isn't clear to me is if the deletion operation can ever fail after the Azure SDK has given us an `LROPoller`.

bmc-msft: This does not change the semantics of the function.  Previously, this was a blocking operation without ever checking the result.   

 `result()` needs to be called to get the result.

ranweiler: Understood re `result()`, but I guess my point is that before, after calling this, our deletion request either succeeded or failed (though we couldn't tell which). Now neither may have happened. I agree that from a caller perspective, this difference probably isn't actionable, due to the lack of any result value or exception.

I still think it would be prudent to rename this to emphasize that this just _attempts_ an operation. I'm happy that the uncertainty is at least more obvious in the implementation now, though.

ranweiler (approved):

PR #1016, opened 2021-06-23
initial support for running the agent locally on OSX

bmc-msft:
This adds basic support to run the local agent on OS X.

The primary feature not supported on OS X is coverage collection.


src/ci/agent.sh
```diff
@@ -41,8 +41,11 @@ cargo fmt -- --check
 # RUSTSEC-2021-0065: a dependency anymap is no longer maintained
 cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2020-0016 --ignore RUSTSEC-2020-0036 --ignore RUSTSEC-2019-0036 --ignore RUSTSEC-2021-0065
 cargo-license -j > data/licenses.json
+
+# on OSX, we only build the agent for now
```
bmc-msft, line 45:
```suggestion
```

src/ci/agent.sh
```diff
@@ -41,8 +41,11 @@ cargo fmt -- --check
...
 cargo build --release --locked
 cargo clippy --release -- -D warnings
+
```
bmc-msft, line 48:
```suggestion
```

src/agent/coverage/examples/coverage.rs
```diff
@@ -94,6 +94,16 @@ fn input_command(argv: &[String], input: &Path) -> Command {
...
 }
 
+#[cfg(target_os = "macos")]
+fn record(
+    _cache: &mut ModuleCache,
+    _filter: CmdFilter,
+    _cmd: Command,
+    _timeout: Duration,
+) -> Result<Coverage> {
+    unimplemented!("coverage recording is not supported on osx");
```
ranweiler, line 104:
```suggestion
    unimplemented!("coverage recording is not supported on macOS");
```
We should use this naming throughout.

src/agent/onefuzz/Cargo.toml
```diff
@@ -44,14 +44,16 @@ onefuzz-telemetry = { path = "../onefuzz-telemetry"}
...
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(target_family = "unix")'.dependencies]
 cpp_demangle = "0.3"
 nix = "0.21"
+
+[target.'cfg(target_os = "linux")'.dependencies]
 pete = "0.5"
-proc-maps = "0.1"
 rstack = "0.3"
+proc-maps = "0.1"
```
ranweiler, line 59: Was this re-ordered on purpose, or just incidentally?

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -192,6 +197,15 @@ impl<'a> Tester<'a> {
         Ok(crash)
     }
 
+    #[cfg(target_os = "macos")]
+    async fn test_input_debugger(
+        &self,
+        _args: Vec<String>,
+        _env: HashMap<String, String>,
+    ) -> Result<Option<CrashLog>> {
+        bail!("running application under a debugger is not supported on darwin");
```
ranweiler, line 206:
```suggestion
        bail!("running application under a debugger is not supported on macOS");
```

src/agent/onefuzz/src/machine_id.rs
```diff
@@ -112,6 +116,42 @@ pub async fn get_os_machine_id() -> Result<Uuid> {
...
+        .stderr(Stdio::null())
+        .output()
+        .await
+        .context("ioreg failed to run")?;
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+
+    for line in stdout.split('\n') {
+        // "IOPlatformUUID" = "GUID_HERE"
+        if line.contains("IOPlatformUUID") {
```
ranweiler, line 138: I suggest pulling this into a helper function, unit testing.

src/agent/onefuzz/src/machine_id.rs
```diff
@@ -112,6 +116,42 @@ pub async fn get_os_machine_id() -> Result<Uuid> {
...
+            let quoted = line.split(" = ").last().ok_or_else(|| {
+                anyhow!("unable to find machine_id from IOPlatformUUID: {}", line)
+            })?;
```
ranweiler, line 141: Are sure that a line contains `IOPlatformUUID` iff it is in the format we're trying parse?

src/agent/onefuzz/Cargo.toml
bmc-msft: incidental

src/agent/onefuzz/src/machine_id.rs
bmc-msft: The `target_os` specific `get_os_machine_id` is unit tested already.  Does the underlying parsing need something more specific?

https://github.com/microsoft/onefuzz/blob/b5388d360c956f69e4b49923acdb6ebc9c772716/src/agent/onefuzz/src/machine_id.rs#L142-L145

src/agent/onefuzz/src/machine_id.rs
bmc-msft: What are you asking?

ranweiler: We invoke `ioreg`. We expect one of the output lines to contain a string of the form `"IOPlatformUUID" = "<some-uuid>"`, with quotes around both key and value. If a line contains the key string (`IOPlatformUUID`, not quoted), then we assume it is the k/v line, and (strictly) parse it as such. If parsing fails on any line we try to parse, we bail.

Since we fail fast, if the output of `ioreg` ever emitted a line with the substring `IOPlatformUUID` _before_ the k/v line we care about, then our parse could fail, even if the line we want occurs later.

I'm assuming that's not what happens in practice where you tested, but for all I know, it is possible, which is why I'm asking.

To some extent, this is why I asked about a unit test: the input would be a snapshot of the expected output of `ioreg`, and then the parsing function (which is pure) can be applied against it. 

I don't think this is necessary, but if we wanted to be defensive, would could have a "what if it changed" test case that simulates the condition I just mentioned. This would force us to change the parsing implementation to stop parsing on first success, but keep trying on the remaining output lines if parsing fails for some line with a matching substring.

src/agent/onefuzz/src/machine_id.rs
ranweiler: Addressed in other comment.

src/agent/onefuzz/src/machine_id.rs
bmc-msft: We verify the functionality works as a whole.  I don't think the additional complexity buys us much.

ranweiler: That's fine, your call!

ranweiler (approved):

PR #1023, opened 2021-06-25
ensure VM IDs are unique before calling Azure reimage/delete APIs

bmc-msft:
Fixes an issue where Azure is now enforcing unique IDs for reimaging & deletion.

```
Exception: HttpResponseError: (InvalidParameter) InstanceIds may not be repeated.
Code: InvalidParameter
Message: InstanceIds may not be repeated.
Target: instanceIds
```

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/vmss.py
+++ b/src/api-service/__app__/onefuzzlib/azure/vmss.py
@@ -153,6 +153,7 @@ def check_can_update(name: UUID) -> Any:
 
 def reimage_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
     check_can_update(name)
+    # Azure requires VM IDs be unique
     vm_ids = list(set(vm_ids))
 
     resource_group = get_base_resource_group()
@@ -178,6 +179,7 @@ def reimage_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
 
 def delete_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
     check_can_update(name)
+    # Azure requires VM IDs to be unique
     vm_ids = list(set(vm_ids))
 
     resource_group = get_base_resource_group()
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/azure/vmss.py
```diff
@@ -153,6 +153,8 @@ def check_can_update(name: UUID) -> Any:
 
 def reimage_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
```
ranweiler, line 154: I'm assuming you already checked, but just in  case: are there are any other places where we pass lists of IDs (VM or otherwise) to Azure, where we should defensively dedupe / pass around sets?

src/api-service/__app__/onefuzzlib/azure/vmss.py
```diff
@@ -177,6 +179,8 @@ def reimage_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
 
 def delete_vmss_nodes(name: UUID, vm_ids: List[UUID]) -> Optional[Error]:
```
ranweiler, line 180: I wonder if these methods should instead take `Set[UUID]`.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/azure/vmss.py
bmc-msft: Yep, I checked.

src/api-service/__app__/onefuzzlib/azure/vmss.py
bmc-msft: Eh, it pushes the set requirement up the stack, but I don't think it has any negative impact.

Done

ranweiler: Yeah, I think either one is fine. A list is nice ergonomically, because they're sort of the lingua franca container in Python, while requiring a set is more precise, and makes it more likely that callers avoid building a collection with duplicates in the first place.

ranweiler (approved):

PR #1024, opened 2021-06-25
bump github actions cache key IDs

bmc-msft:
With the Actions VMs including a new release of rust, we need to bump the cache identifier to get compile cache benefit


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           ~/.cargo/registry
           ~/.cargo/git
           ~/.cargo/bin
-        key: rust-${{ runner.os }}-2020-11
+        key: rust-${{ runner.os }}-2021-06-25-01
```
ranweiler, line 22: Could we set this value using a variable, with a comment, and then just update it in one place?

ranweiler (approved):

.github/workflows/ci.yml
bmc-msft: Yep.

PR #1033, opened 2021-06-30
Use sync pointers for updating `ModuleCache` off-thread

### LINKED ISSUE #1031: Coverage task does not maintain internal invariant on error
OneFuzz: 2.23.1
OS: Windows

In the coverage task, the (blocking, non-async) recording itself occurs on a new OS thread, via `spawn_blocking()`.

To avoid needing to use `sync` primitives, we just `Option::take()` the `TaskContext`'s `ModuleCache`, and pass it to the worker thread, restoring it on completion. However, if recording fails, we now retry some number of times, as of #978. But recording can fail while the worker thread owns the `ModuleCache`, which it will drop on its return with `Err`. We will then early-exit `TaskContext::record_impl()`, without restoring the `Option<ModuleCache>`, which can result in a panic.

Switch to using threadsafe smart pointers, instead.

ranweiler:
Closes #1031.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -4,6 +4,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -322,40 +321,35 @@ impl<'a> TaskContext<'a> {
     }
 }
 
-struct Recorded {
```
ranweiler, line 325: We're no longer passing around `ModuleCache` ownership, so this type is redundant.

bmc-msft (approved):

PR #1038, opened 2021-07-02
only remove authenticated users access to ssh keys if they have permissions

bmc-msft:


src/agent/onefuzz-supervisor/src/commands.rs
```diff
@@ -70,12 +70,37 @@ pub async fn add_ssh_key(key_info: SshKeyInfo) -> Result<()> {
...
+                    .arg(&admins)
+                    .stdin(Stdio::null())
+                    .stdout(Stdio::piped())
+                    .stderr(Stdio::piped())
+                    .spawn()
+                    .context("icacls remove failed to start")?
+                    .wait_with_output()
+                    .await
+                    .context("icalcs remove failed to run")?;
+                if !result.status.success() {
```
chkeita, line 94: would it make sense to retry the operation? I am thinking we could be hitting some transient issue

bmc-msft: I didn't add this due to transient issue, but this prevents the failure that was induced by a different Azure VM extension that added an SSH key.

chkeita (approved):

PR #1051, opened 2021-07-07
automatically clean PII after retention period

nharper285:
This PR removes PII from Jobs, Tasks, and Repros after 18 months.  

This PR also removes notifications tied to a container that has not been used in a task for 18 months.  This is done due to notifications having arbitrarily complex mechanisms for storing PII (typically the "assignee").


src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+        # logging.info(notification)
+
+    for job in Job.search(raw_unchecked_filter=time_filter):
+        logging.info("Retention Timer Job Search")
+        if job.user_info is not None:
+            # logging.info(
+            #     "Found job %s older than 18 months. Scrubbing user_info.",
+            #     job.job_id,
+            # )
+            job.user_info.upn = "None"
```
bmc-msft, line 66:
```suggestion
            job.user_info.upn = None
```

nharper285: Ack'd and changed!

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+    # because editing the upn for tasks will change the timestamp
+    for notification in Notification.search(raw_unchecked_filter=time_filter):
+        logging.info("Retention Timer Notification Search")
+        logging.info(
+            "Found notification %s older than 18 months. Checking related tasks.",
+            notification.notification_id,
+        )
+        container = notification.container
+        timestamp_list = []
+        for task in Task.search():
```
bmc-msft, line 34: It's my understanding, you want the logic to be:  "delete notifications older than X where no tasks were used since Y".

You should be able to query tasks once, rather than each time you query the notification.  Additionally, you should filter the tasks to only query tasks within the window you care about it.

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+            container_str = str(task.config.containers)
```
bmc-msft, line 35: Unless you have a strong reason, you should let the type system help you.  Casting will hide errors from you.  Consider a notification container of `foo` and a task with crashes container of `food`.

The code:
```python
if container in container_str:
```
is effectively:
```python
if "foo" in "[TaskContainers(type=<ContainerType.crashes: 'crashes'>, name='food')]":
```
This evaluates to True, but isn't what you want.

Consider TaskConfig.containers:
https://github.com/microsoft/onefuzz/blob/db66a1d3c5d5e44e3780c8a1cc3f349e31f680ce/src/pytypes/onefuzztypes/models.py#L192

Consider TaskContainers.name
https://github.com/microsoft/onefuzz/blob/db66a1d3c5d5e44e3780c8a1cc3f349e31f680ce/src/pytypes/onefuzztypes/models.py#L181-L183

You can have a typed list and use that:
```python
task_containers = [x.name for x in task.config.containers]
if container in task_containers:
```

This would effectively be:
```python
if Container("foo") in [Container("food")]:
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+            if container in container_str:
+                # Need to make sure there isn't a task still using the container.
+                if task.state == TaskState.stopped:
+                    timestamp_list.append(task.timestamp)
```
bmc-msft, line 39: While in a previous comment, I mentioned tasks in the range you care about should be part of the initial query, this is additional complexity that isn't needed.

Once a single task is meets the criteria (within the time window and with the specified container), the notification can be ignored.

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            if youngest < time_retained:
+                logging.info(
+                    "All related tasks are older than 18 months."
+                    + " Deleting Notification %s.",
+                    notification.notification_id,
+                )
+                notification.delete()
+
+    for job in Job.search(
```
bmc-msft, line 58: For a system that is always in use, this will progressively get slower.  Perhaps the time filter can be "older than 180 days, newer than 200 days" as to not recheck every job older than 180 days each time this is run?

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+                "Found job %s older than 18 months. Scrubbing user_info.",
+                job.job_id,
+            )
+            job.user_info.upn = None
+            job.save()
+
+    for task in Task.search(
+        query={"state": [TaskState.stopped]}, raw_unchecked_filter=time_filter
+    ):
+        logging.info("Retention Timer Task Search")
```
bmc-msft, line 73: this gets logged for every task in the search

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+            logging.info(
+                "Found task %s older than 18 months. Scrubbing user_info.",
+                task.task_id,
+            )
+            task.user_info.upn = None
+            task.save()
+
+    for repro in Repro.search(raw_unchecked_filter=time_filter):
+        logging.info("Retention Timer Repro Search")
+        logging.info(repro)
```
bmc-msft, line 84:
```suggestion
```

src/api-service/__app__/timer_retention/readme.md
```diff
@@ -0,0 +1,11 @@
+# TimerTrigger - Python
```
bmc-msft, line 1: This looks like a generic README.  Perhaps instead of the template documentation, you can include your intent for the function?

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+                    "All related tasks are older than 18 months."
+                    + " Deleting Notification %s.",
+                    notification.notification_id,
+                )
+                notification.delete()
+
+    for job in Job.search(
+        query={"state": [JobState.stopped]}, raw_unchecked_filter=time_filter
+    ):
+        logging.info("Retention Timer Job Search")
```
bmc-msft, line 61: Logging this message on every Job returned is not useful.

```suggestion
```

src/api-service/__app__/timer_retention/__init__.py
bmc-msft:
```suggestion
```

bmc-msft (dismissed):
see comments

src/api-service/__app__/timer_retention/readme.md
nharper285: Added by mistake!

src/api-service/__app__/timer_retention/__init__.py
nharper285: I will remove all logs like this.

src/api-service/__app__/timer_retention/__init__.py
nharper285: Removing

src/api-service/__app__/timer_retention/__init__.py
nharper285: Removing

src/api-service/__app__/timer_retention/__init__.py
nharper285: Okay so the reason I do a blanket search on tasks each time is because I don't want to only query Task.search("older than 18 months") if there are potentially tasks that are newer that still reference the notification/container. 

I.e., Let's say I have a notification that references container-123. I have a set of tasks older than 18 months and a set of tasks younger than 18 months. If I only query Task.search("older than 18 months"), I might inappropriately delete the notification. 

Unless, of course, I'm misunderstanding you're feedback and you're suggesting I flip the loops such that I iterate over the tasks and _then_ iterate over notifications?

src/api-service/__app__/timer_retention/__init__.py
nharper285: This is good feedback. I'll address appropriately.

src/api-service/__app__/timer_retention/__init__.py
nharper285: By ignored do you mean deleted?

src/api-service/__app__/timer_retention/__init__.py
nharper285: This makes sense!

src/api-service/__app__/timer_retention/__init__.py
bmc-msft: Right now, you are calling `Task.search` for every notification.

If you have 300 notifications, you are calling `Task.search` 300 times.  Task.search, as shown here, returns all tasks that have ever existed.

I'm suggesting changing the process to the following:

Get the containers used by all tasks newer than 18 months, call this Set A.  (Only do this once, not once per notification).
For each notification older than 18 months, if the notification.container is not in Set A, clean up the notification.

nharper285: AH!

Excellent okay I understand. Thanks Brian.

src/api-service/__app__/timer_retention/__init__.py
bmc-msft: Your process is currently the psuedo-code:
```
for notification in Notifications.search(...):
   timestamps = []
   for task in Task.search():
      if task containers contain notification.container:
         timestamps.append(tiemstamp)
   if timestamps:
       if get_newest_timestamp < window:
           notification.delete()
```

As soon as you've found one task that is newer than the window, you can stop processing the list.   You don't need to check the rest of the tasks at that point.  Consider if the first of 30000 tasks is newer.  Checking all 30000 tasks is wasteful.

There are three optimizations here, two are discussed in another comment.

1. Perform `Task.search` once to check Notifications.
2. Include the date range you care about in the Task.search.  This puts reduces how many tasks you have to search and the data retrieved from Azure Tables.
3. As soon as the first task includes a container, move onto the next notification, since this one can stay.
4. instead of keeping a list of tasks, keep the set of container names.

This would result in something like this:
```
recent_tasks = Task.search(newer_than_18_months)
recent_containers = get_containers(recent_tasks)
for notification in Notification.search(older_than_18_months):
    if notification.container in recent_containers:
        continue
    notification.delete()
```

nharper285: Just pushed a PR that does basically this. Will address logging PRs in future commits.

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    time_retained_newer = now - SEARCH_EXTENT
+
+    time_filter = (
+        f"Timestamp lt datetime'{time_retained_older.isoformat()}' and "
+        f"Timestamp gt datetime'{time_retained_newer.isoformat()}'"
+    )
+    time_filter_newer = f"Timestamp gt datetime'{time_retained_older.isoformat()}"
+
+    # Collecting 'still relevant' task containers.
+    used_containers = []
```
bmc-msft, line 31: If you have 3 tasks that use container `foo`, then this will be `[Container("foo"), Container("foo"), Container("foo")]`

If you use a set, this will be: `{Container("foo")}`

nharper285: Good point. Addressing.

nharper285:
> see comments

Responded!

src/api-service/__app__/requirements.txt
```diff
@@ -1,34 +1,34 @@
-azure-common~=1.1.25
```
bmc-msft, line 1: Please revert this change.  It's hard to tell if you actually changed anything here.

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            "Found notification %s older than 18 months. Checking related tasks.",
+            notification.notification_id,
+        )
+        container = notification.container
+        if container not in used_containers:
+            logging.info(
+                "All related tasks are older than 18 months."
+                + " Deleting Notification %s.",
+                notification.notification_id,
+            )
```
bmc-msft, line 49:
```suggestion
            logging.info("deleting unused notification: %s", notification.notification_id)
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    # Collecting 'still relevant' task containers.
+    used_containers = set()
+    for task in Task.search(raw_unchecked_filter=time_filter_newer):
+        task_containers = {x.name for x in task.config.containers}
+        used_containers.update(task_containers)
+
+    # You have to do notification before task,
+    # because editing the upn for tasks will change the timestamp
+    for notification in Notification.search(raw_unchecked_filter=time_filter):
+        logging.info(
```
bmc-msft, line 39: Assuming a notification is old, but still in use, this will get logged once a day forever.  This makes it a debug log, which won't get logged to app insights by default.

```suggestion
        logging.debug(
```

src/api-service/__app__/requirements.txt
nharper285: Done!

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+            notification.delete()
+
+    for job in Job.search(
+        query={"state": [JobState.stopped]}, raw_unchecked_filter=time_filter
+    ):
+        if job.user_info is not None and job.user_info.upn is not None:
+            logging.info(
+                "Found job %s older than 18 months. Scrubbing user_info.",
+                job.job_id,
+            )
```
bmc-msft, line 57:
```suggestion
            logging.info("removing PII from job: %s", job.job_id)
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+            job.save()
+
+    for task in Task.search(
+        query={"state": [TaskState.stopped]}, raw_unchecked_filter=time_filter
+    ):
+        if task.user_info is not None and task.user_info.upn is not None:
+            logging.info(
+                "Found task %s older than 18 months. Scrubbing user_info.",
+                task.task_id,
+            )
```
bmc-msft, line 68:
```suggestion
            logging.info("removing PII from task: %s", task.task_id)
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+            task.save()
+
+    for repro in Repro.search(raw_unchecked_filter=time_filter):
+        if repro.user_info is not None and repro.user_info.upn is not None:
+            logging.info(
+                "Found repro entry for task %s on node %s that is older "
+                + "than 18 months. Scrubbing user_info.",
+                repro.task_id,
+                repro.vm_id,
+            )
```
bmc-msft, line 79:
```suggestion
            logging.info("removing PII from repro: %s", repro.vm_id)
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    # because editing the upn for tasks will change the timestamp
+    for notification in Notification.search(raw_unchecked_filter=time_filter):
+        logging.debug(
+            "Found notification %s older than 18 months. Checking related tasks.",
+            notification.notification_id,
+        )
+        container = notification.container
+        if container not in used_containers:
+            logging.info(
+                "Deleting unused notification: %s", notification.notification_id
```
bmc-msft, line 46:
```suggestion
                "deleting expired notification: %s", notification.notification_id
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    for task in Task.search(raw_unchecked_filter=time_filter_newer):
+        task_containers = {x.name for x in task.config.containers}
+        used_containers.update(task_containers)
+
+    # You have to do notification before task,
+    # because editing the upn for tasks will change the timestamp
+    for notification in Notification.search(raw_unchecked_filter=time_filter):
+        logging.debug(
+            "Found notification %s older than 18 months. Checking related tasks.",
+            notification.notification_id,
```
bmc-msft, line 41:
```suggestion
            "checking expired notification for removal: %s",
            notification.notification_id,
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+from ..onefuzzlib.jobs import Job
+from ..onefuzzlib.notifications.main import Notification
+from ..onefuzzlib.repro import Repro
+from ..onefuzzlib.tasks.main import Task
+
+RETENTION_POLICY = datetime.timedelta(days=(18 * 30))
+SEARCH_EXTENT = datetime.timedelta(days=(20 * 30))
+
+
+def main(mytimer1: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
```
bmc-msft, line 17:
```suggestion
def main(mytimer: func.TimerRequest, dashboard: func.Out[str]) -> None:  # noqa: F841
```

src/api-service/__app__/timer_retention/function.json
```diff
@@ -0,0 +1,17 @@
+{
+  "bindings": [
+    {
+      "direction": "in",
+      "name": "mytimer1",
```
bmc-msft, line 5:
```suggestion
      "name": "mytimer",
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
+import datetime
```
bmc-msft, line 1:
```suggestion
#!/usr/bin/env python
#
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

import datetime
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    time_filter_newer = f"Timestamp gt datetime'{time_retained_older.isoformat()}'"
+
+    # Collecting 'still relevant' task containers.
+    used_containers = set()
+    for task in Task.search(raw_unchecked_filter=time_filter_newer):
+        task_containers = {x.name for x in task.config.containers}
+        used_containers.update(task_containers)
+
+    # You have to do notification before task,
+    # because editing the upn for tasks will change the timestamp
```
bmc-msft, line 37:
```suggestion
```

src/api-service/__app__/timer_retention/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    time_retained_older = now - RETENTION_POLICY
+    time_retained_newer = now - SEARCH_EXTENT
+
+    time_filter = (
+        f"Timestamp lt datetime'{time_retained_older.isoformat()}' "
+        f"and Timestamp gt datetime'{time_retained_newer.isoformat()}'"
+    )
+    time_filter_newer = f"Timestamp gt datetime'{time_retained_older.isoformat()}'"
+
+    # Collecting 'still relevant' task containers.
```
bmc-msft, line 30:
```suggestion
    # Collecting 'still relevant' task containers.
    # NOTE: This must be done before potentially modifying tasks otherwise
    # the task timestamps will not be useful.
```

bmc-msft (approved):

Issue #1028, opened 2021-06-30
Supervisor stuck restarting forever if startup fails


ranweiler:
If something causes the supervisor agent to fail on startup, it will be restarted over and over again, and the node will appear to be stuck in `init`. On Linux, this is because the systemd service `Restart` value is `always`.

Observed on Linux, OneFuzz 2.23.1, where a missing shared library in a non-default OS image triggers this (see #1027).

ranweiler:
@bmc-msft restarting is useful, and I'm not suggesting we stop. But the user-facing outcome that this bug describes is definitely bad.

I believe the default restart timeout on systemd is 100ms. In the situation that motivated this bug report, where startup failed immediately, that meant we were restarting around 10 times a second on every node in a brand new scaleset, without starting up successfully even one time.

A few independent changes we could consider:
- Limit the total restart count _until_ we have had at least one successful supervisor startup, then _always_ restart
- Limit the restart _retry_ count. IOW, only limit the number of _consecutive_ failed restarts, but allow infinite total restarts
- Increase the restart timeout, say, to 500ms or 1s. This would reduce log spam, reduce the consumption of the attempt budget (if added), and is probably more appropriate in response to agent failures due to upstream service or network errors

At a high level, I think the most important user-facing thing we're missing is:
- Ensure that a managed node has had a successful initial supervisor heartbeat within some interval after node creation. >= 24 hours is too long, IMO, but was observed by a user. After the timeout, shut down the node and mark it as being in a non-viable state, in a way that users can see when querying the scaleset/nodes using the API.

bmc-msft:
The supervisor heartbeat expiration window is set to 1 hour to compensate for VMs taking an extremely long time to launch (such as patch updates on install).

https://github.com/microsoft/onefuzz/blob/a1502b7d58fd57d31ab9cf8aa0934ba51c3b2eac/src/api-service/__app__/onefuzzlib/workers/nodes.py#L36

ranweiler:
Great. Then based on the user report, it seems like either that (1) does not trigger in the "supervisor never starts" scenario, or (2) it does, and causes us to reimage the doomed node, which will then fail to start up again, which means we have an "outer" infinite loop.

PR #1054, opened 2021-07-09
Fix get_dead_nodes query

chkeita:
partial fix for  #1028


bmc-msft:
I had not considered using Timestamp (which is effectively "last updated") as a solution.  Interesting.

Can this include updating the log & Error messages to differentiate between a node with an expired heartbeat and a node that never generated a heartbeat?

I'm think about this:
https://github.com/microsoft/onefuzz/blob/48978ce881ed3e1f9fce1b6f49f05043f18002fb/src/api-service/__app__/onefuzzlib/workers/scalesets.py#L389-L395

and this:
https://github.com/microsoft/onefuzz/blob/48978ce881ed3e1f9fce1b6f49f05043f18002fb/src/api-service/__app__/onefuzzlib/workers/scalesets.py#L399-L403

bmc-msft (approved):

PR #1055, opened 2021-07-09
Update onefuzz supervisor service restart behavior

chkeita:
- Increase the delay between service restart from 100ms to 1s

partial fix for #1028

### DIFF
--- a/src/runtime-tools/linux/onefuzz.service
+++ b/src/runtime-tools/linux/onefuzz.service
@@ -1,14 +1,12 @@
 # onefuzz systemd startup script
 [Unit]
 Description=onefuzz
-StartLimitIntervalSec=20
-StartLimitBurst=10
 
 [Service]
 Type=simple
 ExecStart=/bin/bash /onefuzz/tools/linux/run.sh
 Restart=always
-RestartSec=1
+RestartSec=30
 
 [Install]
 WantedBy=default.target
--- a/src/runtime-tools/linux/onefuzz.service
+++ b/src/runtime-tools/linux/onefuzz.service
@@ -1,12 +1,14 @@
 # onefuzz systemd startup script
 [Unit]
 Description=onefuzz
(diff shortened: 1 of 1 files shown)


bmc-msft:
I'm on board with increasing the restart delay is good.  We can probably increase the delay even more.  On Windows, the delay is 30s.

For the bursting, will it stop restarting eventually?  If so, I think we probably don't want that.  If the reason the app is failing is due to Azure, restarting forever will enable eventual recovery.

src/runtime-tools/linux/onefuzz.service
```diff
@@ -1,12 +1,14 @@
 # onefuzz systemd startup script
 [Unit]
 Description=onefuzz
+StartLimitIntervalSec=20
```
ranweiler, line 4: This would be great if it only throttled within the interval, and had no effect on retry policy after a cool-off period. But to echo @bmc-msft, I think we should remove these settings, since:

> Units which are started more than `burst` times within an interval time span are **not permitted to start any more.**

> Note that units which are configured for `Restart=`, and which reach the start limit are **not attempted to be restarted anymore** [...]

([source](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval))

Fortunately, with an increased `RestartSec`, I think this is probably redundant anyway.

src/runtime-tools/linux/onefuzz.service
```diff
@@ -1,12 +1,14 @@
...
+StartLimitBurst=10
 
 [Service]
 Type=simple
 ExecStart=/bin/bash /onefuzz/tools/linux/run.sh
-StartLimitInterval=0
 Restart=always
+RestartSec=1
```
ranweiler, line 11:
```suggestion
RestartSec=30
```

ranweiler (approved):

PR #1056, opened 2021-07-09
Enable autoscaling based on worksets

bmc-msft:
This builds on past PRs for autoscaling, as well as #1046, #1047, #1049, and #1050.

For scaling down, this uses the existing `ShrinkQueue` mechanism, which has been used to scale down scalesets.

For scaling up, this orders scalesets by scaleset_id, then increases the size to each scalesets max size.  This should hopefully reduce fragmentation.  The actual increase happens through the existing resize.

Note: AutoScaleConfig.scaleset_size from the original autoscaling component is no longer used.  it's left here for backwards compat with existing CLIs.


src/api-service/__app__/onefuzzlib/workers/autoscale.py
```diff
@@ -0,0 +1,282 @@
...
+        to_add,
+        [x.scaleset_id for x in scalesets],
+    )
+
+    config = pool.autoscale
+    if not config:
+        raise Exception(f"scaling up a non-autoscaling pool: {pool.name}")
+    region = config.region or get_base_region()
+
+    set_shrink_queues(pool, scalesets, 0)
```
chkeita, line 95: Is this needed when scaling up ?

bmc-msft: Yes.  This empties the shrink queue.

chkeita (approved):

Issue #1042, opened 2021-07-06
onefuzz-supervisor failed. Failed to start due to pre-existing workset config


jagunter:
## Information

- Onefuzz version: 2.24.0
- OS: Windows latest

## Provide detailed reproduction steps (if any)

Rerun job `be0217be-e0f5-4c41-95a2-2cfaa4f7cc72`.

### Expected result

Supervisor to not fail.

### Actual result

Supervisor failed. Digging into logs (see below) it looks like the culprit is

```
        "message": "saving failure: failed to start due to pre-existing workset config: C:\\onefuzz\\workset_context.json",
```

## Logs

These are the logs for the machine where the onefuzz supervisor failed, obtained from

```
onefuzz debug logs query 'union isfuzzy=true availabilityResults, requests, exceptions, pageViews, traces, customEvents, dependencies | where timestamp > datetime("2021-07-01T23:19:07.467Z") and timestamp < datetime("2021-07-02T18:47:00.000Z") | where * has "62e61806-86a8-4416-a237-9f3c741be61d" | order by timestamp desc | take 100' --timespan 5d --query '[].{message: message, timestamp: timestamp}'
```

```
    {
        "message": "error running supervisor agent: failed to start due to pre-existing workset config: C:\\onefuzz\\workset_context.json",
        "timestamp": "2021-07-02T18:46:54.905Z"
    },
    {
        "message": "saving failure: failed to start due to pre-existing workset config: C:\\onefuzz\\workset_context.json",
        "timestamp": "2021-07-02T18:46:54.905Z"
    },
    {
        "message": "node 'done' with error: machine_id:62e61806-86a8-4416-a237-9f3c741be61d, data:{\"error\": \"onefuzz-supervisor failed\"}",
        "timestamp": "2021-07-02T18:46:54.8321135Z"
    },
    {
        "message": "sending event: EventType.node_state_updated - {\"machine_id\": \"62e61806-86a8-4416-a237-9f3c741be61d\", \"scaleset_id\": \"1339d1c1-5d28-4481-8de0-d25eaec5a188\", \"pool_name\": \"***\", \"state\": \"done\"}",
        "timestamp": "2021-07-02T18:46:54.8211118Z"
    },
    {
        "message": "node state update: 62e61806-86a8-4416-a237-9f3c741be61d from:NodeState.busy to:NodeState.done",
        "timestamp": "2021-07-02T18:46:54.8204499Z"
    },
    {
[...]

PR #1062, opened 2021-07-13
attempt to log panic stack trace to the supervisor failure log

bmc-msft:
While investigating #1042, I found a gap in logs that is indicative of the supervisor panicking.  

This PR adds a panic handler that reuses the existing 'failure log' mechanism to record the panic and backtrace.  The failure log mechanism enables recording runtime errors to a file, which is sent to the service on restart.


src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -36,6 +36,7 @@ pub mod debug;
 pub mod done;
 pub mod failure;
 pub mod heartbeat;
+pub mod logpanic;
```
ranweiler, line 39: I'd suggest just naming this module `panic`.

ranweiler (approved):

PR #1072, opened 2021-07-15
Add `generic_analysis` example that performs coverage analysis

bmc-msft:
This adds an example script and tool that enables LLVM source-based coverage using the `generic_analysis` task.

This provides:
1. sample python script that launches the template and then the analysis task
1. sample `analysis_exe` wrapper script that launches the LLVM coverage tools
1. sample libfuzzer target for the example
1. walk through submitting the jobs and inspecting the results


src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
+# Performing Source-Coverage using Analysis Tasks
+
+The `generic_analysis` task can be used to perform user-defined analysis for every input.
```
ranweiler, line 3:
```suggestion
The `generic_analysis` task can be used to perform a user-defined analysis of a target executable for every test input from some storage container.  
```

src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
+# Performing Source-Coverage using Analysis Tasks
+
+The `generic_analysis` task can be used to perform user-defined analysis for every input.  
+
+Running an application compiled with [LLVM's source-based code coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) with each input can be used to generate source based coverage information.
```
ranweiler, line 5:
```suggestion
Running an application compiled with [LLVM's source-based code coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) with each input can be used to generate source based coverage information. In this case, the core of the analysis is performed via compiled-in target instrumentation, but this is not required.
```

src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+This example demonstrates using `generic_analysis` and the LLVM source coverage tools to provide source-based coverage on every input for a job.
+
+* [source-coverage-libfuzzer.py](source-coverage-libfuzzer.py): A wrapper that will launch a standard libfuzzer basic job *with* a source-based coverage task.  (used below)
```
ranweiler, line 9:
```suggestion
* [source-coverage-libfuzzer.py](source-coverage-libfuzzer.py): A wrapper that will launch a standard `libfuzzer basic` job *with* a source-based coverage task.  (used below)
```

src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
...
+* [source-coverage.py](source-coverage.py): A wrapper that will launch a new job comprised of a source-based coverage task
+* [setup](setup): a basic libfuzzer target that builds with and without source coverage enabled
```
ranweiler, line 11:
```suggestion
* [setup](setup): a basic libFuzzer target that builds with and without source coverage enabled
```

src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
...
+* [tools/source-coverage.sh](tools/source-coverage.sh): a script that wraps llvm-profdata and llvm-cov to perform the source analysis
```
ranweiler, line 12:
```suggestion
* [tools/source-coverage.sh](tools/source-coverage.sh): a script that wraps `llvm-profdata` and `llvm-cov` to perform the source analysis
```

src/cli/examples/llvm-source-coverage/README.md
```diff
@@ -0,0 +1,134 @@
+# Performing Source-Coverage using Analysis Tasks
```
ranweiler, line 1:
```suggestion
# Collecting Source Coverage using Analysis Tasks
```

ranweiler (approved):

src/cli/examples/llvm-source-coverage/README.md
ranweiler: Maybe we should link to this doc from [docs/custom-analysis.md](https://github.com/microsoft/onefuzz/blob/main/docs/custom-analysis.md).

src/cli/examples/llvm-source-coverage/README.md
bmc-msft: Can you explain your intent for this addition?  The compiled-in instrumentation is required for this example.

src/cli/examples/llvm-source-coverage/README.md
bmc-msft: (I added this to custom-analysis.md)

src/cli/examples/llvm-source-coverage/README.md
ranweiler: This is just a suggestion for emphasizing the big picture of "analysis tasks" while explaining this specific task (and is not essential).

The "analyzer" in this example is just a bit odd for the following reason: it doesn't perform analysis per se. It is just a script that mostly just invokes the `target_exe` and juggles files that the `target_exe` itself generates. Per its name, and in typical scenarios, I think we expect that the `analyzer_exe` is usually a program that _creates_ the analysis files. This larger example is just a bit unusual in the way that libFuzzers are unusual: the actual _analyzer_ is baked into the target here, just like a libFuzzer has the fuzzer baked into the target as instrumentation.

The suggested comment about that might be more confusing than clarifying in context, though.

bmc-msft: In this case, the analyzer_exe does create the files by calling first the fuzzer, then llvm-profdata, then llvm-cov.

ranweiler: So, to be clear, there's no edit needed here, and I think this is one of the best docs we (will) have. The above was just my original rationale about emphasizing what's special/slightly unusual in this example.

PR #1074, opened 2021-07-15
Group membership check 

chkeita:
This PR enabl the restriction of certain api to some user based on their principalid or their group membership

## Api Restriction rules
We provide the ability to create a set of rules that specify which group has access to an api
The set of rules is a list of `RuleDefinition` defined set in the property `api_access_rules` of the Instance configuration
``` json 
  [
      {
          "methods": ["get", "post"],
          "endpoint": ["/api/jobs"],
          "allowed_groups": ["<group_id>"]
      }
  ]
  ```
We store and manage in the `RequestAccess` class defined in #1420

## Rules enforcement
Those rules are enforced when the request is received by the service. 
A call to `check_access` determine if the request is authorized or not. 

## Group membership information retrieval 
The group membership is retrived from azure call to [msgraph](https://docs.microsoft.com/en-us/graph/api/user-list-transitivememberof?view=graph-rest-1.0&tabs=http) in the  `AzureADGroupMembership` class
This call requires [specific permissions](https://docs.microsoft.com/en-us/graph/api/user-list-transitivememberof?view=graph-rest-1.0&tabs=http#permissions) which depending on the envirement can only be granted by an admin. 
We also provide for testing purposes a `StaticGroupMembership` class to provide the access restriction functionality without queryin azure.
The class is initialized with the group membership information stored in the Instance configuration in the field `group_membership`

## Testing
A functional test is provided in `api_restriction_test.py`

### CODE REVIEW

File: src/api-service/functional_tests/api_restriction_test.py [hunk 1]
```diff
@@ -0,0 +1,178 @@
...
+    cli = get_default_cli()
+    cli.logging_cls
+    cli.invoke(args, out_file=open(os.devnull, "w"))
+    if cli.result.result:
+        return cli.result.result
+    elif cli.result.error:
+        raise cli.result.error
+
+
+class APIRestrictionTests:
```
chkeita, line 28: will move to check-pr

src/api-service/__app__/onefuzzlib/request.py
```diff
@@ -6,36 +6,81 @@
...
+    else:
+        return None
+
+
+@cached(ttl=60)
+def get_rules_cached() -> Optional[RequestAccess]:
+    return get_rules()
 
 
 def check_access(req: HttpRequest) -> Optional[Error]:
```
chkeita, line 46: Currently this call is made when parsing a request. So no check happens when a request has no paramters.
This needs to be moved to the call_if function that gets called on all request

src/api-service/__app__/onefuzzlib/azure/group_membership.py
```diff
@@ -0,0 +1,51 @@
...
+
+        groups = self.get_groups(member_id)
+        for g in group_ids:
+            if g in groups:
+                return True
+
+        return False
+
+    def get_groups(self, member_id: UUID) -> List[UUID]:
+        """Gets all the groups of the provided member"""
```
stishkin, line 27: still todo ?

stishkin: or is it base class definition ?

chkeita: This is a Protocol class. It works like interfaces here 
You can see the implementation below in `AzureADGroupMembership` and  `StaticGroupMembership`

src/api-service/__app__/onefuzzlib/request.py
```diff
@@ -6,36 +6,81 @@
...
-    if "ONEFUZZ_AAD_GROUP_ID" in os.environ:
-        message = "ONEFUZZ_AAD_GROUP_ID configuration not supported"
-        logging.error(message)
-        return Error(
-            code=ErrorCode.INVALID_CONFIGURATION,
-            errors=[message],
-        )
+    if "NO_REQUEST_ACCESS_RULES_CACHE" in os.environ:
```
stishkin, line 47: Does this setting needs to be documented ?  Is it exposed to users or is it just for internal testing ?

src/api-service/__app__/onefuzzlib/request.py
```diff
@@ -6,36 +6,81 @@
...
+        rules = get_rules()
     else:
+        rules = get_rules_cached()
+
+    if not rules:
```
stishkin, line 52: The check is that if member id is in rule.allowed_group_ids -> then they are allowed access

If there are no rules, doesn't it mean that no-one should be able to have access, since there are zero allowed_group_ids ?

src/api-service/functional_tests/api_restriction_test.py
```diff
@@ -0,0 +1,178 @@
...
+        if instance_config.api_access_rules is None:
+            instance_config.api_access_rules = {}
+
+        instance_config.api_access_rules["/api/jobs"] = ApiAccessRule(
+            allowed_groups=[group_id],
+            methods=["GET"],
+        )
+
+        self.onefuzz.instance_config.update(instance_config)
+        restart_instance(self.instance_name, self.resource_group)
```
stishkin, line 86: to rules come to effect we need to restart backend function app. Do I read code correctly ?

stishkin: will this need to be done in a production scenario ?

src/api-service/__app__/onefuzzlib/request.py
chkeita: yes this is of internal testing. not caching will result in poor perfromance since it will add a query to the db for every call

src/api-service/functional_tests/api_restriction_test.py
chkeita: the rules are cached so you can either wait for the cache to expire or restart the server for new rules to be used. 
i do not expect we will need to do this in production.

src/api-service/__app__/onefuzzlib/request.py
chkeita: Nope, No rules means there are no restriction on the API.

stishkin: So we are not validating if user is allowed to access, we are validating whether user is NOT allowed ?

stishkin: Because down below there is check for "is_member" to mean that user has access and that seems like a contradiciton

src/api-service/functional_tests/api_restriction_test.py
stishkin: Purely from user experience:
- When does the cache expire ? 
-  Is cache expiration controlled by user ?

src/api-service/__app__/onefuzzlib/request.py
stishkin: Discussed offline

src/api-service/functional_tests/api_restriction_test.py
stishkin: Discussed offline

stishkin (approved):

PR #1075, opened 2021-07-15
check change event type before container name

bmc-msft:
This allows us to ignore events that are not file creation faster.

### DIFF
--- a/src/api-service/__app__/queue_file_changes/__init__.py
+++ b/src/api-service/__app__/queue_file_changes/__init__.py
@@ -29,10 +29,10 @@ def file_added(event: Dict, fail_task_on_transient_error: bool) -> None:
 def main(msg: func.QueueMessage, dashboard: func.Out[str]) -> None:
     event = json.loads(msg.get_body())
     last_try = msg.dequeue_count == MAX_DEQUEUE_COUNT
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/queue_file_changes/__init__.py
```diff
@@ -29,10 +29,10 @@ def file_added(event: Dict, fail_task_on_transient_error: bool) -> None:
 def main(msg: func.QueueMessage, dashboard: func.Out[str]) -> None:
     event = json.loads(msg.get_body())
     last_try = msg.dequeue_count == MAX_DEQUEUE_COUNT
-    if event["topic"] not in corpus_accounts():
+    if event["eventType"] != "Microsoft.Storage.BlobCreated":
```
ranweiler, line 32: Could you add a short comment emphasizing why this order is preferable?

ranweiler (approved):

src/api-service/__app__/queue_file_changes/__init__.py
bmc-msft:
```suggestion
    # check type first before calling Azure APIs
    if event["eventType"] != "Microsoft.Storage.BlobCreated":
```

PR #1076, opened 2021-07-15
sync batch processed results after every 10 inputs

bmc-msft:
When we have a analysis task with a large number of initial inputs, it may take an excessively long time before results show up in the analysis container.

This PR makes it such that we sync the analysis container after every 10 inputs.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -286,8 +286,8 @@ impl<'a> TaskContext<'a> {
                     if entry.file_type().await?.is_file() {
                         self.record_input(&entry.path()).await?;
                         count += 1;
-                        // make sure we save & sync coverage every 10 inputs
 
+                        // make sure we save & sync coverage every 10 inputs
                         if count % 10 == 0 {
                             self.save_and_sync_coverage().await?;
                         }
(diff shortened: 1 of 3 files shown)


chkeita (approved):

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -286,6 +286,11 @@ impl<'a> TaskContext<'a> {
                     if entry.file_type().await?.is_file() {
                         self.record_input(&entry.path()).await?;
                         count += 1;
+                        // make sure we save & sync coverage every 10 inputs
+
```
ranweiler, line 290:
```suggestion

```

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -286,6 +286,11 @@ impl<'a> TaskContext<'a> {
                     if entry.file_type().await?.is_file() {
                         self.record_input(&entry.path()).await?;
                         count += 1;
+                        // make sure we save & sync coverage every 10 inputs
```
ranweiler, line 289:
```suggestion

                        // make sure we save & sync coverage every 10 inputs
```

ranweiler (approved):

PR #1078, opened 2021-07-16
Remove `Stream` impl for `DirectoryMonitor`

ranweiler:
The manual implementation of `Stream` for `DirectoryMonitor` misused the `Waker` provided by `poll_next()`. In practice, this resulted in telling the executor to poll the stream again _immediately_ after any non-terminal `try_recv()`. The end result was equivalent to calling `try_recv()` in a tight loop. This pegged agent CPU usage as soon as we started monitoring directories for new files.

Now, we take a much simpler approach. The private `into_async()` shim function spawns a thread that blocks on new (synchronous) `Receiver` messages, and then immediately sends them to an unbounded async `Receiver`. All subsequent API surface can thus be effortlessly async.

`DirectoryMonitor` now exports a public `next_file()` async method, which is equivalent in practice to implementing `Stream`. We aren't using any `Stream` combinators, we don't lose anything in doing this. We can also use this to impl `Stream` again later, if needed.

- [x] Locally tested that agent directory monitoring use sites pick up new files
- [x] Confirmed that agent CPU usage drops from 100% to < 1%.
- [ ] Manually tested tasks on deployed instance

### DIFF
--- a/src/agent/onefuzz-agent/src/local/common.rs
+++ b/src/agent/onefuzz-agent/src/local/common.rs
@@ -4,11 +4,9 @@ use anyhow::Result;
 use backoff::{future::retry, Error as BackoffError, ExponentialBackoff};
 use clap::{App, Arg, ArgMatches};
 use flume::Sender;
-use onefuzz::jitter::delay_with_jitter;
 use onefuzz::{blob::url::BlobContainerUrl, monitor::DirectoryMonitor, syncdir::SyncedDir};
 use path_absolutize::Absolutize;
 use reqwest::Url;
-use std::task::Poll;
 use std::{
     collections::HashMap,
     env::current_dir,
@@ -275,17 +273,12 @@ impl DirectoryMonitorQueue {
         let handle: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
             let mut monitor = DirectoryMonitor::new(directory_path_clone.clone());
             monitor.start()?;
-            loop {
-                match monitor.poll_file() {
-                    Poll::Ready(Some(file_path)) => {
-                        let file_url = Url::from_file_path(file_path)
-                            .map_err(|_| anyhow!("invalid file path"))?;
-                        queue.enqueue(file_url).await?;
-                    }
-                    Poll::Ready(None) => break,
-                    Poll::Pending => delay_with_jitter(Duration::from_secs(1)).await,
-                }
+
+            while let Some(file_path) = monitor.next_file().await {
+                let file_url = Url::from_file_path(file_path).map_err(|_| anyhow!("invalid file path"))?;
+                queue.enqueue(file_url).await?;
             }
+
             Ok(())
         });
 
--- a/src/agent/onefuzz-agent/src/local/common.rs
+++ b/src/agent/onefuzz-agent/src/local/common.rs
@@ -275,7 +275,8 @@ impl DirectoryMonitorQueue {
             monitor.start()?;
 
             while let Some(file_path) = monitor.next_file().await {
-                let file_url = Url::from_file_path(file_path).map_err(|_| anyhow!("invalid file path"))?;
+                let file_url =
+                    Url::from_file_path(file_path).map_err(|_| anyhow!("invalid file path"))?;
                 queue.enqueue(file_url).await?;
             }
 
--- a/src/agent/onefuzz/src/monitor.rs
+++ b/src/agent/onefuzz/src/monitor.rs
@@ -2,7 +2,10 @@
 // Licensed under the MIT License.
 
 use std::path::PathBuf;
-use std::sync::{self, mpsc::{Receiver as SyncReceiver, RecvError}};
+use std::sync::{
+    self,
+    mpsc::{Receiver as SyncReceiver, RecvError},
+};
 use std::thread::{self, JoinHandle};
 use std::time::Duration;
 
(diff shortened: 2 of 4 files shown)


src/agent/onefuzz/src/monitor.rs
```diff
@@ -49,48 +55,60 @@ impl DirectoryMonitor {
...
+                    // Other filesystem event. Ignore.
+                }
             }
         }
     }
 }
 
-impl Stream for DirectoryMonitor {
-    type Item = PathBuf;
+fn into_async<T: Send + 'static>(
```
ranweiler, line 88: Document, though internal. Note that the `JoinHandle` is inessential, and we could just drop it.

Thought: in doesn't matter in our usage here, but maybe we should use `tokio::task::spawn_blocking()` and return its `JoinHandle`, instead. By using `thread::spawn()`, we return a `JoinHandle` whose `join()` method is sync and blocks the async runtime, which means we'll almost certainly never want to use it.

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -275,17 +273,13 @@ impl DirectoryMonitorQueue {
...
-                    Poll::Ready(Some(file_path)) => {
-                        let file_url = Url::from_file_path(file_path)
-                            .map_err(|_| anyhow!("invalid file path"))?;
-                        queue.enqueue(file_url).await?;
-                    }
-                    Poll::Ready(None) => break,
-                    Poll::Pending => delay_with_jitter(Duration::from_secs(1)).await,
-                }
+
+            while let Some(file_path) = monitor.next_file().await {
```
ranweiler, line 277: Double-check that this is an equivalent or improved substitution. I think this is what the old code really wanted. The only thing lost here is the jittered delay, but I think we'd prefer yielding to the async runtime vs. manually polling anyway.

bmc-msft: I agree with your assessment.

src/agent/onefuzz/src/monitor.rs
bmc-msft: It would make the code more consistent, but the impact is minimal.

This has no mechanism for communicating errors beyond having the channel close.  From the `DirectoryMonitor` callers perspective, this looks the same as the directory being deleted.

src/agent/onefuzz-agent/src/local/common.rs
chkeita: looks good

src/agent/onefuzz/src/monitor.rs
```diff
@@ -1,33 +1,39 @@
...
         let dir = dir.into();
-        let (tx, rx) = mpsc::channel();
+        let (notify_sender, notify_receiver) = sync::mpsc::channel();
         let delay = Duration::from_millis(100);
-        let watcher = notify::watcher(tx, delay).unwrap();
+        let watcher = notify::watcher(notify_sender, delay).unwrap();
+
+        // We can drop the thread handle, and it will continue to run until it
+        // errors or we drop the async receiver.
+        let (notify_events, _handle) = into_async(notify_receiver);
```
chkeita, line 32: do we need this indirection ? Why not just read from the the `notify_sender` ?

ranweiler: `notify_receiver` is not async, so to block on receiving messages (and thus receive OS filesystem notification pushes), we need to do it off-thread. The `notify::Watcher` requires a `std::sync::mpsc` channel, so we can't just pass an async `Sender` to the `Watcher`.

chkeita: got it.
In that case i think we should use [try_rcv](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.try_recv) in the implementation of `into_async` to allow us to yield the task when there is nothing instead of blocking

chkeita: Actually I think we can even eliminated the spawned task and channel by using a logic similar to the code that you deleted in `src/agent/onefuzz-agent/src/local/common.rs` .

src/agent/onefuzz/src/monitor.rs
ranweiler: Oh, only just now saw this comment. I changed it anyway because it literally only required renaming.

src/agent/onefuzz/src/monitor.rs
ranweiler: We are still yielding, because we only block in a worker thread.

This lets us (1) use the FS notify-push blocking API that we get from the `notify` crate in a way that doesn't block our async scheduling, but (2) use an async-aware `UnboundedReceiver` in the impl of `DirectoryMonitor::next_file()` (which is now an `async` method). So our API _does_ yield to the async runtime if there are no new messages, as desired.

bmc-msft (approved):

src/agent/onefuzz/src/monitor.rs
chkeita: As per our discussions. The goal with my proposed change is remove the allocation of an OS thread which could potentially be blocking. 
My assumption is that by using the tokio task we consume much less resource that allocating the OS thread.

ranweiler: Agreed. I'll revisit this again to try and bridge the sync/async receivers with less overhead.

PR #1080, opened 2021-07-16
add a short sleep in the supervisor's event loop to reduce CPU usage

bmc-msft:
Without this sleep, once the `Agent.run` loop is in the `busy` state, it is turned into a tight loop calling `wait4` of the running agents.  This sleep adds a small window to allow the rest of the system to work.  Empirical testing shows this has a significant reduction in CPU use.

This sleep should be removed once the `worker_runner` monitoring is turned into something event driven.


bmc-msft:
Note, this is based off of #1078 , which should be merged before this PR.

The only file with changes unique to this PR is `src/agent/onefuzz-supervisor/src/agent.rs`

ranweiler (approved):

chkeita (approved):

PR #1084, opened 2021-07-19
Replace notifications by default

bmc-msft:
This PR adds two features:

1. On creating notifications for containers that already have notification configs, this replaces the existing notifications by default
2. Instead of returning all containers, then iterating over all containers, enable searching the notifications by container name

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -692,11 +692,17 @@ class Notifications(Endpoint):
     endpoint = "notifications"
 
     def create(
-        self, container: primitives.Container, config: models.NotificationConfig
+        self,
+        container: primitives.Container,
+        config: models.NotificationConfig,
+        *,
+        replace_existing: bool = True,
     ) -> models.Notification:
         """Create a notification based on a config file"""
 
-        config = requests.NotificationCreate(container=container, config=config.config)
+        config = requests.NotificationCreate(
+            container=container, config=config.config, replace_existing=replace_existing
+        )
         return self._req_model("POST", models.Notification, data=config)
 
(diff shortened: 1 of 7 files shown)


src/cli/onefuzz/api.py
```diff
@@ -692,11 +692,17 @@ class Notifications(Endpoint):
     endpoint = "notifications"
 
     def create(
-        self, container: primitives.Container, config: models.NotificationConfig
+        self,
+        container: primitives.Container,
+        config: models.NotificationConfig,
+        *,
+        replace_existing: bool = True,
```
chkeita, line 699: I think the default should be false to prevent accidentally deleting in the case where multiple notification are created on a container

bmc-msft, line 699:
```suggestion
        replace_existing: bool = False,
```

bmc-msft: Addressed per conversation

src/pytypes/onefuzztypes/requests.py
```diff
@@ -41,6 +41,11 @@ class JobSearch(BaseRequest):
 
 class NotificationCreate(BaseRequest, NotificationConfig):
     container: Container
+    replace_existing: bool = Field(default=True)
```
chkeita, line 44: I suggest also setting this default to `False`

chkeita (approved):

src/pytypes/onefuzztypes/requests.py
bmc-msft:
```suggestion
    replace_existing: bool = Field(default=False)
```

PR #1085, opened 2021-07-19
delete secret on object delete

bmc-msft:
As is, when we delete a notification, we don't delete the secrets saved in KeyVault.

This wires in the secrets being deleted upon deleting the parent object.


chkeita (approved):

src/api-service/tests/test_secrets.py
```diff
@@ -36,7 +38,72 @@ def hider(secret_data: SecretData) -> SecretData:
...
+
+
+class Nested(BaseModel):
+    a: WithSecret
+    b: WithDict
+    c: WithList
+
+
 class TestSecret(unittest.TestCase):
+    def test_delete(self) -> None:
```
ranweiler, line 61: 🙏

src/api-service/__app__/onefuzzlib/secrets.py
```diff
@@ -80,3 +80,8 @@ def delete_secret(secret_url: str) -> None:
     (vault_url, secret_name) = parse_secret_url(secret_url)
     keyvault_client = get_keyvault_client(vault_url)
     keyvault_client.begin_delete_secret(secret_name)
+
+
+def delete_secret_data(data: SecretData) -> None:
```
ranweiler, line 85: Maybe it would be even clearer to name this something like:
```suggestion
def delete_remote_secret_data(data: SecretData) -> None:
```

src/api-service/__app__/onefuzzlib/orm.py
```diff
@@ -249,6 +249,27 @@ def hide_secrets(data: B, hider: Callable[[SecretData], SecretData]) -> B:
     return data
 
 
+def delete_secrets(data: B, deleter: Callable[[SecretData], None]) -> None:
```
ranweiler, line 252: This doesn't actually delete unless the callback does deletion, right?

Maybe this should just be:
```suggestion
def visit_secrets(data: B, visitor: Callable[[SecretData], None]) -> None:
```

Then the meaning is clear at the use sites, which might look like:
```python
visit_secrets(self, delete_remote_secret_data)
```

ranweiler (approved):

src/api-service/__app__/onefuzzlib/orm.py
ranweiler: Since we're not renaming this, I suggest just adding a function/doc comment that emphasizes that is is the responsibility of the `deleter` callback to perform any actual deletion.

ranweiler (approved):

PR #1090, opened 2021-07-20
use enum for local commands

bmc-msft:
As is, we use static strings for sub-command names.  Moving to an enum means we can rely on the compiler to verify we've implemented everything correctly for the subcommands.


ranweiler (approved):
We may also consider migrating to `structopt` in the future, too. But that's a bigger change, and this is big improvement in the meantime.

Issue #925, opened 2021-05-25
Ensure `{input}` is present in `target_options` or `target_env`


ranweiler:
Enforce this at the service level, at least.

PR #1097, opened 2021-07-21
verify `{input}` is used by target_env or target_options in coverage task

bmc-msft:


src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -245,6 +249,23 @@ impl<'a> TaskContext<'a> {
...
+        let input = PlaceHolder::Input.get_string();
+
+        for entry in &self.config.target_options {
+            if entry == &input {
+                return true;
+            }
+        }
+        for (k, v) in &self.config.target_env {
+            if k == &input || v == &input {
+                return true;
```
chkeita, line 262: We should use contains instead of equality for the check to match the logic in expand and cover the case where the placeholder could be in quotes.
```suggestion
        for entry in &self.config.target_options {
            if entry.contains(&input) {
                return true;
            }
        }
        for (k, v) in &self.config.target_env {
            if k == &input || v.contains(&input) {
                return true;
```

chkeita (approved):

src/agent/onefuzz/src/expand.rs
```diff
@@ -469,4 +469,13 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_expand_in_string() -> Result<()> {
+        let result = Expand::new()
+            .input_path("src/lib.rs")
+            .evaluate_value("a {input} b")?;
+        assert!(result.contains("src/lib.rs"));
```
bmc-msft, line 478:
```suggestion
        assert!(result.contains("lib.rs"));
```

ranweiler:
See also: #925. I think we'd prefer to validate this at the service layer, though, when creating the task.

bmc-msft:
@ranweiler, this was more specifically to check for the local fuzzing mode.  I agree the service should handle this on task creation which is why I left #925 open.

Issue #1095, opened 2021-07-20
Node reimaged due to expired heartbeat


jagunter:
## Information

- Onefuzz version: 2.27.0
- OS: Windows

## Provide detailed reproduction steps (if any)

Scheduled job `2056f74a-d2c3-478b-a6f5-e8335d33543c` to run for 1 week. However after a day it was stopped with the error:
    
```
    {"code": 468, "errors": ["node reimaged due to expired heartbeat", "scaleset_id:8ab21c8b-a504-4d78-b185-681c55d00fe9 machine_id:97da016a-10db-4b25-bd0a-d4830727bbbf", "last heartbeat:2021-07-20 21:23:11+00:00"]}
``` 

### Expected result

Task to run to completion

### Actual result

Node to which task was assigned was reimaged before task completion.

bmc-msft:
This was due to a "503 Service Unavailable" error coming back from the service after multiple retries.

Investigating why the service had issues.

bmc-msft:
The Azure Functions integration into SignalR failed, which caused the agent-commands to fail, which caused the supervisor to stop.

At the moment, we're using the Azure Functions integration for SignalR, rather than handling the communication ourselves.  

We'll investigate making the supervisor more resilient to the service having issues as well as making the SignalR integration more resilient.

PR #1098, opened 2021-07-21
enable the supervisor to handle longer service outages

bmc-msft:
The current API retry code at the agent level will fail if the service is down for more than 25 seconds.  In the case of longer outages such as those seen in #1095, it would be useful if the supervisor was more resilient while running existing tasks.

This PR enables the supervisor's `poll commands` and `claim commands` interactions with the service to be fallible at the transport layer (https), but not at the content layer (json).


chkeita (approved):

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -202,10 +202,18 @@ impl Coordinator {
...
-            .await
-            .context("PollCommands")?
+        let response = self.send_request(request).await.context("PollCommands");
+
+        // Treat communication issues with the service as `no commands
+        // available`.  This adds resilency to the supervisor during longer
+        // outages.  Given poll_commands runs on a 10 second cycle, this should
+        // provide eventual recovery.
+        if let Err(response) = response {
+            error!("error polling the serivce for commands: {:?}", response);
```
ranweiler, line 212:
```suggestion
            error!("error polling the service for commands: {:?}", response);
```

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -202,10 +202,18 @@ impl Coordinator {
...
         let request = self.client.get(url).json(&request);
 
-        let pending: PendingNodeCommand = self
-            .send_request(request)
-            .await
-            .context("PollCommands")?
+        let response = self.send_request(request).await.context("PollCommands");
+
+        // Treat communication issues with the service as `no commands
+        // available`.  This adds resilency to the supervisor during longer
```
ranweiler, line 208:
```suggestion
        // available`.  This adds resiliency to the supervisor during longer
```

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -202,10 +202,18 @@ impl Coordinator {
...
+        // outages.  Given poll_commands runs on a 10 second cycle, this should
+        // provide eventual recovery.
+        if let Err(response) = response {
+            error!("error polling the serivce for commands: {:?}", response);
+            return Ok(None);
```
ranweiler, line 213: I feel uneasy about equivocating "nothing pending" with "can't talk to the service".

Can we define and return a new concrete error type, and have callers inspect that? This would let the caller know that a backoff or even hard failure is warranted after some limit that they can control.

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -219,7 +227,14 @@ impl Coordinator {
             let url = self.registration.dynamic_config.commands_url.clone();
             let request = self.client.delete(url).json(&request);
 
-            self.send_request(request).await.context("ClaimCommand")?;
+            let response = self.send_request(request).await.context("ClaimCommand");
+
+            // similar polling available commands, this treats issues claiming
```
ranweiler, line 232: This seems a little more tricky than the above, particularly if were to happen repeatedly. I could imagine this getting a supervisor into a situation where it would infinitely pop a command, fail to claim it, then re-pop the command message after the visibility timeout.

src/agent/onefuzz-supervisor/src/coordinator.rs
bmc-msft: addressed

src/agent/onefuzz-supervisor/src/coordinator.rs
bmc-msft: addressed based on conversations offline.

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -172,6 +172,13 @@ impl ICoordinator for Coordinator {
     }
 }
 
+pub enum PollCommandResult {
+    None,
+    Command(NodeCommand),
```
ranweiler, line 177: The following are only style suggestions.

I had originally imagined introducing types something like the following:
```rust
// Ideally would impl `Error` via `thiserror`.
enum PollCommandError {
    RequestFailed(Error),
    ClaimFailed(Error),
    Other(Error),
}

// Would be returned by `poll_commands()`, as the `Ok` variant of `Result<PollCommandResult>`.
type PollCommandResult = Result<Option<NodeCommand>, PollCommandError>;
```

It is more verbose, but I think a bit more idiomatic in its use of `Result`/`Option`.

Short of that, it might be nice to merely reorganize this type as:
```rust
pub enum PollCommandResult {
    Pending(Option<NodeCommand>),
    RequestFailed(Error),
    ClaimFailed(Error),
}
```
IMO it is reasonable to factor some error cases into a non-`Result<_, _>` enum like we are, but it is still preferable not to flatten what is logically just an `Option`.

src/agent/onefuzz-supervisor/src/coordinator.rs
```diff
@@ -219,11 +233,17 @@ impl Coordinator {
             let url = self.registration.dynamic_config.commands_url.clone();
             let request = self.client.delete(url).json(&request);
 
-            self.send_request(request).await.context("ClaimCommand")?;
+            let response = self.send_request(request).await.context("ClaimCommand");
+
+            // similar polling available commands, this treats issues claiming
+            // commands as `no commands available`
+            if let Err(response) = response {
+                return Ok(PollCommandResult::ClaimFailed(response));
```
ranweiler, line 241: Add rationale here, e.g.:
```suggestion
                    // If we failed to claim two commands in a row, it means the service is up (since 
                    // we received the commands we're trying to claim), but something else is going 
                    // wrong, consistently. This is suspicious, and less likely to be a transient 
                    // service or networking error, so bail.
                    return Ok(PollCommandResult::ClaimFailed(response));
```

ranweiler (approved):

src/agent/onefuzz-supervisor/src/coordinator.rs
bmc-msft: added

src/agent/onefuzz-supervisor/src/coordinator.rs
bmc-msft: addressed

ranweiler (approved):

PR #1100, opened 2021-07-21
relay SignalR integrations through a storage queue

bmc-msft:
The SignalR integration from Azure Functions does not have automatic retry.  When the SignalR instance has issues, all other APIs fail.

To make the service resilient to SignalR outages, this bounces SignalR events through an Azure Storage queue.

NOTE: This PR does not remove the integration from all of the functions.  That is intended to be done as a follow-on PR.


src/api-service/__app__/onefuzzlib/events.py
```diff
@@ -5,34 +5,26 @@
...
 from .azure.creds import get_instance_id, get_instance_name
+from .azure.queue import send_message
+from .azure.storage import StorageType
 from .webhooks import Webhook
 
-EVENTS: Queue = Queue()
 
+def queue_signalr_event(event_message: EventMessage) -> None:
+    event = json.loads(event_message.json())
+    message = json.dumps({"target": "events", "arguments": [event]}).encode()
```
chkeita, line 22: how about defining a type for the envelope to reduce the number of conversion ?
``` python 
class SignalREvent(BaseModel):
    target: str
    arguments: List[EventMessage]
```

bmc-msft: Added.

chkeita (approved):

PR #1103, opened 2021-07-22
process all expired nodes rather than those not already marked for deletion

bmc-msft:
This makes sure debug_keep_node is reset and the rest of the reimage processing occurs regardless of reimage_requested and delete_requested being set.

Without this, nodes that are marked `debug_keep_node` do not get reimaged/deleted.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/workers/nodes.py
+++ b/src/api-service/__app__/onefuzzlib/workers/nodes.py
@@ -456,12 +456,9 @@ class Node(BASE_NODE, ORMMixin):
         time_filter = "Timestamp lt datetime'%s'" % (
             (datetime.datetime.utcnow() - NODE_REIMAGE_TIME).isoformat()
         )
-        # skip any nodes already marked for reimage/deletion
         for node in cls.search(
             query={
                 "scaleset_id": [scaleset_id],
-                "reimage_requested": [False],
-                "delete_requested": [False],
             },
             raw_unchecked_filter=time_filter,
         ):
--- a/src/api-service/__app__/onefuzzlib/workers/pools.py
+++ b/src/api-service/__app__/onefuzzlib/workers/pools.py
@@ -227,7 +227,6 @@ class Pool(BASE_POOL, ORMMixin):
             delete_queue(self.get_pool_queue(), StorageType.corpus)
             ShrinkQueue(self.pool_id).delete()
             logging.info("pool stopped, deleting: %s", self.name)
-            self.state = PoolState.halt
             self.delete()
             return
 


src/api-service/__app__/onefuzzlib/workers/pools.py
```diff
@@ -227,7 +227,6 @@ def halt(self) -> None:
             delete_queue(self.get_pool_queue(), StorageType.corpus)
             ShrinkQueue(self.pool_id).delete()
             logging.info("pool stopped, deleting: %s", self.name)
-            self.state = PoolState.halt
```
ranweiler, line 230: Why do we need to stop doing this?

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -456,12 +456,9 @@ def reimage_long_lived_nodes(cls, scaleset_id: UUID) -> None:
         time_filter = "Timestamp lt datetime'%s'" % (
             (datetime.datetime.utcnow() - NODE_REIMAGE_TIME).isoformat()
         )
-        # skip any nodes already marked for reimage/deletion
         for node in cls.search(
             query={
                 "scaleset_id": [scaleset_id],
-                "reimage_requested": [False],
```
ranweiler, line 463: IIUC: when `debug_keep_node` is set for a node, it would not show up as a result for this query. This is because `debug_keep_node` (apparently?) implies that either `reimage_requested` or `delete_requested` is `True`. This query requires that both `_requested` attrs are `False`, so the `debug_keep_node` node would be filtered out.

Where does this happen?

I'm a big confused, and maybe misreading something. I'm mostly seeing us _skip_ `delete()` calls if `debug_keep_node` is set, so I'd expect the `_requested` fields to be `False` already.

src/api-service/__app__/onefuzzlib/workers/pools.py
bmc-msft: Immediately below this, we call `self.delete()`  This is effectively a NOOP.

src/api-service/__app__/onefuzzlib/workers/nodes.py
bmc-msft: Consider:

1. Task starts on node A.
2. User sets `debug_keep_node` for node A.
3. Task finishes on node A.  At this point, the system sets `reimage_requested` to True.

Currently, the scaleset management reimaging does something equivalent to:
```
for node in Nodes.search(reimage_requested=True):
    if node.debug_keep_node:
        continue
    vmss.reimage_node(node)
```

We could solve this in one of two fashions, either:
1. Update the reimage/delete code to *also* query based on timestamp and ignore `debug_keep_node` based on node TTL windows.
2. Make the `mark old nodes as to_reimage` also reset `debug_keep_node`.

Consider this code in a larger context:

https://github.com/microsoft/onefuzz/blob/56675029107111332baca94163ee5ae109517748/src/api-service/__app__/onefuzzlib/workers/nodes.py#L459-L473

The new code checks all nodes that are expired, resets the `debug_keep_node` if set, and sets the node to reimage/delete.  If the node is already set to reimage/delete, `to_reimage()` is effectively a noop.  However, this will always unset `debug_keep_node`.

src/api-service/__app__/onefuzzlib/workers/pools.py
ranweiler: That makes more sense. So this incidental cleanup, unrelated to the larger fix.

src/api-service/__app__/onefuzzlib/workers/nodes.py
ranweiler: I see now, thank you! So the the flow that _typically_ led us to the bug was actually the reimaging case, specifically.

ranweiler (approved):

PR #1104, opened 2021-07-22
allow pools & scalesets set to `shutdown` to `halt`

bmc-msft:
Currently, if a pool or scaleset is set to `shutdown`, it cannot be set to `halt`.

While moving from `halt` to `shutdown` would cause issues, moving from `shutdown` to `halt` is fine.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/workers/pools.py
+++ b/src/api-service/__app__/onefuzzlib/workers/pools.py
@@ -135,8 +135,7 @@ class Pool(BASE_POOL, ORMMixin):
     def init(self) -> None:
         create_queue(self.get_pool_queue(), StorageType.corpus)
         ShrinkQueue(self.pool_id).create()
-        self.state = PoolState.running
-        self.save()
+        self.set_state(PoolState.running)
 
     def schedule_workset(self, work_set: WorkSet) -> bool:
         # Don't schedule work for pools that can't and won't do work.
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/workers/pools.py
```diff
@@ -239,6 +230,17 @@ def halt(self) -> None:
 
         self.save()
 
+    def set_state(self, state: PoolState) -> None:
+        if self.state == state:
+            return
+
+        # scalesets should never leave the `halt` state
```
ranweiler, line 237: This feels a little off, in that I expect `set_state()` to only be responsible for (1) avoiding spurious "updates" of cycle transitions, and (2) managing saving remote state.

Here, we're also dropping requests for illegal state transitions. It'd be nice if we had a higher-level method that was responsible for such enforcement, then called this method to execute the actual state change.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -635,22 +635,11 @@ def reimage_nodes(self, nodes: List[Node]) -> None:
                 node.delete()
 
     def set_shutdown(self, now: bool) -> None:
-        if self.state in [ScalesetState.halt, ScalesetState.shutdown]:
-            return
-
-        logging.info(
```
ranweiler, line 641: Do we really want to discard this logging?

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -807,6 +796,10 @@ def set_state(self, state: ScalesetState) -> None:
         if self.state == state:
             return
 
+        # scalesets should never leave the `halt` state
```
ranweiler, line 799: Apply comment about `Pool.set_state()`. We're being consistent, at least.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
bmc-msft: set_state includes logging because of the event generation

src/api-service/__app__/onefuzzlib/workers/pools.py
ranweiler: Per offline discussion, we'll revisit this concern holistically by investigating a more structured use of state machines.

ranweiler (approved):

Issue #264, opened 2020-11-04
Store verbose, bounded agent logs on VM nodes


ranweiler:
The agents generate log telemetry, which we log in moderation to avoid excessive bandwidth and storage account usage.

However, higher log levels are extremely useful for debugging. Fan out local copies of higher log levels from the supervisor and worker agents on individual nodes. Ensure the logs are bounded, to avoid filling up the disk. If possible, use existing OS mechanisms that provide this out of the box.

chkeita:
addressed in #1721

PR #1108, opened 2021-07-28
record supervisor stdout & stderr and report it on failure

bmc-msft:
This adds unbounded capture of stdout and stderr of the supervisor to disk.  On restarting the supervisor after failure, the supervisor will read all available logs and provide them as context to the service to help debug the failures.

While we want to eventually include bounded logs (see #264), currently we are experiencing an issue where the supervisor panics in some situations without recording logs effectively.

This is a stop-gap that should eventually be replaced by bounded logging.


src/runtime-tools/linux/run.sh
```diff
@@ -19,6 +19,7 @@ echo core | sudo tee /proc/sys/kernel/core_pattern
 echo 0 | sudo tee /proc/sys/kernel/randomize_va_space
 # set core dumping to default behavior
 echo 1 | sudo tee /proc/sys/fs/suid_dumpable
+RUNID=$(uuidgen)
```
ranweiler, line 22: I double-checked, and confirmed that this is also installed on a stock Ubuntu 20.04 image. ✔️

ranweiler (approved):

bmc-msft:
This doesn't capture the error codes/signals used on program exit.  A follow on PR will address this.

PR #1109, opened 2021-07-29
redirect output of supervisor by executing a child process

bmc-msft:
This adds unbounded capture of stdout and stderr of the supervisor to disk.  On restarting the supervisor after failure, the supervisor will read all available logs and provide them as context to the service to help debug the failures.

While we want to eventually include bounded logs (see #264), currently we are experiencing an issue where the supervisor panics in some situations without recording logs effectively.

This is a stop-gap that should eventually be replaced by bounded logging.

Unfortunately, the previous effort (#1108) did not capture the state if the process was terminated via signal, etc.   Now, if the user provides `--redirect-output <FILE>`, the supervisor will fork a child redirecting stdout/stderr to file.   If the program terminates with an error, we append the error to a file.

For example, if you stop the supervisor using `stop-process <PID>` on windows, we'll now see the following in `c:\onefuzz\logs\onefuzz-supervisor-<GUID>.txt`:

```
[2021-07-29T20:21:10Z INFO  onefuzz_supervisor::agent] no work available
[2021-07-29T20:21:40Z INFO  onefuzz_supervisor::agent] no work available
[2021-07-29T20:22:10Z INFO  onefuzz_supervisor::agent] no work available
onefuzz-supervisor child failed: ExitStatus { code: Some(-1), signal: None, success: false }
```


src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -84,12 +90,58 @@ fn version() {
...
 fn licenses() -> Result<()> {
-    use std::io::{self, Write};
     io::stdout().write_all(include_bytes!("../../data/licenses.json"))?;
     Ok(())
 }
 
+fn redirect(opt: RunOpt) -> Result<()> {
+    let log_path = opt
+        .redirect_output
+        .expect("redirect should only be called with log_path");
```
chkeita, line 100: We can just pass the redirect output as a parameter and avoid the check here.

chkeita (approved):

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -14,6 +14,13 @@ function log ($message) {
   Write-Host -ForegroundColor Yellow $timestamp $message @args
 }
 
+function Setup-Silent-Notification {
+   # https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/registry-entries-for-silent-process-exit
```
ranweiler, line 18: 👍

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -14,6 +14,13 @@ function log ($message) {
...
+   log "installing registry key for silent termination notification of onefuzz-supervisor"
+   reg import c:\onefuzz\tools\win64\onefuzz-silent-exit.reg
```
ranweiler, line 20: I see where this is committed, but maybe we should also add a script to re-generate it, with comments that would naturally document the expected contents.

src/agent/onefuzz-supervisor/src/main.rs
```diff
@@ -84,12 +90,58 @@ fn version() {
...
+    info!("redirecting output: {}", log_path.display());
+
+    let stdout = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(&log_path)
+        .context("unable to open log file")?;
+    let stderr = stdout.try_clone().context("unable to clone log handle")?;
```
ranweiler, line 108: So, we're writing interleaving stderr/stdio. Maybe it would be better to write them to separate log files.

ranweiler: (maybe better as a separate PR?)

ranweiler (approved):

src/agent/onefuzz-supervisor/src/main.rs
bmc-msft: Updated.

src/agent/onefuzz-supervisor/src/main.rs
bmc-msft: Given we need to pull out other arguments out of `opt`, that felt like it reduced clarity.

src/runtime-tools/win64/onefuzz.ps1
bmc-msft: This was hand built using regedit following the above documentation for format/guidance.

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -14,6 +14,13 @@ function log ($message) {
...
+   log "done import registry key"
```
ranweiler, line 21:
```suggestion
   log "done importing registry key"
```

src/runtime-tools/win64/onefuzz.ps1
ranweiler: Got it. I also just looked at the file in a non-diff view, and it is more self-describing than I expected, so this is fine as-is.

ranweiler: One last question about this: we set the `ReportingMode` to `0x4`, which is documented as "When silent exit is detected, a pop-up notification is displayed."

Why is that the desired behavior? It seems very oriented towards interaction with a logged-on user. Does the same setting govern some other functionality?

bmc-msft: If there is a local user logged in, the local user will get a popup.  It's always logged though.

PR #1120, opened 2021-08-02
don't overload `list` builtin

bmc-msft:
This removes overloading the `list` built-in.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/containers.py
+++ b/src/api-service/__app__/onefuzzlib/azure/containers.py
@@ -140,7 +140,7 @@ def create_container(
         read=True,
         write=True,
         delete=True,
-        list=True,
+        list_=True,
     )
 
 
@@ -164,7 +164,7 @@ def get_container_sas_url_service(
     read: bool = False,
     write: bool = False,
     delete: bool = False,
-    list: bool = False,
+    list_: bool = False,
     delete_previous_version: bool = False,
     tag: bool = False,
 ) -> str:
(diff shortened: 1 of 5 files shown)


src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -164,7 +164,7 @@ def get_container_sas_url_service(
     read: bool = False,
     write: bool = False,
     delete: bool = False,
-    list: bool = False,
+    list_: bool = False,
```
chkeita, line 167: Here is an alternative proposal but i am ok with these changes.
since we have `ContainerPermission` type we can use that instead 

``` python
def get_container_sas_url_service(
    client: ContainerClient,
    *,
     permissions: List[ContainerPermission]:
    
    ...
    sas = generate_container_sas(
            account_name,
            container_name,
            account_key=account_key,
            permission=ContainerSasPermissions(
            read = ContainerPermission.Read in  permissions,
            write =ContainerPermission.Write in  permissions,
            delete =ContainerPermission.Delete in  permissions, 
            list = ContainerPermission.Delete in  permissions.
            delete_previous_version=delete_previous_version,
            tag=tag,
        ),

    
```

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/containers.py
bmc-msft: I think that would be useful, though should be a follow-up PR.

PR #1129, opened 2021-08-10
don't retry on service-level errors

bmc-msft:
Service level errors respond with a json dictionary of the format: `{code: int, errors: List[str]}`.

For 1f purposes, we should never retry on service level errors.

### DIFF
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -260,6 +260,20 @@ class Backend:
                 if response.status_code not in retry_codes:
                     break
 
+                if response.status_code == 401:
+                    try:
+                        as_json = json.loads(response.content)
+                        if (
+                            isinstance(as_json, dict)
+                            and "code" in as_json
+                            and "errors" in as_json
+                        ):
+                            raise Exception(
+                                f"request failed: application error - {as_json['code']} {as_json['errors']}"
+                            )
+                    except json.decoder.JSONDecodeError:
+                        pass
+
                 LOGGER.info("request bad status code: %s", response.status_code)
             except requests.exceptions.ConnectionError as err:
                 LOGGER.info("request connection error: %s", err)
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -220,6 +220,21 @@ class Backend:
         self.save_cache()
         return access_token
 
+    def check_application_error(self, response: requests.Response) -> None:
+        if response.status_code == 401:
+            try:
+                as_json = json.loads(response.content)
+                if (
+                    isinstance(as_json, dict)
+                    and "code" in as_json
+                    and "errors" in as_json
+                ):
+                    raise Exception(
+                        f"request failed: application error - {as_json['code']} {as_json['errors']}"
+                    )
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/backend.py
```diff
@@ -260,6 +260,20 @@ def request(
                 if response.status_code not in retry_codes:
                     break
 
+                if response.status_code == 401:
+                    try:
+                        as_json = json.loads(response.content)
+                        if (
```
ranweiler, line 266: I assume there's no reasonable way to reuse the models for this (e.g. a shared base class that defines the required fields)

In the absence of that, can we just pull this checking logic out into something like a `looks_like_application_error()` predicate?

ranweiler (approved):

src/cli/onefuzz/backend.py
bmc-msft: Moved the logic into check_application_error

src/cli/onefuzz/backend.py
```diff
@@ -220,6 +220,21 @@ def check_msal_error(value: Dict[str, Any], expected: List[str]) -> None:
         self.save_cache()
         return access_token
 
+    def check_application_error(self, response: requests.Response) -> None:
```
ranweiler, line 223: This is a nice improvement.

I'm going to elaborate on what I had in mind for the sake of review, but do feel free to take it or leave it.

1. We don't use `self` in any way, nor should we need to, semantically. So this could just be a function.
2. I imagined this as a predicate with no side-effects (here, throwing).

So this could be impl'd as:
```python
def is_application_error(response: Response) -> bool:
    if response.status_code == 401:
        try:
            as_json = json.loads(response.content)
            if (
                isinstance(as_json, dict)
                and "code" in as_json
                and "errors" in as_json
            ):
                return True
        except json.decoder.JSONDecodeError:
            pass

    return False
```
This could be easily unit-tested (though what you haven't isn't really any harder to test, if a function).

Then the _caller_ could choose to throw if an app error:
```python
if is_application_error(response):
    as_json = json.loads(response.content)
    raise Exception(f"request failed: application error - {as_json['code']} {as_json['errors']}")
```
I think it is fine to just re-parse the body in the exceptional case, though you could return it if you cared. In that case I'd suggest the function still be pure, but be called something like `parse_application_error()` and return an `Optional[dict[str, str]]`.

ranweiler (approved):

src/cli/onefuzz/backend.py
ranweiler: I'll note that there is an established pattern of effectful `check_` functions in this codebase, so although I'd strongly encourage suggestion (1), I don't think declining (2) is unreasonable in context, although I don't prefer it.

bmc-msft: This mirrors the pattern of another function in this same file, `check_msal_error` almost exactly.

As such, I'd prefer to leave them alone rather than handling similarly scoped errors in two different ways or rewriting both as part of this PR.

I hear your feedback that it doesn't touch self and promoted it out of `Backend` as well as made `check_msal_error` not a nested function.

PR #1142, opened 2021-08-11
add srcview crate

jopletchMSFT:

`srcview` is a crate which ingests debug info (currently only PDB) and execution traces and then generates source/line based coverage report which can be visualized elsewhere (ADO, VSCode). Currently the only supported report format is Cobertura. The intent is that fuzzer authors can use these reports to identify gaps in their coverage and address them, as well as track the fuzzers efficacy over time.

## PR Checklist
* [x] Tests added/passed

## Info on Pull Request

The crate is reasonably documented -- `cargo doc --open` is probably the best current source of info.

This PR includes all the standalone srcview code, documentation, and tests. It does _not_ include any integration work with the rest of the 1f repo.

In terms of specific integration work I'm aware of:
- We will  need to expose the allow and denylist regex's to the user via fuzzer job configuration
- While we should be able to convert between coverage formats via `iter.map(|(module, offset)| ModOff::new(module, offset)).collect()`, we might want to introduce a ModOff collection type instead of just using a Vec. Open to feedback here.

## Validation Steps Performed

The examples and documentation show usage and validation.

### DIFF
--- a/src/agent/srcview/tests/srcview.rs
+++ b/src/agent/srcview/tests/srcview.rs
@@ -1,0 +1,67 @@
+use std::env;
+use std::path::PathBuf;
+
+use srcview::{ModOff, SrcLine, SrcView};
+
+fn test_srcview() -> SrcView {
+    let root = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let pdb_path: PathBuf = [&root, "res", "example.pdb"].iter().collect();
+
+    let mut srcview = SrcView::new();
+    srcview.insert("example.exe", pdb_path).unwrap();
+
+    srcview
+}
+
+#[test]
+fn modoff() {
+    let srcview = test_srcview();
+
+    let good_modoff = ModOff::new("example.exe", 0x6f70);
+    assert_eq!(
+        srcview.modoff(&good_modoff),
+        Some(SrcLine::new("E:\\1f\\coverage\\example\\example.c", 3))
+    );
+
+    let bad_offset = ModOff::new("example.exe", 0x4141);
+    assert_eq!(srcview.modoff(&bad_offset), None);
+
+    let bad_module = ModOff::new("foo.exe", 0x4141);
+    assert_eq!(srcview.modoff(&bad_module), None);
+}
+
+#[test]
+fn symbol() {
+    let srcview = test_srcview();
+
+    let good: Vec<&SrcLine> = srcview.symbol("example.exe!main").unwrap().collect();
+
+    assert_eq!(
+        good,
+        vec![
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 3),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 4),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 5),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 6),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 7),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 10),
+            &SrcLine::new("E:\\1f\\coverage\\example\\example.c", 11),
+        ]
+    );
+
+    assert!(srcview.symbol("dosenotexist").is_none());
+}
+
+#[test]
+fn path() {
+    let srcview = test_srcview();
+
+    let good: Vec<usize> = srcview
+        .path_lines("E:\\1f\\coverage\\example\\example.c")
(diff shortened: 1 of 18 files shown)


src/agent/srcview/tests/srcview.rs
```diff
@@ -0,0 +1,67 @@
+use std::env;
+use std::path::PathBuf;
+
+use srcview::{ModOff, SrcLine, SrcView};
+
+fn test_srcview() -> SrcView {
+    let root = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let pdb_path: PathBuf = [&root, "res", "example.pdb"].iter().collect();
```
ranweiler, line 8: I think we want to avoid committing binary files (here, the `.pdb`) to the repo. This is a pain for your tests, but we don't have a great solution for this yet. Our approach has been to build binary artifacts on-demand in CI, but that doesn't work for some of these _very_ binary-oriented coverage scenarios.

There's one existing example where we have tests that use a fixed binary. We gate with a build flag: https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-agent/src/tasks/fuzz/supervisor.rs#L290

And then use an env var to look up the binary: https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz-agent/src/tasks/fuzz/supervisor.rs#L298

That integration test is currently disabled. We need to revisit binary test artifact management for it and some other tests, too. For example, I have done some manual binary analysis for sancov-related things, and I'd like to have regular integration tests that require fixed binaries.

For now, WDYT about just gating the PDB-dependent integration tests behind a feature flag, and opening an issue to properly address this?

ranweiler (commented):
We need to add this header to the top of all source files:

```
// Copyright (c) Microsoft Corporation.
```

src/agent/srcview/tests/srcview.rs
ranweiler: (Opened: #1143)

jopletchMSFT: I can do this, but I'm not sure I understand the suggestion.

I can feature gate the PDB dependent tests, but then I'm unclear if the PDB still gets checked in:
- If the PDB is checked in, I'm not sure what purpose the feature gate has as the PDB still ends up in the repo.
- If the PDB is not checked in, then I very much dislike it and would prefer to delete the tests -- the tests are only meaningful when combined with the reference PDB, and without it I would say are worse than not existing.

While I'm pretty sure this is understood, without the PDB there would effectively be no tests beyond compilation in the doc examples.

ranweiler: I was just thinking we should _not_ check in the PDB, but add a binary artifact integration test feed in a future PR (closing #1143), and properly re-enable your tests then. The PDB you use would then get added to whatever feed we use for binary test artifacts. So, we would have some unused (in CI) tests for a bit, but not forever. And in the interim, tests that can only be run manually still seem valuable. We'd probably just want to document e.g. the SHA256 hash of the test PDB.

Does something like that seem reasonable?

jopletchMSFT:
I don't quite understand the current CI failure:
```
+ cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2020-0016 --ignore RUSTSEC-2020-0036 --ignore RUSTSEC-2019-0036 --ignore RUSTSEC-2021-0065
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 323 security advisories (from /home/runner/.cargo/advisory-db)
    Updating crates.io index
error: not found: Couldn't load Cargo.lock: I/O operation failed: parse error: parse error: ambiguous dependency: nom
Error: Process completed with exit code 2.
```

the `src/agent/Cargo.lock` changes it presumably doesn't like as there are now two nom versions (5 and 6) present? Is this something you've encountered before?

src/agent/srcview/tests/srcview.rs
jopletchMSFT: Sure, the only part that's still unclear to me is where the PDB lives in the interim.

ranweiler: In the interim, I was thinking just on your workstation? Open to ideas. And if you really hate it, we could totally just remove the tests for now, and add them back later. Just trying to reduce tedium on your end.

jopletchMSFT:
Attaching example.pdb here because I cannot keep track of files to save my life.

[example.pdb.txt](https://github.com/microsoft/onefuzz/files/6971392/example.pdb.txt)

ranweiler (approved):

PR #1146, opened 2021-08-11
add POC srcview binary 

bmc-msft:
This PR adds a POC executable for the `srcview` that we could release.  This PR is more of a starting point for discussion than what we should land on.

This adds the following:
1. moves to using `anyhow` for errors.  The ability to add context to errors generated is extremely valuable.
2. uses `structopt` to perform argument parsing
3. bundles all of the existing examples as subcommands (primarily to validate the functionality works.)
4. includes a `licenses` subcommand, which as a requirement from CELA for all user-focused executables we release.
5. replaces the existing unwrap/expect/panic calls that inhibit using srcview as a library in a long-running application with bubbling out errors with `Result`.


src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,150 @@
...
+    srcview.insert(&sys, &opts.pdb_path)?;
+
+    // Convert our ModOffs to SrcLine so we can draw it
+    let coverage: Vec<SrcLine> = modoffs
+        .into_iter()
+        .filter_map(|m| srcview.modoff(&m))
+        .collect();
+
+    // Generate our report, filtering on our example path
+    let r = Report::new(&coverage, &srcview, Some(r"E:\\1f\\coverage\\example"))?;
```
bmc-msft, line 143: path filtering will need to get promoted into an argument.  

For future work, we might want to consider normalizing the path filtering code with the coverage library's path filtering code.

src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    // Format it as cobertura and display it
+    let formatted = r.cobertura(Some(r"E:\\1f\\coverage\\"))?;
```
bmc-msft, line 146: This path will need to get promoted to an argument as well.

src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,150 @@
...
+    stdout().write_all(include_bytes!("../../../data/licenses.json"))?;
+    Ok(())
+}
+
+fn dump_srcloc(opts: SrcLocOpt) -> Result<()> {
+    let modoff_data = fs::read_to_string(&opts.modoff_path)
+        .with_context(|| format!("unable to read modoff_path: {}", opts.modoff_path.display()))?;
+    let modoffs = ModOff::parse(&modoff_data)?;
+    let mut srcview = SrcView::new();
+    srcview.insert("example.exe", &opts.pdb_path)?;
```
bmc-msft, line 78: we probably want to make this an optional argument

jopletchMSFT: The way I handle this from the code that was dump_cobertura is probably a bit better -- creating four variants for bare, dll, exe, and sys. Im unsure if we should be strict (modoff must be "foo.exe+1234") or relaxed (we accept foo.exe+1234 and foo+1234). The latter is more common from the windbg case, but there are disambiguation problems if a process has both foo.exe and foo.dll in its address space.

jopletchMSFT: Thinking about this slightly more, youre right, it should just be an argument. There's no reason we'd want to enforce that pdb name match the module it describes, despite that being convention. This also solves us from having to create multiple pdbcache's per base name without refcounting or similar.

bmc-msft: I went with the model where it's an optional argument.  If a value is not provided, it uses your original method of file stem + known extensions.

src/agent/srcview/src/report.rs
```diff
@@ -198,10 +205,10 @@ impl Report {
 
             if let Some(root) = anc.last() {
                 // at this point we know we've computed this
-                let dircov = *self.dircov.get(root).unwrap();
-
-                // we don't really care if we're overwriting it
-                overall.insert(root.to_path_buf(), dircov);
+                if let Some(dircov) = self.dircov.get(root) {
```
ranweiler, line 208: It seems like `Some` is expected here. Should we warn if `None`?

src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use anyhow::{format_err, Context, Result};
+use srcview::{ModOff, Report, SrcLine, SrcView};
+use std::fs::{self};
```
jopletchMSFT, line 6: nit: is there a reason to use self here instead of `use std::fs` ?

src/agent/srcview/src/report.rs
```diff
@@ -347,7 +355,7 @@ impl Report {
 
         let unixtime = SystemTime::now()
             .duration_since(UNIX_EPOCH)
-            .expect("system time before unix epoch, wtf")
```
jopletchMSFT, line 350: :(

src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// Generate a Cobertuna XML coverage report
+///
+/// Example:
+///   srcview cobertuna ./res/example.pdb res/example.txt
+///             --include-regex "E:\\\\1f\\\\coverage\\\\"
+///             --filter-regex "E:\\\\1f\\\\coverage\\\\"
+///             --module-name example.exe
+///
+/// In this example, only files that live in E:\1f\coverage are included and
```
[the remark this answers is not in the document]
jopletchMSFT: nit: should we note that as these are regex's, backslashes must be double escaped (as this example does) to prevent them from being treated as backreferences?

jopletchMSFT (approved):
Nice changes 👍

src/agent/srcview/src/bin/srcview.rs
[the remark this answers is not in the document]
bmc-msft: @jopletchMSFT, you've nailed the reasoning.

The regular expression to match `e:\foo` is `e:\\foo`, as a single backslash is a configuration marker for regular expressions.  This can include back references, in the case of `\1`, or things like "any white space" in the case of `\s`.

Since this is coming from the shell, we also need to escape the shell's interpolation of \.

In practice:
* providing `e:\1f\coverage`  gives the regex engine "e:1fcoverage".
* providing `e:\\1f\\coverage` gives the regex engine  `e:\1f\coverage`, which interprets  `\1` is an undefined back reference.

To have what we want, a regex that matches `e:\1f\coverage`, you must provide `e:\\\\1f\\\\coverage`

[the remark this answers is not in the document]
bmc-msft: This is why I suggested investigating potentially reusing the coverage mechanism for selecting modules/etc.

src/agent/srcview/src/report.rs
bmc-msft: panic bad?

Also, we have to pass the `GeoPol` tool's analysis.   For reference: https://aka.ms/NExTGeoPol

src/agent/srcview/src/bin/srcview.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    Ok(())
+}
+
+fn licenses() -> Result<()> {
+    stdout().write_all(include_bytes!("../../../data/licenses.json"))?;
+    Ok(())
+}
+
+fn add_common_extensions(srcview: &mut SrcView, pdb_path: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
bmc-msft: I've added documentation on why it's used, but this isn't something we should be using in the context of TKO or 1f, where we'll have explicit knowledge of the modules of interest.

ranweiler (approved):

PR #1170, opened 2021-08-23
add command to give basic information on a pool

bmc-msft:
The command `onefuzz status pool pool-name` now provides something akin to:

```
❯ onefuzz status pool default
{
    "node_state": {
        "free": 5,
        "busy": 1
    },
    "tasks": {
        "312c725b-a3ca-4718-97b0-793b9a5ca27a": 1
    }
}
```

### DIFF
--- a/src/cli/onefuzz/cli.py
+++ b/src/cli/onefuzz/cli.py
@@ -490,18 +490,33 @@ class Builder:
             level += 1
 
 
+def render(result: Any) -> Any:
+    """Convert arbitrary result streams into something filterable with jmespath"""
+    if isinstance(result, BaseModel):
+        return render(result.dict(exclude_none=True))
+    if isinstance(result, Model):
+        return render(result.as_dict())
+    if isinstance(result, list):
+        return [render(x) for x in result]
+    if isinstance(result, dict):
+        return {render(k): render(v) for (k, v) in result.items()}
+    if isinstance(result, Enum):
+        return result.name
+    if isinstance(result, UUID):
+        return str(result)
+    if isinstance(result, (int, float, str)):
+        return result
+
+    logging.debug(f"unable to render type f{type(result)}")
+
+    return result
+
+
 def output(result: Any, output_format: str, expression: Optional[Any]) -> None:
     if isinstance(result, bytes):
         sys.stdout.buffer.write(result)
     else:
-        if isinstance(result, list) and result and isinstance(result[0], BaseModel):
-            # cycling through json resolves all of the nested BaseModel objects
-            result = [json.loads(x.json(exclude_none=True)) for x in result]
-        if isinstance(result, BaseModel):
-            # cycling through json resolves all of the nested BaseModel objects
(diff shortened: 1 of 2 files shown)


src/cli/onefuzz/cli.py
```diff
@@ -490,18 +490,33 @@ def print_nested_help(self, args: argparse.Namespace) -> None:
             level += 1
 
 
+def render(result: Any) -> Any:
```
ranweiler, line 493: So am I correct in reading the intent of this as: walk the argument, and turn it into something that is similar to what you'd get from a call to  `json.loads()`?

src/cli/onefuzz/status/cmd.py
```diff
@@ -20,9 +21,35 @@ def short(value: UUID) -> str:
...
 
+    def pool(self, *, name: str) -> PoolStatus:
+        pool = self.onefuzz.pools.get(name)
+        nodes = self.onefuzz.nodes.list(pool_name=pool.name)
+
+        tasks = {}
+        node_state = {}
+        for node in nodes:
+            if node.state not in node_state:
+                node_state[node.state] = 0
```
ranweiler, line 40: These are a good use case for `defaultdict`.

src/cli/onefuzz/status/cmd.py
```diff
@@ -20,9 +21,35 @@ def short(value: UUID) -> str:
     return str(value).split("-")[0]
 
 
+class PoolStatus(BaseModel):
+    node_state: Dict[NodeState, int]
```
ranweiler, line 25: Can we document the fields?

ranweiler (approved):

src/cli/onefuzz/status/cmd.py
bmc-msft: Unfortunately, pydantic makes `defaultdict` extremely unergonomic.

ranweiler: Ah, good to know.

src/cli/onefuzz/cli.py
bmc-msft: Yes.  The previous implementation was extremely naive.  It couldn't handle the results as provided by this instance, specifically `Dict[UUID, int]`.

ranweiler: Cool. Maybe just add a short comment emphasizing the sort of thing we're trying to build, for posterity? Based only on function name, I'd expect `render()` to return a string, when the real meaning here is more something like `as_json_like_dict()`.

src/cli/onefuzz/status/cmd.py
bmc-msft: added

src/cli/onefuzz/cli.py
bmc-msft: I renamed it to normalize.  It's not always a dict in response.  As an example, it can take a list of UUIDs and return a list of strings.

ranweiler (approved):

PR #1172, opened 2021-08-23
unless the user is using registration.py directly, no longer log credentials

bmc-msft:


src/deployment/deploy.py
```diff
@@ -894,23 +886,27 @@ def update_registration(self) -> None:
...
-        logger.info(
-            "Update your CLI config via: onefuzz config --endpoint "
-            "https://%s.azurewebsites.net --authority %s --client_id %s %s %s",
-            self.application_name,
-            self.cli_config["authority"],
-            self.cli_config["client_id"],
-            client_secret_arg,
-            multi_tenant_domain,
-        )
+            cmd += ["--tenant_domain", str(self.multi_tenant_domain)]
```
ranweiler, line 905: Not new, but this should be `"--multi_tenant_domain"`.

src/deployment/registration.py
```diff
@@ -402,6 +402,8 @@ def create_and_display_registration(
     registration_name: str,
     approle: OnefuzzAppRole,
     subscription_id: str,
+    *,
+    display_secret: bool = False,
```
ranweiler, line 406: We should document this.

ranweiler: And consider giving it a scarier name.

ranweiler (approved):

src/deployment/deploy.py
bmc-msft: no, it shouldn't.  The CLI doesn't take a `multi_tenant_domain`.   The CLI takes `tenant_domain`.

PR #1180, opened 2021-08-24
enable binding to the IP for a specific NIC on the proxy

bmc-msft:
This is an initial step towards enabling express route for the proxy


src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -236,6 +240,9 @@ def is_alive(self) -> bool:
         return True
 
     def get_forwards(self) -> List[Forward]:
+        if self.private_ip is None:
```
ranweiler, line 243: We call `get_forwards()` on `save_proxy_config()` and `is_used()`. Is there ever a time we want a `Proxy` object, but it is okay for `self.private_ip` to be `None`?

If not, would it be better to raise an error when we get the `private_ip` in the first place?

src/api-service/__app__/onefuzzlib/azure/ip.py
```diff
@@ -132,22 +132,21 @@ def create_public_nic(resource_group: str, name: str, location: str) -> Optional
     return None
 
 
-def get_public_ip(resource_id: str) -> Optional[str]:
+def get_public_private_ip(resource_id: str) -> Tuple[Optional[str], Optional[str]]:
```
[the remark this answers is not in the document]
ranweiler: That class (`dataclass`?) might in turn be a good place to document the different possible conditions wrt `None`, and when they are valid.

src/api-service/__app__/onefuzzlib/proxy.py
bmc-msft: When the Proxy object is initially created, but we're waiting for Azure to allocate the VMs, NICs, and IPs, the private_ip can be none.

PR #1216, opened 2021-08-31
add support for expanding `machine_id`

### LINKED ISSUE #1209: Request to include machine_id as a command replacement field
We're experimenting with recording custom metrics and traces. Since we're going through Application Insights directly, we don't have the traditional OneFuzz properties on the messages that we would normally have. We were able to get `job_id` and `task_id` through command replacements, but `machine_id` does not appear to be available. Inclusion of this would help track which machine from the same job+task was responsible for a particular metric or trace.

bmc-msft:
resolves #1209

### DIFF
--- a/src/agent/onefuzz/src/expand.rs
+++ b/src/agent/onefuzz/src/expand.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use crate::sha256::digest_file_blocking;
+use crate::{machine_id::get_machine_id, sha256::digest_file_blocking};
 use anyhow::{Context, Result};
 use onefuzz_telemetry::{InstanceTelemetryKey, MicrosoftTelemetryKey};
 use std::path::{Path, PathBuf};
@@ -41,6 +41,7 @@ pub enum PlaceHolder {
     ReportsDir,
     JobId,
     TaskId,
+    MachineId,
     CrashesContainer,
     CrashesAccount,
     MicrosoftTelemetryKey,
@@ -73,6 +74,7 @@ impl PlaceHolder {
             Self::ReportsDir => "{reports_dir}",
             Self::JobId => "{job_id}",
             Self::TaskId => "{task_id}",
+            Self::MachineId => "{machine_id}",
             Self::CrashesContainer => "{crashes_container}",
             Self::CrashesAccount => "{crashes_account}",
             Self::MicrosoftTelemetryKey => "{microsoft_telemetry_key}",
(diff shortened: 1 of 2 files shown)


src/agent/onefuzz/src/expand.rs
```diff
@@ -486,4 +510,13 @@ mod tests {
         assert!(result.contains("lib.rs"));
         Ok(())
     }
+
+    #[test]
+    fn test_expand_machine_id() -> Result<()> {
```
ranweiler, line 515: 👍

src/agent/onefuzz/src/expand.rs
```diff
@@ -108,9 +110,29 @@ impl<'a> Expand<'a> {
...
+            ExpandedValue::Mapping(Box::new(Expand::machine_id)),
+        );
+
         Self { values }
     }
 
+    // by making this a Mapping rather than when creating an Expand instance,
+    // users only pay for fetching the machine_id when it's used in expansion
+    fn machine_id(&self, _format_str: &str) -> Option<ExpandedValue<'a>> {
+        let rt = tokio::runtime::Runtime::new().unwrap();
```
ranweiler, line 124: When might this panic?

src/agent/onefuzz/src/expand.rs
```diff
@@ -108,9 +110,29 @@ impl<'a> Expand<'a> {
...
+        let result = rt.block_on(get_machine_id());
```
ranweiler, line 125: I don't like creating a short-lived runtime here, but I understand why you did it.

But since this operation truly is fallible, why not return an error, instead of sinking errors into `None`?

src/agent/onefuzz/src/expand.rs
bmc-msft: This will panic if we can't allocate a new tokio runtime.  From my brief investigation, this should only happen in OOM situations, where we'll panic all over the place as is.

src/agent/onefuzz/src/expand.rs
bmc-msft: The rest of the Expand infrastructure doesn't support failable operations.  It would be pretty large change to propagate Result throughout all of the `Expand` and all of it's callers.

src/agent/onefuzz/src/expand.rs
ranweiler: I think this is comment-worthy (and the motivation: the existing machine ID-getting code uses async IO, and there's not a really a good way to factor out a blocking version).

src/agent/onefuzz/src/expand.rs
ranweiler: Hmm, apparently we similarly use `ok()` for input file hashes in [this](https://github.com/microsoft/onefuzz/blob/main/src/agent/onefuzz/src/expand.rs#L114) private method.

The `evaluate_` methods are already fallible, since the template string may be malformed. Then the trick is just making sure that the callbacks in `Mapping` placeholders have the right return type. But since they're private, I don't think that requires any visible change for users of `Expand`.

ranweiler: Vetted my thought here: #1217

If we choose to take that PR, then you can just use `?` as needed.

ranweiler (approved):

PR #1247, opened 2021-09-15
Added support for --readonly_inputs for libfuzzer basic template

jordyn:

_What is this about?_

Added support for --readonly_inputs for libfuzzer basic template

### DIFF
--- a/src/cli/onefuzz/templates/libfuzzer.py
+++ b/src/cli/onefuzz/templates/libfuzzer.py
@@ -236,6 +236,7 @@ class Libfuzzer(Command):
         wait_for_files: Optional[List[ContainerType]] = None,
         extra_files: Optional[List[File]] = None,
         existing_inputs: Optional[Container] = None,
+        readonly_inputs: Optional[Container] = None,
         dryrun: bool = False,
         notification_config: Optional[NotificationConfig] = None,
         debug: Optional[List[TaskDebugFlag]] = None,
@@ -257,6 +258,9 @@ class Libfuzzer(Command):
         # verify containers exist
         if existing_inputs:
             self.onefuzz.containers.get(existing_inputs)
+
+        if readonly_inputs:
+            self.onefuzz.containers.get(readonly_inputs)
 
         if dryrun:
             return None
@@ -288,6 +292,7 @@ class Libfuzzer(Command):
             ContainerType.coverage,
             ContainerType.unique_inputs,
             ContainerType.regression_reports,
+            ContainerType.readonly_inputs
         )
 
         if existing_inputs:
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -288,6 +292,7 @@ def basic(
             ContainerType.coverage,
             ContainerType.unique_inputs,
             ContainerType.regression_reports,
+            ContainerType.readonly_inputs
```
bmc-msft, line 295: This will create a `readonly_inputs` container for you, which I don't think is what you want.  You should be ok removing this here.

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -296,6 +301,10 @@ def basic(
         else:
             helper.define_containers(ContainerType.inputs)
 
+        if readonly_inputs:
+            self.onefuzz.containers.get(readonly_inputs)
+            helper.containers[ContainerType.readonly_inputs] = readonly_inputs
+
```
bmc-msft, line 307: Since you want `readonly_inputs` as a container used for the fuzzing tasks, you also need to add it to this list:

https://github.com/microsoft/onefuzz/blob/6870b4c4939b18ab26cb0a7e868470d205400680/src/cli/onefuzz/templates/libfuzzer.py#L104-L108

bmc-msft (changes requested):
In addition to the other changes, all python code must be appropriately formatted/linted.

This includes:

* `black`
* `flake8`
* `isort --profile black`
* `mypy`

src/cli/onefuzz/templates/libfuzzer.py
jordyn: Updated and see the container being pulled in: INFO:onefuzz:using container: oft-readonly-inputs-<guid>

src/cli/onefuzz/templates/libfuzzer.py
jordyn: Is it fine if a readonly container isn't defined when we run:

fuzzer_containers = [
            (ContainerType.setup, containers[ContainerType.setup]),
            (ContainerType.crashes, containers[ContainerType.crashes]),
            (ContainerType.inputs, containers[ContainerType.inputs]),
            (ContainerType.readonly_inputs, containers[ContainerType.readonly_inputs]),
        ]

bmc-msft: Instead of always it to fuzzer_containers, you could instead only add it to fuzzer_containers if the user specifies it.

jordyn:
> In addition to the other changes, all python code must be appropriately formatted/linted.
> 
> This includes:
> 
> * `black`
> * `flake8`
> * `isort --profile black`
> * `mypy`

Let me know what else is needed to close this out.

bmc-msft (approved):

PR #1265, opened 2021-09-22
move to using released appinsights version

bmc-msft:
appinsights 0.2 was released, enabling rustls by default.  As such, we
can now track the released versions rather than a specific git hash.

### DIFF
--- a/src/agent/onefuzz-telemetry/src/lib.rs
+++ b/src/agent/onefuzz-telemetry/src/lib.rs
@@ -63,7 +63,7 @@ impl fmt::Display for InstanceTelemetryKey {
     }
 }
 
-pub type TelemetryClient = appinsights::TelemetryClient<appinsights::InMemoryChannel>;
+pub type TelemetryClient = appinsights::blocking::TelemetryClient;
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -63,7 +63,7 @@ impl fmt::Display for InstanceTelemetryKey {
     }
 }
 
-pub type TelemetryClient = appinsights::TelemetryClient<appinsights::InMemoryChannel>;
+pub type TelemetryClient = appinsights::blocking::TelemetryClient;
```
bmc-msft, line 66: Note, the InMemoryChannel is now default and not required to be specified as a type anymore.

ranweiler (approved):

bmc-msft:
A bunch of integration tests are breaking (beyond the known issues with windows libfuzzer).  I will dig into this to see if it's this PR or a wider issue.

bmc-msft:
not functional as is, more investigation on this needed, potentially moving to tracing pipeline as discussed offline.

PR #1268, opened 2021-09-22
enable configurable virtual network ranges

bmc-msft:
In order to enable express route changes, we need to allow administrators to configure the virtual network IP ranges used.

1f creates virtual networks with the name of the region.  To maintain compatibility with existing deployments as well as enabling continued support to not break existing deployments, this  

Without configuration, this change will continue to use `region` as the network name as long as the network config is the defaults.  Otherwise, it will use `region-<namespaced-guid-here>` based on the settings in the config.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/network.py
+++ b/src/api-service/__app__/onefuzzlib/azure/network.py
@@ -4,6 +4,7 @@
 # Licensed under the MIT License.
 
 import logging
+import uuid
 from typing import Optional, Union
 
 from msrestazure.azure_exceptions import CloudError
@@ -11,36 +12,63 @@ from onefuzztypes.enums import ErrorCode
 from onefuzztypes.models import Error
 from onefuzztypes.primitives import Region
 
+from ..config import InstanceConfig
 from .creds import get_base_resource_group
-from .subnet import create_virtual_network, delete_subnet, get_subnet_id
+from .subnet import create_virtual_network, get_subnet_id
+
+# these were original address spaces chosen for 1f.  However, moving forwards
+# these are configurable in the Instance Config.  Network names will be
+# calculated from the address_space/subnet *except* if they are the original
+# values.  This allows backwards compatability to existing configs if you don't
+# change the network configs.
+ORIGINAL_ADDRESS_SPACE = "10.0.0.0/8"
+ORIGINAL_SUBNET = "10.0.0.0/16"
+
+# This was generated randomly and should be preserved moving forwards
+NETWORK_GUID_NAMESPACE = uuid.UUID("372977ad-b533-416a-b1b4-f770898e0b11")
 
 
 class Network:
     def __init__(self, region: Region):
         self.group = get_base_resource_group()
         self.region = region
+        self.network_config = InstanceConfig.fetch().network_config
+
+        if (
+            self.network_config.address_space == ORIGINAL_ADDRESS_SPACE
+            and self.network_config.subnet == ORIGINAL_SUBNET
+        ):
+            self.name: str = self.region
+        else:
+            network_id = uuid.uuid5(
+                NETWORK_GUID_NAMESPACE,
+                "|".join(
+                    [self.network_config.address_space, self.network_config.subnet]
+                ),
+            )
+            self.name = f"{self.region}-{network_id}"
 
     def exists(self) -> bool:
         return self.get_id() is not None
 
     def get_id(self) -> Optional[str]:
-        return get_subnet_id(self.group, self.region)
+        return get_subnet_id(self.group, self.name, self.region)
 
     def create(self) -> Union[None, Error]:
         if not self.exists():
-            result = create_virtual_network(self.group, self.region, self.region)
+            result = create_virtual_network(
+                self.group, self.name, self.region, self.network_config
+            )
             if isinstance(result, CloudError):
                 error = Error(
                     code=ErrorCode.UNABLE_TO_CREATE_NETWORK, errors=[result.message]
                 )
                 logging.error(
-                    "network creation failed: %s- %s",
+                    "network creation failed: %s:%s- %s",
+                    self.name,
                     self.region,
                     error,
                 )
                 return error
 
         return None
-
-    def delete(self) -> None:
-        delete_subnet(self.group, self.region)
--- a/src/api-service/__app__/onefuzzlib/azure/subnet.py
+++ b/src/api-service/__app__/onefuzzlib/azure/subnet.py
@@ -10,15 +10,16 @@ from typing import Any, Optional, Union, cast
 from azure.core.exceptions import ResourceNotFoundError
 from msrestazure.azure_exceptions import CloudError
 from onefuzztypes.enums import ErrorCode
-from onefuzztypes.models import Error
+from onefuzztypes.models import Error, NetworkConfig
+from onefuzztypes.primitives import Region
 
 from .network_mgmt_client import get_network_client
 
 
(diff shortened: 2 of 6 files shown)


src/api-service/__app__/onefuzzlib/azure/network.py
```diff
@@ -4,43 +4,68 @@
...
 
 class Network:
     def __init__(self, region: Region):
         self.group = get_base_resource_group()
         self.region = region
+        self.network_config = InstanceConfig.fetch().network_config
+
+        # Network names will be calculated from the address_space/subnet
+        # *except* if they are the original values.  This allows backwards
+        # compatability to existing configs if you don't change the network
```
ranweiler, line 31:
```suggestion
        # compatibility to existing configs if you don't change the network
```

src/api-service/__app__/onefuzzlib/azure/network.py
```diff
@@ -4,43 +4,68 @@
...
+NETWORK_GUID_NAMESPACE = uuid.UUID("372977ad-b533-416a-b1b4-f770898e0b11")
 
 
 class Network:
     def __init__(self, region: Region):
         self.group = get_base_resource_group()
         self.region = region
+        self.network_config = InstanceConfig.fetch().network_config
+
+        # Network names will be calculated from the address_space/subnet
```
ranweiler, line 29: 👍

src/api-service/__app__/onefuzzlib/azure/network.py
```diff
@@ -4,43 +4,68 @@
...
+        if (
+            self.network_config.address_space
+            == NetworkConfig.__fields__["address_space"].default
+            and self.network_config.subnet == NetworkConfig.__fields__["subnet"].default
+        ):
+            self.name: str = self.region
+        else:
+            network_id = uuid.uuid5(
+                NETWORK_GUID_NAMESPACE,
+                "|".join(
```
ranweiler, line 42: In the CLI, we use `:` as the delimiter when creating namespaced UUIDs from multiple components.

I can't think of a real reason why it'd matter, but should we prefer one or the other?

src/api-service/__app__/onefuzzlib/azure/subnet.py
```diff
@@ -10,15 +10,16 @@
...
 
-def get_subnet_id(resource_group: str, name: str) -> Optional[str]:
+def get_subnet_id(resource_group: str, name: str, subnet_name: str) -> Optional[str]:
     network_client = get_network_client()
     try:
-        subnet = network_client.subnets.get(resource_group, name, name)
+        subnet = network_client.subnets.get(resource_group, name, subnet_name)
         return cast(str, subnet.id)
     except (CloudError, ResourceNotFoundError):
         logging.info(
```
ranweiler, line 25: Should we update this error message to include the newly-passed `subnet_name`, too?

src/api-service/__app__/onefuzzlib/request.py
```diff
@@ -92,8 +92,8 @@ def not_ok(
         )
 
 
-def redirect(location: str) -> HttpResponse:
-    return HttpResponse(status_code=302, headers={"Location": location})
+def redirect(url: str) -> HttpResponse:
```
ranweiler, line 95: I assume this change is just for reader clarity, since variable name previously referred the HTTP notion of `Location` header, whereas elsewhere in this PR, we are interacting with Azure APIs that use "location" to refer to an Azure region.

src/api-service/__app__/onefuzzlib/azure/subnet.py
```diff
@@ -10,15 +10,16 @@
...
 from onefuzztypes.enums import ErrorCode
-from onefuzztypes.models import Error
+from onefuzztypes.models import Error, NetworkConfig
+from onefuzztypes.primitives import Region
 
 from .network_mgmt_client import get_network_client
 
 
-def get_subnet_id(resource_group: str, name: str) -> Optional[str]:
+def get_subnet_id(resource_group: str, name: str, subnet_name: str) -> Optional[str]:
```
ranweiler, line 19: What's the difference between `name` and `subnet_name`?

src/api-service/__app__/onefuzzlib/azure/network.py
```diff
@@ -4,43 +4,68 @@
...
+                ),
+            )
+            self.name = f"{self.region}-{network_id}"
 
     def exists(self) -> bool:
         return self.get_id() is not None
 
     def get_id(self) -> Optional[str]:
-        return get_subnet_id(self.group, self.region)
+        return get_subnet_id(self.group, self.name, self.name)
```
ranweiler, line 52: I think the repetition of `self.name` in the arguments here is intentional, but please confirm.

src/api-service/__app__/onefuzzlib/azure/network.py
```diff
@@ -4,43 +4,68 @@
...
+                    "network creation failed: %s:%s- %s",
+                    self.name,
                     self.region,
                     error,
                 )
                 return error
 
         return None
-
-    def delete(self) -> None:
```
ranweiler, line 45: Why are we removing this?

src/pytypes/onefuzztypes/models.py
```diff
@@ -794,6 +794,11 @@ class Task(BaseModel):
     user_info: Optional[UserInfo]
 
 
+class NetworkConfig(BaseModel):
+    address_space: str = Field(default="10.0.0.0/8")
```
ranweiler, line 798: Maybe we should pass these values via named constants.

src/pytypes/onefuzztypes/models.py
```diff
@@ -835,9 +840,8 @@ class InstanceConfig(BaseModel):
     allow_pool_management: bool = Field(default=True)
 
     allowed_aad_tenants: List[UUID]
-
+    network_config: NetworkConfig = Field(default=NetworkConfig())
```
ranweiler, line 843: It probably shouldn't matter in practice, but I think we really want `default_factory=NetworkConfig`, or similar.

src/api-service/__app__/onefuzzlib/azure/network.py
bmc-msft: It is intentional.  Right now, the subnet id is the same as the network name.

src/api-service/__app__/onefuzzlib/azure/network.py
bmc-msft: it's not actually used.

src/api-service/__app__/onefuzzlib/azure/subnet.py
bmc-msft: right now, nothing.  we have only one subnet name.

src/api-service/__app__/onefuzzlib/request.py
bmc-msft: Correct.  This change was clarity, based on grep to find everywhere we used `location`.

src/pytypes/onefuzztypes/models.py
bmc-msft: I thought about that, but didn't want the named constants used elsewhere, rather always just use the defaults from the model directly.

src/api-service/__app__/onefuzzlib/azure/network.py
bmc-msft: I chose to use a different value as `:` is used in IPv6 addresses.

ranweiler (approved):

PR #1270, opened 2021-09-23
fix multi-tenant deployment

### LINKED ISSUE #1269: Multi-Tenant Auth is broken for 3.0.0
## Information

- Onefuzz version: 3.0.0
- OS: Windows

1. Deploy multi-tenant instance of OneFuzz
2. Configure endpoint with multi-tenant params (--tenant_domain, etc.)
3. Login

### Expected result

Authenticate to app registration with mutli-tenant credentials. 

### Actual result

Can not log in.

chkeita:
When deploying a multitenant instance there was a mismatch between the identifier uri used to configure the application registration and the value we use to authenticate the cli client. This caused #1269 

To test: 
1- deploy on a multi tenant domain 
2- send a onefuzz request

Expected result: The request should be successful

closes #1269


stishkin (approved):

src/deployment/deploy.py
```diff
@@ -251,6 +251,24 @@ def check_region(self) -> None:
...
+            return "https://%s.azurewebsites.net" % self.application_name
+
+    def get_identifier_uri(self) -> str:
+        if self.multi_tenant_domain:
+            return "api://%s/%s" % (
+                self.multi_tenant_domain,
+                self.application_name,
+            )
+        else:
+            return "api://%s.azurewebsites.net" % self.application_name
```
ranweiler, line 270: Can we link to the docs that define this format?

src/deployment/deploy.py
```diff
@@ -251,6 +251,24 @@ def check_region(self) -> None:
     def create_password(self, object_id: UUID) -> Tuple[str, str]:
         return add_application_password(object_id, self.get_subscription_id())
 
+    def get_url(self) -> str:
```
ranweiler, line 254: Can we unit test `get_url()` and `get_identifier_uri()`?

ranweiler: Are there more specific names we should maybe choose, instead of `get_url()`? That one is really generic. It would be helpful to even just add a doc/function comment about _what_ UR{L,I}s are supposed to be returned here and below.

ranweiler (commented):
Could we update the PR description to describe (1) what we believe the problem was, and its scope, (2) why it fixes the linked issue, and (3) how one could test it?

bmc-msft:
I don't know if the error from integration testing is unique to this PR or not, but I get the following when testing a new single-tenant deployment.

```
INFO:deploy:app password created
INFO:deploy:deploying arm template: azuredeploy.json

  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/base_polling.py", line 483, in run
    self._poll()
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/base_polling.py", line 522, in _poll
    raise OperationFailed("Operation failed or canceled")
azure.core.polling.base_polling.OperationFailed: Operation failed or canceled

During handling of the above exception, another exception occurred:

  File "deploy.py", line 1091, in <module>
    main()
  File "deploy.py", line 1085, in main
    state[1](client)
  File "deploy.py", line 540, in deploy_template
    raise error
  File "deploy.py", line 516, in deploy_template
    result = client.deployments.begin_create_or_update(
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/_poller.py", line 255, in result
    self.wait(timeout)
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/tracing/decorator.py", line 83, in wrapper_use_tracer
    return func(*args, **kwargs)
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/_poller.py", line 275, in wait
    raise self._exception # type: ignore
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/_poller.py", line 192, in _start
    self._polling_method.run()
  File "/tmp/tmplye2wlof/deploy-venv/lib/python3.8/site-packages/azure/core/polling/base_polling.py", line 501, in run
    raise HttpResponseError(
azure.core.exceptions.HttpResponseError: (DeploymentFailed) At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.
Code: DeploymentFailed
Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.
Exception Details:      (BadRequest) {
          "error": {
            "code": "HashConflictOnDifferentRoleAssignmentIds",
            "message": "A hash conflict was encountered for the role Assignment ID. Please use a new Guid."
          }
        }
        Code: BadRequest
        Message: {
          "error": {
            "code": "HashConflictOnDifferentRoleAssignmentIds",
            "message": "A hash conflict was encountered for the role Assignment ID. Please use a new Guid."
          }
        }
```

bmc-msft (changes requested):
see concerns

src/deployment/deploy.py
chkeita: i have not found a doc for this particular format, we derived it from the default value proposed by azure when we create the registration

src/deployment/deploy.py
chkeita: I am not exactly sure what is worth testing here. The point was just to extract to a function the logic for building the url. I will update get_url to get instance_url.

src/deployment/deploy.py
chkeita: renamed and added comment

src/deployment/deploy.py
chkeita: added comments

ranweiler: Thank you, those comments help a lot!

In the PR description, you say "the identifier was incorrect". I was thinking we could add tests that check that the identifiers look correct, in each case. This would then naturally document examples of "correct" for both the single and multi-tenant cases.

ranweiler: (Maybe the identifiers were always well-formed, but were just using the wrong sometimes, and that's why my question is confusing?)

chkeita: incorrect might not have been a good word for it. I updated the description to use mismatched instead. What is important is that the value in the request matches the value in the configuration. 
The integration test (assuming that we will have them for a multi tenant instance will catch this kind of issue)

nharper285:
I've tested this and it works with both mutli-tenant credentials.

chkeita:
> I don't know if the error from integration testing is unique to this PR or not, but I get the following when testing a new single-tenant deployment.

I don't think this is related to this PR. I just had a successful deployment

bmc-msft (approved):

PR #1271, opened 2021-09-23
add check-pr from personal repo

bmc-msft:
Adding the tool I use for automating execution of integration tests on a new deployment for a given pr.


src/utils/check-pr/check-pr.py
```diff
@@ -0,0 +1,319 @@
...
+
+        cmd = ["az", "group", "delete", "-n", self.instance, "--yes", "--no-wait"]
+        print(cmd)
+        subprocess.call(cmd)
+        print("done")
+
+    def run(self, *, merge_on_success: bool = False) -> None:
+        release_filename = "release-artifacts.zip"
+        self.downloader.get_artifact(
+            "microsoft/onefuzz",
```
chkeita, line 249: This is getting the artifact from the onefuzz repo right? can we make it parameter so that we can use on other repo

src/utils/check-pr/check-pr.py
```diff
@@ -0,0 +1,319 @@
...
+            "ci.yml",
+            self.branch,
+            self.pr,
+            "release-artifacts",
+            release_filename,
+        )
+
+        test_filename = "integration-test-artifacts.zip"
+        self.downloader.get_artifact(
+            "microsoft/onefuzz",
```
chkeita, line 259: same as my previous comment, this should configurable

src/utils/check-pr/check-pr.py
bmc-msft: added

src/utils/check-pr/check-pr.py
bmc-msft: added

chkeita (approved):

PR #1286, opened 2021-09-27
use flake8 on deployment

bmc-msft:
All of the rest of the python codebases within the repo use flake8 as
part of the lint process.  This aligns the deployment code with the
practice.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,10 +309,11 @@ jobs:
         set -ex
         cd src/deployment
         python -m pip install --upgrade pip
-        pip install mypy isort black types-requests
+        pip install mypy isort black types-requests flake8
         mypy .
         isort --profile black . --check
         black . --check
+        flake8 *.py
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -309,10 +309,11 @@ jobs:
         set -ex
         cd src/deployment
         python -m pip install --upgrade pip
-        pip install mypy isort black types-requests
+        pip install mypy isort black types-requests flake8
         mypy .
         isort --profile black . --check
         black . --check
+        flake8 *.py
```
ranweiler, line 316: Is it important that this arg is `*.py` and not `.`?

ranweiler (approved):

.github/workflows/ci.yml
bmc-msft: Yes.  flake8 was picking up on the mypy cache and processing all of that content, which isn't intended.

PR #1295, opened 2021-09-28
Remove references to `playground` instance

### LINKED ISSUE #1281: Update getting started doc to remove reference to playground instance
We are removing the playground instance. According to the telemetry it's not being used. Further for those internal customers who want to try out Onefuzz and have access to the instance, it's likely to create some confusion because we are creating different ways for teams to interact with the system. 

## Provide a description of requested docs changes
In docs/getting-started.md we need to remove the reference to the playground instance and update the example in the "Connecting to your instance" section use a different name. 

_What is the purpose and what should be changed?_
Remove confusion.

ranweiler:

Update the `getting-started` doc to remove redundant references to the `playground` trial instance.

## PR Checklist
* [x] Applies to work item: Closes #1281
* [ ] ~~Tests added/passed~~
* [ ] ~~Requires documentation to be updated~~
* [ ] ~~I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx~~

## Info on Pull Request

See summary.

## Validation Steps Performed

Doc update, N/A.


docs/getting-started.md
```diff
@@ -51,25 +51,16 @@ pip install ./onefuzz*.whl
...
 ```
 
 From here, you can use Onefuzz.
 
-#### Using Onefuzz Playground
-
-Microsoft employees are welcome to use MSR's Playground instance of OneFuzz for
-experimentation purposes, but we reserve the right to stop or limit fuzzing
-jobs. [Come talk to us](https://aka.ms/fuzzingatmicrosoft) about running OneFuzz
-for your group!
```
bmc-msft, line 71: You probably want to move this message somewhere else in the getting started guide, since we have a number of Microsoft people looking to use 1f that don't know about "Fuzzing @ Microsoft"

ranweiler: Yeah, I don't like doing that, but I agree that it would be most pragmatic.

bmc-msft (approved):

PR #1300, opened 2021-09-28
Fix authentication with client secret

chkeita:

Fixes the logic of retrieving the access token when a `client_secret` is configured


stishkin (approved):

src/cli/onefuzz/backend.py
```diff
@@ -199,11 +199,13 @@ def client_secret(self, scopes: List[str]) -> Any:
 
         for scope in scopes:
             result = self.app.acquire_token_for_client(scopes=[scope])
+            if "error" not in result:
```
ranweiler, line 202: Could you please add comments that document what is going on here?

ranweiler: I think what's happening is: if any scope enables token acquisition (`acquire_token_for_client()` succeeded), then we're done, and we can stop looking for one that will get us a token. We don't care if some scopes fail to enable token acquisition, because we only require that the input `scopes` be _candidate_ sources for token acquisition. We don't require that the caller has ensured that _every_ scope is valid.

In other words, the semantics of this function from the caller perspective are: "Here are a bunch of scopes I have (with no promises about them). Try to get me an access token for the app using anything that will work."

This was wrong because in the old impl, if there were multiple scopes, and the first once succeeded, then the second one failed, Python loop and variable scoping rules are such that the failed would would be the `result` that we check at the end, so we'd throw an exception, even though one of our scopes was valid.

Is that accurate?

chkeita: yes we were supposed to break as soon as we found a token

ranweiler (approved):

Issue #1297, opened 2021-09-28
Multiple onefuzz deployments under single subscription


stmh-infosec:
I have been trying to deploy different versions of onefuzz (3.0.0 and 2.16.0) under a single subscription. The deployment succeeds (I do change the client app registration name i.e: instead of `onefuzz-cli`, I name it version specific eg: `onefuzz-cli-3-0-0`). When I install the cli and configure it, i am not able to use the cli to create pools, create vms and schedule jobs. I get the below error:

ERROR:cli:command failed: error: invalid_client
'AADSTS7000215: Invalid client secret is provided.
Trace ID: <redacted>
Correlation ID: <redacted>
Timestamp: 2021-09-28 20:43:14Z'

Can I have multiple instances of onefuzz running under the single subscription ? If yes, could you let me know why my cli is failing to connect with the instance ?

chkeita:
Yes, it is possible to deploy multiple instances under the same subscription.

Do you have a client_secret in your config? You can run `onefuzz config` and check if `client_secret` is set
If that is the case, you can rerun the command to set your config with the reset option 
`onefuzz config --endpoint <endpoint> --authority <authority> --client_id <client_id> --reset true`
This will  remove the `client_secret` value

stmh-infosec:
Yea, I have tried that. But with deployment of 3.0.0. I have been getting the below error when i execute `onefuzz versions check --exact`

WARNING:nsv-backend:failed to get access token with scope https://<<websitename>>/.default
ERROR:cli:command failed: error: invalid_resource
The resource principal named <<website-name>> was not found in the tenant named Default Directory. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You might have sent your authentication request to the wrong tenant.

This is a normal deployment with no permission issues during deployment. I am the owner of the subscription and the account.

chkeita:
are you using a client_secret in your config?

ranweiler:
@stmh-infosec, you shouldn't have to rename the `onefuzz-cli` app registration. The original should have worked fine for all instances, and its existence should not have caused a deployment error. If it caused errors on your first deployment attempts, and that's why you renamed it, that is noteworthy.

Please share the exact order of operations you did, e.g.
1. Deployed 2.16 instance
2. Changed name of `onefuzz-cli` app registration
3. Deployed 3.0 instance

(or re-arranged as appropriate, depending on what you did)

[...]

PR #1304, opened 2021-09-29
Disable storageclient logs when setting admins 

chkeita:

When checking the existence of a table. The `storageclient` module logs an error message and throws an exception. 
The exception is handled internally bu the message shows up in our deployment logs (see #1297).
This PR disables the logs for that module.


src/deployment/deploy.py
```diff
@@ -562,12 +562,15 @@ def assign_scaleset_identity_role(self) -> None:
         )
 
     def apply_migrations(self) -> None:
+        logger.info("applying database migrations")
+        self.results["deploy"]["func-storage"]["value"]
```
ranweiler, line 566: What is the purpose of this line?

src/deployment/set_admins.py
```diff
@@ -5,12 +5,20 @@
 
 import argparse
 import json
+from time import sleep
```
ranweiler, line 8:
```suggestion
```

src/deployment/set_admins.py
```diff
@@ -5,12 +5,20 @@
...
 from azure.common.client_factory import get_client_from_cli_profile
 from azure.cosmosdb.table.tableservice import TableService
 from azure.mgmt.storage import StorageManagementClient
+import logging
+
+## Disable logging from storageclient. This modules display and error message
+## when a resource is not found even if the exception is raised and handled internally.
+## This happen when a table does not exist. An error message is displayed but the exception is
+## handled by the library.
+logging.getLogger('azure.cosmosdb.table.common.storageclient').disabled = True
```
ranweiler, line 21: We don't control this name, so we should ensure that `getLogger` returns an object. Otherwise the `.disabled` attr lookup could fail and throw an exception at module load time.

src/deployment/set_admins.py
```diff
@@ -5,12 +5,20 @@
...
+from time import sleep
 from typing import List, Optional
 from uuid import UUID
 
 from azure.common.client_factory import get_client_from_cli_profile
 from azure.cosmosdb.table.tableservice import TableService
 from azure.mgmt.storage import StorageManagementClient
+import logging
+
+## Disable logging from storageclient. This modules display and error message
```
ranweiler, line 17:
```suggestion
## Disable logging from storageclient. This module displays an error message
```

src/deployment/set_admins.py
```diff
@@ -5,12 +5,20 @@
...
+## when a resource is not found even if the exception is raised and handled internally.
```
ranweiler, line 18: Should we open an issue in the upstream library?

src/deployment/set_admins.py
ranweiler: What other log messages do we lose by disabling this logger completely?

Is there a way we can avoid disabling it, for debugging deployments? E.g. maybe at higher verbosity levels, we should include logs from dependencies.

src/deployment/set_admins.py
chkeita: it is already being discussed
https://github.com/Azure/azure-storage-python/issues/516

src/deployment/deploy.py
chkeita: copy paste error

src/deployment/set_admins.py
chkeita: I update the code to only disable the logging when we check for the table existence.

ranweiler (approved):

Issue #1263, opened 2021-09-22
Windows integration tests fail due to mis-compiled test fuzzers


ranweiler:
Windows integration tests build libFuzzers using the latest LLVM and Windows SDK. We install LLVM [here](https://github.com/microsoft/onefuzz/blob/8cc70188957f60db62dc279ee1f406ffd0a0da04/.github/workflows/ci.yml#L410). The SDK version is inherited from the GitHub Actions environment. A recent update to the Windows SDK (either [here](https://github.com/actions/virtual-environments/commit/b913c5ba8707a668ae8c82fbff78a3d50ca9b855) or maybe [here](https://github.com/actions/virtual-environments/commit/3703015712cec213f081e031bde670d29a3f6836#diff-816219d3c9f30fcc1c40f88ca826c1662bf99eded781c1b419dd2dfb577ae227R237)) looks to have surfaced a mis-compilation in the Windows ASAN runtime static library that is shipped with LLVM. As a result, ASAN-instrumented targets exit abnormally, which our integration tests catch via a failed binary validity check.

In practice, the test failure looks like:

> `ERROR:integration:task failed: windows-libfuzzer - libfuzzer_fuzz (code=<ErrorCode.TASK_FAILED: 468> errors=['task failed. exit_status:code=1 signal=None success=False', '', '[2021-09-15T18:08:22Z INFO onefuzz_agent::tasks::config] agent ready, dispatching task\n[2021-09-15T18:08:22Z INFO onefuzz_agent::tasks::config] task_start event_type:libfuzzer_fuzz\n[2021-09-15T18:08:25Z ERROR onefuzz_agent::managed::cmd] error running task: fuzzer does not respond to \'-help=1\'. output:Output { status: ExitStatus(ExitStatus(255)), stdout: "", stderr: "" }\nError: fuzzer does not respond to \'-help=1\'. output:Output { status: ExitStatus(ExitStatus(255)), stdout: "", stderr: "" }\n']) - 7a9a9766-5088-49f0-826b-39570d8a6e03`

Ways we can work around this:
- Implement binary test artifacts (see: #1143), which would let us integration test with known-good artifacts
- In CI, patch LLVM post-install, for example, by replacing the faulty ASAN runtime static lib with a known good one.
- Downgrade the Windows SDK to a known "good" version (one which doesn't trigger the mis-compile's edge case).

PR #1310, opened 2021-09-30
Update `nix`, patch dependents

ranweiler:
Temporarily depend on patched `pete`, `ctrlc` crates so we only use the latest `nix`.

We can depend on upstream `ctrlc` after https://github.com/Detegr/rust-ctrlc/pull/81 is merged and/or released.

Replaces #1303.

- [x] Integration tests

### DIFF
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -18,3 +18,6 @@ members = [
 
 [profile.release]
 lto = "thin"
+
+[patch.crates-io]
+pete = { git = "http://github.com/ranweiler/pete", rev = "efeb0769b4376d22a33328553c31ce331a7261ad" }
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -19,5 +19,6 @@ members = [
 [profile.release]
 lto = "thin"
 
+# Remove after `pete 0.8` released.
 [patch.crates-io]
 pete = { git = "http://github.com/ranweiler/pete", rev = "efeb0769b4376d22a33328553c31ce331a7261ad" }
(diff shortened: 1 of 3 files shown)


src/agent/Cargo.toml
```diff
@@ -18,3 +18,11 @@ members = [
 
 [profile.release]
 lto = "thin"
+
+
+[patch.crates-io]
+# Remove after `Detegr/rust-ctrlc#81` is released.
+ctrlc = { git = "http://github.com/ranweiler/rust-ctrlc", rev = "7535c9306557ef69bc07401ba653aa5d7d625e2f" }
+
```
chkeita, line 26: can we create an issue to keep track of those two evens
- release of  `Detegr/rust-ctrlc#81`
- release of `ranweiler/pete#71`

ranweiler: I was thinking that we'll already get update PRs for free from Dependabot. Do you think that's enough?

chkeita (approved):

src/agent/Cargo.toml
bmc-msft: You won't get PRs from dependabot when you've pinned to specific git hashes.

bmc-msft: Oh, nevermind.  That feature got added to dependabot!

`Dependabot can handle git sources, and will create PRs to keep them pointing to the latest commit`

ranweiler: @bmc-msft, on top of that, this will be a good live test of what to expect, all things considered. This is a particularly interesting test case because it isn't just a git source, but a _patched_ git source!

bmc-msft: Yep.  More so since you control one of said repos, it means you can set a todo item for yourself to check within 24 hours of you making a commit and then release of pete.

mgreisen (approved):

ranweiler:
Integration tests passed except known failures from #1263.

PR #1311, opened 2021-10-01
Add more setup info to `check-pr` docs

ranweiler:
- Describe what permissions to grant the GitHub API `PAT`
- Document required env vars
- Recommend the use of `virtualenv` and `direnv`
- Note that `-h` can be used for more usage info

### DIFF
--- a/src/utils/check-pr/README.md
+++ b/src/utils/check-pr/README.md
@@ -1,8 +1,20 @@
 # PR integration test tool
 
-Requirements:
+## Requirements
 
-* Python packages (See `requirements.txt`)
+* Python >= 3.7, and dependencies (see `requirements.txt`)
 * [az-cli](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli).
 * [azcopy](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10)
 * [azure-functions-core-tools](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v3%2Clinux%2Ccsharp%2Cportal%2Cbash%2Ckeda#install-the-azure-functions-core-tools)
+
+## Setup
+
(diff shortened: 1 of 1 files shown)


src/utils/check-pr/README.md
```diff
@@ -1,8 +1,20 @@
...
 
-* Python packages (See `requirements.txt`)
+* Python >= 3.7, and dependencies (see `requirements.txt`)
 * [az-cli](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli).
 * [azcopy](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10)
 * [azure-functions-core-tools](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v3%2Clinux%2Ccsharp%2Cportal%2Cbash%2Ckeda#install-the-azure-functions-core-tools)
+
+## Setup
+
+1. In GitHub, generate a personal access token (PAT) with the `public_repo` scope.
```
ranweiler, line 12: Some `check-pr` functionality may require more scopes (unsure). This is definitely enough to run integration tests on a PR.

bmc-msft: Note, the only scope I use is `workflow`

ranweiler: This makes me wonder if `workflow` implies `public_repo`.

The API docs state that Artifacts access requires the [`[public]_repo` scope](https://docs.github.com/en/rest/reference/actions#artifacts).

chkeita (approved):

src/utils/check-pr/README.md
ranweiler: Confirmed: when creating a GitHub PAT, if you check `workflow`, it auto-checks the entire `repo` scope _category_ (and disables un-checking), so it is much more permissive than `public_repo`.

PR #1314, opened 2021-10-04
Add a username to the resource group name created by check-pr utility.

mgreisen:

_What is this about?_
This change adds a username to the resource group name created by the check-pr utility. This makes it easy to identify who's resources these are. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
A small change to the check-pr.py script.

## Validation Steps Performed
Ran the script on both wsl2 and powershell to verify it works on both windows and linux printing out the created resource group name. 
Ran the script on wsl2 to verify the resource group name is created correctly. 

Run the script and look for the resource group name.

### DIFF
--- a/src/utils/check-pr/check-pr.py
+++ b/src/utils/check-pr/check-pr.py
@@ -279,14 +279,14 @@ class Deployer:
 def main() -> None:
     # Get a name that can be added to the resource group name
     # to make it easy to identify the owner
-    cmd = ["az", "ad", "signed-in-user", "show", "--query", "mailNickname"]
-    name = subprocess.run(cmd, capture_output=True, encoding='UTF-8')
+    cmd = ["az", "ad", "signed-in-user", "show", "--query", "mailNickname", "-o", "tsv"]
+    name = subprocess.check_output(cmd, encoding="UTF-8")
 
-    # The result from az includes quotes and a newline
+    # The result from az includes a newline
     # which we strip out.
-    name = name.stdout.strip()[1:-1]
+    name = name.strip()
 
-    default_instance = f'pr-check-{name}-%s' % uuid.uuid4().hex
+    default_instance = f"pr-check-{name}-%s" % uuid.uuid4().hex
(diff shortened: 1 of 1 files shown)


src/utils/check-pr/check-pr.py
```diff
@@ -277,7 +277,15 @@ def run(self, *, merge_on_success: bool = False) -> None:
...
 def main() -> None:
-    default_instance = "pr-check-%s" % uuid.uuid4().hex
+    # Pull out a username to include in the resource group name
+    # USER is used in wsl2, USERNAME is used under windows
+    if "USER" in os.environ:
+        user=os.environ["USER"]
+    elif "USERNAME" in os.environ:
+        user=os.environ["USERNAME"]
+    else:
+        user=""
```
bmc-msft, line 287: In the failure case, this will resolve to pr-check--GUID, which is probably not undesired.

Given this requires the `az` command be logged in, perhaps `az ad signed-in-user show --query userPrincipalName` and remove everything after `@` would be better.

bmc-msft: If you're OK requiring service accounts that deploy using check-pr, then this would work without having to split:  `az ad signed-in-user show --query mailNickname`

src/utils/check-pr/check-pr.py
```diff
@@ -277,7 +277,16 @@ def run(self, *, merge_on_success: bool = False) -> None:
 
 
 def main() -> None:
-    default_instance = "pr-check-%s" % uuid.uuid4().hex
+    # Get a name that can be added to the resource group name
+    # to make it easy to identify the owner
+    cmd = ["az", "ad", "signed-in-user", "show", "--query", "mailNickname"]
+    name=subprocess.run(cmd, capture_output=True, encoding='UTF-8')
```
bmc-msft, line 283: You should use `subprocess.check_output` to validate the command completed successfully.  Additionally, you can get the output without quotes by using the flags: `-o tsv`

Issue #1315, opened 2021-10-04
Instance config update API to support NSG configuration


stishkin:
Backend implementation of instance config update with `nsg-settings` `allow-rules` 

Make sure to check on the backend that admins only are allowed to run the command.

In case service tags need to be stored separately:
```json
		"nsg_configuration" : {
			"allow_service_tags" : [],
			"allow_IPs" : []
		}
```

If service tags and IPs can be easily stored in the same field, then use following format
```json
"nsg_configuration" {
			"allow" : []
}
```

PR #1331, opened 2021-10-06
Creating InstanceConfig Attributes for NSG Refactor

nharper285:

_What is this about?_
This work has to do with InstanceConfig updates related to the NSG refactor. 

## PR Checklist
* [ ] Applies to work item: #1315 
* [ ] ~CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.~
* [ ] ~Tests added/passed~
* [ ] ~Requires documentation to be updated~

## Info on Pull Request

_What does this include?_
Changes to models.py related to NSG src IP allow rules

## Validation Steps Performed
Deploying test build and testing value config. 

See above.

### DIFF
--- a/docs/webhook_events.md
+++ b/docs/webhook_events.md
@@ -649,6 +649,9 @@ Each event will be submitted via HTTP POST to the user provided URL.
             "address_space": "10.0.0.0/8",
             "subnet": "10.0.0.0/16"
         },
+        "nsg_config": {
+            "allow_IPs": []
+        },
         "proxy_vm_sku": "Standard_B2s"
     }
 }
@@ -759,6 +762,9 @@ Each event will be submitted via HTTP POST to the user provided URL.
                 "network_config": {
                     "$ref": "#/definitions/NetworkConfig"
                 },
+                "nsg_config": {
+                    "$ref": "#/definitions/NetworkSecurityGroupConfig"
+                },
                 "proxy_vm_sku": {
                     "default": "Standard_B2s",
                     "title": "Proxy Vm Sku",
@@ -813,6 +819,27 @@ Each event will be submitted via HTTP POST to the user provided URL.
                 }
             },
             "title": "NetworkConfig",
+            "type": "object"
+        },
+        "NetworkSecurityGroupConfig": {
+            "properties": {
+                "allow_IPs": {
+                    "default": [],
+                    "items": {
+                        "type": "string"
+                    },
+                    "title": "Allow Ips",
+                    "type": "array"
+                },
+                "allow_service_tags": {
+                    "items": {
+                        "type": "string"
+                    },
+                    "title": "Allow Service Tags",
+                    "type": "array"
+                }
+            },
+            "title": "NetworkSecurityGroupConfig",
             "type": "object"
         }
     },
@@ -5830,6 +5857,9 @@ Each event will be submitted via HTTP POST to the user provided URL.
                 "network_config": {
                     "$ref": "#/definitions/NetworkConfig"
                 },
+                "nsg_config": {
+                    "$ref": "#/definitions/NetworkSecurityGroupConfig"
+                },
                 "proxy_vm_sku": {
                     "default": "Standard_B2s",
                     "title": "Proxy Vm Sku",
@@ -5935,6 +5965,27 @@ Each event will be submitted via HTTP POST to the user provided URL.
                 }
             },
             "title": "NetworkConfig",
+            "type": "object"
+        },
+        "NetworkSecurityGroupConfig": {
+            "properties": {
+                "allow_IPs": {
+                    "default": [],
+                    "items": {
+                        "type": "string"
+                    },
+                    "title": "Allow Ips",
+                    "type": "array"
+                },
+                "allow_service_tags": {
+                    "items": {
+                        "type": "string"
+                    },
+                    "title": "Allow Service Tags",
+                    "type": "array"
+                }
+            },
+            "title": "NetworkSecurityGroupConfig",
             "type": "object"
         },
         "NoReproReport": {
--- a/src/pytypes/onefuzztypes/models.py
+++ b/src/pytypes/onefuzztypes/models.py
@@ -799,6 +799,11 @@ class NetworkConfig(BaseModel):
     subnet: str = Field(default="10.0.0.0/16")
 
 
+class NetworkSecurityGroupConfig(BaseModel):
+    allow_service_tags: Optional[List[str]]
+    allow_IPs: List[str] = Field(default=[])
+
+
 class KeyvaultExtensionConfig(BaseModel):
     keyvault_name: str
     cert_name: str
@@ -841,6 +846,9 @@ class InstanceConfig(BaseModel):
 
     allowed_aad_tenants: List[UUID]
     network_config: NetworkConfig = Field(default_factory=NetworkConfig)
+    nsg_config: NetworkSecurityGroupConfig = Field(
+        default_factory=NetworkSecurityGroupConfig
+    )
     extensions: Optional[AzureVmExtensionConfig]
     proxy_vm_sku: str = Field(default="Standard_B2s")
 


src/pytypes/onefuzztypes/models.py
```diff
@@ -799,6 +799,11 @@ class NetworkConfig(BaseModel):
     subnet: str = Field(default="10.0.0.0/16")
 
 
+class NetworkSecurityGroupConfig(BaseModel):
+    allow_service_tags: Optional[List[str]]
+    allow_IPs: List[str] = Field(default=[])
```
ranweiler, line 804: Python attribute names are conventionally always lowercase. See [here](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables).

I wonder if it would make more sense to write `allowed_` instead of `allow_`. The latter sounds more like an imperative, and at a glance, I expect the value to be boolean. To me, using the adjective `allowed_` builds more of a noun phrase, which better matches the fact that this field is a "list of allowed IPs". WDYT?

If we applied both of my suggestions, this would become:
```suggestion
    allowed_ips: List[str] = Field(default=[])
```

src/pytypes/onefuzztypes/models.py
```diff
@@ -799,6 +799,11 @@ class NetworkConfig(BaseModel):
     subnet: str = Field(default="10.0.0.0/16")
 
 
+class NetworkSecurityGroupConfig(BaseModel):
+    allow_service_tags: Optional[List[str]]
```
ranweiler, line 803: If you choose to apply my suggestion below, then this should become:
```suggestion
    allowed_service_tags: Optional[List[str]]
```
(they should match, whatever you go with)

src/pytypes/onefuzztypes/models.py
ranweiler: Oh, and I just noted that in other models, we have `allowed_aad_tenants` for a list-valued field, so my suggestion is more consistent with that.

ranweiler (commented):
Can you please update the PR description to include the specific scenarios you tested, and what it concretely means to "test" them?

bmc-msft (changes requested):

src/pytypes/onefuzztypes/models.py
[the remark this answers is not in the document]
nharper285: No _we do_ want the list to be empty. An empty list means that no src IPs are allowed through the NSG.

nharper285: Okay updating!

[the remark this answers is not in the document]
bmc-msft: Doesn't that leave the service by default as "nobody can connect to the service"?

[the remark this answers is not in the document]
nharper285: "Nobody can connect to debug proxy", which is the behavior we agreed on in our previous conversation. We are likely making the 'set_nsg_rules' param in deploy.py required so that admins know that they'd be locking down by default.

ranweiler: Also, warning, there is a general Python gotcha here we should be careful of: the default value `[]` is mutable. This can cause issues in some situations when you mutate the default value.

Example with plain old functions:
```python
>>> def foo(values=[], x=None):
...     if x is not None:
...             values.append(x)
...     print(values)
...

>>> foo()
[]

# This will mutate the specific instance of `list` that is `foo`'s default arg!
>>> foo(x='hi')
['hi']

# We're printing the default value, which was mutated.
# Our default value isn't a new, empty list. It is a reference to a specific list.
>>> foo()
['hi']
```
Pydantic appears to handle this for us by deep-copying under the hood, but the fact that this works is NOT a given.

ranweiler: (here's where Pydantic has code to do what we really want in this situation: https://github.com/samuelcolvin/pydantic/blob/master/pydantic/fields.py#L375-L376)

bmc-msft: Note, `Field(default_factory=list)` will work here.

bmc-msft: We use this pattern in the CLI a few times:

````
onefuzz/backend.py
93:    features: Set[str] = Field(default_factory=set)

onefuzz/job_templates/cache.py
26:    entries: Dict[str, EndpointCache] = Field(default_factory=dict)
```

docs/webhook_events.md
```diff
@@ -649,6 +649,9 @@ Each event will be submitted via HTTP POST to the user provided URL.
             "address_space": "10.0.0.0/8",
             "subnet": "10.0.0.0/16"
         },
+        "nsg_config": {
```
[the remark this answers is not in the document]
bmc-msft: @stishkin The build pipeline requires the documentation is always up to date.

https://github.com/microsoft/onefuzz/blob/58eea1e51a84efe8531b3f2d72cfad9e74514cb7/src/ci/onefuzztypes.sh#L28-L29

src/pytypes/onefuzztypes/models.py
nharper285: Updated! Thanks for the explanation!

src/pytypes/onefuzztypes/models.py
```diff
@@ -799,6 +799,11 @@ class NetworkConfig(BaseModel):
     subnet: str = Field(default="10.0.0.0/16")
 
 
+class NetworkSecurityGroupConfig(BaseModel):
+    allowed_service_tags: Optional[List[str]]
+    allowed_IPs: List[str] = Field(default_factory=list)
```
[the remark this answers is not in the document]
nharper285: Whoops I missed this.

src/pytypes/onefuzztypes/models.py
```diff
@@ -799,6 +799,11 @@ class NetworkConfig(BaseModel):
     subnet: str = Field(default="10.0.0.0/16")
 
 
+class NetworkSecurityGroupConfig(BaseModel):
```
ranweiler, line 802: Zooming out just a little bit: do you think we should name this something like `ProxyNetworkSecurityGroupConfig`?

src/pytypes/onefuzztypes/models.py
```diff
@@ -841,6 +846,9 @@ class InstanceConfig(BaseModel):
 
     allowed_aad_tenants: List[UUID]
     network_config: NetworkConfig = Field(default_factory=NetworkConfig)
+    nsg_config: NetworkSecurityGroupConfig = Field(
```
ranweiler, line 849: Should we name this field `proxy_nsg_config`?

src/pytypes/onefuzztypes/models.py
ranweiler: This is probably less important than the name of the field on the `InstanceConfig`.

ranweiler: I think this is fine as-is, and we can even imagine it being re-used for something that isn't proxy-specific.

src/pytypes/onefuzztypes/models.py
nharper285: Personally I think we can keep it more general in case we have use for it in other contexts.

ranweiler: Yeah, maybe! OTOH, that's a bit far out, and it's easy to just add another (more specific) field. Keeping it as-is for that reason kind of makes the bet that (1) we'll want another NSG, (2) that this unknown future NSG will want to have the _exact same settings_ as the proxy NSG. My guess is that is unlikely. I think there's also the risk of `network_config` (for the VMSS vnet) and `nsg_config` seeming related, when the NSG specifically _not_ for the VMSS network! So the extra specificity here might actually be a really helpful distinguisher.

To contrast with my reasoning elsewhere, where I thought it made more sense to keep the config class name as-is, and more general: if we _did_ add another NSG, which needed a config, "an NSG config" is a more generic concept. The fields of an an NSG config are determined by what Azure NSG resources provide, not their purpose in OneFuzz. In other words, the role of the NSG (which we provide) is unrelated to the configuration values (which we don't provide).

nharper285: I think that's a fair point. 

Do you want both to be changed? I.e. proxy_nsg_config & ProxyNetworkSecurityGroupConfig?

nharper285: Updated.

ranweiler: 👍 to what @bmc-msft said, that's exactly where my own thinking landed.

ranweiler (approved):

nharper285:
@ranweiler, tested upgrading old instance with changes: 
image

Issue #1316, opened 2021-10-04
Update service deployment scripts to support setting of NSG configuration 


stishkin:
Update `deploy.py` and `registration.py` scripts with new **required** parameter --nsg-configuration. This used to create nsg_configuration entry in instance config to include a list of allowed NSG rules. Empty list means - no-one is allowed to connect to debug or repro proxies. Setting ["*"] as allowed IPs let's anyone connect.

TODO: needs investigation if passing a file path to NSG rules is better than supporting a JSON blob from command line. 

(Storing rules in a file means they can be stored in version control...)

PR #1334, opened 2021-10-07
Refactor set_admins into configuration library and update deployment parameters to include required nsg-rules

nharper285:

_What is this about?_
This PR is intended to introduce functionality to the deploy.py file that requires nsg-rules during deployment. 
In addition, I've refactored much of the set_admins code into a more general configuration.py file that acts as a new library of instance_config-based functionality. The nsg_rule update functionality has been included in this file. 

## PR Checklist
* [x] Applies to work item: #1316
* [ ] ~CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.~
* [ ] ~Tests added/passed~
* [ ] ~Requires documentation to be updated~

## Info on Pull Request

_What does this include?_
Changes to deploy.py: 
* New required param (nsg_config) that takes in a list of IPs to act as a default IP allow list. 
* New file, configuration.py, that includes code from set_admins.py that deals with instance_config updates. Also includes nsg_config update functions
* Removed instance_config functions from set_admins.py

## Validation Steps Performed
Deploying test build and testing value config. 

More specifically, to test and validate this work, I am downloading the 'custom' release-artifacts from the below build and deploying them to my subscription. I then test the following using the 'custom' CLI from release-artifiacts: 
* [ ] Run >onefuzz instance_config update @C:\body.json where body.json includes a list of individual IPs and IP ranges
* [ ] Run >onefuzz instance_config update @C:\body.json where body.json includes updated list of service tags
* [ ] Upgraded (not just new) instances work, as specified by the existing test checklist


src/deployment/configuration.py
```diff
@@ -0,0 +1,109 @@
...
+            "admins": json.dumps(admins_as_str),
+        },
+    )
+
+
+def update_nsg(
+    table_service: TableService,
+    resource_group: str,
+    nsg_ip_rules: List[str],
+    # nsg_tag_rules: Optional[List[str]],
```
stishkin, line 82: nsg_ip_rules here is the JSON blob that includes list of IPs and list of Service Tags ?

src/deployment/deploy.py
```diff
@@ -569,12 +570,22 @@ def apply_migrations(self) -> None:
         migrate(table_service, self.migrations)
 
     def set_instance_config(self) -> None:
-        logger.info("setting instance config")
+        logger.info("setting instance configfdfdfdfd")
```
stishkin, line 573: configfdfdfdfd ?

src/deployment/requirements.txt
```diff
@@ -9,3 +9,4 @@ pyfunctional==1.4.3
 pyopenssl==19.1.0
 adal~=1.2.5
 idna<3,>=2.5
+
```
bmc-msft, line 12:
```suggestion
```

src/deployment/deploy.py
bmc-msft: Time to fix now that you're looking for review?

src/deployment/configuration.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    # if nsg_tag_rules:
+    #     table_service.insert_or_merge_entity(
+    #         TABLE_NAME,
+    #         {
+    #             "PartitionKey": resource_group,
+    #             "RowKey": resource_group,
+    #             "admins": json.dumps(nsg_ip_rules),
+    #         },
+    #     )
```
bmc-msft, line 104: Why is this commented out?

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         set -ex
         ls artifacts
         cd src/cli
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip==21.2.4
```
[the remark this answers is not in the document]
nharper285: I meant to add this locally. I'll remove.

src/deployment/configuration.py
nharper285: Removed.

src/deployment/deploy.py
nharper285: Fixed.

src/deployment/requirements.txt
nharper285: Done.

src/deployment/configuration.py
```diff
@@ -0,0 +1,99 @@
...
+            "RowKey": resource_group,
+            "admins": json.dumps(admins_as_str),
+        },
+    )
+
+
+def update_nsg(
+    table_service: TableService,
+    resource_group: str,
+    allowed_rules: str,
```
ranweiler, line 79: What if this instead took a `List[NsgRule]`, where `NsgRule` was a parsed IP or service tag? Then the parsing can be decoupled from the updating.

src/deployment/configuration.py
```diff
@@ -0,0 +1,99 @@
...
+) -> None:
+    create_if_missing(table_service)
+    
+    rules_as_str = allowed_rules.split(" ")
```
ranweiler, line 83: I'm pretty sure this doesn't do what we want on various inputs with extra spaces.

Example:
```python
>>> rules = 'a    b   c d    e'

# What we probably want.
>>> rules.split()

# What we have.
['a', 'b', 'c', 'd', 'e']
>>> rules.split(' ')
['a', '', '', '', 'b', '', '', 'c', 'd', '', '', '', 'e']
```

src/deployment/configuration.py
```diff
@@ -0,0 +1,99 @@
...
+
+    nsg_config = {
```
ranweiler: Hmm, didn't we have a type that we can use here, instead of a `dict`?

src/deployment/configuration.py
```diff
@@ -0,0 +1,99 @@
...
+        if not table_service.exists(TABLE_NAME):
+            table_service.create_table(TABLE_NAME)
+    finally:
+        enable_storage_client_logging()
+
+
+def update_allowed_aad_tenants(
+    table_service: TableService, resource_group: str, tenants: List[UUID]
+) -> None:
+    create_if_missing(table_service)
```
ranweiler, line 46: The fact that we are constantly calling `create_if_missing()` suggests that maybe we should have a wrapper class that has as member a known-to-exist (via the constructor) `TableService`. It could also encapsulate a specific `resource_group`. Then we only need to ensure the existence of the resource once.

src/deployment/configuration.py
ranweiler: Also, the naming here is confusing. I think you're just following a previous pattern, but given that this is a `list` obtained by parsing a string, calling it `.._as_str` is misleading.

nharper285: I can remove the param and use the default .split()

src/deployment/configuration.py
nharper285: Yes, see previous comment. I currently cannot use any onefuzztypes because they aren't available during the package build stage.

ranweiler: Can we fix that, and make `onefuzztypes` available?

src/deployment/configuration.py
```diff
@@ -0,0 +1,98 @@
...
+    table_service: TableService,
+    resource_group: str,
+    allowed_rules: str,
+) -> None:
+    create_if_missing(table_service)
+    
+    rules_as_str = allowed_rules.split(" ")
+
+    nsg_config = {
+        "allowed_ips": rules_as_str
```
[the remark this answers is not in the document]
stishkin: that was the original idea. To have tags and IPs separate. I'm fine with that, just concat the lists when you set the rules in the NSG.

src/deployment/configuration.py
nharper285: I'll look into some restraints for tags and definitely work on some for IPs/IP ranges!

ranweiler: Given that we'd be using it in `configuration`, we'd want to define it there (or else we'd have a circular dependency).

src/deployment/configuration.py
ranweiler: @nharper285, it would also be fine (probably better) to address that kind of change as a separate PR.

src/deployment/configuration.py
```diff
@@ -0,0 +1,170 @@
...
+        try:
+            ipaddress.ip_network(value)
+            return
+        except ValueError:
+            pass
+
+        self.is_tag = True
+
+
+## Disable logging from storageclient. This module displays an error message
```
stishkin, line 85: Commented out code - was it like this before ?

src/deployment/configuration.py
```diff
@@ -0,0 +1,170 @@
...
+            if not table_service.exists(TABLE_NAME):
+                table_service.create_table(TABLE_NAME)
+        finally:
+            self.enable_storage_client_logging()
+
+
+class NsgRule:
+
+    rule: str
+    is_tag: bool
```
stishkin, line 51: If I use

rule2 = NsgRule("127.0.0.*")

Then I get "is_tag" set to True

Is that expected ?

nharper285: Is this a valid NSG rule by NSG standards? Like can you put this into an NSG? 

I agree it should not be marked as "is_tag", but if anything it should just be invalid. 

@ranweiler and I talked a good bit about how much validation we want to do at the CLI level, and this might not be the best place for this kind of check. Instead, this kind of validation could be done at the NSG level.

src/deployment/configuration.py
nharper285: It's been moved into the INstanceConfigClient class. I removed the commented out code.

Issue #1319, opened 2021-10-04
NSG deployment on a creation of new debug/repro proxy.


stishkin:
Deploy NSG (if there is no NSG in the region where proxy is being deployed) on every new debug or repro proxy deployed.
NSG must be deployed in the same region as the proxy in order to associate NSG with proxy's NIC.

Update NSG rules from Instance Config. In case of error then handle it same way as when proxy deployment fails. 

[Might need to separate this work item into two, in order handle debug and repro proxies separately]

PR #1340, opened 2021-10-08
NSG deployment on a creation of new debug/repro proxy.

stishkin:

_What is this about?_

## PR Checklist
* [X] Applies to work item: #1319 
* [X] Tests added/passed

## Info on Pull Request

_What does this include?_
- New nsg.py interface for NSG Azure API
- Association/Dissociation of NSG on VM creation/deletion (and integration into debug_proxy and repro VMs)

## Validation Steps Performed

- Deploy new scaleset and when proxy is created, NSG is also created in the same region. Proxy's NIC is associated with NSG (as of this PR NSG allows all connection through, thus preserving existing traffic flow)
- Deploy a repro VM, the NIC of the repro VM gets associated with NSG in the same region
- Delete repro VM, repro VM's NIC is dissociated from NSG 
- Connect to debug or repro VM to validate the connections are not blocked


src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,311 @@
...
+                    nic.name,
+                    err,
+                )
+            ],
+        )
+
+    return None
+
+
+class NSG(BaseModel):
```
nharper285, line 274: This should be in models.py, right? I understand that VM.py has it's own model class defined in the same file, but it seems to me like we should be keeping all the models in models.py

src/api-service/__app__/onefuzzlib/azure/vm.py
```diff
@@ -181,16 +187,20 @@ def has_components(name: str) -> bool:
     return False
 
 
-def delete_vm_components(name: str) -> bool:
+def delete_vm_components(name: str, nsg: Optional[NSG]) -> bool:
```
nharper285, line 190: I'm not sure it makes sense to pass int he entire NSG. Is there a way to do this where the VMs location is used to retrieve the NSG for that region?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,311 @@
...
+        return cast(
+            NetworkSecurityGroup,
+            network_client.network_security_groups.get(resource_group, name),
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        logging.debug("nsg %s does not exist %s", name, err)
+        return None
+
+
+def create_nsg(name: str, location: Region) -> Union[None, Error]:
```
nharper285, line 44: 'name' is the region, right? Is this appropriate? It's not really the NSG for the region, it's the NSG for the proxy _for a particular region_. As some people mentioned in reviewing my PR, we might leave room for future NSGs which are more general than proxy. 

I'm just proposing that we maybe name the NSG something like 'proxy-nsg-GUID' or something of that nature.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+            return None
+        return Error(
+            code=ErrorCode.UNABLE_TO_CREATE,
+            errors=["Unable to create nsg %s due to %s" % (name, err)],
+        )
+    return None
+
+
+def update_nsg(nsg: NetworkSecurityGroup) -> Union[None, Error]:
```
nharper285, line 73: What exactly is the purpose of update if it's not updating NSG rules? Maybe I'm missing the more general purpose of this function elsewhere.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,311 @@
...
+    for src in sources:
+        security_rules.append(
+            SecurityRule(
+                name="Allow" + str(priority),
+                protocol=SecurityRuleProtocol.TCP,
+                source_port_range="*",
+                destination_port_range="*",
+                source_address_prefix=src,
+                destination_address_prefix="*",
+                access=SecurityRuleAccess.ALLOW,
```
nharper285, line 140: This is a more general question, but do we believe that we should have a new storage table entry for NSG? Since there are potentially multiple NSGs because of different regions, does it make sense to make a new table entry?

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -126,6 +144,9 @@ def set_failed(self, error: Error) -> None:
 
     def extensions_launch(self) -> None:
         vm = self.get_vm()
+        if isinstance(vm, Error):
+            self.set_failed(vm)
+            return
```
nharper285, line 149: What was the reason behind these changes?

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: I based this on vm.py code. Possible that I should move this to models.py, I'll take a look.

src/api-service/__app__/onefuzzlib/azure/vm.py
stishkin: VM's NIC is associated with an NSG. From within VM code VM does not know which NSG it is associated with. 
In case there are multiple NSGs for a region - VM needs to know which one it needs to dissociate from.

Unless we have explicit agreement that there is never more than one NSG per region

nharper285: Okay yeah I suppose we'd have to say that there's only one NSG per pregion.

src/api-service/__app__/onefuzzlib/azure/nsg.py
nharper285: I guess I was making a more general point and using this function as a place to make it. 

The question is - 'do we want to name the NSG the region?' Referring to your point above, if there are multiple NSGs per region, the this doesn't make sense.

src/api-service/__app__/onefuzzlib/proxy.py
stishkin: get_vm is ether getting existing VM or creating a new one (and we don't know what is going to happen when we call the function) and since get_vm might create a new VM and therefore we might need to associate that VM with NSG and this operation might fail

src/api-service/__app__/onefuzzlib/azure/nsg.py
nharper285: Would set_rules not be one of these changes that we'd want to apply?

src/api-service/__app__/onefuzzlib/azure/nsg.py
bmc-msft: My decision point was "If this is intended to be visible to the user in any fashion, then it should go in onefuzztypes. Otherwise, it shouldn't."

bmc-msft: Broadly scoped, this means anything that was events, requests, and responses go in onefuzztypes.  everything else does not.

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: From NSG class point of view - it is not supposed to know how we are naming NSGs. 
I am assigning NSG name to be region since it is one resource per region. Exactly in the same manner as VNETs. They are named same as the region (but in the code they accept name and region separately)

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: that's the only way NSG update works, get Azure nsg structure, change fields and pass it back. So if we change rules - then that will be applied. If there are other NSG manipulation tasks we decide to add, we'd be calling this function to apply them

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,311 @@
...
+                    err,
+                )
+            ],
+        )
+
+    return None
+
+
+class NSG(BaseModel):
+    name: Union[UUID, str]
```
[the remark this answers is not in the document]
stishkin: Yes, I'll remove UUID 

From within NSG - it does not know what name it will get, so I'll keep it as string

src/api-service/__app__/onefuzzlib/azure/vm.py
```diff
@@ -215,6 +225,7 @@ class VM(BaseModel):
     sku: str
     image: str
     auth: Authentication
+    nsg: Optional[NSG]
```
bmc-msft, line 228: This  appears to always be provided now.  Does this need to be optional?

src/api-service/__app__/onefuzzlib/azure/vm.py
```diff
@@ -248,10 +259,11 @@ def create(self) -> Union[None, Error]:
             self.image,
             self.auth.password,
             self.auth.public_key,
+            self.nsg,
         )
 
     def delete(self) -> bool:
-        return delete_vm_components(str(self.name))
+        return delete_vm_components(str(self.name), self.nsg)
```
bmc-msft, line 266: If there is one NSG per region, and now multiple debug VMs get spun up, then 1 gets deleted, won't cause issues?

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -68,19 +69,36 @@ class Proxy(ORMMixin):
...
         return ("region", "proxy_id")
 
-    def get_vm(self) -> VM:
+    def get_vm(self) -> Union[VM, Error]:
+        nsg = NSG(
+            name=self.region,
+            region=self.region,
+        )
+
+        result = nsg.create()
```
bmc-msft, line 78: `get_vm` doesn't actually create VMs or any of the resources behind a VM at this time.    This doesn't behave like any of the other underlying resources associated with a VM, like nics, IPs, etc.

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -158,6 +179,9 @@ def extensions_launch(self) -> None:
 
     def stopping(self) -> None:
         vm = self.get_vm()
```
bmc-msft, line 181: This feels weird, considering `get_vm` comment earlier, where during the VM deletion process, we'll create an NSG, then delete it... is that intended?

src/api-service/__app__/onefuzzlib/proxy.py
bmc-msft: `get_vm` doesn't create a VM if it doesn't exist.  you have to call `VM.create()` to do that.

src/api-service/__app__/onefuzzlib/proxy.py
stishkin: NSG does not get deleted when VM is deleted, VM's NIC is dissociated from NSG. To avoid a window where VM is in the open - we need to dissociate right before NIC gets deleted.

stishkin: No, NIC has to be dissociated first. 
But even if NIC is available - the actual virtual machine is deleted prior to the point where we reach NIC deletion (line 195). 
So worse case we end up with stand alone "public" NIC without a VM

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+
+    logging.debug("getting nsg: %s", name)
+    network_client = get_network_client()
+    try:
+        return cast(
+            NetworkSecurityGroup,
+            network_client.network_security_groups.get(resource_group, name),
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        logging.debug("nsg %s does not exist %s", name, err)
```
ranweiler, line 39:
```suggestion
        logging.debug("nsg %s does not exist: %s", name, err)
```

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+    resource_group = get_base_resource_group()
+
+    logging.debug("getting nsg: %s", name)
+    network_client = get_network_client()
+    try:
+        return cast(
+            NetworkSecurityGroup,
+            network_client.network_security_groups.get(resource_group, name),
+        )
+    except (ResourceNotFoundError, CloudError) as err:
```
ranweiler, line 38: Why does `CloudError` imply that the NSG does not exist? Could that just be a transient Azure API error?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+
+    logging.info("creating nsg %s:%s:%s", resource_group, location, name)
+    network_client = get_network_client()
+
+    params: Dict = {
+        "location": location,
+    }
+
+    if "ONEFUZZ_OWNER" in os.environ:
+        params["tags"] = {"OWNER": os.environ["ONEFUZZ_OWNER"]}
```
ranweiler, line 54: Newline to make it clear that this `if` is unrelated to the `try/except` that follows:
```suggestion
        params["tags"] = {"OWNER": os.environ["ONEFUZZ_OWNER"]}
        
```

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+    try:
+        network_client.network_security_groups.begin_create_or_update(
+            resource_group, name, params
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        if "The request failed due to conflict with a concurrent request" in str(err):
```
ranweiler, line 60: How do we know that the concurrent request was (1) a request to create the same NSG, (2) one that succeeded? In other words, why are we justified in returning `None`?

Example concern: we're invoking `begin_create_or_update()`. Can this request fail due to a concurrent request to _delete_ the NSG? If that were true, the concurrently-serviced `delete` would delete the NSG, but we'd return `None` here, implying the NSG creation request was successful.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+            logging.debug(
+                "create NSG had conflicts with concurrent request, ignoring %s", err
+            )
+            return None
+        return Error(
+            code=ErrorCode.UNABLE_TO_CREATE,
+            errors=["Unable to create nsg %s due to %s" % (nsg.name, err)],
```
ranweiler, line 90:
```suggestion
            errors=["Unable to update nsg %s due to %s" % (nsg.name, err)],
```

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+
+    logging.info("updating nsg %s:%s:%s", resource_group, nsg.location, nsg.name)
+    network_client = get_network_client()
+
+    try:
+        network_client.network_security_groups.begin_create_or_update(
+            resource_group, nsg.name, nsg
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        if "The request failed due to conflict with a concurrent request" in str(err):
```
ranweiler, line 83: Since this is getting repeated, with a hardcoded string, can we factor this logic out into a predicate function like `is_concurrent_request_conflict(err: CloudError) -> bool`?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+            )
+            return None
+        return Error(
+            code=ErrorCode.UNABLE_TO_CREATE,
+            errors=["Unable to create nsg %s due to %s" % (nsg.name, err)],
+        )
+    return None
+
+
+def delete_nsg(name: str) -> Any:
```
ranweiler, line 95: What can actually be returned here? Can we inspect the response and return a `Union` instead?

We really want to avoid `Any`, and if we can't, we need to document what's going on.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+    logging.info(
+        "setting allowed incoming connection sources for nsg: %s %s",
+        resource_group,
+        name,
+    )
+    security_rules = []
+    min_priority = 100
+    max_priority = 4096
```
ranweiler, line 119: Is this an inclusive upper bound?

If so, and the set of priorities is `{100, 101, ..., 4095, 4096}`, then don't we want `max_rule_count = max_priority - min_priority + 1`?

Also, I suggest computing `max_rule_count` explicitly as a local var, and using it both in the check and in the error message.

Ideally we'd also do this input sanitization earlier in the function, before making multiple Azure calls.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+            errors=["cannot update nsg rules. nsg %s not found" % name],
+        )
+
+    logging.info(
+        "setting allowed incoming connection sources for nsg: %s %s",
+        resource_group,
+        name,
+    )
+    security_rules = []
+    min_priority = 100
```
ranweiler, line 118: Can we link to docs where we derive this values?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+                source_port_range="*",
+                destination_port_range="*",
+                source_address_prefix=src,
+                destination_address_prefix="*",
+                access=SecurityRuleAccess.ALLOW,
+                priority=priority,  # between 100 and 4096
+                direction="Inbound",
+            )
+        )
+        priority += 1
```
ranweiler, line 144:
```suggestion
        # Will not exceed `max_priority` due to earlier check of `len(sources)`.
        priority += 1
```

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+
+    nsg.security_rules = security_rules
+    return update_nsg(nsg)
+
+
+def clear_all_rules(name: str) -> Union[None, Error]:
+    return set_allowed(name, [])
+
+
+def get_all_rules(name: str) -> Any:
```
ranweiler, line 154: Apply earlier comment about `Any`. How can we avoid this, or why can't we?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+    logging.info("associating nic %s with nsg: %s %s", nic.name, resource_group, name)
+
+    nic.network_security_group = nsg
+    network_client = get_network_client()
+    try:
+        network_client.network_interfaces.begin_create_or_update(
+            resource_group, nic.name, nic
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        if "The request failed due to conflict with a concurrent request" in str(err):
```
ranweiler, line 196: Apply earlier comments about identifying concurrent request conflicts, and our success conditions.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+    return None
+
+
+class NSG(BaseModel):
+    name: str
+    region: Region
+
+    @validator("name", allow_reuse=True)
+    def check_name(cls, value: str) -> str:
+        if len(value) > 80:
```
ranweiler, line 279: Where does this value come from? Are there docs we can link to?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,309 @@
...
+            raise ValueError("NSG name too long")
+        return value
+
+    def create(self) -> Union[None, Error]:
+        if self.get() is not None:
```
ranweiler, line 284: I think we're using `get()` as an existence check and skipping the create call if it succeeds, but this warrants a comment.

ranweiler: Important to note: is this existence check necessary to avoid an error? Is it an optimization? &c

src/api-service/__app__/onefuzzlib/proxy.py
```diff
@@ -90,6 +93,23 @@ def init(self) -> None:
...
+                name=self.region,
+                region=self.region,
+            )
+
+            result = nsg.create()
+            if isinstance(result, Error):
+                self.set_failed(result)
+                return
+
+            result = nsg.set_allowed_sources(["*"])
```
ranweiler, line 106: This determines important security-relevant defaults, right? We need to extensively document such choices, and ideally also use a named constant.

In this case, don't we want the default behavior to be "deny all incoming IPs"?

src/api-service/__app__/onefuzzlib/repro.py
```diff
@@ -86,6 +87,20 @@ def init(self) -> None:
...
+            nsg = NSG(
+                name=vm.region,
+                region=vm.region,
+            )
+            result = nsg.create()
+            if isinstance(result, Error):
+                self.set_failed(result)
+                return
+
+            result = nsg.set_allowed_sources(["*"])
```
ranweiler, line 99: Apply earlier comment about documenting, and incoming IP defaults. Don't we actually want to ensure that we reject all incoming sources, by default?

src/api-service/__app__/onefuzzlib/proxy.py
nharper285: I think this might have been a placeholder that he was using while waiting for my changes from last week.

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: https://github.com/microsoft/onefuzz/blob/19ecba43cb964e4c3f48b30a8b285d09efbc5d31/src/api-service/__app__/onefuzzlib/azure/vm.py#L114

I'm using vm.py as template. Since that code been validated and battle-tested, nsg is following same pattern

src/api-service/__app__/onefuzzlib/proxy.py
stishkin: This is a default for this PR, since the point of this PR is just to add NSG. Applying NSG rules is coming up down the road.
As per @nharpher285 comment

src/api-service/__app__/onefuzzlib/azure/nsg.py
ranweiler: In that case, (1) my comment applies there too, (2) maybe we should have a general predicate that encodes this type of exception checking into something like a Python [context manager](https://docs.python.org/3/reference/datamodel.html#context-managers).

If you look at the history of that file, it originally _only_ caught `CloudError`, and then `ResourceNotFoundError` was added when the external `azure` library was upgraded.

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: Opened work item: 
https://github.com/microsoft/onefuzz/issues/1349

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: https://github.com/microsoft/onefuzz/issues/1349

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: https://github.com/microsoft/onefuzz/issues/1350

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: Actually there are only 1000 rules are allowed, so all that logic is a bit different

src/api-service/__app__/onefuzzlib/azure/nsg.py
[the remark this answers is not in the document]
stishkin: https://github.com/microsoft/onefuzz/issues/1354

bmc-msft (approved):

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,333 @@
...
+                source_port_range="*",
+                destination_port_range="*",
+                source_address_prefix=src,
+                destination_address_prefix="*",
+                access=SecurityRuleAccess.ALLOW,
+                priority=priority,  # between 100 and 4096
+                direction="Inbound",
+            )
+        )
+        # Will not exceed `max_rule_count` due to earlier check of `len(sources)`.
```
ranweiler, line 165:
```suggestion
        # Will not exceed `max_rule_count` or max NSG priority (4096) due to earlier check of `len(sources)`.
```

ranweiler (approved):

Issue #1321, opened 2021-10-04
Handle Proxy deletion through CLI 


stishkin:
If CLI supports proxy deletion command, then need to delete NSG for the same region that debug proxy was deleted for.

Repro proxy deletion should not influence NSG lifetime.

PR #1358, opened 2021-10-13
Delete NSG if no resources associated with it

stishkin:

_What is this about?_

## PR Checklist
* [X] Applies to work item: #1321 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Perform on timer check for unused NSGs and delete them

## Validation Steps Performed

Deploy OneFuzz, add scaleset to a unique region, and then delete scaleset from the region.
Expected behavior - debug proxy is deleted from the above region, and NSG is deleted as well.


src/api-service/__app__/timer_proxy/__init__.py
```diff
@@ -50,3 +51,10 @@ def main(mytimer: func.TimerRequest) -> None:  # noqa: F841
     for region in regions:
         if all(x.outdated for x in proxies if x.region == region):
             Proxy.get_or_create(region)
+
+    # if there are NSGs with name same as the region that they are allocated
+    # and have no NIC associated with it then delete the NSG
+    for nsg in NSG.list_nsgs():
+        if not (nsg.location in regions and nsg.location == nsg.name):
```
ranweiler, line 58: I don't think this conditional expr matches the comment. If my NSG has a name that doesn't match the location, won't we always delete it if it has no NIC?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -72,6 +72,15 @@ def create_nsg(name: str, location: Region) -> Union[None, Error]:
     return None
 
 
+def list_nsgs() -> List[NetworkSecurityGroup]:
+    resource_group = get_base_resource_group()
+    network_client = get_network_client()
+    nsgs = []
```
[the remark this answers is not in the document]
stishkin: I'll try out your suggestions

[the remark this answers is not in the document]
ranweiler: Ah, got it. What you did is consistent with our preference for `for` loops, which is good in context. In this case, since we _only_ append elements, I think it is possibly clearer to use my suggestion. It is also idiomatic for this codebase, since we do use list comprehensions for simple cases.

IMO, the `for` loop approach is best in this codebase if we have any nontrivial logic in each loop iterator. It would definitely break existing conventions to start using `map()` and `filter()` combinators a bunch.

src/api-service/__app__/timer_proxy/__init__.py
```diff
@@ -50,3 +51,10 @@ def main(mytimer: func.TimerRequest) -> None:  # noqa: F841
     for region in regions:
         if all(x.outdated for x in proxies if x.region == region):
             Proxy.get_or_create(region)
+
+    # if there are NSGs with name same as the region that they are allocated
+    # and have no NIC associated with it then delete the NSG
+    for nsg in list_nsgs():
```
ranweiler, line 57: I think it would be valuable to factor out the logic that decides to delete the region or not as a pure fucnction, then unit test it.

src/api-service/tests/test_nsg.py
```diff
@@ -0,0 +1,31 @@
...
+# Licensed under the MIT License.
+
+from typing import Set, Tuple
+
+import pytest
+from onefuzztypes.primitives import Region
+
+from __app__.onefuzzlib.azure.nsg import ok_to_delete
+
+NsgOkToDeleteTestCase = Tuple[Set[Region], str, str, bool]
```
ranweiler, line 13:
```suggestion
# Active regions, NSG region, NSG name, expected result
NsgOkToDeleteTestCase = Tuple[Set[Region], str, str, bool]
```

src/ci/proxy.sh
```diff
@@ -12,9 +12,11 @@ mkdir -p artifacts/proxy
 cd src/proxy-manager
 cargo fmt -- --check
 cargo clippy --release -- -D warnings
-# RUSTSEC-2020-0016: a dependency net2 (pulled in from tokio) is deprecated
-# RUSTSEC-2021-0065: a dependency anymap is no longer supported
-cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2020-0016 --ignore RUSTSEC-2021-0065
+# RUSTSEC-2020-0016: a dependency `net2` (pulled in from `tokio`) is deprecated
```
ranweiler, line 15: I suggest you update the target PR (feature branch) with these unrelated changes before merging, to keep the feature branch's history easy to understand.

ranweiler (approved):

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -95,6 +101,10 @@ def update_nsg(nsg: NetworkSecurityGroup) -> Union[None, Error]:
     return None
 
 
+def ok_to_delete(regions: Set[Region], nsg_location: str, nsg_name: str) -> bool:
+    return nsg_location not in regions and nsg_location == nsg_name
```
chkeita, line 105: I would rename these to give better context 

```suggestion
def ok_to_delete(active_regions: Set[Region], nsg_region: str, nsg_name: str) -> bool:
    return nsg_region not in active_regions and nsg_region == nsg_name
```

chkeita (approved):

PR #1371, opened 2021-10-19
Fix validation of `target_exe` blob name

### LINKED ISSUE #1362: libfuzzer template fails when target_exe is sub-folder inside setup_dir
- Onefuzz version: 3.2.0
- OS: Linux

libfuzzer template throws error when target_exe is specified from a sub-folder inside setup_dir.
Specifying the target_exe in the root of the setup_dir works, but the target executable (setup_dir/bin/exec-sgx) needs its dependent libraries from (setup_dir/lib) so it cannot be in root folder.

I tried --target_exe "bin/exec-sgx"  and --target_exe "./bin/exec-sgx" but both fails with different message.

**Command:** 
onefuzz template libfuzzer basic mystikos exec-sgx 12345 mystikos-pool --target_exe "bin/exec-sgx" --duration 1 --setup_dir /home/mystikos/build/myst_onefuzz_setup

**ERROR:cli:command failed: local file (bin/exec-sgx) is not within the setup directory (/home/mystikos/build/myst_onefuzz_setup)**

**Command:** 
onefuzz template libfuzzer basic mystikos exec-sgx 12345 mystikos-pool --target_exe "./bin/exec-sgx" --duration 1 --setup_dir /home/mystikos/build/myst_onefuzz_setup

**ERROR:cli:command failed: request did not succeed: HTTP 400 - {"code": 450, "errors": ["target_exe must be a canonicalized relative path"]}**

ranweiler:
### Summary
Our validation of `target_exe` blob names spuriously failed when the target was in a nested (virtual) directory. Fix validation to support this case, and check additional Azure blob storage naming criteria. Also, add unit tests.

Closes #1362.

### Explanation

A bug-witnessing `target_exe` value is `bin/fuzz.exe`. This should be accepted.

But consider:
```
>>> import ntpath, posixpath
>>> target_exe = 'bin/fuzz.exe'
>>> posixpath.relpath(target_exe)
'bin/fuzz.exe'
>>> ntpath.relpath(target_exe)
'bin\\fuzz.exe'
```
Our validation required that _each_ `relpath()` invocation above is equivalent to the identity function. But the output of the last `relpath()` call differs from the input.

### Testing

- [x] Can create a job whose `target_exe` is in the root of `setup_dir`
- [x] Can create a job whose `target_exe` is in a nested directory, like `{setup_dir}/bin/fuzz.exe` (from Linux, on Linux pool)
- [x] Can create a job whose `target_exe` is in a nested directory, like `{setup_dir}\\bin\\fuzz.exe` (from Windows, on Windows pool)

The last test case ~~ensures~~ [1] that the CLI normalizes relativized paths as POSIX-style, which is consistent with the Blob Storage Service's use of forward slash (`/`) as the virtual directory delimiter.

[1] It does not. This test passes, because the Windows-style paths still (happen to) work on Windows VMs. Linux jobs created from Windows with this path style fail at task runtime. Captured in #1373.

### DIFF
--- a/src/api-service/tests/test_task_config.py
+++ b/src/api-service/tests/test_task_config.py
@@ -31,6 +31,16 @@ BLOB_NAME_TEST_CASES = [
     ("  ", False),
     ("/".join("a" * 255), False),
     ("a" * 1025, False),
+    # Paths with invalid segments.
+    ("a.", False),
+    ("a..", False),
+    ("a./b", False),
+    ("a/b./c", False),
+    ("a./", False),
+    ("a../", False),
+    ("a./b/", False),
+    ("a/b./c/", False),
+    ("a//", False),
 ]
 
 
--- a/src/api-service/tests/test_task_config.py
+++ b/src/api-service/tests/test_task_config.py
@@ -7,7 +7,6 @@ import pytest
 
 from __app__.onefuzzlib.tasks.config import is_valid_blob_name
 
-
 BLOB_NAME_TEST_CASES = [
     # Valid
     ("fuzz.exe", True),
--- a/src/api-service/tests/test_task_config.py
+++ b/src/api-service/tests/test_task_config.py
@@ -3,9 +3,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+from typing import Tuple
+
 import pytest
 
 from __app__.onefuzzlib.tasks.config import is_valid_blob_name
+
+BlobNameTestCase = Tuple[str, bool]
+
 
 BLOB_NAME_TEST_CASES = [
     # Valid
@@ -44,7 +49,7 @@ BLOB_NAME_TEST_CASES = [
 
 
 @pytest.mark.parametrize("blob_name_test_case", BLOB_NAME_TEST_CASES)
-def test_is_valid_blob_name(blob_name_test_case):
+def test_is_valid_blob_name(blob_name_test_case: BlobNameTestCase) -> None:
     blob_name, expected = blob_name_test_case
 
     is_valid = is_valid_blob_name(blob_name)
--- a/src/api-service/tests/test_task_config.py
+++ b/src/api-service/tests/test_task_config.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import pytest
+
+from __app__.onefuzzlib.tasks.config import is_valid_blob_name
+
+
+BLOB_NAME_TEST_CASES = [
+    # Valid
+    ("fuzz.exe", True),
+    ("bin/fuzz.exe", True),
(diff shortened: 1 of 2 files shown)


src/api-service/tests/test_task_config.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import pytest
+
+from __app__.onefuzzlib.tasks.config import is_valid_blob_name
+
+
+BLOB_NAME_TEST_CASES = [
```
stishkin, line 11: Looking at the spec that you included:

Avoid blob names that end with a dot (.), a forward slash (/), or a sequence or combination of the two. No path segments should end with a dot (.).

Looks like is_valid_blob_name("abcd.")    will return true

src/api-service/tests/test_task_config.py
```diff
@@ -0,0 +1,43 @@
...
+    # Valid
+    ("fuzz.exe", True),
+    ("bin/fuzz.exe", True),
+    ("/".join("a" * 254), True),
+    ("a" * 1024, True),
+    # Invalid (absolute)
+    ("/fuzz.exe", False),
+    ("/bin/fuzz.exe", False),
```
chkeita, line 19: can we add some test with windows style absolute path: c:\..

src/api-service/tests/test_task_config.py
ranweiler: Good idea, will add.

src/api-service/tests/test_task_config.py
ranweiler: I had those originally, but I think it starts getting confusing, and probably isn't where we want our API boundary.

Here's what I'm thinking.
1. The service should expect that blob names as specified by Azure (and not arbitrary, OS-dependent filesystem names). Any client-local filesystem delimiting should be normalized to a representation using only forward slashes (i.e. are POSIX-style). Any other legal characters after that forward-slash normalization should be treated as just part of the name of some path component. Note that an (escaped) backslash (`\`) is a valid part of a POSIX filename.
2. The CLI should use the client OS to help construct valid relative blob names, and should be responsible for normalizing them as POSIX-style.

The old could would actually throw exceptions if Windows absolute paths ever got to the service.

I can still add those test cases if you think it helps for documentation. But the idea is that I want to make it so the service does not have to try to detect the path delimiter style that is being used.

stishkin (approved):

src/api-service/tests/test_task_config.py
chkeita: I don't think the cli currently rewrites the target_exe path so it is possible to receive the windows absolute path.
Plus having the validation on the service side makes it independent from the client code

chkeita: Ok i just looked a the code and we do indeed rewrite the target_exe path. But my second point still stands.

Issue #1349, opened 2021-10-12
Investigate better handling of Azure CloudError when creating Azure resources


stishkin:
See PR discussion here for background information:
https://github.com/microsoft/onefuzz/pull/1340#discussion_r727325069 

AB#35961

PR #1407, opened 2021-10-28
Feature/proxy nsg

stishkin:

_What is this about?_

Add NSG functionality to OneFuzz

Merging feature branch into main


src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,480 @@
...
+def delete_nsg(name: str) -> bool:
+    # NSG can be only deleted if no other resource is associated with it
+    resource_group = get_base_resource_group()
+
+    logging.info("deleting nsg: %s %s", resource_group, name)
+    network_client = get_network_client()
+
+    try:
+        network_client.network_security_groups.begin_delete(resource_group, name)
+        return True
```
chkeita, line 122: it looks we need the result of the operation here to confirm the deletion 
```suggestion
        delete_nsg = network_client.network_security_groups.begin_delete(resource_group, name)
        return delete_nsg.done()
```

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: When we call begin_delete we get Poller object, that if we call "done()" most likely will return false, since it is along running operation and deletion is in progress. [Poller object "done" documentation](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.polling.lropoller?view=azure-python#done--)

So we need to wait for it to be done, I don't think we want to block the execution of the azure function to wait for deletion to complete.

The behavior in the PR aligns with the current implementation across OneFuzz azure module. For example:
[vmss](https://github.com/microsoft/onefuzz/blob/6cdb5e17dbcd7b0fca2e06580842083c7a8fe419/src/api-service/__app__/onefuzzlib/azure/vmss.py#L69)

[vm](https://github.com/microsoft/onefuzz/blob/6cdb5e17dbcd7b0fca2e06580842083c7a8fe419/src/api-service/__app__/onefuzzlib/azure/vm.py#L160)

The  check for whether nsg deletion call should retried happens in parallel inside of `timer_proxy`. And this behavior is aligned with existing behavior from across OneFuzz azure module

chkeita: after discussion and code diving. I update my suggestion to change the name of the function to `begin_delete` to comunicate that a `true` return value does not necessarily mens that the resource was deleted

src/deployment/configuration.py
```diff
@@ -0,0 +1,172 @@
...
+) -> None:
+    tags_as_str = [x.rule for x in allowed_rules if x.is_tag]
+    ips_as_str = [x.rule for x in allowed_rules if not x.is_tag]
+    nsg_config = {"allowed_service_tags": tags_as_str, "allowed_ips": ips_as_str}
+    # create class initialized by table service/resource group outside function that's checked in deploy.py
+    config_client.table_service.insert_or_merge_entity(
+        TABLE_NAME,
+        {
+            "PartitionKey": config_client.resource_group,
+            "RowKey": config_client.resource_group,
```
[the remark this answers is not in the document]
nharper285: Yes this is pulled directly from the original script.

src/deployment/configuration.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+
+def parse_rules(proxy_config: Dict[str, str]) -> List[NsgRule]:
+    allowed_ips = proxy_config["allowed_ips"]
+    allowed_service_tags = proxy_config["allowed_service_tags"]
+
+    nsg_rules = []
+    if "*" in allowed_ips:
+        nsg_rule = NsgRule("*")
```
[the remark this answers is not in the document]
nharper285: Resolved.

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,480 @@
...
+    try:
+        network_client.network_security_groups.begin_create_or_update(
+            resource_group, nsg.name, nsg
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        if is_concurrent_request_error(str(err)):
+            logging.debug(
+                "create NSG had conflicts with concurrent request, ignoring %s", err
+            )
+            return None
```
ranweiler, line 97: Sorry, please remind me: why is this not an error case?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,480 @@
...
+            errors=["Unable to update nsg %s due to %s" % (nsg.name, err)],
+        )
+    return None
+
+
+def ok_to_delete(active_regions: Set[Region], nsg_region: str, nsg_name: str) -> bool:
+    return nsg_region not in active_regions and nsg_region == nsg_name
+
+
+def is_one_fuzz_nsg(nsg_region: str, nsg_name: str) -> bool:
```
ranweiler, line 109:
```suggestion
def is_onefuzz_nsg(nsg_region: str, nsg_name: str) -> bool:
```
(and update use sites)

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,480 @@
...
+            )
+            return None
+        return Error(
+            code=ErrorCode.UNABLE_TO_CREATE,
+            errors=["Unable to update nsg %s due to %s" % (nsg.name, err)],
+        )
+    return None
+
+
+def ok_to_delete(active_regions: Set[Region], nsg_region: str, nsg_name: str) -> bool:
```
ranweiler, line 105: Please document what "ok" is supposed to mean in this context.

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,480 @@
...
+    return nsg_region not in active_regions and nsg_region == nsg_name
+
+
+def is_one_fuzz_nsg(nsg_region: str, nsg_name: str) -> bool:
+    return nsg_region == nsg_name
+
+
+def start_delete_nsg(name: str) -> bool:
```
ranweiler, line 113: Please document the meaning of `True` vs. `False` return values.

src/api-service/__app__/onefuzzlib/azure/subnet.py
```diff
@@ -16,21 +17,41 @@
...
+            if subnet.name == subnet_name:
+                return subnet
+
+    return None
+
+
+def get_subnet_id(resource_group: str, name: str, subnet_name: str) -> Optional[str]:
+    subnet = get_subnet(resource_group, name, subnet_name)
+    if subnet:
+        return cast(str, subnet.id)
```
ranweiler, line 50: Instead of a `cast`, can gate this on a check to `isinstance(subnet.id, str)`? That should convince `mypy` and also actually ensure the type is what we think it is.

src/utils/check-pr/check-pr.py
```diff
@@ -189,6 +189,7 @@ def deploy(self, filename: str) -> None:
         subprocess.check_call(f"python -mvenv {venv}", shell=True)
         pip = venv_path(venv, "pip")
         py = venv_path(venv, "python")
+        config = os.getcwd() + "/config.json"
```
ranweiler, line 192: We should build paths with functions like [`os.path.join()`](https://docs.python.org/3/library/os.path.html#os.path.join), not string concat.

src/deployment/deploylib/tests/test_deploy_config.py
```diff
@@ -0,0 +1,90 @@
...
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import unittest
+from typing import Any, List
+
+from deploylib.configuration import NetworkSecurityConfig
+
+
+class TestNetworkSecurityConfig:
```
ranweiler, line 12: Is this actually used anywhere?

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: I opened bug to address this when I submitted PR into feature branch.

Investigate better handling of Azure CloudError when creating Azure resources #1349

NSG implementation follows same implementation pattern as other Azure OneFuzz components (for exampe: https://github.com/microsoft/onefuzz/blob/c82c84d1a4fa6f5c9faf57a605fa83afa2f7fbf6/src/api-service/__app__/onefuzzlib/azure/vm.py#L115)

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,484 @@
...
+def get_nsg(name: str) -> Optional[NetworkSecurityGroup]:
+    resource_group = get_base_resource_group()
+
+    logging.debug("getting nsg: %s", name)
+    network_client = get_network_client()
+    try:
+        nsg = network_client.network_security_groups.get(resource_group, name)
+        return cast(NetworkSecurityGroup, nsg)
+    except (ResourceNotFoundError, CloudError) as err:
+        logging.debug("nsg %s does not exist: %s", name, err)
```
ranweiler, line 42: Should this be at a higher log level?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,484 @@
...
+    logging.info("updating nsg %s:%s:%s", resource_group, nsg.location, nsg.name)
+    network_client = get_network_client()
+
+    try:
+        network_client.network_security_groups.begin_create_or_update(
+            resource_group, nsg.name, nsg
+        )
+    except (ResourceNotFoundError, CloudError) as err:
+        if is_concurrent_request_error(str(err)):
+            logging.debug(
```
ranweiler, line 94: Again, is this the right log level? How should we log when catching these kinds of errors?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,484 @@
...
+            ],
+        )
+
+    priority = min_priority
+    for src in all_sources:
+        security_rules.append(
+            SecurityRule(
+                name="Allow" + str(priority),
+                protocol="*",
+                source_port_range="*",
```
ranweiler, line 176: Should we be restricting ports, too?

src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -0,0 +1,484 @@
...
+            code=ErrorCode.UNABLE_TO_UPDATE,
+            errors=[
+                "network interface and nsg have to be in the same region.",
+                "nsg %s %s, nic: %s %s"
+                % (nsg.name, nsg.location, nic.name, nic.location),
+            ],
+        )
+
+    if nic.network_security_group and nic.network_security_group.id == nsg.id:
+        return None
```
ranweiler, line 228: How about a log message at some level, like:
```suggestion
        logging.info("NIC %s and NSG %s already associated, not updating", nic.name, name)
        return None
```
This also acts as documentation of what is going on.

src/deployment/deploylib/configuration.py
```diff
@@ -0,0 +1,228 @@
...
+        if not isinstance(proxy_config, Dict):
+            raise Exception(
+                "Inner Configuration is not a Dictionary. Please provide Valid Config."
+            )
+        if len(proxy_config.keys()) == 0:
+            raise Exception(
+                "Empty Inner Configuration File Provided. Please Provide Valid Config."
+            )
+        if (
+            None in proxy_config.keys()
```
ranweiler, line 81: What is the purpose of these `None in ..` checks? cc @nharper285?

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: is it a requirement at this point ? 

If request is coming from trusted IP  - might want to ssh/RDP to a debug VM, and have a debugger listening on a port of their choosing, no ?

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: Yup, fixing this one

src/api-service/__app__/onefuzzlib/azure/nsg.py
ranweiler: Just thinking aloud. WDYT?

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: This error is ignored, since there is a request in progress already. In this specific case "debug" makes sense

src/api-service/__app__/onefuzzlib/azure/nsg.py
ranweiler: There are several other places with this pattern, that I didn't comment on. Might want to check all your new `except` blocks.

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: Existing components log with debug for concurrent requests:
https://github.com/microsoft/onefuzz/blob/c82c84d1a4fa6f5c9faf57a605fa83afa2f7fbf6/src/api-service/__app__/onefuzzlib/azure/vmss.py#L376

https://github.com/microsoft/onefuzz/blob/c82c84d1a4fa6f5c9faf57a605fa83afa2f7fbf6/src/api-service/__app__/onefuzzlib/azure/vm.py#L117

ranweiler: That makes sense.

src/deployment/deploylib/configuration.py
stishkin: I'll fix it

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: my thinking that currently we just want to control IPs only, since IP is trusted and OneFuzz is a "dev tool" trusted people  should be able to do what they think is needed to be done.

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: Done, checked them all. The ones where we are "ignoring" the exceptions, I kept those as debug, since other existing components log as debug in the same situation

src/api-service/__app__/onefuzzlib/azure/nsg.py
ranweiler: I think that since these NSGs apply to ephemeral proxy VMs, and we only support public key auth with short-lived keys, this is a solid starting point, and is a strict improvement on the status quo. We can investigate further restriction later (will open an issue for it).

ranweiler: Captured in: #1436.

ranweiler (approved):

Issue #1328, opened 2021-10-04
User Assignment Required Set By Default


nharper285:
For service principals created by the OneFuzz platform during deployment, we default to setting "User Assignment Required" to `false`. In the deployment script, this is controlled by the parameter `app_role_assignment_required`.

Proposed change: set `app_role_assignment_required` to `true` by default, and add the user who deploys a given instance of OneFuzz as a pre-authorized user. The owner/deployer could then add additional users, as necessary. We could enable this functionality by including a deployment parameter that takes in an authenticated group, as chosen by the deploying user.   

AB#39994145

PR #1425, opened 2021-11-02
SP Guest Account Access locked down by default and deploying user added as admin. 

nharper285:

_What is this about?_
For service principals created by the OneFuzz platform during deployment, we default to setting "User Assignment Required" to `false`. In the deployment script, this is controlled by the parameter `app_role_assignment_required`.

Change: set `app_role_assignment_required` to `true` by default, and add the user who deploys a given instance of OneFuzz as a pre-authorized user. The owner/deployer could then add additional users, as necessary. We could enable this functionality by including a deployment parameter that takes in an authenticated group, as chosen by the deploying user. Theoretically, this could just use the 'admins' parameter at deployment time to add users.  

## PR Checklist
* [x] Applies to work item: #1328 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
deploy.py: `app_role_assignment_required` set to true and includes new call to add_user()
registration.py: add_user function that takes in principal id(s) and add them to the SP. 

## Validation Steps Performed

* [x] User is added by default to SP and `app_role_assignment_required` set to true during deployment.
* [x] Test deploying from service principal
* [x] [Optional] List of admins specified by set_admins is added.


src/deployment/registration.py
```diff
@@ -646,33 +646,34 @@ def set_app_audience(
...
-            "run deploy.py again : \n%s",
-            query,
-        )
-        err_str = "Unable to add user to SP using Microsoft Graph Query API. \n"
-        raise Exception(err_str)
+    for id in principalIds:
+        http_body = {
+            "principalId": id,
+            "resourceId": objectId,
+            "appRoleId": "00000000-0000-0000-0000-000000000000",
```
chkeita, line 655: is this supposed to be an approle of the onefuzz instance ?

src/deployment/registration.py
```diff
@@ -646,33 +646,34 @@ def set_app_audience(
         raise Exception(err_str)
 
 
-def add_user(objectId: str, principalId: str) -> None:
+def add_users(objectId: str, principalIds: List[str]) -> None:
```
chkeita, line 649: i would keep the function as is and call it in a loop instead

src/deployment/registration.py
nharper285: I'll be honest I'm not entirely sure. The information I have present is all I need to lock down the SP and add the user. If i need to specify an appRoleId for best practices, I can look into doing that.

src/deployment/registration.py
nharper285: Can do!

src/deployment/registration.py
chkeita: this is where we define [the roles](https://github.com/microsoft/onefuzz/blob/fb13fb45c1c1d888625fc5f20c9bed53831fd3e0/src/deployment/deploy.py#L295) of the application .
let's add a new role to that list and use it here.
you can look at [assign_app_role](https://github.com/microsoft/onefuzz/blob/fb13fb45c1c1d888625fc5f20c9bed53831fd3e0/src/deployment/deploy.py#L295) for an example of how to look for a specific app role and assign it

src/deployment/deploy.py
```diff
@@ -594,6 +602,47 @@ def assign_scaleset_identity_role(self) -> None:
...
+                    if (
+                        "service principal was not updated with proper appRoleAssignmentRequired value (True)."
+                        not in str(err)
+                    ):
+                        raise err
+                logger.warning("udating service principal failed with an error.")
+
+            # Assign Roles and Add Users
+            roles = [
+                x["id"] for x in app["appRoles"] if x["displayName"] == "UserAssignment"
```
chkeita, line 637:
``` suggestion
                x["id"] for x in app["appRoles"] if x["displayName"] == OnefuzzAppRole.UserAssignment.value
```

src/deployment/deploy.py
```diff
@@ -309,6 +311,14 @@ def setup_rbac(self) -> None:
                 "isEnabled": True,
                 "value": OnefuzzAppRole.ManagedNode.value,
             },
+            {
+                "allowedMemberTypes": ["Application"],
```
chkeita, line 315: this should be user

nharper285: Fixed!

src/deployment/deploy.py
nharper285: Fixed!

src/deployment/deploy.py
```diff
@@ -594,6 +602,47 @@ def assign_scaleset_identity_role(self) -> None:
...
+        app = get_application(
+            display_name=self.application_name,
+            subscription_id=self.get_subscription_id(),
+        )
+        user = get_signed_in_user(self.subscription_id)
+
+        if app:
+            sp = get_service_principal(app["appId"], self.subscription_id)
+            # Update appRoleAssignmentRequired if necessary
+            if not sp["appRoleAssignmentRequired"]:
```
chkeita, line 616: I don't think we should be changing the value of this when we create an assignment. 
The user might have intentionally change that value. 
Instead let's print a warning message indicating that this value is not set and unassigned users can still access the service

nharper285: Agreed and updated.

src/deployment/deploy.py
```diff
@@ -594,6 +602,38 @@ def assign_scaleset_identity_role(self) -> None:
...
+                logger.warning(
+                    "The Service Principal does not have 'appRoleAssignmentRequired' set to True."
+                    + " This means that any authenticated user can access the principal."
```
chkeita, line 619:
```suggestion 
                    + " This means that any authenticated user can access the service."
```

src/deployment/deploy.py
```diff
@@ -594,6 +602,38 @@ def assign_scaleset_identity_role(self) -> None:
...
+            subscription_id=self.get_subscription_id(),
+        )
+        user = get_signed_in_user(self.subscription_id)
+
+        if app:
+            sp = get_service_principal(app["appId"], self.subscription_id)
+            # Update appRoleAssignmentRequired if necessary
+            if not sp["appRoleAssignmentRequired"]:
+                logger.warning(
+                    "The Service Principal does not have 'appRoleAssignmentRequired' set to True."
```
chkeita, line 618: i would simplify the whole warning like this:

> the service is not currently configured to require a role assignment to access it. 
This means that any authenticated user can access the service.
To change this behavior enble `Assignment required?` on the service principal in the azure portal

The problem is `appRoleAssignmentRequired` is the value you set in the msgraph rest api but in the portal it is called a `Assignment required`

nharper285: Done!

Issue #1244, opened 2021-09-13
Writing the same crash file multiple times can result in corrupted file in container


jagunter:
## Information

- Onefuzz version: 2.33.1
- OS: Windows

## Provide detailed reproduction steps (if any)

1. Start job
2. Create and write file in a temp directory
3. Move file to crashes folder to be uploaded to crashes container by OneFuzz
4. Repeat, replacing the file in the crashes folder.

In our example we only replaced it once to trigger the issue, but we expect this was (un)fortunate timing on our part. Repeatedly replacing the file may work better to trigger the behavior.

Examples from OneFuzz instance:

* Job id: `3e0ed3a3-dcfe-441b-89e4-87914b958554`
* Crashes container: `oft-crashes-47f2815a04915f93857618ea4a946f68` . The corrupted file is the one whose file size is 80896 bytes (about 20% smaller than the other files).

### Expected result

Upload to fail. In other cases where we've had issues like this (e.g. the inputs container) we've encountered an `azcopy` error indicating the file had changed between start and end of upload.

### Actual result

The uploaded `.tar` file was corrupted. The archive length was smaller than the actual file length within the archive.

### Impact

For what it's worth, we were not intending to write the crash file multiple times. There was a separate race condition on our side that we have corrected, so we are no longer impacted by this. Reporting in case there are other fuzzers that could trip over this.

stishkin:
Would it make sense for OneFuzz to change permissions on the file when it starts uploading it to prevent it from being overwritten ? 

Or it is better to allow the file be overwritten and OneFuzz would fail the upload ?

jagunter:
Most to least preferable:

1. Succeed upload, retrying in background as necessary (reasonable limits apply). OneFuzz environment is ideally transparent to fuzzer
2. Fail fast and loud, clearly indicating the problem
3. Silently fail while corrupting the destination file

mgreisen:
@stishkin can this be closed now?

stishkin:
@mgreisen - this work is not done. I did not merge my PR. I closed it, since what I implemented was not correct

PR #1427, opened 2021-11-02
Writing the same crash file multiple times can result in corrupted file in container

stishkin:

_What is this about?_

## PR Checklist
* [X] Applies to work item: #1244 

## Info on Pull Request

_What does this include?_
Compare file size and modification date before and after az copy ran. If ether file size changed, or last modification date changed, then re-upload file to storage container.

## Validation Steps Performed

Manual validation

### DIFF
--- a/src/agent/onefuzz/src/az_copy.rs
+++ b/src/agent/onefuzz/src/az_copy.rs
@@ -7,6 +7,7 @@ use backoff::{self, future::retry_notify, ExponentialBackoff};
 use std::{
     ffi::{OsStr, OsString},
     fmt,
+    fs::Metadata,
     path::Path,
     process::Stdio,
     sync::atomic::{AtomicUsize, Ordering},
@@ -130,6 +131,42 @@ fn should_always_retry(err: &anyhow::Error) -> bool {
     false
 }
 
+fn file_change_detected(
+    file: &OsStr,
+    pre_copy: Result<Metadata, std::io::Error>,
+    post_copy: Result<Metadata, std::io::Error>,
+) -> bool {
+    match (pre_copy, post_copy) {
+        (Ok(pre_meta), Ok(post_meta)) => {
+            let pre_len = pre_meta.len();
+            let post_len = post_meta.len();
+            if pre_meta.len() != post_meta.len() {
+                info!(
+                    "File length changed from {} to {} for file {:?}",
+                    pre_len, post_len, file
+                );
+                true
+            } else {
+                match (pre_meta.modified(), post_meta.modified()) {
+                    (Ok(pre_modified), Ok(post_modified)) => {
+                        if pre_modified != post_modified {
+                            info!(
+                                "File modified timestamp changed from {:?} to {:?} for file {:?}",
+                                pre_modified, post_modified, file
+                            );
+                            true
+                        } else {
+                            false
+                        }
+                    }
+                    (_, _) => false,
+                }
+            }
+        }
+        (_, _) => false,
+    }
+}
+
 async fn retry_az_impl(mode: Mode, src: &OsStr, dst: &OsStr, args: &[&str]) -> Result<()> {
     let attempt_counter = AtomicUsize::new(0);
     let failure_counter = AtomicUsize::new(0);
@@ -137,6 +174,7 @@ async fn retry_az_impl(mode: Mode, src: &OsStr, dst: &OsStr, args: &[&str]) -> R
     let operation = || async {
         let attempt_count = attempt_counter.fetch_add(1, Ordering::SeqCst);
         let mut failure_count = failure_counter.load(Ordering::SeqCst);
+        let pre_copy = fs::metadata(src).await;
         let result = az_impl(mode, src, dst, args).await.with_context(|| {
             format!(
                 "azcopy {} attempt {} failed.  (failure {})",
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz/src/az_copy.rs
```diff
@@ -130,13 +131,50 @@ fn should_always_retry(err: &anyhow::Error) -> bool {
...
+                true
+            } else {
+                match (pre_meta.modified(), post_meta.modified()) {
+                    (Ok(pre_modified), Ok(post_modified)) => {
+                        if pre_modified != post_modified {
+                            info!(
+                                "File modified timestamp changed from {:?} to {:?} for file {:?}",
+                                pre_modified, post_modified, file
+                            );
+                            true
```
chkeita, line 157: isn't this enough to check that the file is modified. Why do we need to check the length first ?
it seems to me that the whole function could just be 

```rust
match (pre_meta.modified(), post_meta.modified()) {
    (Ok(pre_modified), Ok(post_modified)) => pre_modified != post_modified, // + logging
    _ => false
}

```

stishkin: It depends how you copy.

Here is my F# test function:


```

let copy( n ) =
    files
    |> Seq.map (fun _ ->
        let i = System.Random().Next(Seq.length files)
        Seq.item i files
    ) |> Seq.map(fun file ->
        let path = @"C:\Users\statis\Documents\repos\onefuzz\az-copy-test\test-files\src.tar"
        System.IO.File.Copy(file.FullName, path, true)
        let fi = System.IO.FileInfo(path)
       
       //need to uncomment this to update last modified time
       //fi.LastWriteTimeUtc <- System.DateTime.UtcNow
       
        printfn "Last modified time: %A" fi.LastWriteTimeUtc

        sprintf "%s last access time: %A and size: %A" file.FullName fi.LastWriteTime ((float fi.Length)/(1024.0*1024.0))
    )

Prints:

Last modified time: 11/1/2021 9:20:10 PM
Last modified time: 11/1/2021 9:58:08 PM
Last modified time: 11/1/2021 9:18:47 PM
Last modified time: 11/1/2021 9:20:24 PM
Last modified time: 11/1/2021 9:18:27 PM
Last modified time: 11/1/2021 9:57:15 PM
Last modified time: 11/1/2021 9:20:47 PM
Last modified time: 11/1/2021 9:20:24 PM
Last modified time: 11/1/2021 9:58:08 PM
Last modified time: 11/1/2021 9:20:00 PM
Last modified time: 11/1/2021 9:18:03 PM
Last modified time: 11/1/2021 9:58:08 PM

```

I created files on 11/1/2021 and I'm running it on 11/2/2021

chkeita: not sure i get the point here. i see different value for last modified how does that show that it is not enough to check the modified value to determine if there was a change ?

stishkin: This is 12 different files being copied over and over a file named src.tar

and I created those files manually on 11/1, after a copy src.tar still has 11/1 timestamp. Even though copy is actually happening on 11/2. 

Which means that checking time stamp itself might not be sufficient.

chkeita: ok but i thought we were comparing the timestamp before and after the copy ?
so what is important here is that LastWriteTimeUtc before the copy is different from LastWriteTimeUtc after the copy ( if we are not copying the exact same file)
i think that what is happening here is that copy is just moving the file with all its metadata.

chkeita: also i think you should test it with rust. I am not sure we can assume that the copy implementation is the same on both language

stishkin: F# script is calling binary produced with Rust compiler. To test - I build Rust az-copy code into a separate executable. And I'm using F# to drive test scenario.

src/agent/onefuzz/src/az_copy.rs
```diff
@@ -130,13 +131,50 @@ fn should_always_retry(err: &anyhow::Error) -> bool {
     false
 }
 
+fn file_change_detected(
+    file: &OsStr,
+    pre_copy: Result<Metadata, std::io::Error>,
```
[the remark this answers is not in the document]
ranweiler: More to the point: these type signatures surprise me. If either of our `metadata()` calls failed, something has gone horribly wrong, right? Why can't this function just assume that we were able to get metadata? In my mind, this shouldn't even be called if we can't get file metadata before or after uploading.

src/agent/onefuzz/src/az_copy.rs
```diff
@@ -130,13 +131,50 @@ fn should_always_retry(err: &anyhow::Error) -> bool {
...
+                            true
+                        } else {
+                            false
+                        }
+                    }
+                    (_, _) => false,
+                }
+            }
+        }
+        (_, _) => false,
```
ranweiler, line 166: 1. In Rust, you can just use `_` to pattern match on the entire tuple, which makes more sense since we're just discarding the match completely, and we aren't disambiguating anything by matching on tuple positions.
2. I think it'd be more idiomatic Rust to do an early return in the one case where we're checking values, and just default to `false` as the non-early return path. You could do an `if let` to just match on that one situation, but as noted above, I think this function should probably just take two `Metadata` values, or maybe `&Metadata`.

src/agent/onefuzz/src/az_copy.rs
```diff
@@ -145,16 +183,24 @@ async fn retry_az_impl(mode: Mode, src: &OsStr, dst: &OsStr, args: &[&str]) -> R
...
-                    failure_count = failure_counter.fetch_add(1, Ordering::SeqCst);
-                }
-                if failure_count >= RETRY_COUNT {
-                    Err(backoff::Error::Permanent(err))
-                } else {
-                    Err(backoff::Error::Transient(err))
+        let post_copy = fs::metadata(src).await;
+
+        if file_change_detected(src, pre_copy, post_copy) {
+            Err(backoff::Error::Transient(anyhow::Error::msg(
```
ranweiler, line 189: Why shouldn't these failures subject to our retry limiting policies (the retry counter + `should_always_retry`)?

stishkin:
After discussion this needs a bit more work. Closing for now

PR #1435, opened 2021-11-04
address PR Comments

stishkin:
Addressing PR comments

### DIFF
--- a/src/deployment/deploylib/configuration.py
+++ b/src/deployment/deploylib/configuration.py
@@ -64,7 +64,7 @@ class NetworkSecurityConfig:
             raise Exception(
                 "Empty Configuration File Provided. Please Provide Valid Config."
             )
-        if "proxy_nsg_config" not in config.keys():
+        if "proxy_nsg_config" not in config:
             raise Exception(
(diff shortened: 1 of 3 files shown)


src/deployment/deploylib/configuration.py
```diff
@@ -78,8 +78,7 @@ def parse_nsg_json(self, config: Any) -> None:
                 "Empty Inner Configuration File Provided. Please Provide Valid Config."
             )
         if (
-            None in proxy_config.keys()
-            or "allowed_ips" not in proxy_config.keys()
+            "allowed_ips" not in proxy_config.keys()
```
nharper285, line 81: This is done again a few lines above on line 67 and needs to be removed, as well.

ranweiler: Also, with Python dicts, `k in d` checks if a value `k` occurs as a _key_ in a dict `d`.

Example:
```python
>>> d = {'k': 'v'}
>>> 'k' in d
True
>>> 'v' in d
False
```
So the calls to `keys()` are unnecessary.

nharper285 (approved):

Issue #1411, opened 2021-10-29
Task fails due to "generate inputs failed"


stishkin:
## Information

- OS: Windows 

Issue happened when running `check-pr.py` script.

``` shell
ERROR:integration:error log: error running task: generate inputs failed

Caused by:
    0: generator failed to run: C:\onefuzz\7c682750-67d1-4150-8244-37e8e9eef84a\task_tools_0\radamsa.exe
    1: process failed: Output { status: ExitStatus(ExitStatus(1)), stdout: "", stderr: "" }
ERROR:integration:error log: process (stderr) generator: Error: cannot write to 'C:\Windows\Temp\.tmpDuRUY4/input-ed900944bf30926cb64c475478c3bcf3aa083f7a830723f0003d123fc166b0b7.txt'

```

![image](https://user-images.githubusercontent.com/7307755/139467356-32f52aff-2074-42af-9c6c-4f42eaa2e47f.png)

ranweiler:
Possibly-related: https://github.com/microsoft/onefuzz/issues/956

PR #1444, opened 2021-11-10
suspend process before killing it

chkeita:
This is a fix for #1411. The error was caused by the handling of the stop message. The task seems to log an error after the being killed by the supervisor. These errors are cause the integration test to fail.
the fix is to suspend the child process before killing it so that no error messages gets forwarded to appinsight

How to test:
run check-pr.py

### DIFF
--- a/src/agent/onefuzz-supervisor/src/worker.rs
+++ b/src/agent/onefuzz-supervisor/src/worker.rs
@@ -254,7 +254,7 @@ impl SuspendableChild for Child {
     }
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "macos"))]
 impl SuspendableChild for Child {
     fn suspend(&mut self) -> Result<()> {
         use nix::sys::signal;
--- a/src/agent/onefuzz-supervisor/src/worker.rs
+++ b/src/agent/onefuzz-supervisor/src/worker.rs
@@ -243,7 +243,6 @@ trait SuspendableChild {
     fn suspend(&mut self) -> Result<()>;
 }
 
-
 #[cfg(target_os = "windows")]
 impl SuspendableChild for Child {
     fn suspend(&mut self) -> Result<()> {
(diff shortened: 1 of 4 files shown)


src/integration-tests/integration-test.py
```diff
@@ -504,6 +504,9 @@ def check_jobs_impl() -> Tuple[bool, str, bool]:
 
             for job_id in to_remove:
                 if stop_on_complete_check:
+                    # send the stop marker here to ignore errors after the jobs stops
```
bmc-msft, line 507: I highly recommend against this change.

There are two issues here.
1. This means errors in job shutdown won't be seen.  In the past, this functionality has caught a number of issues including VM reimaging issues and job dependency issues.
2. This prevents checking logs for issues with live repros, which happen *after* jobs are shut down.

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -239,6 +239,33 @@ impl IWorkerRunner for WorkerRunner {
...
 }
 
+trait SuspendableChild {
+    fn suspend(&mut self) -> Result<()>;
+}
+
+#[cfg(target_os = "windows")]
+impl SuspendableChild for Child {
+    fn suspend(&mut self) -> Result<()> {
+        let result = unsafe { winapi::um::debugapi::DebugActiveProcess(self.id() as u32) };
```
ranweiler, line 249: Interesting. For others, can you please document why we're using this as a workaround?

src/agent/onefuzz-supervisor/src/worker.rs
```diff
@@ -358,6 +385,12 @@ impl IWorkerChild for RedirectedChild {
     fn kill(&mut self) -> Result<()> {
         use std::io::ErrorKind;
 
+        // trying to gracefully kill the child process.
```
ranweiler, line 388:
```suggestion
        // Try to gracefully kill the child process to avoid spurious error telemetry;
```

ranweiler (approved):

PR #1447, opened 2021-11-11
Change logging from `error` to `debug` in get_nsg method

stishkin:
Change logging from `error` to `debug`
1) having error logged fails `integration-test.py`
2) Method returns `None` and caller should decide whether this is an error or not

## Info on Pull Request

_What does this include?_
Update log message method

## Validation Steps Performed

Run `check-pr.py` to validate that `get_nsg` does not trigger `check-pr.py` failure when logs are inspected at the end of the run

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/nsg.py
+++ b/src/api-service/__app__/onefuzzlib/azure/nsg.py
@@ -39,7 +39,7 @@ def get_nsg(name: str) -> Optional[NetworkSecurityGroup]:
         nsg = network_client.network_security_groups.get(resource_group, name)
         return cast(NetworkSecurityGroup, nsg)
     except (ResourceNotFoundError, CloudError) as err:
-        logging.error("nsg %s does not exist: %s", name, err)
+        logging.debug("nsg %s does not exist: %s", name, err)
         return None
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/azure/nsg.py
```diff
@@ -39,7 +39,7 @@ def get_nsg(name: str) -> Optional[NetworkSecurityGroup]:
         nsg = network_client.network_security_groups.get(resource_group, name)
         return cast(NetworkSecurityGroup, nsg)
     except (ResourceNotFoundError, CloudError) as err:
-        logging.error("nsg %s does not exist: %s", name, err)
+        logging.debug("nsg %s does not exist: %s", name, err)
```
chkeita, line 42: do we even need to log this ? it seems like the return type already indicates that we expect that it might not exist

chkeita (approved):

src/api-service/__app__/onefuzzlib/azure/nsg.py
stishkin: just in case. we can remove it later

Issue #1260, opened 2021-09-22
MS Teams Notification - Issue - Incomplete Alert metadata


rthenmozhis:
## Information

- Onefuzz version: onefuzz-2.33.1
- OS: Ubuntu 20.04

## Provide detailed reproduction steps (if any)

1. Created a notification with a incoming webhook for unique crash reports
2. Got successful alerts to the MS team channel 
3. But issue with alert information :  incomplete and blanked out data on the alert

### Expected result

 Call Stack shown with all metadata 

### Actual result

Incomplete Alert ( Please find attached Screeshot)
![image](https://user-images.githubusercontent.com/82480246/134265345-2612d95c-2c7c-4e94-83b3-3ef35aca1ad0.png)

puhley:
I have identified the root cause of this issue. The bug exists in the file:

`https://github.com/microsoft/onefuzz/blob/main/src/api-service/__app__/onefuzzlib/notifications/teams.py`

Within the _code_block_ function, line 29 attempts to return strings representing a Markdown code block:

````
return "\n```%s\n```\n" % data
````

The problem is that there needs to be a newline ("\n") before the "%s". Assuming that "%s" represents the _data_ string "foo", the existing code will generate the following two lines (with the newlines interpreted):

````
```foo
```
````

This is not a valid Markdown code block. The same behavior seen in the Teams screenshot above can be reproduced in GitHub's Markdown.  For instance, I will add those lines below:

```foo
```

In order to be rendered correctly in Markdown, the data needs to be on a separate line from the backticks. Therefore, the function should be returning the following lines (with the newlines interpreted):
````
```
foo
```
```` 

I can create a pull request with the extra newline inserted before the "%s" to separate the backticks from the data.

PR #1449, opened 2021-11-12
Add newline to correct Markdown format

puhley:

This pull request is to address issue #1260. The existing code currently produces invalid Markdown which causes rendering issues in the Teams notifications. A picture of the rendering issue is included within the issue report. This code change will convert the output into a valid Markdown code block which properly renders in Microsoft Teams.

## PR Checklist
* [x] Applies to work item: #1260 
* [ ] Tests added/passed
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #1260

## Info on Pull Request

The _code_block_ function should include a newline after the first three back ticks in order to properly start the code block section. Without the newline, the existing code produces three back ticks followed immediately by the data. In Markdown, a word adjacent to the initial three back ticks is assumed to be a language value for the purposes of syntax highlighting. Therefore, the Markdown parser will ignore the invalid data and produces an empty code block. The proposed change adds a single new line so that the produced output is a valid Markdown code block. 

## Validation Steps Performed

The Markdown spec was used to confirm that data following the first three back ticks is not allowed. This code was tested with Microsoft Teams within our environments.  The flaw can also be reproduced if you attempt to use the original function output within GitHub's Markdown preview feature.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/notifications/teams.py
+++ b/src/api-service/__app__/onefuzzlib/notifications/teams.py
@@ -26,7 +26,7 @@ def markdown_escape(data: str) -> str:
 
 def code_block(data: str) -> str:
     data = data.replace("`", "``")
-    return "\n```%s\n```\n" % data
+    return "\n```\n%s\n```\n" % data
 
 
 def send_teams_webhook(


ranweiler (approved):

ranweiler:
Thanks for the bugfix, @puhley. Manually validating and running integration tests.

ranweiler:
Confirmed fix, integration tests passed modulo known log noise.

Issue #1443, opened 2021-11-08
Incorrect handling of custom images


puhley:
## Information

- Onefuzz version: 3.2.0
- OS: Azure Function App

## Provide detailed reproduction steps (if any)

1. The documentation for using custom OS images within OneFuzz scale sets (https://github.com/microsoft/onefuzz/blob/main/docs/custom-images.md) says that ti accepts Azure Compute Gallery resource IDs, such as: `/subscriptions/MYSUBSCRIPTION/resourceGroups/MYGROUP/providers/Microsoft.Compute/galleries/MYGALLERY/images/MYDEFINITION/versions/MYVERSION`
2. When creating a VM scale set, the _get_os_ function gets called to identify the operating system type of the image within _create_vmss_  (_onefuzzlib/azure/vmss.py_ - line 317). The _get_os_ function of _image.py_ in the _onefuzzlib_ library (https://github.com/microsoft/onefuzz/blob/main/src/api-service/__app__/onefuzzlib/azure/image.py) attempts to fetch the image information using "client.images.get" (line 25). Unfortunately, that Azure SDK method fetches definitions from Azure Images and not Azure Compute Galleries.
3. The correct code for supporting Azure Compute Galleries would be:

`name = client.gallery_images.get(parsed["resource_group"], parsed["name"], parsed["child_name_1"]).os_type.lower()`

4. There is also a second issue in the same section of code. The existing code for retrieving the image from Azure Images tries to access the 'name' property of the 'os_type' which does not exist. This should instead be a call to '.lower()'. Both issues are described in further detail in the Actual Result section below. 

### Expected result

It does not appear to be possible to create a scale set from an Azure Image. Therefore, the code needs to be updated to correctly handle Compute Galleries. Assuming that the Azure Image approach within _get_os_ is necessary for other workflows within OneFuzz,  the os_type reference to the  'name' property should be replaced with a call to '.lower()'.

### Actual result

The code throws a ResourceNotFound error because it is given an Azure Compute Gallery resource ID and it is expecting an Azure Images resource ID. The code needs to be updated to leverage _client.gallery_images.get()_ instead in order to support Azure Compute Galleries (https://docs.microsoft.com/en-us/python/api/azure-mgmt-compute/azure.mgmt.compute.v2021_07_01.operations.galleryimagesoperations?view=azure-python#get-resource-group-name--gallery-name--gallery-image-name----kwargs-). 

[...]

PR #1450, opened 2021-11-13
Add Compute Gallery images support

puhley:

The documentation for using custom images while fuzzing states that Compute Gallery images are supported. However, the existing code for handling custom images does not properly address Compute Gallery resource IDs. The code only handles references to Azure Images. This pull request adds support for Compute Gallery resource IDs as discussed in issue #1443 .  In addition, it corrects an error within the existing code where an attempt to retrieve a "name" parameter is performed on the image "os_type". The "os_type" is a string and does not have a "name" parameter.  This pull request removes the invalid "name" reference and converts the string using ".lower()" similar to the default code path.

## PR Checklist
* [X] Applies to work item: #1443 
* [ ] Tests added/passed
* [X] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #1443 

## Info on Pull Request

This pull request includes a check to determine whether the resource ID refers to a "galleries" resource type. If the resource ID is confirmed to be a Compute Gallery reference, then it will use the corresponding ComputeManagementClient methods and properties to retrieve the OS type. If an Azure Image reference is provided, then the code has been changed to provide the lower case value of the OS type. Since it was unknown what existing dependencies relied on the previous behavior, the code defaults to the existing approach of assuming an Azure Image reference was given.

## Validation Steps Performed

This was tested locally within what is assumed to be a simple deployment environment. More robust testing is welcomed.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/image.py
+++ b/src/api-service/__app__/onefuzzlib/azure/image.py
@@ -19,6 +19,9 @@ from .compute import get_compute_client
 @cached(ttl=60)
 def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
+    # The dict returned here may not have any defined keys.
+    # 
+    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134
     parsed = parse_resource_id(image)
     if "resource_group" in parsed:
         if parsed["type"] == "galleries":
--- a/src/api-service/__app__/onefuzzlib/azure/image.py
+++ b/src/api-service/__app__/onefuzzlib/azure/image.py
@@ -33,6 +33,7 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
                 return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
         else:
             try:
+                # See: https://docs.microsoft.com/en-us/rest/api/compute/images/get
                 name = client.images.get(
                     parsed["resource_group"], parsed["name"]
                 ).storage_profile.os_disk.os_type.lower()
--- a/src/api-service/__app__/onefuzzlib/azure/image.py
+++ b/src/api-service/__app__/onefuzzlib/azure/image.py
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
     parsed = parse_resource_id(image)
     if "resource_group" in parsed:
-        try:
-            name = client.images.get(
-                parsed["resource_group"], parsed["name"]
-            ).storage_profile.os_disk.os_type.name
-        except (ResourceNotFoundError, CloudError) as err:
-            return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        if parsed["type"] == "galleries":
+            try:
+                name = client.gallery_images.get(
+                    parsed["resource_group"], parsed["name"], parsed["child_name_1"]
+                ).os_type.lower()
+            except (ResourceNotFoundError, CloudError) as err:
+                return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        else:
+            try:
+                name = client.images.get(
+                    parsed["resource_group"], parsed["name"]
+                ).storage_profile.os_disk.os_type.lower()
+            except (ResourceNotFoundError, CloudError) as err:
+                return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
     else:
         publisher, offer, sku, version = image.split(":")
         try:
--- a/src/api-service/__app__/onefuzzlib/azure/image.py
+++ b/src/api-service/__app__/onefuzzlib/azure/image.py
@@ -20,13 +20,13 @@ from .compute import get_compute_client
 def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
     # The dict returned here may not have any defined keys.
-    # 
-    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134
+    #
+    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501 
     parsed = parse_resource_id(image)
     if "resource_group" in parsed:
         if parsed["type"] == "galleries":
             try:
-                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage
+                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage  # noqa: E501 
                 name = client.gallery_images.get(
                     parsed["resource_group"], parsed["name"], parsed["child_name_1"]
                 ).os_type.lower()
--- a/src/api-service/__app__/onefuzzlib/azure/image.py
+++ b/src/api-service/__app__/onefuzzlib/azure/image.py
@@ -21,12 +21,12 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
     # The dict returned here may not have any defined keys.
     #
-    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501 
+    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501
     parsed = parse_resource_id(image)
     if "resource_group" in parsed:
         if parsed["type"] == "galleries":
             try:
-                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage  # noqa: E501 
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
     parsed = parse_resource_id(image)
```
ranweiler, line 22: Not a problem with your PR, but let's document the caller's burden / lack of guarantees:
```suggestion
    # The dict returned here may not have any defined keys.
    # 
    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134
    parsed = parse_resource_id(image)
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
...
     if "resource_group" in parsed:
-        try:
-            name = client.images.get(
-                parsed["resource_group"], parsed["name"]
-            ).storage_profile.os_disk.os_type.name
-        except (ResourceNotFoundError, CloudError) as err:
-            return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        if parsed["type"] == "galleries":
```
ranweiler, line 24: Thinking through this: I think we've been assuming that if the `"resource_group"` key is present, we'll at least have a _key_ in `parsed` for every return key described in the `parse_resource_id()` docs. Unfortunately, that's not valid, and per the very weak contract of our parameter type, we can actually get an _empty_ dict.

Example:
```python
>>> rid_provider = '/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/my-resource-group/providers/Microsoft.Web/sites/my-site-name'
>>> rid_no_provider = '/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/my-resource-group'
>>> parse_resource_id(rid_provider)
{'children': '',
 'name': 'my-site-name',
 'namespace': 'Microsoft.Web',
 'resource_group': 'my-resource-group',
 'resource_name': 'my-site-name',
 'resource_namespace': 'Microsoft.Web',
 'resource_parent': '',
 'resource_type': 'sites',
 'subscription': '11111111-1111-1111-1111-111111111111',
 'type': 'sites'}
>>> parse_resource_id(rid_no_provider)
{'resource_group': 'my-resource-group',
 'subscription': '11111111-1111-1111-1111-111111111111'}
```

This is not a great situation, and we should probably add types (in a `Union`) to explicitly represent the different expected cases, and allow clearer dispatching at this site.

ranweiler: (this can be done by us in a follow-up PR)

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
...
+            try:
+                name = client.gallery_images.get(
```
ranweiler, line 26:
```suggestion
                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage
                name = client.gallery_images.get(
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
...
+                    parsed["resource_group"], parsed["name"], parsed["child_name_1"]
+                ).os_type.lower()
+            except (ResourceNotFoundError, CloudError) as err:
+                return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        else:
+            try:
+                name = client.images.get(
```
ranweiler, line 33:
```suggestion
                # See: https://docs.microsoft.com/en-us/rest/api/compute/images/get
                name = client.images.get(
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
...
+                    parsed["resource_group"], parsed["name"]
+                ).storage_profile.os_disk.os_type.lower()
```
ranweiler, line 35: 👍

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -21,12 +21,20 @@ def get_os(region: Region, image: str) -> Union[Error, OS]:
...
+            except (ResourceNotFoundError, CloudError) as err:
```
ranweiler, line 36: I get the motivation, but I'm not sure it really makes sense to scope these `except` blocks to only these Azure client calls. Not an issue for this PR, though.

ranweiler (commented):
Thanks for this, @puhley. I have some small suggested changes, and a few comments, but the latter are more for us to follow up with.

src/api-service/__app__/onefuzzlib/azure/image.py
ranweiler: Captured in #1464.

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -19,14 +19,27 @@
 @cached(ttl=60)
 def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
+    # The dict returned here may not have any defined keys.
+    #
```
ranweiler, line 23:
```suggestion
    #
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -19,14 +19,27 @@
 @cached(ttl=60)
 def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
+    # The dict returned here may not have any defined keys.
+    # 
+    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134
```
ranweiler, line 24:
```suggestion
    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501 
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -19,14 +19,27 @@
...
     if "resource_group" in parsed:
-        try:
-            name = client.images.get(
-                parsed["resource_group"], parsed["name"]
-            ).storage_profile.os_disk.os_type.name
-        except (ResourceNotFoundError, CloudError) as err:
-            return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        if parsed["type"] == "galleries":
+            try:
+                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage
```
ranweiler, line 29:
```suggestion
                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage  # noqa: E501 
```

ranweiler (approved):

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -19,14 +19,27 @@
...
     if "resource_group" in parsed:
-        try:
-            name = client.images.get(
-                parsed["resource_group"], parsed["name"]
-            ).storage_profile.os_disk.os_type.name
-        except (ResourceNotFoundError, CloudError) as err:
-            return Error(code=ErrorCode.INVALID_IMAGE, errors=[str(err)])
+        if parsed["type"] == "galleries":
+            try:
+                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage  # noqa: E501
```
ranweiler, line 29:
```suggestion
                # See: https://docs.microsoft.com/en-us/rest/api/compute/gallery-images/get#galleryimage  # noqa: E501
```

src/api-service/__app__/onefuzzlib/azure/image.py
```diff
@@ -19,14 +19,27 @@
 @cached(ttl=60)
 def get_os(region: Region, image: str) -> Union[Error, OS]:
     client = get_compute_client()
+    # The dict returned here may not have any defined keys.
+    #
+    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501
```
ranweiler, line 24:
```suggestion
    # See: https://github.com/Azure/msrestazure-for-python/blob/v0.6.3/msrestazure/tools.py#L134  # noqa: E501
```

ranweiler:
Running integration tests.

Porges:
@puhley out of interest, what type (distro/version) of Linux images are you deploying? We might need to make our agent binary "less portable" in order to comply with some security policies, so I'm trying to gather information on what distros are currently in use.

PR #1497, opened 2021-11-30
[Bug Fix] Adding Check for SP During User Role Assignment.

nharper285:

_What is this about?_
I introduced a bug when I implemented the new function for assigning user access. If a user tried upgrading an instance with an SP, then deployment would fail when this function was called due to lack of permissions.
 
## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Changes to deploy.py assign_user_access()

## Validation Steps Performed
check-pr run. 

Deploy with a service principal.

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -615,7 +615,7 @@ class Client:
     def assign_user_access(self) -> None:
         if self.upgrade:
             logger.info(
-                "Deploying w/ Service Principal. Skipping assignment of the user role."
+                "Upgrading: Skipping assignment of current user to app role"
             )
             return
         logger.info("assinging user access to service principal")
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -45,6 +45,8 @@ from azure.storage.blob import (
     ContainerSasPermissions,
     generate_container_sas,
 )
+from msrest.serialization import TZ_UTC
+
 from deploylib.configuration import (
     InstanceConfigClient,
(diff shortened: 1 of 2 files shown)


mgreisen (approved):

src/deployment/deploy.py
```diff
@@ -613,6 +613,11 @@ def assign_scaleset_identity_role(self) -> None:
         )
 
     def assign_user_access(self) -> None:
+        if self.results["client_id"]:
```
chkeita, line 616: I don't understand this condition
why skipping when the client_id is set ? Shouldn't we update that client_id instead ?

nharper285: image
This was the offending code. You can't call this function as an SP.

nharper285: Okay @chkeita, I actually just changed it to self.upgrade. The original logic did not work properly because client_id and client_secret are set regardless.

src/deployment/deploy.py
```diff
@@ -613,6 +613,11 @@ def assign_scaleset_identity_role(self) -> None:
         )
 
     def assign_user_access(self) -> None:
+        if self.upgrade:
+            logger.info(
+                "Deploying w/ Service Principal. Skipping assignment of the user role."
```
chkeita, line 618:
```suggestion
                "Upgrading: Skipping assignment of current user to app role"
```

chkeita (approved):

Issue #1454, opened 2021-11-16
Spurious error telemetry when stopping Windows generator tasks


ranweiler:
## Information

- Onefuzz version: 3.2.0 (latest)
- OS: Windows

## Provide detailed reproduction steps (if any)

1. Create a Windows generator task, such as via the `radamsa basic` template.
2. Wait until the job's `generic_generator` fuzz task starts (is in `running`)
2. Stop the job with `onefuzz template stop`.

### Expected result

The job is shut down, without any error telemetry.

### Actual result

We send error-level telemetry, such as:
```
error running task: generate inputs failed

Caused by:
    0: generator failed to start: C:\onefuzz\ee26eeba-21b1-4da7-804c-233bd831d25b\task_tools_0\radamsa.exe
    1: The media is write protected. (os error 19)
```
This causes spurious integration test failures.

ranweiler:
Using the above steps, I've been able to repro this several times in a row.

ranweiler:
Other errors elicited by the same repro procedure (all from a single job):

#### `generic_analysis`
```
error running task: poller failed

Caused by:
    0: QueueClient.pop
    1: storage queue pop failed
    2: request attempt 6 failed
    3: HTTP status client error (404 Not Found) for url (<REDACTED>)
```

#### `generic_crash_report`
```
error running task: poller failed

Caused by:
    0: QueueClient.pop
    1: storage queue pop failed
    2: request attempt 6 failed
    3: HTTP status client error (404 Not Found) for url (<REDACTED>)
```

#### `generic_generator`
```
error running task: OS Error 19 (FormatMessageW() returned error 19) (os error 19) at path "C:\\Windows\\TEMP\\.tmpr3WyJK"
```

ranweiler:
The `template stop` CLI command is used internally in the integration tests. It is invoked indirectly [here](https://github.com/microsoft/onefuzz/blob/bd5aa7911265378e0c014c70bde17419bcf3de60/src/integration-tests/integration-test.py#L506-L507), when we stop test jobs that have had their checks completed. This is set to `True` in the call to `Run.test()`, [here](https://github.com/microsoft/onefuzz/blob/bd5aa7911265378e0c014c70bde17419bcf3de60/src/integration-tests/integration-test.py#L918). The inner call the `stop` template command is [here](https://github.com/microsoft/onefuzz/blob/bd5aa7911265378e0c014c70bde17419bcf3de60/src/integration-tests/integration-test.py#L664).

[...]

PR #1505, opened 2021-12-01
Integration tests reliability fixes

chkeita:

Multiple fixes for #1454
- Check if the node is in the don state before reimaging. This gives the node the opportunity to process the `StopCommand` before we reimage it 
- Log node done event as warning when the node is deleted from the database
- Log heartbeat from deleted nodes as warning

## Validation Steps Performed

run check-pr


stishkin (approved):

ranweiler (commented):
Because of way the existing code is written, I find it hard to see how the change to the default behavior of `stop_if_complete()` interact with various other flows in the system.

Generally: what other system behaviors are determined by that function, and how did we test that certain error cases are still handled correctly?

Specifically: I have a vague concern that there may be error scenarios where we really want to reimage an unresponsive node, but we've now made ourselves rely on its cooperation, and so reimaging will never happen. I don't have an example, but I'd like to think through and test some cases like that.

chkeita:
@ranweiler 

The change i introduced make it so that we don't force the state to `done` when stopping a task. We rely on the machine to report the state for that.
for the case of unresponsive node, the state is forced to [`done`](https://github.com/microsoft/onefuzz/blob/e4e4671038fc5909126fa5b449142b53bf261443/src/api-service/__app__/onefuzzlib/workers/scalesets.py#L410) [before](https://github.com/microsoft/onefuzz/blob/e4e4671038fc5909126fa5b449142b53bf261443/src/api-service/__app__/onefuzzlib/workers/nodes.py#L384) reimaging so my change will not prevent the reimage

ranweiler (approved):

PR #1523, opened 2021-12-14
Bump `z3-sys` to 0.7.1

ranweiler:
- Update to latest release of `z3-sys`
- Install Z3 headers on build machines, as is now required

The `z3-sys` crate used to vendor pre-generated modules from Z3 headers. Now, the Z3 header to use can be chosen by library users, but it almost must be available at compile-time, as of https://github.com/prove-rs/z3.rs/commit/d0fc517d8eaf7242bb3e89e413b4c97d64d346db.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Linux Prereqs
       run: |
         sudo apt-get -y update
-        sudo apt-get -y install libssl1.0-dev libunwind-dev
+        sudo apt-get -y install libssl1.0-dev libunwind-dev libz3-dev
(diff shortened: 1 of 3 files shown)


mgreisen (approved):

stishkin (approved):

ranweiler:
The missing header-generated items are real. Ubuntu 18.04 ships Z3 4.4.1, which was released in October 2015. This version doesn't have, for example, `Z3_SEQ_SORT`. This FFI enum mismatch was a real lurking bug in `z3-sys` (which they now surface) by requiring the linked-against header.

raghavthind2005 (approved):

PR #1531, opened 2021-12-16
pin mypy to 0.910

stishkin:
With latest release of mypy it broke the build.
See following github items for details:

https://github.com/samuelcolvin/pydantic/issues/3528
https://github.com/python/mypy/issues/6617#issuecomment-892438903
https://github.com/samuelcolvin/pydantic/pull/3175#issuecomment-914897604


mgreisen:
If this works, be sure to create an item to remove it once the actual fix is in.

mgreisen (approved):

Issue #1380, opened 2021-10-20
Integrate source-based coverage reporting for Linux


ranweiler:
After #1379 is implemented, add it to the generic `coverage` task, using the output format defined by the work in #1378.

ranweiler:
Closed by:
- #1535
- #1572

Issue #1378, opened 2021-10-20
Integrate source-based coverage reporting for Windows


ranweiler:
Our existing `srcview` crate enables emitting source coverage from PE/PDB targets. Integrate this into the `coverage` task.

PR #1533, opened 2021-12-16
Added function for converting coverage source format to Cobertura xml

hayleycall:

Added function to take in SourceCoverage struct from coverage source format and convert to a format that aligns with Cobertura guidelines as xml output. 

## PR Checklist
* [x] Applies to work item: Related to these #1378, #1380.
* [X] Tests added/passed

## Info on Pull Request
Adds function for converting source format, specifically a SourceCoverage struct to Cobertura XML.

## Validation Steps Performed

Test added to manually create XML format vs. reading in a SourceCoverage struct and creating the XML format from that.

### DIFF
--- a/src/agent/coverage/src/cobertura.rs
+++ b/src/agent/coverage/src/cobertura.rs
@@ -25,9 +25,19 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
             .attr("lines-covered", "0")
             .attr("lines-valid", "0")
             .attr("branches-covered", "0")
+            .attr("branches-valid", "0")
             .attr("complexity", "0")
             .attr("version", "0.1")
             .attr("timestamp", &format!("{}", unixtime)),
+    )?;
+
+    emitter.write(XmlEvent::start_element("packages"))?;
+    emitter.write(
+        XmlEvent::start_element("package")
+            .attr("name", "0")
+            .attr("line-rate", "0")
+            .attr("branch-rate", "0")
+            .attr("complexity", "0"),
     )?;
 
     emitter.write(XmlEvent::start_element("classes"))?;
@@ -60,6 +70,8 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
     }
 
     emitter.write(XmlEvent::end_element())?; // classes
+    emitter.write(XmlEvent::end_element())?; // package
+    emitter.write(XmlEvent::end_element())?; // packages
     emitter.write(XmlEvent::end_element())?; // coverage
 
     Ok(String::from_utf8(backing)?)
@@ -123,9 +135,19 @@ mod tests {
                 .attr("lines-covered", "0")
                 .attr("lines-valid", "0")
                 .attr("branches-covered", "0")
+                .attr("branches-valid", "0")
                 .attr("complexity", "0")
                 .attr("version", "0.1")
                 .attr("timestamp", &format!("{}", unixtime)),
+        )?;
+
+        _emitter_test.write(XmlEvent::start_element("packages"))?;
+        _emitter_test.write(
+            XmlEvent::start_element("package")
+                .attr("name", "0")
+                .attr("line-rate", "0")
+                .attr("branch-rate", "0")
+                .attr("complexity", "0"),
         )?;
 
         _emitter_test.write(XmlEvent::start_element("classes"))?;
@@ -182,6 +204,8 @@ mod tests {
         _emitter_test.write(XmlEvent::end_element())?; // lines
         _emitter_test.write(XmlEvent::end_element())?; // class
         _emitter_test.write(XmlEvent::end_element())?; // classes
+        _emitter_test.write(XmlEvent::end_element())?; // package
+        _emitter_test.write(XmlEvent::end_element())?; // packages
         _emitter_test.write(XmlEvent::end_element())?; // coverage
 
         assert_eq!(source_coverage_result?, String::from_utf8(backing_test)?);
--- a/src/agent/coverage/src/cobertura.rs
+++ b/src/agent/coverage/src/cobertura.rs
@@ -90,9 +90,6 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
         }
         emitter.write(XmlEvent::end_element())?; // lines
 
-        // emitter.write(XmlEvent::end_element())?; // method
-        // emitter.write(XmlEvent::end_element())?; // methods
-
         emitter.write(XmlEvent::end_element())?; // class
     }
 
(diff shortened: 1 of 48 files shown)


src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,245 @@
...
+        _emitter_test.write(XmlEvent::end_element())?; // classes
+        _emitter_test.write(XmlEvent::end_element())?; // package
+        _emitter_test.write(XmlEvent::end_element())?; // packages
+        _emitter_test.write(XmlEvent::end_element())?; // coverage
+
+        // let check = match source_coverage_result {
+        //     Ok(source_coverage_result) => source_coverage_result,
+        //     Err(err) => return Err(err),
+        // };
+        assert_eq!(source_coverage_result.unwrap(), String::from_utf8(backing_test)?);
```
hayleycall, line 241: Unsure if using .unwrap() is bad code. Checking an error some other way (similar to what I have commented out above may be better).

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,245 @@
...
+                .attr("filename", &file.file)
+                .attr("lines-valid", "0")
+                .attr("lines-covered", "0")
+                .attr("line-rate", "0")
+                .attr("branches-valid", "0")
+                .attr("branches-covered", "0")
+                .attr("branch-rate", "0")
+                .attr("complexity", "0"),
+        )?;
+        // emitter.write(XmlEvent::start_element("methods"))?;
```
hayleycall, line 67: I removed methods from the XML output for now because it included things source format doesn't currently have such as signature.

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,245 @@
...
+        .create_writer(&mut backing);
+
+    let unixtime = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .context("system time before unix epoch")?
+        .as_secs();
+
+    emitter.write(
+        XmlEvent::start_element("coverage")
+            .attr("lines-valid", "0")
```
hayleycall, line 23: For now I kept it as simple as possible with only file, line, and counts- the exact members the SourceCoverage struct has, other things have been hard coded as 0.

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,255 @@
...
+        _emitter_test.write(XmlEvent::end_element())?; // package
+        _emitter_test.write(XmlEvent::end_element())?; // packages
+        _emitter_test.write(XmlEvent::end_element())?; // coverage
+
+        // let check = match source_coverage_result {
+        //     Ok(source_coverage_result) => source_coverage_result,
+        //     Err(err) => return Err(err),
+        // };
+        assert_eq!(
+            source_coverage_result.unwrap(),
```
ranweiler, line 249: In a test, calling `unwrap()` is okay. This is because failed `assert`'s are just "helpful panics", and a panic is the standard mechanism of test function failure (other tests will still run). Howeer, we should prefer `expect()` (which also panics, but lets add a helpful error string).

In this situation, your function under test returns a `Result`, so you can actually use `?` (like you do with `String::from_utf8()`).

If you had an `Option`, you couldn't just `?` in this test function because it returns a `Result`, so an `expect()` would be fine.

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,255 @@
...
+        _emitter_test.write(XmlEvent::end_element())?; // line
+
+        _emitter_test.write(XmlEvent::end_element())?; // lines
+        _emitter_test.write(XmlEvent::end_element())?; // class
+        _emitter_test.write(XmlEvent::end_element())?; // classes
+        _emitter_test.write(XmlEvent::end_element())?; // package
+        _emitter_test.write(XmlEvent::end_element())?; // packages
+        _emitter_test.write(XmlEvent::end_element())?; // coverage
+
+        // let check = match source_coverage_result {
```
ranweiler, line 244: Go ahead and get rid of any stray commented code. You'll also need to run `cargo fmt` on everything, if you haven't already.

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,255 @@
...
+            .attr("branches-valid", "0")
+            .attr("branches-covered", "0")
+            .attr("branch-rate", "0")
+            .attr("complexity", "0"),
+    )?;
+    emitter.write(XmlEvent::start_element("classes"))?;
+    // loop through files
+    let files: Vec<SourceFileCoverage> = source_coverage.files;
+    for file in files {
+        emitter.write(
```
ranweiler, line 55: Should we omit `class`, too? Are these required sections?

src/agent/coverage/src/cobertura.rs
ranweiler: Makes sense. Before merging, we should explore what happens if we _omit_ the zeroed fields, when it comes to HTML report rendering. To start, we're most interested in the output as rendered in [ADO](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-code-coverage-results?view=azure-devops).

src/agent/coverage/src/cobertura.rs
```diff
@@ -0,0 +1,255 @@
...
+            .attr("complexity", "0")
+            .attr("version", "0.1"),
+    )?;
+    emitter.write(XmlEvent::start_element("sources"))?;
+    emitter.write(XmlEvent::start_element("source"))?;
+    emitter.write(XmlEvent::characters(""))?;
+    emitter.write(XmlEvent::end_element())?; // source
+    emitter.write(XmlEvent::end_element())?; // sources
+
+    emitter.write(XmlEvent::start_element("packages"))?;
```
ranweiler, line 39: Again: required? Let's figure that out.

src/agent/coverage/src/cobertura.rs
ranweiler: Based on [this schema](https://github.com/danielpalme/ReportGenerator), I think zero (or more) elements are valid for these fields we don't have real data for.

What matters most, though, is how [ReportGenerator](https://github.com/danielpalme/ReportGenerator) renders the XML as HTML. This is because it is what [powers the ADO task](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/PublishCodeCoverageResultsV2/ThirdPartyNotices.txt#L7).

Let's investigate what happens in ADO.

src/agent/coverage/src/cobertura.rs
ranweiler: Based on [this schema](https://github.com/danielpalme/ReportGenerator), I think zero (or more) elements are valid for these fields we don't have real data for.

What matters most, though, is how [ReportGenerator](https://github.com/danielpalme/ReportGenerator) renders the XML as HTML. This is because it is what [powers the ADO task](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/PublishCodeCoverageResultsV2/ThirdPartyNotices.txt#L7).

Let's investigate what happens in ADO.

src/agent/coverage/src/cobertura.rs
hayleycall: Class and method both required.

src/agent/coverage/src/cobertura.rs
hayleycall: Source and package elements not required.

src/agent/coverage/src/cobertura.rs
hayleycall: *I was incorrect, only the attributes are required

stishkin (approved):

chkeita (approved):

mgreisen (approved):

mgreisen (approved):

chkeita (approved):

PR #1535, opened 2021-12-17
Emit source-line coverage reports in `coverage` task

### LINKED ISSUE #1378: Integrate source-based coverage reporting for Windows

ranweiler:
In the coverage task, write a new `source-coverage.json` file to the `coverage` container, next to the existing `coverage.json`.

Consumers that subscribe to updates of this blob storage container will now need to make sure they check which blob was updated.

Closes #1378, #1380.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -12,6 +12,7 @@ use async_trait::async_trait;
 use coverage::block::CommandBlockCov;
 use coverage::cache::ModuleCache;
 use coverage::code::{CmdFilter, CmdFilterDef};
+use coverage::debuginfo::DebugInfo;
 use onefuzz::expand::{Expand, PlaceHolder};
 use onefuzz::syncdir::SyncedDir;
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -339,11 +344,24 @@ impl<'a> TaskContext<'a> {
...
             .await
             .with_context(|| format!("writing coverage to {}", path.display()))?;
+
+        let path = self.config.coverage.local_path.join(SOURCE_COVERAGE_FILE);
+        let src_coverage = {
+            let mut debuginfo = self
+                .debuginfo
+                .lock()
+                .map_err(|e| anyhow::format_err!("{}", e))?;
+            self.coverage.source_coverage(&mut *debuginfo)?
```
ranweiler, line 358: This method potentially does a lot of blocking IO. We should call it via `spawn_blocking()`.

hayleycall (approved):

PR #1538, opened 2021-12-17
Use symbol path when searching for PDBs for source coverage

ranweiler:
On Windows, when we search for PDB debuginfo for PEs, use our existing `dbghelp`-based mechanism, which uses the symbol search path.

Also, if we don't find out, try to use the PE, instead of failing immediately. If it lacks debuginfo (it usually will), we'll now return `None` instead of failing, which is our intended semantics.

Tested locally using the `src-cov` example against a binary. I tested 4 cases:
1. No PDB available at all
2. PE-described absolute-path PDB in original compilation directory, no other copies
3. PDB in same dir as PDB (only)
4. PDB in non-PE dir, included in `_NT_SYMBOL_PATH`

As expected, the PDB is not found in (1), and is found in (2) - (4). Not-committed debug logging verified that the resolved PDB was the expected one in each case. When found, source coverage is correctly-generated.

### DIFF
--- a/src/agent/coverage/src/debuginfo.rs
+++ b/src/agent/coverage/src/debuginfo.rs
@@ -7,11 +7,13 @@ use std::io;
 use std::path::{Path, PathBuf};
 
 use anyhow::Result;
-use goblin::pe::PE;
 use symbolic::{
     debuginfo::Object,
     symcache::{SymCache, SymCacheWriter},
 };
+
+#[cfg(windows)]
+use goblin::pe::PE;
 
 #[cfg(windows)]
 use symbolic::debuginfo::pe;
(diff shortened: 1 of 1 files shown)


stishkin (approved):

stishkin (approved):

src/agent/coverage/src/debuginfo.rs
```diff
@@ -76,18 +82,25 @@ impl ModuleDebugInfo {
...
+        {
+            // If our module is a PE file, the debug info will be in the PDB.
             //
-            // TODO: Find PDB using `dbghelp::`SymGetSymbolFile()`.
-            let pdb = module.with_extension("pdb");
-            data = fs::read(&pdb)?.into_boxed_slice();
+            // We will need a similar check to support split DWARF.
+            let is_pe = pe::PeObject::test(&data);
+            if is_pe {
+                let pe = PE::parse(&data)?;
```
ranweiler, line 97: This is wasteful, but unavoidable, since we need to get at the debug info header in the PE.

We use the `goblin::pe::PE` because the `PeObject` from `symbolic` does not expose enough codeview info. I opened an issue about that upstream.

ranweiler: Also: since we're caching loaded debuginfo, we only ever do this once per PE module.

PR #1539, opened 2021-12-17
Avoid leaks

ranweiler:
Before, we leaked unused file and symcache data if object or symcache parsing failed, or if the object parsed, but lacked debuginfo. This is because we leaked the box, but then used the `Try` operator (`?`) to early return on those errors.

Now, we save off raw pointers while we have access to a mutable reference to the leaked data. If we return early due to a failure, we also reconstruct the `Box` from the raw pointer, to leverage its `Drop` impl and free the unused data (after ensuring it doesn't even have any remaining local references).


src/agent/coverage/src/debuginfo.rs
```diff
@@ -103,20 +103,72 @@ impl ModuleDebugInfo {
...
-        let object = Object::parse(data)?;
+        // Save a raw pointer to the file data. If object parsing fails, or
+        // there is no debuginfo, we will use this to avoid leaking memory.
+        let data_ptr = data as *mut _;
+
+        let object = match Object::parse(data) {
+            Ok(object) => {
+                if !object.has_debug_info() {
+                    // Drop the object to drop its static references to the leaked data.
+                    drop(object);
```
ranweiler, line 118: If we fail to do this before rehydrating/dropping the `Box`, it won't necessarily cause observable memory corruption on drop, but _is_ undefined behavior, since the internal references will be dangling.

PR #1541, opened 2021-12-18
Adding AutoConfig Properties to AzSecPack Extensions

nharper285:

_What is this about?_
We are required to enable autoconfig for AzSecPack deployments. These changes included the addition of the relevant property. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Changes to the AzSecPack extensions in extension.py

## Validation Steps Performed

check-pr.py
manual deployments and tests.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/extension.py
+++ b/src/api-service/__app__/onefuzzlib/extension.py
@@ -150,7 +150,7 @@ def azmon_extension(
         "type": "AzureMonitorLinuxAgent",
         "typeHandlerVersion": "1.9",
         "autoUpgradeMinorVersion": False,
-        "settings": {},
+        "settings": {"GCS_AUTO_CONFIG": True},
         "protectedsettings": {
             "configVersion": config_version,
             "moniker": moniker,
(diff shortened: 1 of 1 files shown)


src/api-service/__app__/onefuzzlib/extension.py
```diff
@@ -150,7 +150,7 @@ def azmon_extension(
         "type": "AzureMonitorLinuxAgent",
         "typeHandlerVersion": "1.9",
         "autoUpgradeMinorVersion": False,
```
mgreisen, line 152: In this extension we have `autoUpgradeMinorVersion` set to False, but on line 175, we set that extension autUpgradeMinorVersion to True. Do we think that AzureMonitorLinuxAgent could reboot on upgrade and AzureSecurityLinuxAgent wouldn't?

nharper285: AzureMonitorLinuxAgent was set to False because a previous upgrade caused issues based off agent naming conventions. The current version is a valid and supported version. 

AzureSecurityLinuxAgent doesn't have this issue. 

If the agent restarted there would be zero problems. Both agents support restarts and can re-establish their connections with the paired agent. I do it all the time manually on the nodes.

mgreisen (approved):

PR #1560, opened 2022-01-04
Adding new InstanceConfig value for VMSS & VM tags

nharper285:

_What is this about?_
For our prod instance to run the most up-to-date version of AzSecPack, we need to be able to add tags to the VMSS and VMs running in an instance. I've added two new instance_config fields that allow instance operators to specify vmss and vm tags. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Changes to: 
- vm.py
- vmss.py
- models.py

## Validation Steps Performed

Running check-pr and manual tests that include new instance_config tags.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/vmss.py
+++ b/src/api-service/__app__/onefuzzlib/azure/vmss.py
@@ -25,7 +25,6 @@ from onefuzztypes.enums import OS, ErrorCode
 from onefuzztypes.models import Error
 from onefuzztypes.primitives import Region
 
-from ..config import InstanceConfig
 from .compute import get_compute_client
 from .creds import (
     get_base_resource_group,
@@ -247,8 +246,6 @@ def create_vmss(
     tags: Dict[str, str],
 ) -> Optional[Error]:
 
-    instance_config = InstanceConfig.fetch()
-
     vmss = get_vmss(name)
     if vmss is not None:
         return None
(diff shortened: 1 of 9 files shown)


stishkin (approved):

src/api-service/__app__/onefuzzlib/azure/vmss.py
```diff
@@ -246,6 +247,8 @@ def create_vmss(
     tags: Dict[str, str],
 ) -> Optional[Error]:
 
+    instance_config = InstanceConfig.fetch()
```
chkeita, line 250: I would suggest passing the additional tags in the tag parameter instead of fetching the instance config here. 
This will put the tag generation in one place making it easier to understand what tags are being added to the VM

nharper285: I'm not sure I know what you mean? Lower down in the file, I include both the parameter tags and the instance_config tags.

chkeita (approved):

src/api-service/__app__/onefuzzlib/repro.py
```diff
@@ -71,10 +75,12 @@ def get_vm(self) -> VM:
             sku=vm_config.sku,
             image=vm_config.image,
             auth=self.auth,
+            tags=tags,
         )
 
     def init(self) -> None:
-        vm = self.get_vm()
+        config = InstanceConfig.fetch()
```
bmc-msft, line 82: Since you only use the config within self.get_vm(), why not make the call there?

bmc-msft: oh, nevermind, I see it later on.

PR #1572, opened 2022-01-10
Integrate Cobertura format and output as file in coverage container

hayleycall:

Outputs coverage file as Cobertura XML format.

## PR Checklist
* [X] Applies to work item: Related to these #1378, #1380.
* [X] Tests added/passed: test in previous #1533 PR.

## Info on Pull Request

Continues work of outputting improved source coverage as file(s) alongside block coverage file. This work completes the integration of having a third file be outputted to the coverage container, this file being in Cobertura xml format. It calls the function that converts source coverage data to Cobertura format and outputs it as a file. This file will be able to be used as input in ReportGenerator tool.

## Validation Steps Performed

A cobertura-coverage.txt file should be created in coverage container in Cobertura XML format with same data as source-coverage.json file.

### DIFF
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -370,8 +370,7 @@ impl<'a> TaskContext<'a> {
             .local_path
             .join(COBERTURA_COVERAGE_FILE);
         let cobertura_source_coverage = cobertura(src_coverage)?;
-        let text = &cobertura_source_coverage;
-        fs::write(&path, &text)
+        fs::write(&path, &cobertura_source_coverage)
             .await
             .with_context(|| format!("writing cobertura source coverage to {}", path.display()))?;
 
--- a/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
@@ -13,6 +13,7 @@ use coverage::block::CommandBlockCov;
 use coverage::cache::ModuleCache;
 use coverage::code::{CmdFilter, CmdFilterDef};
 use coverage::debuginfo::DebugInfo;
+use coverage::cobertura::cobertura;
 use onefuzz::expand::{Expand, PlaceHolder};
 use onefuzz::syncdir::SyncedDir;
 use onefuzz_telemetry::{warn, Event::coverage_data, EventData};
@@ -30,6 +31,7 @@ use crate::tasks::heartbeat::{HeartbeatSender, TaskHeartbeatClient};
 const MAX_COVERAGE_RECORDING_ATTEMPTS: usize = 2;
 const COVERAGE_FILE: &str = "coverage.json";
 const SOURCE_COVERAGE_FILE: &str = "source-coverage.json";
+const COBERTURA_COVERAGE_FILE: &str = "cobertura-coverage.json";
 const MODULE_CACHE_FILE: &str = "module-cache.json";
 
 const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(5);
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -30,6 +31,7 @@ use crate::tasks::heartbeat::{HeartbeatSender, TaskHeartbeatClient};
 const MAX_COVERAGE_RECORDING_ATTEMPTS: usize = 2;
 const COVERAGE_FILE: &str = "coverage.json";
 const SOURCE_COVERAGE_FILE: &str = "source-coverage.json";
+const COBERTURA_COVERAGE_FILE: &str = "cobertura-coverage.txt";
```
ranweiler, line 34:
```suggestion
const COBERTURA_COVERAGE_FILE: &str = "cobertura-coverage.xml";
```

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -362,6 +364,13 @@ impl<'a> TaskContext<'a> {
             .await
             .with_context(|| format!("writing source coverage to {}", path.display()))?;
 
+        let path = self.config.coverage.local_path.join(COBERTURA_COVERAGE_FILE);
+        let cobertura_source_coverage = cobertura(src_coverage)?;
+        let text = &cobertura_source_coverage;
+        fs::write(&path, &text)
+        .await
```
ranweiler: Heads up, you'll need to run `cargo fmt` (CI will complain).

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
```diff
@@ -362,6 +364,13 @@ impl<'a> TaskContext<'a> {
             .await
             .with_context(|| format!("writing source coverage to {}", path.display()))?;
 
+        let path = self.config.coverage.local_path.join(COBERTURA_COVERAGE_FILE);
+        let cobertura_source_coverage = cobertura(src_coverage)?;
+        let text = &cobertura_source_coverage;
```
ranweiler, line 373: You should be able to just use `&cobertura_source_coverage` in the next line, if you like.

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
hayleycall: Just remembered!

src/agent/onefuzz-agent/src/tasks/coverage/generic.rs
hayleycall: I like separating them out- is it bad design if I do?

ranweiler: Nope, not necessarily. There's a lot of subjectivity.

Here's the rationale of why I would personally choose not to do this. 
- The value we are constructing and naming is just borrow of the value `cobertura_source_coverage`. It truly is just a restricted pointer to the original thing. There's no complex, expensive, tricky, or _fallible_ conversion happening, either. Dedicating a new variable name (and a new line) adds one more thing to keep in our heads when we read ("what is the meaning of the variable `text`?"), without really giving us much for maintaining that distinction. 
- Furthermore, the syntactic weight (taking up an entire line) makes it seem like something important is going on, especially because introducing a new variable for this scenario is uncommon in Rust. But that's not true: we're just doing a regular old borrow, then immediately using it exactly once. So in my mind, the intent is made clearer (at a lower cognitive cost) just by constructing the reference exactly when we pass it to the function.
- The local name `text` is also much less descriptive than `cobertura_source_coverage`. That could be fixed, but what would we rename it to? The meaning of this variable would be best described by something like `cobertura_source_coverage_ref`. But at that point, we could just call it `&cobertura_source_coverage`, which expresses the same idea using the syntax dedicated for exactly that purpose. It also keeps our borrow duration as short as possible, which is always what we want.

ranweiler (approved):

PR #1589, opened 2022-01-19
Bump backoff from 0.3.0 to 0.4.0 in /src/agent, resolve dependabot block

hayleycall:

Resolves dependabot PR 1526. This PR updates backoff version from 0.3.0 to 0.4.0 and updates new fields for Backoff::Error::Transient. 

## PR Checklist
* [X] Tests added/passed

## Info on Pull Request

Update Backoff version and change uses of Backoff::Error::Transient.


src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -299,10 +299,11 @@ pub async fn wait_for_dir(path: impl AsRef<Path>) -> Result<()> {
         if path.as_ref().exists() {
             Ok(())
         } else {
-            Err(BackoffError::Transient(anyhow::anyhow!(
-                "path '{:?}' does not exist",
-                path.as_ref()
-            )))
+            let max_retry_period: Option<Duration> = Some(WAIT_FOR_MAX_WAIT);
```
chkeita, line 302: The fix [suggested in the doc](https://github.com/ihrwein/backoff#03x---04x) is a better fit here this should keep the existing behavior (same applies to the other changes)
``` suggestion 
       Err(BackoffError::transient {
              err: anyhow::anyhow!("path '{:?}' does not exist", path.as_ref()),
              retry_after: max_retry_period,
          })
```

src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -299,10 +299,11 @@ pub async fn wait_for_dir(path: impl AsRef<Path>) -> Result<()> {
...
+            Err(BackoffError::Transient {
+                err: anyhow::anyhow!("path '{:?}' does not exist", path.as_ref()),
```
ranweiler, line 304: Values of type `&Path` do not impl the `Display` trait, but they provide a [`display()` method](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.display) which returns a thin wrapper type that does impl `Display`, so we can use `{}` instead of `{:?}`. This will result in my more readable error messages with (for example) less escaping of path separators.

Note that this is specific to `Path`/`PathBuf`.
```suggestion
                err: anyhow::anyhow!("path '{}' does not exist", path.as_ref().display()),
```

hayleycall: Fixed!

src/agent/onefuzz-agent/src/local/common.rs
hayleycall: This change causes this error: error[E0599]: no variant named `transient` found for enum `backoff::Error<_>`

chkeita: you need to call the function not the variant. it is a parenthesis `(` instead of bracket `{` after BackoffError::transient

hayleycall: fixed!

PR #1591, opened 2022-01-20
Move the event grid topic creation and subscription to the deployment template

chkeita:
Moves the event grid subscription to the deployment template. This allows us to control the name associated to the event grid topic created so we can reuse it in the bug filing service


src/deployment/deploy.py
```diff
@@ -694,42 +689,50 @@ def set_instance_config(self) -> None:
             tenants.append(tenant)
         update_allowed_aad_tenants(config_client, tenants)
 
-    def create_eventgrid(self) -> None:
-        logger.info("creating eventgrid subscription")
-        src_resource_id = self.results["deploy"]["fuzz-storage"]["value"]
-        dst_resource_id = self.results["deploy"]["func-storage"]["value"]
-
+    def remove_eventgrid(self) -> None:
```
stishkin, line 692: old one has to be removed, can't be re-used ?

chkeita: no we can't reuse it because it was associated with the auto generated topic that we want to remove

src/deployment/azuredeploy.json
```diff
@@ -67,7 +67,8 @@
         "Storage Account Contributor": "17d1049b-9a84-46fb-8f53-869881c3d3ab",
         "Virtual Machine Contributor": "9980e02c-c2be-4d73-94e8-173b1dc7cf3c",
         "Storage Blob Data Reader": "2a2b9908-6ea1-4ae2-8e65-a410df84e7d1",
-        "keyVaultName": "[concat('of-kv-', uniquestring(resourceGroup().id))]"
+        "keyVaultName": "[concat('of-kv-', uniquestring(resourceGroup().id))]",
+        "fuzz-blob-topic-name": "[concat('fuzz-blob-topic-', uniquestring(resourceGroup().id))]"
```
mgreisen, line 71: What's the purpose of the "fuzz-blob-topic-name"? It's strikes me as fairly generic and I'm wondering if its the right name.

src/deployment/azuredeploy.json
```diff
@@ -865,6 +866,44 @@
...
+            "location": "[resourceGroup().location]",
+            "properties": {
+                "source": "[resourceId('Microsoft.Storage/storageAccounts', variables('storageAccountName'))]",
+                "topicType": "microsoft.storage.storageaccounts"
+            }
+        },
+        {
+            "type": "Microsoft.EventGrid/systemTopics/eventSubscriptions",
+            "apiVersion": "2021-12-01",
+            "name": "[concat(variables('fuzz-blob-topic-name'), '/onefuzz1_subscription')]",
```
mgreisen, line 883: Why the "1" in the name "/onefuzz1_subscription", was that for debugging purposes?

src/deployment/deploy.py
```diff
@@ -694,42 +689,50 @@ def set_instance_config(self) -> None:
...
-            )
+
+        old_subscription_name = "onefuzz1"
+
+        def old_subscription_exists() -> bool:
+            try:
+                result = client.event_subscriptions.get(
+                    src_resource_id, old_subscription_name
+                )
+                return bool(result.provisioning_state == "Succeeded")
```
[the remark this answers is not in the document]
chkeita: updated. we will remove if it exist in any state

src/deployment/azuredeploy.json
chkeita: It is the topic created to gather the notification of the blobs in the fuzz storage account.
We subscribe to these events to trigger the bug filling function. 
I am open to suggestions on the name

src/deployment/azuredeploy.json
chkeita: i just followed the existing name. The existing subscription was named `onefuzz1`. I just added the suffix `_subscription`

stishkin (approved):

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
-                event_time_to_live_in_minutes=1440,
-            ),
-        )
+        try:
+            storage_accounts = client.storage_accounts.list_by_resource_group(
+                self.resource_group
+            )
+            for storage_account in storage_accounts:
+                if storage_account.name.startswith("fuzz"):
+                    src_resource_id = storage_account.id
```
ranweiler, line 703:
```suggestion
                    src_resource_id = storage_account.id
                    break
```

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
-                    "Microsoft.Storage.BlobDeleted",
-                ]
-            ),
-            retry_policy=RetryPolicy(
-                max_delivery_attempts=30,
-                event_time_to_live_in_minutes=1440,
-            ),
-        )
+        try:
+            storage_accounts = client.storage_accounts.list_by_resource_group(
```
ranweiler, line 698: It might be useful to factor this entire `try` block out into a function that just returns the user data/`fuzz<..>` storage account.

Then this could be:
```suggestion
            src_resource_id = user_data_storage_account().id
```
(or similar)

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
-        result = client.event_subscriptions.begin_create_or_update(
-            src_resource_id, "onefuzz1", event_subscription_info
-        ).result()
-        if result.provisioning_state != "Succeeded":
-            raise Exception(
-                "eventgrid subscription failed: %s"
-                % json.dumps(result.as_dict(), indent=4, sort_keys=True),
-            )
+
+        old_subscription_name = "onefuzz1"
```
ranweiler, line 714: Suggest documenting version here.

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
+
+        def old_subscription_exists() -> bool:
```
ranweiler, line 716: I'd find it easier to read this if you either (1) computed this value inline, with no variable-capturing closure, or (2) fully factored out this function, so it takes `src_resource_id` as an argument.

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
+            try:
+                client.event_subscriptions.get(src_resource_id, old_subscription_name)
```
ranweiler, line 718: If `get()` ever stopped throwing an exception, and instead started returning `None` if `old_subscription_name` is absent, then this would start silently failing.

Maybe we should defensively do something like;
```
subscription = client.event_subscriptions.get(src_resource_id, old_subscription_name)
return bool(subscription)
```

src/deployment/deploy.py
```diff
@@ -694,42 +689,48 @@ def set_instance_config(self) -> None:
...
+            except ResourceNotFoundError:
+                return False
+
+        if old_subscription_exists():
+            logger.info("removing deprecated event subscription")
+            result = client.event_subscriptions.begin_delete(
+                src_resource_id, old_subscription_name
+            ).result()
+
+            if result and result.provisioning_state != "Succeeded":
```
ranweiler, line 729: If `result` is absent / evals `False`, then does that mean something else has gone wrong? Should we have a check like:
```python
if not result:
    raise Exception(...)
```

src/deployment/deploy.py
chkeita: i expect any error other than `ResourceNotFoundError` to be raised "loudly" here

ranweiler: Right, loud failures are what will happen right now, since we don't try to catch any exception but `ResourceNotFoundError`. We assume other error conditions will fail loudly as uncaught exceptions. That makes sense.

I'm just noting that:
1. Right now we implicitly treat any non-throwing call to `get()` as proof that the subscription exists.
2. So, the upstream library could break us in a way that would pass `mypy` checks: if they made `get()` return an `Optional[_]`, and started returning `None` in some situations (_instead_ of throwing), we'd never see it.

Arguably too defensive, but I was imagining it'd be a big pain to track down if it ever happened!

src/deployment/deploy.py
chkeita: the check on result was actually not necessary. The operation return a polling object.

ranweiler (approved):

PR #1594, opened 2022-01-21
Fix clippy-caused build break

stishkin:
Fixing multiple build breaks caused by clippy


chkeita (approved):

ranweiler (approved):
This is fine for now. I opened #1595 so I can audit them properly.

PR #1610, opened 2022-01-24
Add `fuzzer_target_options` to `libfuzzer basic` template

### LINKED ISSUE #1437: Detect and distinguish dual-use libfuzzer args for fuzzing
When creating libfuzzer jobs, users may add `-runs=N` to `target_options` to try to handle situations like target code that leaks memory. This is fine and expected, and handled by the `libfuzzer_fuzz` task.

However, when a libFuzzer is invoked in test case repro mode, the `-runs=N` option says "_test_ the given input(s) `N` times each". This is definitely not what we want in analysis tasks like `coverage`, `libfuzzer_crash_report`, &c.

Revisit our task arguments, and fix this end-user facing API. Make it easy (at the job template level) to clearly distinguish between libFuzzer args meant for fuzzing only, test case repro, or both. This is more easily avoided outside of templates, where tasks must be created individually.

ranweiler:
Add a new `fuzzer_target_options` argument to the libfuzzer templates to enable setting some libfuzzer arguments only when invoked in fuzzing mode.

Tested the matrix of `{target_options = None, [x=y]} x {fuzzer_target_options = None, [x=y]}}` by creating jobs with the `libfuzzer basic` template. Verified that all 4 cases look as expected by querying the created tasks and examining their configs. In particular, the `coverage` task should always have `{input}` added to it (because it is not LibFuzzer-aware), and in all cases, the `fuzzer_target_options` values should be _appended_ to the job-wide `target_options` for the `libfuzzer_fuzz` task _only_, and omitted for all others. The `dotnet` and `qemu_user` templates were tested with the same matrix.

Closes #1437.


stishkin (approved):

chkeita (approved):

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -567,6 +568,15 @@ def dotnet(
             helper.upload_inputs(inputs)
         helper.wait_on(wait_for_files, wait_for_running)
 
+        # Build `target_options` for the `libfuzzer_fuzz` task.
```
chkeita, line 571: I suggest you move these changes to line 532. it will make it easier to understand the state of `target_options`.
looking at just this is not evident that `target_options` is already initialized at this point and we ca safely append to it.

ranweiler: My goal was to follow the pattern of what we do with the (internal) `coverage_target_options`, where we define the task-specific variants near the task creation call site.

With that in mind, would you still suggest moving it?

chkeita: Maybe it makes more sense to move the `target_options` close to this code then. that will still allow you to follow the pattern. 
Either way, I don't have a strong opinion on this so i'll let you choose to move it or not.

ranweiler: Ya, that might make more sense. I'll revisit it holistically when I context switch back to this PR.

ranweiler: I'm going to leave this where it is, and here's why: I think of our defaulting of a `None`-valued `target_options` to `[]` as logically part of the function "prologue". The only reason we have to do it in our code is that in Python, the mutability of the list object literal in default value position makes it act like a function-static variable.

We also "manually default" other arguments (like `target_env`) near the top, in this "prologue-like" part of the function, so I really think this is the right place to do it.

hayleycall (approved):

PR #881, opened 2021-05-12
update HasState Protocol to alway log state transitions

bmc-msft:

Issue #851, opened 2021-05-03
Consistently use TZ-aware/unaware `datetime` values


ranweiler:
Python `datetime` objects can be TZ-aware or unaware. In the service code, we are inconsistent in our usage of one or another (mostly using TZ-unaware).

If we only perform local calculations with objects of the same awareness, no problems can occur. But if they differ and we compare them, they may fail in (inconsistent) ways:
- `==` comparisons silently fail, and evaluate `False` when comparing two semantically-equal `datetime` objects, but where just one is TZ-aware
- Inequality checks (`<`, &c) throw an _exception_

TZ-awareness appears to not be statically-checkable out of the box. See python/mypy#10067.

We seem to only use TZ-aware code in some parts of the proxy (which caused pain in testing for #839). The quick fix is just to update our usage there. Trying to change the existing (predominantly unaware) code to TZ-aware will likely have a long tail of bugs. A longer-term and safer fix could be to use our own (more statically-constrained) `datetime` wrappers or subclasses, and move everything to those, to prevent mismatches in the future. 

AB#39994539

glyph:
You can statically check TZ-awareness using [this little library I wrote](https://blog.glyph.im/2022/06/dates-and-times-and-types.html) until Mypy fixes the bug and ideally absorbs it.

Porges:
Closing this as service has migrated to C#, where we are using `DateTimeOffset` consistently.

PR #1616, opened 2022-01-25
Fixing VMSS Re-Image 7-Day Timer

nharper285:

_What is this about?_
The 7-Day VMSS Re-Image timer currently does not work. It relies upon the node 'TimeStamp' entry which is actually updated every few minutes when new node heartbeats come in. I am creating a new table entry - created_at - to fix this behavior. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

* [x] Set timer to 15 min and observe timer reimage
* [x] Deploy old instance with scaleset (w/o init_at field). Then deploy update over old instance and confirm scaleset updates with new field.


src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -76,6 +77,8 @@ def on_state_update(
         # successfully.
         node.reimage_requested = False
         node.set_state(state)
+        node.created_at = datetime.datetime.utcnow()
```
chkeita, line 80: i don't think `created_at` should be mutable. The node is created once and reimage multiple times. 
I propose rename this to last_os_deployment or something similar to avoid the confusion.

nharper285: I think that's fair! Made the change.

chkeita (approved):

chkeita:
@noah can you update the description with your validation steps

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -453,7 +453,7 @@ def reimage_long_lived_nodes(cls, scaleset_id: UUID) -> None:
         reasonably up-to-date with OS patches without disrupting running
         fuzzing tasks with patch reboot cycles.
         """
-        time_filter = "Timestamp lt datetime'%s'" % (
+        time_filter = "initialized_at lt datetime'%s'" % (
```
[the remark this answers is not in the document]
nharper285: Since this check is already always false for older nodes, I figured older nodes would have to wait for a post-job reimage to pick up the new field. 

I will test this to see how it behaves.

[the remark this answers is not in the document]
chkeita: you can check [this](https://github.com/microsoft/onefuzz/blob/2ad37bdfb0be0be372940b676861b7320837b082/src/api-service/__app__/onefuzzlib/workers/nodes.py#L124) to see how we take it into account when creating the query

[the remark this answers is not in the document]
nharper285: Sorry Cheick, what do you mean? Are you suggesting I use 'search_outdated'?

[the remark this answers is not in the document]
mgreisen: If we wait for older nodes to pickup jobs before upgrading and reimaging, we risk having idle nodes become really stale because of the way we statically define VMSS sizes.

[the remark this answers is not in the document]
nharper285: I see. I will make sure this is updated.

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -367,8 +367,8 @@ def is_outdated(self) -> bool:
     def is_too_old(self) -> bool:
         return (
             self.scaleset_id is not None
```
ranweiler, line 369: Not for this PR, but presumably this hasn't actually been (logically) optional for quite some time.

src/pytypes/onefuzztypes/models.py
```diff
@@ -568,6 +568,7 @@ class NodeCommandEnvelope(BaseModel):
 
 class Node(BaseModel):
     timestamp: Optional[datetime] = Field(alias="Timestamp")
```
ranweiler, line 570: So logically, what does this mean? Is it `last_modified`, `last_heartbeat`, `last_event`, &c? It'd be good to document exactly what the meaning is, since we've been tripped up by changes that made it ambiguous.

src/pytypes/onefuzztypes/models.py
```diff
@@ -568,6 +568,7 @@ class NodeCommandEnvelope(BaseModel):
...
+    initialized_at: Optional[datetime]
```
ranweiler, line 571: Is the idea here that when a node enters `NodeState.init`, that implies that it has been updated/reimaged per the latest VMSS model?

Do we expect this to only get set once, and on re-init we get a whole new `Node` (with e.g. a new `machine_id`)? Should we call that out in a field comment?

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -35,7 +35,7 @@
 from .shrink_queue import ShrinkQueue
 
 NODE_EXPIRATION_TIME: datetime.timedelta = datetime.timedelta(hours=1)
-NODE_REIMAGE_TIME: datetime.timedelta = datetime.timedelta(days=7)
+NODE_REIMAGE_TIME: datetime.timedelta = datetime.timedelta(days=6)
```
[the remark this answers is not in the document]
nharper285: The queries use this value to define stale nodes. 
image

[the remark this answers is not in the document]
nharper285: Unless you just mean changing from 7 to 6. In that case, I'd say no. I'm changing the value to 6 because some of our s360 issues are based off 7 day timers. For example, antimalware asks us to update our signatures every 7 days - which we do via reimage. 
Since the s360 timer is on 7 days, it makes sense for us to have our timer be one day less to avoid weird overlap.

src/pytypes/onefuzztypes/models.py
nharper285: I'd say it's last_modified... and really, it's 'last time the storage table was modified'.

So maybe 'config_last_modified'

src/pytypes/onefuzztypes/models.py
nharper285: To your first question - yes, Cheick and I were able to confirm that yesterday. 

I'd expect this to only get set once, yes. I can include a field comment.

hayleycall (approved):

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -35,7 +35,7 @@
 from .shrink_queue import ShrinkQueue
 
 NODE_EXPIRATION_TIME: datetime.timedelta = datetime.timedelta(hours=1)
-NODE_REIMAGE_TIME: datetime.timedelta = datetime.timedelta(days=7)
+NODE_REIMAGE_TIME: datetime.timedelta = datetime.timedelta(minutes=10)
```
ranweiler, line 38:
```suggestion
NODE_REIMAGE_TIME: datetime.timedelta = datetime.timedelta(days=6)
```

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -453,15 +453,18 @@ def reimage_long_lived_nodes(cls, scaleset_id: UUID) -> None:
...
         )
-        for node in cls.search(
+        reimage_nodes = cls.search(
             query={
                 "scaleset_id": [scaleset_id],
             },
             raw_unchecked_filter=time_filter,
-        ):
+        )
+        logging.info(f"Number of nodes to reimaged {len(reimage_nodes)}")
```
ranweiler, line 465:
```suggestion
        logging.info(f"Number of nodes to reimage: {len(reimage_nodes)}")
```

src/pytypes/onefuzztypes/models.py
ranweiler, line 570:
```suggestion
    # Last modification time of node metadata.
    timestamp: Optional[datetime] = Field(alias="Timestamp")
```

src/pytypes/onefuzztypes/models.py
```diff
@@ -568,6 +568,7 @@ class NodeCommandEnvelope(BaseModel):
 
 class Node(BaseModel):
     timestamp: Optional[datetime] = Field(alias="Timestamp")
+    initialized_at: Optional[datetime]  # Set only once when a node is intialized
```
ranweiler, line 571:
```suggestion

    # Set only once, when a node is initialized.
    initialized_at: Optional[datetime]  
```

src/api-service/__app__/onefuzzlib/workers/nodes.py
ranweiler: (ICYMI)

src/api-service/__app__/onefuzzlib/agent_events.py
```diff
@@ -75,7 +76,9 @@ def on_state_update(
         # they send 'init' with reimage_requested, it's because the node was reimaged
         # successfully.
         node.reimage_requested = False
+        node.initialized_at = datetime.datetime.utcnow()
```
ranweiler, line 79: Should this be:
```suggestion
        node.initialized_at = datetime.datetime.now(datetime.timezone.utc)
```
The difference (naive vs. TZ-aware) is described here: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

Note that our queries which operate on this field use the TZ-aware form.

src/pytypes/onefuzztypes/models.py
ranweiler: (fixes typo, formats for breathing room next to the other suggested field coment)

src/pytypes/onefuzztypes/models.py
ranweiler: Great, I think the comment will make it clear.

chkeita: Note that this field is not specific to this entity. This is a field is provided by Azure Table and is used by other entities too.
I don't think it is necessary to comment on its meaning here only since it is not logically different from any place where it is being used.

ranweiler: Ah, that's right. Since we're just inheriting it from Azure Table, I agree with @chkeita on this field.

ranweiler: @nharper285, disregard this suggestion, since we're just inheriting it from Azure Table, and commenting further is actually more confusing than helpful (there's nothing special about it).

src/pytypes/onefuzztypes/models.py
bmc-msft, line 571: Instead of making this field optional, why not have it use `Field` to set a date if it doesn't already exist?

chkeita: This would change the meaning of the field. The idea is to record when we received the init message. A none value means that we have not received it yet. if we generate it automatically, we lose that information

src/api-service/__app__/onefuzzlib/agent_events.py
nharper285: Hm I see them used fairly interchangeably in the code. The queries that we use to check the timer are all datetime.datetime.utcnow(), as well.

ranweiler: > Hm I see them used fairly interchangeably in the code. 

That doesn't mean they should be 😄. Take a look at #851.

Example of mixing going awry:
```python
>>> import datetime as dt

# "naive" (not TZ-aware)
>>> now = dt.datetime.utcnow()

# The same time, but TZ-aware
>>> tznow = dt.datetime(2022, 1, 26, 22, 47, 1, 24174, tzinfo=dt.timezone.utc)

# These times are the "same" in many ways
>>> now
datetime.datetime(2022, 1, 26, 22, 47, 1, 24174)
>>> tznow
datetime.datetime(2022, 1, 26, 22, 47, 1, 24174, tzinfo=datetime.timezone.utc)
>>> now.timestamp()
1643237221.024174
>>> tznow.timestamp()
1643237221.024174

# But NOT the same in other ways
>>> now == tznow
False
>>> now <= tznow
  File "<stdin>", line 1, in <module>
TypeError: can't compare offset-naive and offset-aware datetimes
```

> The queries that we use to check the timer are all datetime.datetime.utcnow(), as well.

What about [this?](https://github.com/microsoft/onefuzz/pull/1616/files#diff-a2210def4c65168b052d6da50ae48f32979d2cc6a5bf02f1fcfd2d510f16e7ebR372)

nharper285: Okay wow I see the issue and yes it looks like the query above uses the appropriate format. Did we ever fully address bug #881? That could be a good follow up to this PR.

ranweiler: No, #851 hasn't been addressed yet. I think we do probably want to clean that up to avoid confusing and frustrating errors in the future.

nharper285: Sounds good. I'll make a note of it for our future sync.

PR #1635, opened 2022-01-31
add tool_name and onfuzz_version to CrashReport

### LINKED ISSUE #1611: Add `Version` and `ToolName` to the Report
To help the conversion of our report to the SARIF format, we will need to add the following properties 
- Verion: The version of onefuzz that generated the report
- ToolName: The name of the tool/fuzzer/analyzer that created the report

chkeita:
closes #1611


src/agent/onefuzz-agent/src/tasks/report/crash_report.rs
```diff
@@ -221,6 +227,8 @@ impl CrashReport {
             Some(crash_log.minimized_stack_function_names)
         };
 
+        let onefuzz_version = env!("ONEFUZZ_VERSION").to_string();
```
ranweiler, line 230: Maybe `onefuzz_version` should be a parameter, and we should do the `env!` lookup where we call `CrashReport::new()`. This way the ctor stays a pure function, the `version` can be changed in tests, &c.

src/pytypes/onefuzztypes/models.py
```diff
@@ -213,6 +213,8 @@ class Report(BaseModel):
     asan_log: Optional[str]
     task_id: UUID
     job_id: UUID
+    tool_name: str
+    onefuzz_version: str
```
ranweiler, line 217: Is this `onefuzz_version` in the sense of versioning the crash report, or is it the the tool version? 

Should we also include `tool_version`, or just have that be part of `tool_name` if relevant?

ranweiler (approved):

PR #1647, opened 2022-02-07
Enable scale in protection on VMSS instances

tevoinea:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
As part of an overall feature to allow onefuzz to scale (up and down) based on scheduled work, this PR adds the capability to protect nodes from azure scaling actions only while the nodes are completing work.

## Validation Steps Performed

### DIFF
--- a/src/api-service/__app__/agent_can_schedule/__init__.py
+++ b/src/api-service/__app__/agent_can_schedule/__init__.py
@@ -40,7 +40,7 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
         allowed = False
 
     if allowed:
-        node.acquire_scale_in_protection()
+        allowed = not isinstance(node.acquire_scale_in_protection(), Error)
 
     return ok(CanSchedule(allowed=allowed, work_stopped=work_stopped))
 
--- a/src/api-service/__app__/agent_can_schedule/__init__.py
+++ b/src/api-service/__app__/agent_can_schedule/__init__.py
@@ -39,6 +39,9 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if work_stopped:
         allowed = False
 
+    if allowed:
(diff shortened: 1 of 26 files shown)


stishkin (approved):

src/api-service/__app__/agent_can_schedule/__init__.py
```diff
@@ -39,6 +39,9 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if work_stopped:
         allowed = False
 
+    if allowed:
+        node.acquire_scale_in_protection()
```
chkeita, line 43: This call add some time to the processing of the request. Have you check the behavior of the agent when there is a timeout. 
Thinking about the following scenario. 

- azure is busy so requests to protect nodes are very slow
- an agent sends check if it can schedule work 
- the request times out because azure is slow But the lock was acquired 
- The agent drops the work

How long before we can remove the lock. I see that you reclaim it when we reimage the node. Can you confirm that in this scenario the node is ready to be reclaimed?
would this be a problem with a lot of nodes in the same situation

tevoinea, line 43: TODO: If the protection fails, we shouldn't allow the work to be scheduled

tevoinea: If the protection is acquired but the CanSchedule call times out, the node will retry 5 times every 5 seconds. If none of those are successful, the node will try to claim the next piece of work. If there's no new work available or the node continues to fail to claim new work, the node will idle in a protected state until we reclaim it after 6 days. This implementation does lean towards over-protecting the nodes. If the node has been idle it can definitely be reclaimed.

Also, we can always tune the unlock behavior to be more aggressive if we're noticing resource wastage.

PR #1657, opened 2022-02-15
Fail fast if managed task workers are near-OOM

### LINKED ISSUE #1633: Excessive fuzzer memory usage can cause confusing task failures
Fuzzers (and user `target_exe`'s in general) can hang or OOM. We detect hangs with timeouts, but OOMs can take down a node and/or fail a task due to cross-process OOM-ing.

This can be true even in LibFuzzers, where the `-rss_limit_mb` arg results in a [periodic check](https://github.com/llvm/llvm-project/blob/23091f7d504afde4bba3fc885718a1633746e063/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L304-L311) of memory usage, [triggering a callback](https://github.com/llvm/llvm-project/blob/23091f7d504afde4bba3fc885718a1633746e063/compiler-rt/lib/fuzzer/FuzzerLoop.cpp#L308-L321) on excess usage.

The core problem here is that we need to be able to quickly differentiate OneFuzz or other system bugs from faults in user code. Ideally, any input-triggered OOM should be captured as a faulting input. Failing that, user-provided targets should not be able to indirectly fail a node. In the worst case, the task should still fail cleanly and informatively.

Possible solutions:
- Track _total_ system memory usage, and informatively fast-fail the task if it exceeds some limit
- Run all user-provided binaries with resource limits, via rlimit/job objects

ranweiler:
### Summary
- Add `onefuzz::memory::available_bytes()` to enable checking system-wide memory usage
- In managed task worker runs, heuristically check for imminent OOM conditions and try to exit early

### Testing
- [x] Added unit tests for limited parsing of `/proc/meminfo` on Linux
- [x] Memory querying functionally tested locally on both OSes using `memory` example binary in `onefuzz` crate
- [x] Near-OOM check in `onefuzz-agent` functionally tested by forcing external on-VM OOM
  - [x] Linux
  - [x] Windows

Tested on Windows by using `notmyfault` from SysInternals to leak from the paged memory pool, on a node where a healthy task was in the `running` state. On Linux, disabled OOM-killer, set overcommit to always-allow, and ran small custom program that rapidly leaks heap memory. Checked for task failure, with the reason matching the new error message, as well as telemetry.

Closes #1633.


src/agent/onefuzz-agent/src/managed/cmd.rs
```diff
@@ -23,6 +39,35 @@ pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
     result
 }
 
+// Periodically check available system memory.
+//
+// If available memory drops below the minimum, exit informatively.
+async fn out_of_memory(min_bytes: u64) -> Result<OutOfMemory> {
```
ranweiler, line 45: I could have use the private named constant that we always pass here. I only parameterized this check to make it clear how we'd make  it configurable (and thread the exact limit through to the final error message).

src/agent/onefuzz/examples/memory.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
```
ranweiler, line 1: This is just an example binary for manual testing.

src/agent/onefuzz-agent/src/managed/cmd.rs
stishkin: add that as comment in the code ?

src/agent/onefuzz-agent/src/managed/cmd.rs
```diff
@@ -13,7 +19,17 @@ pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
     let config = Config::from_file(config_path, setup_dir)?;
 
     init_telemetry(config.common());
-    let result = config.run().await;
+
+    let result = tokio::select! {
+        result = config.run() => result,
+
+        // Ignore this task if it returns due to a querying error.
+        Ok(oom) = out_of_memory(MIN_AVAILABLE_BYTES) => {
```
ranweiler, line 27: @chkeita, note how this differs a bit from our earlier discussion.

The supervisor `Agent` is implemented using (dynamic trait object) indirection to allow integration testing with test doubles for its internal components. Trying to add this global resource check as-is would pollute that, and was actually a little awkward (the supervisor really needs to be refactored to be `select`-driven).

This was the most minimal alternative way I could think up to get the behavior change I wanted, which was: "if we are in the `busy` state, running a task, and we're about to OOM, stop the task with an informative error".

src/agent/onefuzz-agent/src/managed/cmd.rs
ranweiler: IMO the actual usage is clear enough, I was more just noting it here for discussion's sake.

Do we think we'll want to make it tunable (i.e. per-node or task) in the future? If not, I can just simplify it by making it NOT parameterized.

ranweiler: (just added a comment anyway)

mgreisen: I'm ok with leaving it tunable.

mgreisen (approved):

src/agent/onefuzz-agent/src/managed/cmd.rs
```diff
@@ -23,6 +43,43 @@ pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
...
+                        available_bytes,
+                        min_bytes,
+                    });
+                }
+            }
+            Err(err) => {
+                warn!("error querying system memory usage: {}", err);
+                return Err(err);
+            }
+        }
```
chkeita, line 67: This seems a bit agressive to stop everything as soon as we are unable to read the memory.
I would suggest raising the error if we fail to read the memory after a small number of attempts

src/agent/onefuzz-agent/src/managed/cmd.rs
```diff
@@ -13,7 +23,17 @@ pub async fn run(args: &clap::ArgMatches<'_>) -> Result<()> {
...
-    let result = config.run().await;
+
+    let result = tokio::select! {
+        result = config.run() => result,
+
+        // Ignore this task if it returns due to a querying error.
+        Ok(oom) = out_of_memory(MIN_AVAILABLE_BYTES) => {
+            // Convert the OOM notification to an error, so we can log it below.
+            let err = format_err!("out of memory: {} bytes available, {} required", oom.available_bytes, oom.min_bytes);
+            Err(err)
```
chkeita, line 34: I think the decision to stop the task when OOM should be configurable and enabled by default.

src/agent/onefuzz-agent/src/managed/cmd.rs
ranweiler: For the record, the current behavior only stops the memory watchdog async task. But agreed, we should only give up the memory check after several failures in a row.

src/agent/onefuzz-agent/src/managed/cmd.rs
ranweiler: Added to the common config, on by default: 46a3d7a899ac90049fab00a3f04d1f3e71e4c222

Can be disabled by [setting to 0](https://github.com/microsoft/onefuzz/pull/1657/commits/46a3d7a899ac90049fab00a3f04d1f3e71e4c222#diff-083902b0f7ce9dceb9ecfb4eb6dda1de4a3595bd2161c306505e925880271b1cR55-R57).

ranweiler: Added: https://github.com/microsoft/onefuzz/pull/1657/commits/0befdc5f5608141a0e8fc86122dc0b37e2c432b0

chkeita (approved):

ranweiler:
- Add `onefuzz::memory::available_bytes()` to enable checking system-wide memory usage
- In managed task worker runs, heuristically check for imminent OOM conditions and try to exit early

PR #1661, opened 2022-02-17
Create autoscale resources for scaleset

tevoinea:

_What is this about?_
This PR creates an auto scale resource for scalesets that uses the pool work queue as the tirgger metric.

## Info on Pull Request

_What does this include?_

* These changes only affect new scalesets, existing scalesets will not be modified
* Scalesets that already have an auto scale resource associated with them will not be modified
* The autoscaling resource is disabled meaning scaleset behavior won't change
* In the interest of keeping this reviewable, hooking up scaling event diagnostics will be a follow up PR

## Validation Steps Performed

1. Create a new scaleset
2. Find your new scaleset n azure portal
3. Under "Scaling", select the "JSON" tab
4. Verify there is 1 object under `.properties.profiles`
5. Verify `.properties.enabled == false`

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
+++ b/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
@@ -27,7 +27,7 @@ from azure.mgmt.monitor.models import (
     ComparisonOperationType,
     MetricStatisticType,
     ScaleDirection,
-    ScaleType
+    ScaleType,
 )
 from azure.core.exceptions import (
     HttpResponseError,
@@ -45,8 +45,11 @@ from .creds import (
 
 # TODO: Look into deleting auto scale resource when deleting scaleset
 
+
 @retry_on_auth_failure()
-def add_auto_scale_to_vmss(vmss: UUID, auto_scale_profile: AutoscaleProfile) -> Optional[Error]:
+def add_auto_scale_to_vmss(
+    vmss: UUID, auto_scale_profile: AutoscaleProfile
+) -> Optional[Error]:
     logging.info("Checking scaleset %s for existing auto scale resources" % vmss)
     client = get_monitor_client()
     resource_group = get_base_resource_group()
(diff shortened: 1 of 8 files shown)


src/api-service/__app__/onefuzzlib/azure/monitor.py
```diff
@@ -1,29 +1,9 @@
-#!/usr/bin/env python
```
tevoinea, line 1: The original file was moved to `log_analytics.py`

src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -0,0 +1,154 @@
...
+from msrestazure.azure_exceptions import CloudError
+
+from .creds import (
+    get_base_region,
+    get_base_resource_group,
+    get_subscription,
+    retry_on_auth_failure,
+)
+
+# TODO: Look into deleting auto scale resource when deleting scaleset
```
tevoinea, line 46: Tested this locally using `onefuzz scalesets shutdown`. Azure automatically deletes the auto scale resource so we don't need to add any additional logic for that case.

src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -0,0 +1,145 @@
...
+
+    scaleset_uri = (
+        "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/virtualMachineScaleSets/%s"  # noqa: E501
+        % (subscription, resource_group, resource_id)
+    )
+
+    params: Dict[str, Any] = {
+        "location": location,
+        "profiles": [profile],
+        "target_resource_uri": scaleset_uri,
```
tevoinea, line 95: When it comes time to enable autoscale by default, we just need to add:

```python
"enabled": True,
```

chkeita (approved):

stishkin (approved):

Issue #1614, opened 2022-01-25
Unify log analytics workspaces


ranweiler:
Right now, an instance has two Log Analytics Workspaces:
- [`logs-wb-..`](https://github.com/microsoft/onefuzz/blob/main/src/deployment/azuredeploy.json#L438-L453)
- [`<instance-name>`](https://github.com/microsoft/onefuzz/blob/main/src/deployment/azuredeploy.json#L392-L398)

One collects data that includes Azure Monitor logs and perf counters. The other gets App Insights data.

If possible, just deploy one workspace, named after the instance, which collects both. This will enable easily creating queries and workbooks that pull data from both workspaces. 

AB#39993881

PR #1679, opened 2022-02-24
Consolidating Log Analytics References & Definitions

nharper285:

_What is this about?_

From issue #1614:
Right now, an instance has two Log Analytics Workspaces:
- [`logs-wb-..`](https://github.com/microsoft/onefuzz/blob/main/src/deployment/azuredeploy.json#L438-L453)
- [`<instance-name>`](https://github.com/microsoft/onefuzz/blob/main/src/deployment/azuredeploy.json#L392-L398)

One collects data that includes Azure Monitor logs and perf counters. The other gets App Insights data.

If possible, just deploy one workspace, named after the instance, which collects both. This will enable easily creating queries and workbooks that pull data from both workspaces.

-----------------------------------------------------------------------------
The [`<instance-name>`](https://github.com/microsoft/onefuzz/blob/main/src/deployment/azuredeploy.json#L392-L398) workspace was created to query App Insights data, as AI did not yet have the ability to do so. I am removing the original reference to this workspace and renaming the monitor-based workspace `[log-wb-...]` to `[instance-name]`.

As a part of this update, VMSS OMSExtensions will now point to the consolidated `[instance-name]` log analytics workspace. Originally, the OMSExtension monitoring pointed to the `[log-wb-...]` workspace. This means that when a running instance is upgraded to include these changes, existing VMSS will initially still point to the old LA. However, timer-workers will eventually update the configs of these VMSS with the correct LA and once the nodes are upgraded and reimaged, they will correctly send VMInsights to the `[instance-name]` LA. 

If customers would like to keep the original `[log-wb-...]` workspace to retain old VMSS logs, that is reasonable.
If the `[log-wb-...]` workspace is deleted immediately then VMInsights for existing VMSS from the last 30 days will no longer be accessible. 

Regardless, log retention for all workspaces is 30 days. This means that 30 days after the last VMSS node updates to point to the `[instance-name]` workspace, there will no longer be logs left to query and the workspace can be safely deleted. 

If customers do not want to wait for the update that will point VMSS nodes to the `[instance-name]` workspace, they can force an upgrade by setting the `needs_config_update` VMSS config property to `true` in the `scaleset` table in the OneFuzz `func...` storage account and then 'upgrading' the scaleset manually. 

## PR Checklist
* [x] Applies to work item: #1614 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Changes to AzureDeploy.json workspace and app insights definitions. 

## Validation Steps Performed

* [x] Manual Deployment
* [x] Verifying Presence of relevant logs
* [ ] Check-Pr
* [x] Verifying Migration Story


src/deployment/azuredeploy.json
```diff
@@ -456,16 +410,16 @@
                 {
                     "apiVersion": "2015-11-01-preview",
                     "location": "[resourceGroup().location]",
-                    "name": "[concat('VMInsights', '(', variables('monitorAccountName'), ')')]",
+                    "name": "[concat('VMInsights', '(', parameters('name'), ')')]",
```
ranweiler, line 413: This concatenated `VMInsights(<instance-name>)` value seems to get recreated in a few places. Can we define it once, in a template `variable()`, and just reuse that?

ranweiler: It may also be worth keeping `monitorAccountName` as a variable, but just making it exactly equal to `parameters('name')`. That way it is easy to change in the future, see where it gets used, &c.

ranweiler (approved):

src/deployment/azuredeploy.json
nharper285: These are both good ideas. I've updated the PR to reflect the suggestion.

src/deployment/azuredeploy.json
```diff
@@ -49,7 +49,8 @@
     "variables": {
         "autoscale_name": "[concat('onefuzz-autoscale-', uniquestring(resourceGroup().id))]",
         "log_retention": 30,
-        "monitorAccountName": "[concat('logs-wb-', uniquestring(resourceGroup().id))]",
+        "monitorAccountName": "[parameters('name')]",
+        "vmInsightsName": "[concat('VMInsights', '(', variables('monitorAccountName'), ')')]",
```
ranweiler, line 53: 👍

src/deployment/azuredeploy.json
```diff
@@ -583,6 +538,46 @@
...
+            "apiVersion": "2020-02-02-preview",
+            "name": "[parameters('name')]",
+            "type": "microsoft.insights/components",
+            "location": "[resourceGroup().location]",
+            "kind": "",
+            "properties": {
+                "ApplicationId": "[parameters('name')]",
+                "Application_Type": "other",
+                "RetentionInDays": "[variables('log_retention')]",
+                "WorkspaceResourceId" : "[resourceId('Microsoft.OperationalInsights/workspaces', parameters('name'))]"
```
ranweiler, line 552: Should this be concatenating `variables('monitorAccountName')`? Might want to double check all these final uses.

nharper285: Ah yes thanks for catching.

And yes, I'm redeploying the updated instance to make sure everything still works properly

src/deployment/azuredeploy.json
```diff
@@ -583,6 +538,46 @@
                 }
             ]
         },
+
+        {
+            "apiVersion": "2020-02-02-preview",
+            "name": "[parameters('name')]",
```
ranweiler, line 544: This is the name of the App Insights resource, right? How about we similarly use a `variable()` here and in all the places where we use `parameters('name')` or `[resourceId('microsoft.insights/components', parameters('name'))]` to refer to it.

nharper285: @ranweiler, I ended up adding new variables for all of the resources name using parameters('name') and updating the references.

ranweiler (approved):

PR #1686, opened 2022-03-02
Abstract out node disposal

### LINKED ISSUE #1685: Create abstraction allowing us to choose node disposal strategy
We would like to be able to choose which node disposal strategy is implemented. This will make it easy for us to switch from our current delete/reimage strategy to one that also allows azure to scale us in.

tevoinea:

Introduces a new way for nodes to be reaped from a scale set. This allows azure auto scale to scale in nodes when appropriate.

## PR Checklist
* [x] Closes #1685
* [ ] Tests added/passed

## Info on Pull Request

This PR includes:
* Abstracting out how nodes are disposed (azure auto-scale scale-in or node delete/reimage)
* The capability for onefuzz to correctly handle when azure scales the nodes in
* Removal of the `resize` state that is no longer necessary
* Tuning auto scale rules according to azure guidance

## Validation Steps Performed

1. Create a new scale set with some nodes
2. Do not submit any jobs
3. Watch auto scale delete the nodes
4. Verify the size of the scale set is accurately represented in azure table
5. Submit a job
6. Watch the scale set grow and items from the pool queue being consumed
7. Repeat step 4
8. Verify the protection policy on the busy nodes is set to "Scale in actions" in the "Instances" tab of the scale set
9. Once the job is complete, watch the nodes scale back in


src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -123,23 +123,47 @@ def create_auto_scale_profile(min: int, max: int, queue_uri: str) -> AutoscalePr
...
                     # The average amount of messages there are in the pool queue
                     time_aggregation=TimeAggregationType.AVERAGE,
                     statistic=MetricStatisticType.COUNT,
-                    # Over the past 10 minutes
-                    time_window=timedelta(minutes=10),
+                    # Over the past 15 minutes
+                    time_window=timedelta(minutes=15),
                     # When there's more than 1 message in the pool queue
-                    operator=ComparisonOperationType.GREATER_THAN,
+                    operator=ComparisonOperationType.GREATER_THAN_OR_EQUAL,
```
tevoinea, line 134: This was a bug, would not spin up new nodes if there is only 1 message in the queue.

src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -123,23 +123,47 @@ def create_auto_scale_profile(min: int, max: int, queue_uri: str) -> AutoscalePr
                 metric_trigger=MetricTrigger(
                     metric_name="ApproximateMessageCount",
                     metric_resource_uri=queue_uri,
-                    # Check every minute
-                    time_grain=timedelta(minutes=1),
+                    # Check every 15 minutes
```
tevoinea, line 126: Some of these timing numbers were turned after reading this guidance on autoscaling: https://docs.microsoft.com/en-us/azure/architecture/best-practices/auto-scaling

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -542,14 +552,11 @@ def resize(self) -> None:
             self.set_shutdown(now=True)
             return
 
-        if size == self.size:
-            self._resize_equal()
-        elif self.size > size:
-            self._resize_grow()
-        else:
-            self._resize_shrink(size - self.size)
```
tevoinea, line 550: We're letting autoscale do the resizing for us. We already handle the case if additional nodes are added in azure.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -495,14 +502,17 @@ def sync_scaleset_size(self) -> None:
...
+            # Azure auto-scaled us or nodes were manually added/removed
+            # New node state will be synced in cleanup_nodes
             logging.info(
                 SCALESET_LOG_PREFIX + "unexpected scaleset size, resizing.  "
                 "scaleset_id:%s expected:%d actual:%d",
                 self.scaleset_id,
                 self.size,
                 size,
             )
-            self.set_state(ScalesetState.resize)
```
tevoinea, line 505: The resize state is no longer necessary since the onefuzz service doesn't handle scaling itself up or down.

src/api-service/__app__/onefuzzlib/azure/auto_scale.py
chkeita: Please add this link as comment to the code

src/pytypes/onefuzztypes/enums.py
```diff
@@ -411,3 +411,8 @@ class UserFieldType(Enum):
     Str = "Str"
     DictStr = "DictStr"
     ListStr = "ListStr"
+
+
+class NodeDisaposalStrategy(Enum):
+    scale_in = "scale_in"
+    reimage = "reimage"
```
chkeita, line 418: Is the reiamge strategy used anywhere? if not we should remove it

src/api-service/__app__/onefuzzlib/azure/auto_scale.py
mgreisen: These are just default values, right? We're (eventually) providing a way for the admin to customize these values? Or do we expect them to use the portal for tweeks.

tevoinea: I think the values that we should include in the cli are:
* `max`/`default` - required to create the scale set in the first place
* `min` - optional but defaults to 1
* `scale-out-amount`/`scale-out-cooldown`/`scale-in-amount`/`scale-in-cooldown` - will vary by scale set and use case so it's convenient to have it configurable. Ex: busier systems will want bigger `scale-{in|out}-amount` values, if nodes take long to set up they'll want longer cooldowns.

We can keep the current values as defaults since I think they're appropriate for a less busy deployment.

chkeita (approved):

Issue #1688, opened 2022-03-02
CI broken on clippy lint


tevoinea:
Broken CI run: https://github.com/microsoft/onefuzz/runs/5397374562?check_suite_focus=true#step:7:555

### Expected result

CI should pass

### Actual result

CI didn't pass

PR #1690, opened 2022-03-03
Clippy fix fell through the cracks

tevoinea:

_What is this about?_

## PR Checklist
* [ ] Applies to work item: #1688

### DIFF
--- a/src/ci/agent.sh
+++ b/src/ci/agent.sh
@@ -34,6 +34,12 @@ if [ X${CARGO_INCREMENTAL} == X ]; then
     cargo clean
 fi
 
+rustc --version
+cargo --version
+cargo audit --version
+cargo clippy --version
+cargo fmt --version
+
 cargo fmt -- --check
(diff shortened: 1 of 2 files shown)


src/ci/agent.sh
```diff
@@ -34,6 +34,12 @@ if [ X${CARGO_INCREMENTAL} == X ]; then
     cargo clean
 fi
 
+rustc --version
```
ranweiler, line 37: Maybe move this to just _above_ the first `cargo` invocation. That way, for example, if `cargo clean` ever failed inconsistently, we could compare tool versions between runs.

tevoinea: Do you mean the whole block of `--version` or just the rustc one?

ranweiler: I'd say all of them (particularly `cargo` per se), just to keep it simple!

chkeita (approved):

stishkin (approved):

mgreisen (approved):

PR #1721, opened 2022-03-22
Upload logs of the agent

### LINKED ISSUE #1711: Provide the ability to download the logs specific to a job
The proposal is to redirect agent logs to a blob and make it available via the cli.

## On the agent
The redirection can happen in [onefuzz-telemetry](https://github.com/microsoft/onefuzz/blob/194e7d0e9ec09fb78dae688cc549b63641a1b69e/src/agent/onefuzz-telemetry/src/lib.rs#L542) by adding a logic to upload to an [append blob](https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-append-blobs). 
We should consider splitting the blob into mutliple files to prevent the creation of single very large blob file. The `x-ms-blob-append-offset` property of the [Append block](https://docs.microsoft.com/en-us/rest/api/storageservices/append-block) can provide the data needed to decide to create a new log file. 

## In the cli
cli work will be done in #1714

chkeita:
Agent side changes to support uploads of logs. (continuation of #1727)

- We create a `TaskLogger` that receive messages broadcasted by the onefuzz-telemtry component. 
- The task logger accumulate a  messages before uploading them to a blob 
- The blob is split as soon as the maximum size configured is reached.

closes #1711


src/api-service/__app__/jobs/__init__.py
```diff
@@ -53,6 +53,31 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if isinstance(user_info, Error):
         return not_ok(user_info, context="jobs create")
 
+    # create the job logs container
+    log_container_sas = create_container(
+        Container("logs"),
```
stishkin, line 58: was this supposed to be per job ?

stishkin: i mean should this include "job id" as well ?

stishkin (approved):

src/agent/onefuzz-agent/src/tasks/config.rs
```diff
@@ -204,6 +206,8 @@ impl Config {
         telemetry::set_property(EventData::Version(env!("ONEFUZZ_VERSION").to_string()));
         telemetry::set_property(EventData::InstanceId(self.common().instance_id));
         telemetry::set_property(EventData::Role(Role::Agent));
+        telemetry::set_property(EventData::Role(Role::Agent));
```
tevoinea, line 209: Was this line accidentally duplicated?

src/agent/onefuzz-agent/src/tasks/task_logger.rs
```diff
@@ -0,0 +1,604 @@
...
+                println!(" {:?}", v);
+            }
+        }
+
+        assert_eq!(x.keys().len(), 3, "expected 3 groups of messages");
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[ignore]
```
tevoinea, line 533: Are these tests marked with `#[ignore]` intended to be run manually by specifying the env var?

tevoinea (approved):

src/agent/onefuzz-agent/src/tasks/config.rs
chkeita: yes

src/agent/onefuzz-agent/src/tasks/task_logger.rs
chkeita: correct.

PR #1727, opened 2022-03-28
Log redirection, service side

chkeita:
This PR sets  up the service side of the log redirection work #1711 

- A log container is created or reused when e create a job
- When scheduling the task we send the log location to the agent
The expected log structure looks lieke
`{fuzzContainer}/logs/{job_id}/{task_id}/{machine_id}/1.log`

the agent side of the work will be done in #1721


src/api-service/__app__/jobs/__init__.py
```diff
@@ -50,6 +53,31 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if isinstance(user_info, Error):
         return not_ok(user_info, context="jobs create")
 
+    # create the job logs container
+    log_container_sas = create_container(
```
stishkin, line 57: Does this mean there in only one container named logs and it is shared across all users and all jobs ?

chkeita: yes one container per storage account. create container will randomly assign the storage account if you have more than one

stishkin: does it mean that multiple jobs from multiple users will share same container for their logs ? 

if that's the case - is there  a reason not to separate at least logs from each job into a separate container ?

chkeita: yes multiple job will share the same container. 
No particular reason to not separate.

stishkin (approved):

stishkin:
One more - what's the container deletion policy ? 

Can any user delete "logs" container ? If so won't the delete other user's logs ?

src/api-service/__app__/jobs/__init__.py
chkeita: updated to add the job_id as part of the container name
@tevoinea FYI This will change the log paths to 
`{fuzzContainer}/logs-{job_id}/{task_id}/{machine_id}/1.log`

chkeita:
> One more - what's the container deletion policy ?
> 
> Can any user delete "logs" container ? If so won't the delete other user's logs ?
@stishkin 
That is true for all container currently today. any body can delete all containers.

tevoinea (approved):

docs/webhook_events
```diff
@@ -0,0 +1,6857 @@
+# Webhook Events
```
tevoinea, line 1: I think this file was accidentally generated (it's missing the `.md` extension)

PR #1728, opened 2022-03-28
Update Cobertura Output

hayleycall:

Modified Cobertura output to include line-rates, covered and valid to have the summary tab show up in the Publish Cde Coverage task in the pipeline. 

Modified Cobertura output to have file directory as package name and only file name as class name for every file. This creates easier to digest output in pipeline.

## PR Checklist
* [X] Tests added/passed

## Info on Pull Request

Modified Cobertura output to include line-rates, covered and valid to have the summary tab show up in the Publish Cde Coverage task in the pipeline. 

Modified Cobertura output to have file directory as package name and only file name as class name for every file. This creates easier to digest output in pipeline.

Includes only changes to Cobertura format.

Refactored Cobertura conversion code to be more readable/debuggable and included additional tests.

## Validation Steps Performed

Test Updated


src/agent/coverage/src/cobertura.rs
```diff
@@ -176,17 +214,28 @@ mod tests {
...
+                .attr("branch-rate", "0")
+                .attr("complexity", "0"),
+        )?;
+
+        _emitter_test.write(XmlEvent::start_element("classes"))?;
 
         _emitter_test.write(
             XmlEvent::start_element("class")
-                .attr("name", "0")
+                .attr("name", "C:/Users/file2.txt")
```
stishkin, line 236: Does this mean that this will not going to work on Linux ?

hayleycall: Good point. This test would only work on Windows. WIll Fix.

src/agent/coverage/src/cobertura.rs
```diff
@@ -18,12 +19,31 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
         .context("system time before unix epoch")?
         .as_secs();
 
+    // compute line rate, lines-covered, lines-valid by summing total lines and summing lines that have at least 1 hit, will do this for overall, and for each file (package and class in this case)
+    let mut total_valid_lines = 0;
```
ranweiler, line 23: This block, and the one below, are doing something very similar. Can you factor this duplicate logic out into a shared function?

src/agent/coverage/src/cobertura.rs
```diff
@@ -18,12 +19,31 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
+    let mut total_hit_lines = 0;
+    let mut total_line_rate = 0_f32;
+    let copy_source_coverage: SourceCoverage = source_coverage.clone();
+    let coverage_files: Vec<SourceFileCoverage> = source_coverage.files;
```
ranweiler, line 27: Did you find this type ascription necessary?

src/agent/coverage/src/cobertura.rs
```diff
@@ -18,12 +19,31 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
+    for file in coverage_files {
+        let locations: Vec<SourceCoverageLocation> = file.locations;
+        for location in locations {
+            total_valid_lines += 1;
+            if &location.count > &0 {
+                total_hit_lines += 1;
+            }
+        }
+    }
+    if total_valid_lines > 0 {
```
ranweiler, line 37: 👍

src/agent/coverage/src/cobertura.rs
```diff
@@ -32,30 +52,47 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
+            if &location.count > &0 {
+                package_hit_lines += 1;
+            }
+        }
+        if package_valid_lines > 0 {
+            package_line_rate = package_hit_lines as f32 / package_valid_lines as f32;
+        }
+        let full_path_file = Path::new(&file.file);
+        let path = full_path_file.parent().unwrap();
+        emitter.write(
```
ranweiler, line 74: Ideally, the data aggregation (line/hit/rate computation) and markup codegen (XML element emitting) will be really clearly separated. If you factor out the aggregation that I mentioned, you could perhaps introduce a new struct return value that would make it easy to do this. The resulting code could be a lot easier to understand or change in the future!

src/agent/coverage/src/cobertura.rs
```diff
@@ -154,9 +192,9 @@ mod tests {
 
         _emitter_test.write(
             XmlEvent::start_element("class")
-                .attr("name", "0")
+                .attr("name", "C:/Users/file1.txt")
```
ranweiler, line 195: This path is a funny mix of Windows and POSIX-style: it has a letter drive for the root, but uses forward slashes!

It makes me wonder: does the XML library we're using handle string escaping? What happens if our inputs are classic Windows paths like `C:\Users\file.txt`?

src/agent/coverage/src/cobertura.rs
```diff
@@ -195,20 +244,22 @@ mod tests {
...
 
+        _emitter_test.write(XmlEvent::end_element())?; // line
         _emitter_test.write(XmlEvent::end_element())?; // lines
         _emitter_test.write(XmlEvent::end_element())?; // class
         _emitter_test.write(XmlEvent::end_element())?; // classes
         _emitter_test.write(XmlEvent::end_element())?; // package
         _emitter_test.write(XmlEvent::end_element())?; // packages
         _emitter_test.write(XmlEvent::end_element())?; // coverage
 
+        //println!("{}", source_coverage_result?);
```
ranweiler, line 260: We should be sure to prune stray / commented dev code, like here and below.

src/agent/coverage/src/cobertura.rs
```diff
@@ -32,30 +52,47 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
+            package_valid_lines += 1;
+            if &location.count > &0 {
+                package_hit_lines += 1;
+            }
+        }
+        if package_valid_lines > 0 {
+            package_line_rate = package_hit_lines as f32 / package_valid_lines as f32;
+        }
+        let full_path_file = Path::new(&file.file);
+        let path = full_path_file.parent().unwrap();
```
ranweiler, line 73: We never want to `unwrap()`/`expect()` unless a panic is _impossible_ due to an earlier check or upheld precondition. When this is true, we need to document the reason.

src/agent/coverage/src/cobertura.rs
hayleycall: Opps, done!

src/agent/coverage/src/cobertura.rs
hayleycall: No, resolved.

src/agent/coverage/src/cobertura.rs
hayleycall: It does handle, string escaping, i.e. C:\\Users

src/agent/coverage/src/cobertura.rs
hayleycall: Seperated out into a couple different functions, wasn't able to completely unduplicate.

src/agent/coverage/src/cobertura.rs
hayleycall: Added in a panic message, is this enough in terms of error checking/messaging?

src/agent/coverage/src/cobertura.rs
hayleycall: Split out into a couple different functions, wasn't able to completely unduplicate.

src/agent/coverage/src/cobertura.rs
```diff
@@ -1,12 +1,56 @@
...
-use crate::source::SourceCoverageLocation;
 use crate::source::SourceFileCoverage;
 use anyhow::Context;
 use anyhow::Error;
 use anyhow::Result;
+use std::path::Path;
 use std::time::{SystemTime, UNIX_EPOCH};
 use xml::writer::{EmitterConfig, XmlEvent};
 
+pub fn compute_line_rate(line_values: Vec<u32>) -> f32 {
```
ranweiler, line 10: Let's use this function to work through a few Rust things.

1. This function will panic if `line_values` has less than 2 elements. We never want to panic in our code, so we should only call functions that can panic if we're sure that the panic case is unreachable for all inputs. In this case, the (notation-hidden) function is [`Index::index()`](https://doc.rust-lang.org/std/ops/trait.Index.html#tymethod.index). However, this gets fixed if we address (2) below.
2. These indices actually have a _fixed_ meaning (count of lines hit, valid). When that's true, a function like this should either take 2 arguments, or take 1 argument that is a meaning-encoding struct. For example, something like `struct LineCounts { hit: u32, count: u32 }`. In this case, I think 2 args is probably more simple and straightforward.
3. We typically don't want to take ownership over a non-`Copy` value unless we need to. If we kept the existing signature, we'd probably want this argument to have type `line_values: &[u32]`. We'd want a `&[u32]` instead of a `&Vec<u32>` because the only `Vec` methods we're using actually come from `Deref`-ing through to slices anyway. It almost never makes sense to pass a `&Vec<T>`, because the most `Vec`-specific methods require a _mutable_ reference to the `Vec`. An example of an exception: when you might want to clone the `Vec` itself inside the function.
4. Since we only support 64-bit, our machine word size is 64 bits, and there's no semantic use in limiting these computations to `u32`/`f32`. I suggest thinking of the "default" numeric types for our purposes as `u64`/`i64`/`f64`. Especially when counting things in Rust, the "usual" numeric type is `usize` (which is also used for indexing), but here I'd suggest an explicit `u64`. If the Cobertura format only supports up to 32-bit numbers, then your current choice is the right one, but we should document it!

src/agent/coverage/src/cobertura.rs
```diff
@@ -1,12 +1,56 @@
...
+    let mut line_rate = 0_f32;
+    if line_values[1] > 0 {
+        line_rate = line_values[1] as f32 / line_values[0] as f32;
+    }
+    line_rate
+}
+
+pub fn compute_line_values_coverage(files: Vec<SourceFileCoverage>) -> (Vec<u32>, f32) {
```
ranweiler, line 18: Apply earlier comment(3) about not passing owned values unless we need to.

Here, we probably want `files: &[SourceFileCoverage]`.

src/agent/coverage/src/cobertura.rs
```diff
@@ -1,12 +1,56 @@
...
+            valid_lines += 1;
+            if location.count > 0 {
+                hit_lines += 1;
+            }
+        }
+    }
+    line_values.push(valid_lines);
+    line_values.push(hit_lines);
+    let line_rate = compute_line_rate(line_values.clone());
+    (line_values, line_rate)
```
ranweiler, line 34: In Rust, we can create types to help us out. They can also make code a little more self-documenting.

When we read a type like `(Vec<u32>, f32)`, you have to read where it came from to know what it means.

On the other hand, suppose we defined a struct like:
```rust
struct LineCoverage {
    pub hit: u64,
    pub valid: u64,
    pub rate: f64,
}

impl LineCoverage {
    pub fn new(hit: u64, valid: u64) -> Self {
        let rate = if valid == 0 {
            0.0
        } else {
            (hit as f64) / (valid as f64)
        };

        Self { hit, valid, rate }
    }
}
```
Now we know what everything means just from , and the rate calculation is encapsulated in the type construction.

src/agent/coverage/src/cobertura.rs
ranweiler: Here's the core question: if the path is a real Windows path, with backslashes, is the resulting XML valid? How about other XML special characters (`<`, `>`, `&`, `'`, `"`)?

If not, do we need to manually invoke an XML escaping function?

src/agent/coverage/src/cobertura.rs
ranweiler: I don't think so, please revisit my first comment.

Think about what this code does right now: if any input has a filename that lacks a parent (for example, `/foo.c`), this code will intentionally crash (with a panic). That means any _other_ code that calls it will crash, too!

The goal is not panicking is to avoid this situation. There are situations where it can make sense to have some code panic on an invalid input, but it needs to be really clearly documented. In this case, not only do we not want to panic, but we probably want to _recover_ and continue gracefully in some way that makes sense for the problem being solved.

src/agent/coverage/src/cobertura.rs
hayleycall: Makes, sense, fixed for this function and all 32-bit types changed to 64.

src/agent/coverage/src/cobertura.rs
hayleycall: Created new type to have line value information.

src/agent/coverage/src/cobertura.rs
hayleycall: Removed this function and replaced with line values type that computes line rate, as suggested in other comment.

src/agent/coverage/src/cobertura.rs
hayleycall: Changed to pass in slice rather than vector.

src/agent/coverage/src/cobertura.rs
hayleycall: Completed, testing full XML output in pipeline to see how invalid message shows up (if it is very visible).

src/agent/coverage/src/cobertura.rs
```diff
@@ -32,30 +90,43 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
-            .attr("complexity", "0"),
-    )?;
+    // path (excluding file name) will be package name for better results with ReportGenerator
+    // class name will be full file path and name
+    for file in &source_coverage.files {
+        let package_line_values = compute_line_values_package(file);
+        let path = Path::new(&file.file);
+        let _file_check = path.file_name();
+        let none_message = "Invalid file format: ".to_owned() + &file.file;
+        let parent_path = match _file_check {
```
ranweiler, line 100: This is all great! I suggest factoring the logic that computes `parent_path` (or its falllback) into its own function.

src/agent/coverage/src/cobertura.rs
```diff
@@ -32,30 +90,43 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
-            .attr("line-rate", "0")
-            .attr("branch-rate", "0")
-            .attr("complexity", "0"),
-    )?;
+    // path (excluding file name) will be package name for better results with ReportGenerator
+    // class name will be full file path and name
+    for file in &source_coverage.files {
+        let package_line_values = compute_line_values_package(file);
+        let path = Path::new(&file.file);
+        let _file_check = path.file_name();
```
ranweiler, line 98: FYI re Rust conventions: we usually only prefix a var name with `_` if you are explicitly ignoring it and not using it, but it's still worth naming. An example of doing this is a drop guard. Since in the current impl, you `match` on `_file_check`, it'd make more sense to name it `file_check` (or probably better, `file_name`), or even just `match` directly on `path.file_name()`.

src/agent/coverage/src/cobertura.rs
```diff
@@ -32,30 +91,47 @@ pub fn cobertura(source_coverage: SourceCoverage) -> Result<String, Error> {
...
-            .attr("name", "0")
-            .attr("line-rate", "0")
-            .attr("branch-rate", "0")
-            .attr("complexity", "0"),
-    )?;
+    // path (excluding file name) will be package name for better results with ReportGenerator
+    // class name will be full file path and name
+    for file in &source_coverage.files {
+        let package_line_values = compute_line_values_package(file);
+        let path = if cfg!(windows) {
```
ranweiler, line 98: This code checks for the `windows` build config option, documented [here](https://doc.rust-lang.org/reference/conditional-compilation.html#unix-and-windows).

This is set when the build option `target_family = "windows"`, which is true when your build target is Windows.

But this approach means that the code here will only work for Windows paths on Windows, and POSIX paths on non-Windows build targets.

We should try for a fix that, no matter where it is being run _from_ (no matter the build target), explicitly makes a decision of the form "my input data uses $OS paths, and so I will analyze their components accordingly".

src/agent/coverage/src/cobertura.rs
hayleycall: Done!

src/agent/coverage/src/cobertura.rs
hayleycall: Converted all paths in standard POSIX paths with path-slash crate.

src/agent/coverage/src/cobertura.rs
hayleycall: Changed to directly match on path.file_name()!

src/agent/coverage/src/cobertura.rs
```diff
@@ -1,12 +1,79 @@
...
+    let mut hit_lines = 0;
+    for location in &file.locations {
+        valid_lines += 1;
+        if location.count > 0 {
+            hit_lines += 1;
+        }
+    }
+    LineValues::new(valid_lines, hit_lines)
+}
+// handle OSs and path formats accordingly
```
ranweiler, line 58:
```suggestion

// handle OSs and path formats accordingly
```

src/agent/coverage/src/cobertura.rs
```diff
@@ -1,12 +1,79 @@
...
+
+// get directory of file if valid file path, otherwise make package name include and error message
+pub fn get_parent_path(path_slash: &str) -> PathBuf {
+    let path = Path::new(&path_slash);
+    let none_message = "Invalid file format: ".to_owned() + path_slash;
+    let parent_path = match path.file_name() {
+        Some(_parent_path) => path.parent().unwrap(),
+        None => Path::new(&none_message),
+    };
+    (parent_path).to_path_buf()
```
ranweiler, line 75:
```suggestion
    parent_path.to_path_buf()
```

src/agent/coverage/src/cobertura.rs
```diff
@@ -176,17 +286,259 @@ mod tests {
...
+        _emitter_test.write(XmlEvent::end_element())?; // package
+        _emitter_test.write(XmlEvent::end_element())?; // packages
+        _emitter_test.write(XmlEvent::end_element())?; // coverage
+
+        assert_eq!(source_coverage_result?, String::from_utf8(backing_test)?);
+        Ok(())
+    }
+
+    #[test]
+    fn test_source_to_cobertura_posix_paths() -> Result<()> {
```
ranweiler, line 405: 👍

ranweiler (approved):
Looking great!

As discussed offline, the last thing I'd love to see in this: right now, in some tests, we currently "manually" dynamically generate the _expected_ XML in-memory, whenever we run the test, and then compare it to the output of the `cobertura()` function.

Instead, we can commit hardcoded XML files for the expected output. We should include cases for Windows, POSIX, and mixed paths. Then use `include_bytes!()`/`include_str!()` to expose those file contents as `&[u8]`/`&str` constants in the `tests` module. The test assertions can then just compare the `cobertura()` function's output against those expected XML files, as strings.

This lets us avoid depending on the XML emitter in the test functions, shortens the test bodies, and lets future readers see a full, human-friendly representation of "good output", without any tricks or extra steps.

src/agent/coverage/src/cobertura.rs
hayleycall: This works as expected. But after some discussion this may not be the best approach.

PR #1737, opened 2022-04-01
Logging

stishkin:
Add logging to app insights and console (logs include correlation id)

### DIFF
--- a/src/ApiService/ApiService/Log.cs
+++ b/src/ApiService/ApiService/Log.cs
@@ -1,0 +1,172 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.DataContracts;
+
+using System.Diagnostics;
+
+namespace Microsoft.OneFuzz.Service;
+
+public interface ILog {
+    void Log(Guid correlationId, String message, SeverityLevel level, IDictionary<string, string> tags, string? caller);
+    void LogEvent(Guid correlationId, String evt, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller);
+    void LogException(Guid correlationId, Exception ex, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller);
+    void Flush();
+}
+
+class AppInsights : ILog {
+    private TelemetryClient telemetryClient =
+            new TelemetryClient(
+                new TelemetryConfiguration(EnvironmentVariables.AppInsights.InstrumentationKey));
+
+    public void Log(Guid correlationId, String message, SeverityLevel level, IDictionary<string, string> tags, string? caller) {
+        tags.Add("Correlation ID", correlationId.ToString());
+        if (caller is not null) tags.Add("CalledBy", caller);
+        telemetryClient.TrackTrace(message, level, tags);
+    }
+    public void LogEvent(Guid correlationId, String evt, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller) {
+        tags.Add("Correlation ID", correlationId.ToString());
+        if (caller is not null) tags.Add("CalledBy", caller);
+        telemetryClient.TrackEvent(evt, properties: tags, metrics: metrics);
+    }
+    public void LogException(Guid correlationId, Exception ex, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller) {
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/Log.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.DataContracts;
+
+using System.Diagnostics;
+
+namespace Microsoft.OneFuzz.Service;
+
+public interface ILog {
```
chkeita, line 11: why create a new logging interface when ILogger already exist ?

tevoinea (approved):

src/ApiService/ApiService/Log.cs
```diff
@@ -0,0 +1,172 @@
...
+    void Log(Guid correlationId, String message, SeverityLevel level, IDictionary<string, string> tags, string? caller);
+    void LogEvent(Guid correlationId, String evt, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller);
+    void LogException(Guid correlationId, Exception ex, IDictionary<string, string> tags, IDictionary<string, double>? metrics, string? caller);
+    void Flush();
+}
+
+class AppInsights : ILog {
+    private TelemetryClient telemetryClient =
+            new TelemetryClient(
+                new TelemetryConfiguration(EnvironmentVariables.AppInsights.InstrumentationKey));
```
chkeita, line 21: can we use a contructor paramter instead of reading from an environment variable here. This makes the code easier to test

src/ApiService/ApiService/Log.cs
stishkin: because ILogger does not have methods that I'd like to have.

No metrics logging, no Event logging. You cannot pass metrics with Exception if you have any.

src/ApiService/ApiService/Log.cs
stishkin: Microsoft.OneFuzz.Service.EnvironmentVariables.LogDestinations <- [| Microsoft.OneFuzz.Service.LogDestination.Console |]

let log = Microsoft.OneFuzz.Service.Log(Guid.NewGuid())
log.Tags.Add("Tag1", "V1")
log.Tags.Add("Tag2", "V2")

log.Info("blah")

src/ApiService/ApiService/Log.cs
```diff
@@ -0,0 +1,172 @@
...
+    public void Flush() {
+        System.Console.Out.Flush();
+    }
+}
+
+public class Log {
+
+    static List<ILog> loggers;
+
+    static Log() {
```
chkeita, line 98: Can we have the logging be an instance method. And use dependency injection to pass around the current logger.

src/ApiService/ApiService/Log.cs
stishkin: just pre-set destination before calling your test.

Passing through constructor - means that every time you create logger object in azure function that is configured to do APpInsights, you have to pass AppInsights in the constructor.


This Log object meant to be an execution flow context type of thing, that tracks correlation ID and tags

src/ApiService/ApiService/Log.cs
stishkin: probably, 

we can look into this later. so far there is no code that dependent on this, so nothing will be broken with your proposed change

src/ApiService/ApiService/Log.cs
chkeita: > Microsoft.OneFuzz.Service.EnvironmentVariables.LogDestinations <- [| Microsoft.OneFuzz.Service.LogDestination.Console |]

This make it hard to have two different instances of loggers 


> Passing through constructor - means that every time you create logger object in azure function that is configured to do APpInsights, you have to pass AppInsights in the constructor.

No you can create the logger once when you start the application then pass it around. That is the whole point of dependency injection. If you look at `QueueNodeHearbeat` you will see that it received an ilogger

stishkin: 1) Why do we need 2 loggers writing to different destinations ? 

2) we need to track correlation ID across the flow of calls, and tags. 
So we know in which order which function got called.

src/ApiService/ApiService/Log.cs
chkeita: can we extend iLogger then ?

chkeita (approved):

PR #1738, opened 2022-04-02
Setting up the ORM

chkeita:
Adds a converter between C# record and [TableEntity](https://docs.microsoft.com/en-us/dotnet/api/azure.data.tables.tableentity?view=azure-dotnet) with the following feature
- Renaming PascalCase property to snake_case column
- Renaming PascalCase enum to snake_case column
- Renaming via `JsonPropertyName` attribute
- Attribute for tagging PartitionKey and RowKey


src/ApiService/ApiService/onefuzzlib/CustomConverterFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    public sealed class CustomEnumConverter<T> : JsonConverter<T> where T : Enum
+    {
+        private readonly JsonNamingPolicy _namingPolicy;
+
+        private readonly Dictionary<string, T> _readCache = new();
+        private readonly Dictionary<T, JsonEncodedText> _writeCache = new();
+
+        // This converter will only support up to 64 enum values (including flags) on serialization and deserialization
+        private const int NameCacheLimit = 64;
```
stishkin, line 41: 64 per enum ? or 64 total values from all enums ?

src/ApiService/ApiService/ApiService.csproj
```diff
@@ -7,6 +7,7 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="CaseExtensions" Version="1.1.0" />
```
stishkin, line 10: let's hope there is no supply chain attack hiding here...

src/ApiService/ApiService/onefuzzlib/CustomConverterFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        private const string ValueSeparator = ", ";
```
stishkin, line 43: can you provide an example of how this is used ?

src/ApiService/ApiService/onefuzzlib/CustomConverterFactory.cs
chkeita: 64 per enum

src/ApiService/ApiService/ApiService.csproj
chkeita: removed

src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -0,0 +1,267 @@
...
+    public interface IStorageProvider
+    {
+        Task<TableServiceClient> GetStorageClient(string? table, string? accounId);
+    }
+
+    public class StorageProvider : IStorageProvider
+    {
+        public async Task<TableServiceClient> GetStorageClient(string? table, string? accounId)
+        {
+            accounId ??= System.Environment.GetEnvironmentVariable("ONEFUZZ_FUNC_STORAGE");
```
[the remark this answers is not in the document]
chkeita: removing it for now, i will re introduce it in a later PR

src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -0,0 +1,265 @@
...
+                }
+
+                return new EntityProperty(f.Name, dbName, f.ParameterType, kind);
+            }).ToArray();
+
+            return new EntityInfo(typeof(T), parameters, BuildConstructerFrom(constructor));
+        });
+    }
+
+    public TableEntity ToTableEntity<T>(T typedEntity)
```
stishkin, line 112: https://github.com/stishkin/onefuzz/blob/44da180e2646d1f10042cfd345b2d45c30e0d98b/src/api-service/__app__/onefuzzlib/orm.py#L301

Python Models have some "filters" specified to make sure that some fields are not stored in the tables. 

Is there plan to add that in next pr ?  Since that will require handling of Ignore attribute (not just Name attrtibute)

src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -0,0 +1,265 @@
...
+            throw new NullReferenceException();
+        }
+        var tableEntity = new TableEntity();
+        var entityInfo = GetEntityInfo<T>();
+        foreach (var prop in entityInfo.properties)
+        {
+            var value = entityInfo.type.GetProperty(prop.name)?.GetValue(typedEntity);
+            if (prop.type == typeof(Guid) || prop.type == typeof(Guid?))
+            {
+                tableEntity.Add(prop.dbName, value?.ToString());
```
stishkin, line 130: does dbName means name inside of the database ? Like a ColumnName ? 

Also should this account for Property Name attribute to allow us rename properties ?

src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
chkeita: no needed in our case here. The code looks at the parameter of the record constructor and maps it to the db. 
`excludes` where used for data that was added after the fact. for those we can use the property syntax of the record.
``` C#
/// the record constructor parameters will save in the db
record MyEntity([PartitonKey] string name, [RowKey] int id)
{
        /// this property will not be saved in the db
	public string AdditionalProperty {get; set;}
}
```

stishkin: so that's a convention we need to remember

src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
chkeita: yep columnName is a better name. 
Yes we support renaming, see `GetEntityInfo`

stishkin (approved):

PR #1739, opened 2022-04-04
Preserve existing outputs patch

puhley:

   When performing a libfuzzer merge operation, the Rust application code would return a `missing field` error that was generated by Serde when it deserialized the message from the Azure Blob Queue. Based on the output from the `onefuzz jobs tasks list` command line, it was clear that the flag was being sent to the API service. However, when monitoring the messages in the Azure Blob Queue, it was apparent that the flag was not being added to the workunits.config portion of the message. 
   
 ```
 {"reboot": false,
 "setup_url": "https://....", 
"script": false, 
"work_units": [
    {"job_id": "ab31f19b-1234-1234-1234-12348c44ccb2", 
     "task_id": "b51b18b1-1234-1234-1234-1234108d5f0b",
      "task_type": "libfuzzer_merge", 
       "config": "{\ *** flag should exist here ***}",
     ...}]
}
```

   This change ensures that the flag is always present in the queue message by adding it as a feature within the `libfuzzer_merge` task definition within `defs.py`. In addition, the `build_task_config` method in `config.py` was updated to check whether `preserve_existing_outputs` is a feature for that task, and adds the flag if present.
   
   There is also a change to `models.py` in `onefuzztypes` to add `preserve_existing_outputs` as an optional flag just because it seemed to be missing. This can be removed from the pull request if it is an issue.

## PR Checklist
* [X] Applies to work item: #1672
* [ ] Tests added/passed
* [X] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #1672 

## Info on Pull Request

This pull requests includes two changes to enable `preserve_existing_outputs`. The first is to add `preserve_existing_outputs` as a task feature to `libfuzzer_merge` in `defs.py`. The second is to check for that feature when creating a task in `config.py`.

This pull request also includes an update to the tasks model adding `preserve_existing_outputs` as an optional flag.

## Validation Steps Performed

To validate this change, run a libfuzzer_merge command twice. The first execution should include the `preserve_existing_outputs` flag to confirm the flag is acknowledged when set. The second execution should omit the flag to test the condition when the setting defaults to the false value.

A valid libfuzzer_merge command would be:

`onefuzz template libfuzzer merge your_project your_name your_build your_pool --existing_inputs your_input_container --output_container your_output_container --target_exe your_libfuzzer_exe --preserve_existing_outputs`

This command should start the job successfully. When running `onefuzz jobs tasks list your_job_id`, you should not see any `missing field` errors in the output. Instead, you should see the details for an executing job.

### CODE REVIEW

File: src/api-service/__app__/onefuzzlib/tasks/config.py [hunk 1]
```diff
@@ -411,6 +411,13 @@ def build_task_config(
             else True
         )
 
+    if TaskFeature.preserve_existing_outputs in definition.features:
+        config.preserve_existing_outputs = (
+            task_config.task.preserve_existing_outputs
+            if task_config.task.preserve_existing_outputs is not None
+            else True
```
ranweiler, line 418: Is this the right default? For example, on the CLI side, we [default to `False`.](https://github.com/microsoft/onefuzz/blob/main/src/cli/onefuzz/templates/libfuzzer.py#L399)

@chkeita, thoughts?

puhley: It is the wrong default. I copied the code for `check_fuzzer_help` {which defaults to True} and forgot to change the default value.

puhley: I've updated the pull request.

ranweiler:
Running integration tests.

ranweiler (approved):

raghavthind2005 (approved):

puhley:
Let me know if there is any more information required for this pull request.

stishkin:
No longer relevant, since Python is not the language of choice anymore.

This on the other hand still needs to be ported to C# code-base

PR #1750, opened 2022-04-05
Add dotnet project to the Continuous integration

chkeita:


mgreisen (approved):

src/ApiService/ApiService/UserCredentials.cs
```diff
@@ -51,12 +51,12 @@ public class UserCredentials {
     }
 
 
-    static async Task<OneFuzzResult<string[]>> GetAllowedTenants() {
-        return OneFuzzResult<string[]>.Ok(Array.Empty<string>());
+    static Task<OneFuzzResult<string[]>> GetAllowedTenants() {
```
stishkin, line 54: this is not a proper implementation, it's just a stub. I'm waiting for Instance Config model to be available

src/ApiService/Tests/Tests.csproj
```diff
@@ -2,8 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
```
stishkin, line 5: this is supposed to be a compiler flag to help identify potential null issues
is it removed because it's not helping ?

stishkin: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis

src/ApiService/ApiService/QueueNodeHearbeat.cs
```diff
@@ -25,7 +25,7 @@ public QueueNodeHearbeat(ILoggerFactory loggerFactory, IStorageProvider storageP
     [Function("QueueNodeHearbeat")]
     public async Task Run([QueueTrigger("myqueue-items", Connection = "AzureWebJobsStorage")] string msg)
     {
-        var hb = JsonSerializer.Deserialize<NodeHeartbeatEntry>(msg, EntityConverter.GetJsonSerializerOptions());
+        var hb = JsonSerializer.Deserialize<NodeHeartbeatEntry>(msg, EntityConverter.GetJsonSerializerOptions()).EnsureNotNull($"wrong data {msg}");
```
stishkin, line 28: Use this ? 

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator

src/ApiService/ApiService/UserCredentials.cs
chkeita: I just changed it  because it was causing a warning since you have an async function with no aait

src/ApiService/Tests/Tests.csproj
chkeita: I know i just removed it only for the tests project because is it was causing a lot of warning.  So i removed it instead of having a lot of null check that would make the tests less readable.

src/ApiService/ApiService/QueueNodeHearbeat.cs
chkeita: no the same thing. I want to convert nullable to a non nullable and throw if it is not the case. 
with the operator the code will look like 
``` C#
var hb = JsonSerializer.Deserialize<NodeHeartbeatEntry>(msg, EntityConverter.GetJsonSerializerOptions()) ?? throw new NullReferenceException($"wrong data {msg}");
```
The extension method is just a shortcut

src/ApiService/Tests/Tests.csproj
chkeita: brought it back https://github.com/microsoft/onefuzz/pull/1753

PR #1762, opened 2022-04-07
Add formatting step to the build And apply default formatting

chkeita:
Check the formatting in the build and apply default formatting to the C# solution using `dotnet format` 
Only formatting is applies **No other code changes**


tevoinea (approved):
Is there a way to auto format on local build? Or maybe it already does that?

chkeita:
@tevoinea i could not find any option to do it on the build. but you can run `dotnet format` from the command line

PR #1765, opened 2022-04-08
Finishing implementation of SendEvent

chkeita:
- Addin more injected types
- Fixing an issue with the log tracer where the same keys was added multiple time to the tags
- Removing UserInfo when writing events
- other bug fixes

### DIFF
--- a/src/ApiService/ApiService/Program.cs
+++ b/src/ApiService/ApiService/Program.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OneFuzz.Service.OneFuzzLib.Orm;
+using ApiService.OneFuzzLib;
 
 namespace Microsoft.OneFuzz.Service;
 
@@ -35,6 +36,10 @@ public class Program
             .AddSingleton<ILogTracerFactory>(_ => new LogTracerFactory(GetLoggers()))
             .AddScoped<ILogTracer>(s => s.GetService<LogTracerFactory>()?.MakeLogTracer(Guid.NewGuid()) ?? throw new InvalidOperationException("Unable to create a logger"))
             .AddSingleton<IStorageProvider>(_ => new StorageProvider(EnvironmentVariables.OneFuzz.FuncStorage ?? throw new InvalidOperationException("Missing account id")))
+            .AddSingleton<INodeOperations, NodeOperations>()
+            .AddSingleton<IEvents, Events>()
+            .AddSingleton<IWebhookOperations, WebhookOperations>()
+            .AddSingleton<IQueue, Queue>()
             .AddSingleton<ICreds>(_ => new Creds())
             .AddSingleton<IStorage, Storage>()
         )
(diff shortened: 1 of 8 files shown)


src/ApiService/ApiService/onefuzzlib/Events.cs
```diff
@@ -60,15 +64,42 @@ public async Task SendEvent(BaseEvent anEvent)
...
         public void LogEvent(BaseEvent anEvent, EventType eventType)
         {
-            //todo
-            //var scrubedEvent = FilterEvent(anEvent);
-            //throw new NotImplementedException();
+            var options = EntityConverter.GetJsonSerializerOptions();
+            options.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
+            options.Converters.Add(new RemoveUserInfo());
+            var serializedEvent = JsonSerializer.Serialize(anEvent, options);
+            _logger.LogInformation($"sending event: {eventType} - {serializedEvent}");
```
stishkin, line 71: using the built-in logger does not log info that we need to have. Correlation ID is not there, custom tags are not there. There is a ton of useful info that we should have and we cannot have due to way built in logger works..

src/ApiService/ApiService/onefuzzlib/Queue.cs
```diff
@@ -16,12 +17,12 @@ public interface IQueue
 public class Queue : IQueue
 {
     IStorage _storage;
-    ILog _logger;
+    ILogger _logger;
 
-    public Queue(IStorage storage, ILog logger)
+    public Queue(IStorage storage, ILoggerFactory loggerFactory)
```
stishkin, line 22: here it should be ILogTraceFactory. 

The factory is set as a member in the object. When a new function triggered on the object - create a new Log intance from the factory and set CorrelationId on it (we could move this into middleware, I don't know how dependency injection will work with middleware though...)

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/Events.cs
chkeita: The version with the LogTracer was not logging when i ran the code locally so reverted that change. 
will bring it back when that the local scenario works

src/ApiService/ApiService/Program.cs
```diff
@@ -33,8 +34,12 @@ public static void Main()
         .ConfigureServices((context, services) =>
             services
             .AddSingleton<ILogTracerFactory>(_ => new LogTracerFactory(GetLoggers()))
-            .AddScoped<ILogTracer>(s => s.GetService<LogTracerFactory>()?.MakeLogTracer(Guid.NewGuid()) ?? throw new InvalidOperationException("Unable to create a logger"))
```
chkeita, line 36: @stishkin this is where i tried to initialized the IlogTracer so that we have one instance with shared for the duration of the request so that every controller involved in the request shares the same correlationId. 
When using this the log did not appear locally when i run it.

stishkin: yup, I think I know how to fix this, I'll submit separate PR for that

PR #1779, opened 2022-04-11
Adding Admin Checks to Node Operations.

nharper285:

_What is this about?_
Node operations are not currently blocked behind admin checks. I'm adding checks to all node operations. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Adding auth checks to node/init.py

## Validation Steps Performed

Deploy instance and check admin.


src/api-service/__app__/node/__init__.py
```diff
@@ -50,6 +50,10 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if isinstance(request, Error):
         return not_ok(request, context="NodeUpdate")
 
+    answer = check_can_manage_pools(req)
```
mgreisen, line 53: It's probably time for a more meaningful name over `check_can_manage_pools`.

nharper285: I think the underlying idea is consistent actually. This check is used for pools, scalesets, and nodes. Scalesets and nodes are just _part_ of the pool so I think it makes sense that you're still checking if _you can manage the pool_.

mgreisen: The problem is that although they might be related the things that are bing handled (scalesets, nodes) _aren't_ pools, so it's miss-leading.

nharper285: If we want to change then we will need to also change the instance_config value - can_manage_pools since this is the actual value that the function is checking against.

mgreisen (approved):

src/api-service/__app__/node/__init__.py
```diff
@@ -50,6 +50,10 @@ def post(req: func.HttpRequest) -> func.HttpResponse:
     if isinstance(request, Error):
         return not_ok(request, context="NodeUpdate")
 
+    answer = check_require_admins(req)
```
tevoinea, line 53: Should we even be trying to parse the request if it doesn't have the right perms?

tevoinea (approved):

src/api-service/__app__/node/__init__.py
nharper285: I would think we'd rather verify that the request is valid in the first place. This is the convention used across our endpoints.

docs/webhook_events.md
```diff
@@ -147,8 +147,7 @@ If webhook is set to have Event Grid message format then the payload will look a
                 "tools",
                 "unique_inputs",
                 "unique_reports",
-                "regression_reports",
-                "logs"
+                "regression_reports"
```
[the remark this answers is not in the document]
nharper285: There should be. Any time we update models.py webhook_events usually to be updated, as it uses the naming conventions from the models. 

So in this case, I've changed an instance_config param and this needs to be reflected in webhook_events. There is a build check for this. There is a script - /pytypes/extra/generate_docs.py - that will update the docs in webhook_events.md based off the current models. 

Regardless, it looks like the right updates did not get pushed properly :) So I _do_ need to fix that!

PR #1780, opened 2022-04-11
switch to our logger

stishkin:
once we start using middleware, can move log instantiation and setting of Correlation ID in there

### DIFF
--- a/src/ApiService/ApiService/EnvironmentVariables.cs
+++ b/src/ApiService/ApiService/EnvironmentVariables.cs
@@ -12,7 +12,11 @@ public static class EnvironmentVariables
 
     static EnvironmentVariables()
     {
+#if DEBUG
+        LogDestinations = new LogDestination[] { LogDestination.AppInsights, LogDestination.Console };
+#else
         LogDestinations = new LogDestination[] { LogDestination.AppInsights };
+#endif
     }
 
     //TODO: Add environment variable to control where to write logs to
--- a/src/ApiService/ApiService/Info.cs
+++ b/src/ApiService/ApiService/Info.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+
+namespace Microsoft.OneFuzz.Service;
+
+public record FunctionInfo(string Name, string ResourceGroup, string? SlotName);
+
+
+public class Info
+{
+
+    private readonly ILogTracerFactory _loggerFactory;
+
+
+    public Info(ILogTracerFactory loggerFactory)
+    {
+        _loggerFactory = loggerFactory;
+    }
+
+    [Function("Info")]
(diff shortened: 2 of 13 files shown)


src/ApiService/ApiService/EnvironmentVariables.cs
```diff
@@ -12,7 +12,11 @@ public static class EnvironmentVariables
 
     static EnvironmentVariables()
     {
+#if DEBUG
```
stishkin, line 15: this is just for now, totally ok to remove later

src/ApiService/ApiService/Info.cs
```diff
@@ -0,0 +1,40 @@
...
+    private readonly ILogTracerFactory _loggerFactory;
+
+
+    public Info(ILogTracerFactory loggerFactory)
+    {
+        _loggerFactory = loggerFactory;
+    }
+
+    [Function("Info")]
+    public async Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)] HttpRequestData req)
```
stishkin, line 23: a test knob to validate logging works as expected

stishkin: also returns some basic function info

src/ApiService/ApiService/QueueFileChanges.cs
```diff
@@ -52,11 +52,11 @@ public QueueFileChanges(ILoggerFactory loggerFactory, IStorageProvider storagePr
             return Task.CompletedTask;
         }
 
-        file_added(fileChangeEvent, lastTry);
+        file_added(log, fileChangeEvent, lastTry);
```
tevoinea, line 55: This pattern of passing the `log` around is only temporary until we move `var log = _loggerFactory.MakeLogTracer(Guid.NewGuid());` into middleware right?

stishkin: yes, that was my thinking.

Correlation ID might come from CLI as well (we should allow that at least), or from a message on a Queue, since we are tracking message progress through the system

src/ApiService/ApiService/Log.cs
```diff
@@ -121,77 +121,80 @@ public interface ILogTracer
...
-
     public void Info(string message)
     {
-        var caller = new StackTrace()?.GetFrame(1)?.GetMethod()?.Name;
-        foreach (var logger in loggers)
+        var caller = GetCaller();
+        foreach (var logger in _loggers)
         {
-            logger.Log(correlationId, message, SeverityLevel.Information, Tags, caller);
+            logger.Log(CorrelationId, message, SeverityLevel.Information, Tags, caller);
```
chkeita, line 146: These calls modify the tags of this instance. It means the logger is not thread safe, is that intentional ?

stishkin: I'll change to a record , so adding  new tags will create  a new logger.

Here logger is just a bag of properties, and ideally you just clone it per thread.

PR #1791, opened 2022-04-13
instance config

stishkin:
Instance config fetch/save


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -168,3 +173,102 @@ String InstanceName
...
+    string Namespace,
+    [property: JsonPropertyName("monitoringGSEnvironment")] string MonitoringGSEnvironment,
+    [property: JsonPropertyName("monitoringGCSAccount")] string MonitoringGCSAccount,
+    [property: JsonPropertyName("monitoringGCSAuthId")] string MonitoringGCSAuthId,
+    [property: JsonPropertyName("monitoringGCSAuthIdType")] string MonitoringGCSAuthIdType
+);
+
+public record AzureVmExtensionConfig(
+    KeyvaultExtensionConfig? Keyvault,
+    AzureMonitorExtensionConfig AzureMonitor
```
nharper285, line 200: In the python code, `AzureSecurityExtensionConfig` and `GenevaExtensionConfig` are both defined models included in `AzureVmExtensionConfig`. These are later used to check to include the extensions or not based off the instance_config passed to the instance. 

How will this happen in the future and what was the thinking behind leaving out these two definitions.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -168,3 +173,102 @@ String InstanceName
...
+        null)
+    { }
+
+
+    /// <summary>
+    /// Check if instance config is valid
+    /// </summary>
+    /// <returns>true, [] if instance config is valid;
+    /// otherwise false, with a list of errors</returns>
+    public (bool, List<string>) CheckInstanceConfig()
```
nharper285, line 265: Did you chose to leave out the check_admins func?

src/ApiService/ApiService/onefuzzlib/Config.cs
```diff
@@ -0,0 +1,66 @@
+﻿using ApiService.OneFuzzLib.Orm;
+using System;
+using System.Threading.Tasks;
```
nharper285, line 3: Can we rename this to InstanceConfig.cs? Config.cs feels a little too generic.

src/ApiService/ApiService/onefuzzlib/Config.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public async Task<InstanceConfig> Fetch()
+    {
+        var key = EnvironmentVariables.OneFuzz.InstanceName ?? throw new Exception("Environment variable ONEFUZZ_INSTANCE_NAME is not set");
+        var config = await GetEntityAsync(key, key);
+        return config;
+    }
+
+    public async Task Save(ILogTracer log, InstanceConfig config, bool isNew = false, bool requireEtag = false)
```
nharper285, line 30: So in the old instance_config app function init.py, the post is where we update the NSG. Are we thinking this would eventually go here? Or do we just need to refactor the instance_config app function entirely and that will be a separate file?

src/ApiService/ApiService/onefuzzlib/Config.cs
stishkin: Was just using 1 to 1 with Python code. 

Renamed to InstanceConfig.cs

src/ApiService/ApiService/OneFuzzTypes/Model.cs
stishkin: For now idea is just to port 1 to 1

If you have any ideas, do take notes and think about them. We'll discuss all those strategies once we have 1 to 1 port complete and we start re-architecture phase

src/ApiService/ApiService/onefuzzlib/Config.cs
stishkin: so that Http Trigger for instance config is still to be written. 

That code will be in the HttpTrigger handler

src/ApiService/ApiService/QueueNodeHearbeat.cs
```diff
@@ -31,16 +31,25 @@ public async Task Run([QueueTrigger("myqueue-items", Connection = "AzureWebJobsS
 
         var node = await _nodes.GetByMachineId(hb.NodeId);
 
+        var log2 = log.AddTag("NodeId", hb.NodeId.ToString());
```
tevoinea, line 34: Can we reuse `log` here?

```suggestion
        log = log.AddTag("NodeId", hb.NodeId.ToString());
```

src/ApiService/ApiService/onefuzzlib/WebhookOperations.cs
```diff
@@ -65,7 +65,7 @@ private void QueueObject(string v, WebhookMessageQueueObj obj, StorageType confi
 
 public interface IWebhookOperations
 {
-    Task SendEvent(EventMessage eventMessage);
+    Task SendEvent(ILogTracer log, EventMessage eventMessage);
```
chkeita, line 68: Where are we with the plan on having the LogTracer in the middleware ?
With the current solution we need to add that new parameter to every method
Can we bring back [this change](https://github.com/microsoft/onefuzz/pull/1765#discussion_r847519396) 
 and rely on the fact that a LogTracer will be generated for each request with a new ID

src/ApiService/ApiService/QueueNodeHearbeat.cs
stishkin: cleaning this up in next PR

src/ApiService/ApiService/Program.cs
```diff
@@ -6,13 +6,36 @@
...
 
 namespace Microsoft.OneFuzz.Service;
 
 public class Program
 {
+    public class LoggingMiddleware : IFunctionsWorkerMiddleware
+    {
+        public async Async.Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+        {
+            //TODO
```
mgreisen, line 20: Are there work items for the TODO's to make sure they don't get forgotten?

stishkin: for now just TODOs

chkeita (approved):

PR #1812, opened 2022-04-18
Check for missing shared libs on failed LibFuzzer `-help` check

### LINKED ISSUE #1806: When LibFuzzer `-help=1` check fails, report any missing shared libraries
Part of #1706.

ranweiler:
Add existing shared library error reporting capabilities to the LibFuzzer helpers used by the actual task worker.

Closes #1806.

### Manual testing
- [x] `linux-libfuzzer-linked-library` w/missing dependency
- [x] `windows-libfuzzer-linked-library` w/missing dependency


stishkin (approved):

ranweiler:
The missing library logic kicks in as expected. There are currently 2 issues with the error messages:
- Windows missing library detection fails on a fresh image's registry. This can be addressed separate from this PR, and our fallback error message is no worse than before.
- In the OneFuzz context, we may over-report missing libraries on Linux. Again, this can be addressed in a separate PR, and appears to be due to env var configuration in the way we invoke the shared library checker.

PR #1813, opened 2022-04-18
Add property based testing

stishkin:
Test all models using property based testing


chkeita (approved):

tevoinea (approved):

src/ApiService/Tests/OrmModelsTest.cs
```diff
@@ -0,0 +1,614 @@
...
+
+
+        public static Gen<Uri> Uri()
+        {
+            return Arb.Generate<IPv4Address>().Select(
+                arg => new Uri($"https://{arg.Item.ToString()}:8080")
+            );
+        }
+
+        public static Gen<WebhookMessageLog> WebhookMessageLog()
```
chkeita, line 35: can you only generate then `Event` an `EventType` fields these based on BaseEvent generator.
This is currently generating event that are no implemented yet

stishkin: I added it anyways

PR #1814, opened 2022-04-19
Limited support for polymorphic deserialization

chkeita:
Allow deserialization of a field based on the value of another


src/ApiService/ApiService/OneFuzzTypes/Events.cs
```diff
@@ -52,12 +50,42 @@ this switch
...
+    }
+}
+
+public class EventTypeProvider : ITypeProvider
+{
+    public Type GetTypeInfo(object input)
+    {
+        return (input as EventType?) switch
+        {
+            EventType.NodeHeartbeat => typeof(EventNodeHeartbeat),
```
stishkin, line 80: can we have one centralized function for this ? so we just update this switch in one place. 

Otherwise we going to forget to update one of those for sure

chkeita: moved the logic to BaseEvent

stishkin (approved):

PR #1821, opened 2022-04-21
Container converter

tevoinea:
The container type wasn't converting properly to and from the DB. This fixes that.


nharper285 (approved):

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -205,6 +205,11 @@ private EntityInfo GetEntityInfo<T>()
 
                 tableEntity.Add(prop.columnName, string.Join(",", values));
             }
+            else if (prop.type == typeof(Container))
```
chkeita, line 208: don't couple the entity converter with the enties themselves.
instead, create jsonconverter on the entity type

chkeita: `BaseEventConverter` is one example of jsonConverter. In your case you can create a converter and apply it to the `Container` type

src/ApiService/Tests/OrmTest.cs
```diff
@@ -276,5 +276,25 @@ public void TestEventSerialization2()
...
         }
+
+        record Entity3(
+            [PartitionKey] int Id,
+            [RowKey] string TheName,
+            Container Container
+        ) : EntityBase();
+
+        [Fact]
+        public void TestContainerSerialization()
```
chkeita, line 287: This Test is probably going to work even without your change. 
The problem is the  default serialization will translate a container to an object 
``` json 
{
   "ContainerName": "<name>"
}
```

but in existing object a container is represented by a simple string 

This test make sure that we can deserialize what **WE** serialize but it does not show that w can deserialize existing objects

chkeita (changes requested):
See comments

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
chkeita: also this fix onlyworks when the entity is saved to the db but not for when we read a json string

chkeita (approved):

Issue #1752, opened 2022-04-05
Add the dotnet AppService to the deployment


chkeita:
This work will integrate the [dotnet azure function](https://github.com/microsoft/onefuzz/tree/main/src/ApiService/ApiService) to the build. 
Here the steps necessary to accomplish this:
1. Update the [build](https://github.com/microsoft/onefuzz/blob/13b2a6ef80b5d4d13fb3fee0cf0c7c636458f917/.github/workflows/ci.yml#L251) to  publish the artifact by following [the same pattern](https://github.com/microsoft/onefuzz/blob/13b2a6ef80b5d4d13fb3fee0cf0c7c636458f917/.github/workflows/ci.yml#L226) used for the python function
2. Add a step similar to [deploy_app](https://github.com/microsoft/onefuzz/blob/9eb7f019a9e6ae24d736a83ad18a6b54f6703a28/src/deployment/deploy.py#L991) in deploy.py  to deploy the dotnet code.

PR #1830, opened 2022-04-22
Updating deploy.py for dotnet deployments #1752

AdamL-Microsoft:

Adding flag and function for '--dotnet_deploy' to deploy.py to specify deploying the dotnet version of the services along with the python ones, with the default (e.g. no --dotnet_deploy flag) still being to only deploy the python versions.

Also something recently changed on the backend of Azure creating a possible race condition for our function app deployments when using bicep templates. In order for the deployment to work and pass all of the application settings to both the python and dotnet function apps at deployment time, we had to reconfigure the .bicep templates to order the app settings to be setup after both of the function apps have already been deployed. 

## PR Checklist
* [x] Applies to work item: #1752

## Info on Pull Request

Updated:
src/deployment/deploy.py
src/deployment/azuredeploy.bicep
src/deployment/bicep-templates/function.bicep

Added:
src/deployment/bicep-templates/function-settings.bicep


src/deployment/deploy.py
```diff
@@ -1238,7 +1280,14 @@ def main() -> None:
         )
         states = rbac_only_states
     else:
-        states = full_deployment_states
+        if args.dotnet_deploy:
+            logger.info(
+                "deploying dotnet services for Azure functions instead of python"
+            )
```
chkeita, line 1286: The option here should be to deploy both C# and python or just python. 
We will need to have those functions running side by side

AdamLeclerc-Microsoft: Gotcha, the default being just building python then, correct?

I'll adjust the --dotnet_deploy flag for building both C# & python if so.

chkeita: "deploy" instead of "build". But yeah that's the idea

src/deployment/deploy.py
```diff
@@ -1238,7 +1281,14 @@ def main() -> None:
         )
         states = rbac_only_states
     else:
-        states = full_deployment_states
+        if args.dotnet_deploy:
+            logger.info(
+                "deploying dotnet and python services for Azure functions"
+            )
+            states = full_deployment_states + [("dotnet-api", Client.deploy_dotnet_app)]
```
chkeita, line 1288: careful here you are adding dotnet at the end of the list. it should be right after the python deployment. Because we have some steps taht occurs after we push the function

AdamLeclerc-Microsoft: good call, I'll revise this

chkeita (approved):

raghavthind2005 (approved):

PR #1831, opened 2022-04-22
OneFuzz CLI Docker container

stishkin:
Dockerize OneFuzz CLI
(Deployment scripts are not included)

### DIFF
--- a/docs/OneFuzz-Docker-CLI.md
+++ b/docs/OneFuzz-Docker-CLI.md
@@ -1,0 +1,33 @@
+# OneFuzz CLI in a Docker container
+
+## Using official release Docker container
+TODO
+
+## Building your own Docker container
+
+Docker file is located in `src` folder.
+
+To buid your own OneFuzz CLI Docker container use following command from `src` folder
+```
+docker build . --tag <CONTAINER_TAG> --build-arg REPO=<GITHUB_REPO> --build-arg PR=<PR> --build-arg GITHUB_TOKEN=<GITHUB_TOKEN>
+```
+where
+
+- <CONTAINER_TAG> - container image tag, it's an optional parameter. It will be used later in the document to explain how to run the container.
+
+- <GITHUB_REPO> - GitHub repository that contains a successfully build pull request that will be used for creating Docker container.
+- <PR> - GitHub pull request number that contains build artifacts to use to create Docker container.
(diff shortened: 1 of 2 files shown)


chkeita (approved):

mgreisen (approved):

docs/OneFuzz-Docker-CLI.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Running OneFuzz CLI Docker container
+
+There are three different scenarios that get enabled with OneFuzz CLI Docker container
+
+1. To have a new OneFuzz CLI session where you need to configure and authenticate every time on Docker container startup use following command `docker run -it <CONTAINER_TAG>`.
+
+2. If you have used OneFuzz CLI in your dev environment, and want to re-use configuration and authentication cache. Run following command (PowerShell example)  `docker run -it -v $env:USERPROFILE\.cache\onefuzz:/root/.cache/onefuzz <CONTAINER_TAG>`. It will mount your OneFuzz cache folder into OneFuzz CLI Docker container.
+
+3. If you have several OneFuzz deployments. You can store OneFuzz configuration per deploymnet in your dev environment by creating a different folder for each OneFuzz deployment and then mountint that folder as OneFuzz CLI cache when running the Docker container.
```
ranweiler, line 32:
```suggestion
3. If you have several OneFuzz deployments. You can store OneFuzz configuration per deployment in your dev environment by creating a different folder for each OneFuzz deployment and then mounting that folder as OneFuzz CLI cache when running the Docker container.
```

ranweiler (approved):
Noting for the record / future work:
- To support multiple deployments, the OneFuzz CLI can already take arguments for the config parameters, like `--endpoint`, `--authority`, &c.
- The best solution for multiple deployment profiles is probably just adding a `--config` argument that loads a JSON config file, and uses a default when omitted.
- Dockerizing the CLI really shouldn't be necessary. On Windows we already use [`pyinstaller`](https://pyinstaller.org/en/stable/), and that should also be usable on Linux. But we should probably also be able to use [`pex`](https://pex.readthedocs.io/en/v2.1.82/whatispex.html#whatispex) on Linux. Long-term, we should move to a self-contained static binary written in Rust. That's not very practical right now due to heavy CLI use of the shared `pytypes` package.

PR #1835, opened 2022-04-22
Singleton context

stishkin:
- Move environment variables to singleton
-  Move some persistent resources into SingletonResources class
- Rename ResultOk to ResultVoid


src/ApiService/ApiService/Program.cs
```diff
@@ -82,10 +82,12 @@ public static void Main()
             .AddScoped<IReports, Reports>()
             .AddScoped<INotificationOperations, NotificationOperations>()
             .AddScoped<IUserCredentials, UserCredentials>()
+            .AddScoped<ICreds, Creds>()
```
chkeita, line 85: I think we can just use Icred as the singleton instead of creating a new entity. Plus the name is more descriptive.

stishkin: HttpClient should stay in Singleton resources ? Or move it to Creds ?

chkeita (approved):

PR #1836, opened 2022-04-23
migrate timer_proxy part 2

chkeita:
Continuing the migration of time_proxy


src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.OneFuzz.Service;
+
+using System.Security.Cryptography;
+
+public class Auth
+{
+    public static Authentication BuildAuth()
+    {
```
stishkin, line 8: you verified it matches Python calling ssh-key gen and it matches ?

src/ApiService/ApiService/onefuzzlib/Network.cs
```diff
@@ -50,7 +44,7 @@ public static async Async.Task<Network> Create(string region, ICreds creds, ICon
             }
             else
             {
-                var networkId = GenerateGuidv5(NETWORK_GUID_NAMESPACE, string.Join("|", networkConfig.AddressSpace, networkConfig.Subnet));
+                var networkId = Faithlife.Utility.GuidUtility.Create(NETWORK_GUID_NAMESPACE, string.Join("|", networkConfig.AddressSpace, networkConfig.Subnet), 5);
```
stishkin, line 47: we need to take a 3rd party dependency to generate network id ?

src/ApiService/ApiService/onefuzzlib/Auth.cs
chkeita: No i did not. I am counting on check pr to validate this

src/ApiService/ApiService/onefuzzlib/Network.cs
chkeita: Yes, There is not way generate a deterministic Guid with the base library. 
If you have a better suggestion let me know

stishkin (approved):

PR #1839, opened 2022-04-25
Refactor timer_repro

tevoinea:


stishkin (approved):

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -5,6 +5,9 @@ namespace Microsoft.OneFuzz.Service;
 public interface IScalesetOperations : IOrm<Scaleset>
 {
     IAsyncEnumerable<Scaleset> Search();
+
+    public IAsyncEnumerable<Scaleset?> SearchByPool(string poolName);
```
chkeita, line 9: the type paramter should not be nullable 
```suggestion
    public IAsyncEnumerable<Scaleset> SearchByPool(string poolName);
```

chkeita (approved):

PR #1848, opened 2022-04-26
breaking the circular dependency between JobOperations and TaskOperat…

chkeita:
Removing ItaskOperations from JobOeprations

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/JobOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/JobOperations.cs
@@ -6,17 +6,15 @@ public interface IJobOperations : IStatefulOrm<Job, JobState> {
     System.Threading.Tasks.Task<Job?> Get(Guid jobId);
     System.Threading.Tasks.Task OnStart(Job job);
     IAsyncEnumerable<Job> SearchExpired();
-    System.Threading.Tasks.Task Stopping(Job job);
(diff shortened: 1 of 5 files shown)


src/ApiService/ApiService/onefuzzlib/JobOperations.cs
```diff
@@ -46,17 +44,17 @@ public class JobOperations : StatefulOrm<Job, JobState>, IJobOperations {
         throw new NotImplementedException();
     }
 
-    public async System.Threading.Tasks.Task Stopping(Job job) {
+    public async System.Threading.Tasks.Task Stopping(Job job, ITaskOperations taskOperations) {
```
stishkin, line 47: For later, should we just have a "mega-factory" being passed in the constructor, and then each object can create what it needs ? That way it will be harder to have circular dependency ?

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/JobOperations.cs
chkeita: yep. I was toying with the idea of passing `IServiceCollection` directly. But i think we should have something a bit more specific to Onefuzz

stishkin: Agreed, let's have OneFuzz Factory of some sort, add AddScoped factory, and all creation happens inside of the factory....

tevoinea (approved):

PR #1849, opened 2022-04-26
Upgrade to latest stable `notify` preview

### LINKED ISSUE #1083: Revisit use of `notify` when 5.0 is released
The latest stable (4.0) release of the `notify` crate provides a `std::sync::mpsc` channel-based API, only. To use this in an async context, we must either adapt the sync `Receiver::recv()` to an async version, via something like `spawn_blocking()`, or manually perform an async poll using `try_recv()` and `tokio::time::sleep()`.

Ideally, we could impl `Stream` on the `DirectoryMonitor` in a way where, when pending, we arrange for the current `Waker` to get called by a `notify`-provided callback on each internal `poll()` event. This is possible in the new, pre-release 5.0 API of `notify`, which provides a notion of "immediate" `Watcher` that invokes a callback. Once `notify 5.0` is released, revisit our usage of it in the `DirectoryMonitor`.

ranweiler:
Update `notify` to its latest stable preview release `5.0.0-pre.14`.

The latest `notify` release on `crates.io` is `4.0.17`. The 4.x series receives very occasional backports (the last one was a year ago), but it is not actively developed, and a `cargo audit` run uncovers many issues that are unlikely to be addressed. This upgrade lets us avoid problematic transitive dependencies, and align our direct dependency on `notify` to its actively-maintained major version.

By upgrading to the most-stable 5.x preview version, we also get to use the new `Watcher` API. This accepts any implementer of the new `EventHandler` trait, which includes closures. This lets us directly send (fallible) watcher events via the synchronous `UnboundedSender::send()` method, instead of having to shim the sync `Sender` from the 4.x API with a background async task.

This PR also changes the `DirectoryMonitor` public API such that `start()` is async. This enables replacing a private use of the blocking `std::fs::canonicalize()` function. It also makes `next_file()` fallible, which reflects the fallibility of internal filesystem watch eventing. Finally, it adds unit tests and an example binary.

Closes #1083.

### DIFF
--- a/src/agent/onefuzz/src/monitor.rs
+++ b/src/agent/onefuzz/src/monitor.rs
@@ -2,36 +2,33 @@
 // Licensed under the MIT License.
 
 use std::path::PathBuf;
-use std::sync::{self, mpsc::Receiver as SyncReceiver};
-use std::time::Duration;
 
-use anyhow::Result;
-use notify::{DebouncedEvent, Watcher};
+use anyhow::{format_err, Result};
+use notify::{Event, EventKind, Watcher};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
-use tokio::task::{self, JoinHandle};
 
 pub struct DirectoryMonitor {
     dir: PathBuf,
-    notify_events: UnboundedReceiver<DebouncedEvent>,
+    notify_events: UnboundedReceiver<notify::Result<Event>>,
     watcher: notify::RecommendedWatcher,
 }
 
 impl DirectoryMonitor {
-    pub fn new(dir: impl Into<PathBuf>) -> Self {
+    pub fn new(dir: impl Into<PathBuf>) -> Result<Self> {
         let dir = dir.into();
-        let (notify_sender, notify_receiver) = sync::mpsc::channel();
-        let delay = Duration::from_millis(100);
-        let watcher = notify::watcher(notify_sender, delay).unwrap();
+        let (sender, notify_events) = unbounded_channel();
+        let event_handler = move |event_or_err| {
+            // A send error only occurs when the channel is closed. No remedial
+            // action is needed (or possible), so ignore it.
+            let _ = sender.send(event_or_err);
+        };
+        let watcher = notify::recommended_watcher(event_handler)?;
 
-        // We can drop the thread handle, and it will continue to run until it
-        // errors or we drop the async receiver.
-        let (notify_events, _handle) = into_async(notify_receiver);
-
-        Self {
+        Ok(Self {
             dir,
             notify_events,
             watcher,
-        }
+        })
     }
 
     pub fn start(&mut self) -> Result<()> {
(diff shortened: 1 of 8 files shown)


src/agent/onefuzz/src/monitor.rs
```diff
@@ -45,62 +42,61 @@ impl DirectoryMonitor {
...
+                }
+            };
 
-            if event.is_none() {
-                // Make sure we stop our `Watcher` if we return early.
-                let _ = self.stop();
-            }
+            match event.kind {
+                EventKind::Create(..) => {
+                    let path = event
```
ranweiler, line 74: Events in the new API always have a `paths` field, which contains a [vector of involved paths](https://docs.rs/notify/5.0.0-pre.14/notify/event/struct.Event.html#structfield.paths), which depend on the [`AccessKind`](https://docs.rs/notify/5.0.0-pre.14/notify/event/enum.AccessKind.html). The guarantees here are undocumented, so we just take care to handle anything.

src/agent/onefuzz/src/monitor.rs
```diff
@@ -2,36 +2,33 @@
...
-        // We can drop the thread handle, and it will continue to run until it
-        // errors or we drop the async receiver.
-        let (notify_events, _handle) = into_async(notify_receiver);
-
-        Self {
+        let (sender, notify_events) = unbounded_channel();
+        let event_handler = move |event_or_err| {
+            // A send error only occurs when the channel is closed. No remedial
+            // action is needed (or possible), so ignore it.
+            let _ = sender.send(event_or_err);
```
ranweiler, line 23: We ignore `Sender` errors on this line, but are still sending `Watcher` errors, so the monitor can report them and shutdown as nicely as possible.

src/agent/onefuzz/src/monitor.rs
chkeita: Why only return the first path here ? Could we receive more than 1 path for each events ?

src/agent/onefuzz/src/monitor.rs
```diff
@@ -45,62 +42,61 @@ impl DirectoryMonitor {
...
+                    return Ok(None);
+                }
+            };
 
-            if event.is_none() {
-                // Make sure we stop our `Watcher` if we return early.
-                let _ = self.stop();
-            }
+            match event.kind {
+                EventKind::Create(..) => {
```
ranweiler, line 73: We should probably scrutinize the `CreateKind` that we're wildcarding out, and only emit events that are known to have kind `CreateKind::File`. Only thing I'm not sure about is how to deal with the `Any`/`Other` variants (will dig into the source and experiment locally).

src/agent/onefuzz/src/monitor.rs
ranweiler: I'll verify as I read the source, but I think there are only multiple paths to support things like `src`/`dest` when `rename()`-ing files (see: https://docs.rs/notify/5.0.0-pre.14/notify/event/enum.EventKind.html#variant.Remove). But for that, something like a tuple of `Option<PathBuf>` values would seem to make more sense, so I'll definitely double-check.

ranweiler: Yeah, on all the platforms I looked at, I'm really only seeing one `PathBuf` ever get pushed to the `paths` field. There definitely isn't anything that would cause `paths` to grow arbitrarily (e.g. debounced coalescing logic, if they were batching paths to several `Create` events that happened in quick succession).

chkeita (approved):

src/agent/onefuzz/src/monitor.rs
ranweiler: Will address this in a followup PR.

src/agent/onefuzz/src/monitor/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    fs::write(&file_a, "aaa").await?;
+    fs::write(&file_b, "bbb").await?;
+    fs::write(&file_c, "ccc").await?;
+
+    assert_eq!(monitor.next_file().await?, Some(canonicalize(&file_a)?));
+    assert_eq!(monitor.next_file().await?, Some(canonicalize(&file_b)?));
+    assert_eq!(monitor.next_file().await?, Some(canonicalize(&file_c)?));
+
+    // TODO: on Windows, `notify` doesn't provide an event for the removal of a
```
ranweiler, line 94: Captured in https://github.com/microsoft/onefuzz/issues/1850.

PR #1859, opened 2022-04-27
Only watch directories, not files

### LINKED ISSUE #1857: `DirectoryMonitor` can be configured to watch a file
The path to watch, which we pass to the `DirectoryMonitor` ctor, should only ever be a directory, not a file.

ranweiler:
If a `DirectoryMonitor` is pointed at a file (and not a directory), fail on invocation of `start()`.

Validated with a unit test and manual testing using the `dir-monitor` example.

Closes #1857.


src/agent/onefuzz/src/monitor.rs
```diff
@@ -40,6 +40,13 @@ impl DirectoryMonitor {
         // Canonicalize so we can compare the watched dir to paths in the events.
         self.dir = fs::canonicalize(&self.dir).await?;
 
+        // Make sure we are watching a directory.
+        //
+        // This check will pass for symlinks to directories.
```
ranweiler, line 45: The unit test that backs up this claim was added in the previous PR.

chkeita (approved):

raghavthind2005 (approved):

PR #1863, opened 2022-04-29
Timer workers (part 1)

stishkin:


src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -20,4 +48,173 @@ public NodeOperations(IStorage storage, ILogTracer log, IServiceConfig config)
...
+
+    public NodeTasksOperations(IStorage storage, ILogTracer log, IServiceConfig config)
+        : base(storage, log, config) {
+        _log = log;
+    }
+
+    public async IAsyncEnumerable<Node> GetNodesByTaskId(Guid taskId, INodeOperations nodeOps) {
+        List<Node> results = new();
+        await foreach (var entry in QueryAsync($"task_id eq '{taskId}'")) {
+            var node = await nodeOps.GetByMachineId(entry.MachineId);
```
chkeita, line 149: this can probably be optimize by query all NodesTasks  then query the all machine in single request

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
stishkin: I'll add this suggestion as TODO in the code.

For now just porting logic 1 to 1 with python

chkeita: @stishkin this is what is meant 

``` C#
var nodetasks = await  QueryAsync($"task_id eq '{taskId}'").Select(x => x.MachineId.ToString()).ToListAsync();
var query = Query.EqualAny("RowKey", nodetasks);
await foreach( var node in nodeOps.QueryAsync(query))
{
    yield return node;
}
```

PR #1865, opened 2022-05-01
Switch to misuse-resistant smart constructor

### LINKED ISSUE #1864: Check for invalid paths in `DirectoryMonitor` constructor
The `DirectoryMonitor` has a fallible constructor, but also does some checks very "late", in the `start()` method. This is a little odd, and really only just provides an opportunity to forget to call `start()` (which will cause `next_file()` to async-block forever).

Fix both of these design quirks:
- Remove `start()`
- Set the OS-level watch on `new()`
- Validate the target path on `new()`

ranweiler:
Use a simpler and less error-prone API for `DirectoryMonitor`.

- Make `new()` ctor async, set watch on construction
- Remove `start()`, validate that target path is a directory in `new()`

Closes #1864.

### DIFF
--- a/src/agent/onefuzz-agent/src/local/common.rs
+++ b/src/agent/onefuzz-agent/src/local/common.rs
@@ -1,5 +1,10 @@
-use crate::tasks::config::CommonConfig;
-use crate::tasks::utils::parse_key_value;
+use std::{
+    collections::HashMap,
+    env::current_dir,
+    path::{Path, PathBuf},
+    time::Duration,
+};
+
 use anyhow::Result;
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-agent/src/local/common.rs
```diff
@@ -294,6 +283,18 @@ impl DirectoryMonitorQueue {
     }
 }
 
+async fn monitor_directory(queue_client: QueueClient, directory: PathBuf) -> Result<()> {
```
ranweiler, line 286: Just split out for readability, no logic changes.

stishkin (approved):

PR #1870, opened 2022-05-02
Refactor agent can schedule

tevoinea:


src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -33,11 +33,18 @@ public class Orm<T> : IOrm<T> where T : EntityBase {
...
         }
 
-        public async IAsyncEnumerable<T> QueryAsync(string? filter = null) {
+        public async IAsyncEnumerable<T> QueryAsync(string? filter = null, int? numResults = null) {
             var tableClient = await GetTableClient(typeof(T).Name);
+            int totalResults = 0;
 
             await foreach (var x in tableClient.QueryAsync<TableEntity>(filter).Select(x => _entityConverter.ToRecord<T>(x))) {
                 yield return x;
+                if (numResults != null) {
```
chkeita, line 42: Can't you just use Take before the select ?

chkeita: And since we already return IAsyncEnumerable we might not event need to add a numResult Parameter. since the caller can use take anyway.
If this is a 1 to one port feel free to ignore this suggestion. But we might want to revisit later

stishkin (approved):

PR #1873, opened 2022-05-02
Build on pushes to `main`

ranweiler:
#1872 disabled triggering the "Build" action on pushes to the `main` branch, which we should preserve. Explicitly enable this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: Build
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


tevoinea (approved):

stishkin (approved):

.github/workflows/ci.yml
```diff
@@ -1,5 +1,10 @@
 name: Build
-on: [pull_request, workflow_dispatch]
+on:
```
ranweiler, line 2: Syntax docs for this scenario: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events

PR #1876, opened 2022-05-02
timer workers (part 2)

stishkin:
there will be part 3

### DIFF
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -86,7 +86,7 @@ public record Node
     [PartitionKey] PoolName PoolName,
     [RowKey] Guid MachineId,
     Guid? PoolId,
-    Version Version,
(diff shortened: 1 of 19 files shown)


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -83,17 +83,20 @@ DateTimeOffset TimeStamp
...
     [RowKey] Guid MachineId,
-    NodeState State,
-    Guid? ScalesetId,
-    DateTimeOffset Heartbeat,
-    string Version,
-    bool ReimageRequested,
-    bool DeleteRequested,
-    bool DebugKeepNode
+    Guid? PoolId,
+    Version Version,
```
chkeita, line 89: This won't work. The version class does not support the format of custom version we use when we deploy from  git. 
example "5.3.0-fe0ae308234cb6e5e39193e20f5819108ff78856"

chkeita (approved):

PR #1886, opened 2022-05-03
Introducing OnefuzzContext to prevent circular references

chkeita:
- Creating a OnefuzzContext object that contains most of our dependencies
- update some classes to use this object instead of direct dependencies

### DIFF
--- a/src/ApiService/ApiService/Program.cs
+++ b/src/ApiService/ApiService/Program.cs
@@ -103,6 +103,7 @@ public class Program {
             .AddScoped<IVmssOperations, VmssOperations>()
             .AddScoped<INodeTasksOperations, NodeTasksOperations>()
             .AddScoped<INodeMessageOperations, NodeMessageOperations>()
+            .AddScoped<IOnefuzzContext, OnefuzzContext>()
 
             .AddSingleton<ICreds, Creds>()
             .AddSingleton<IServiceConfig, ServiceConfiguration>()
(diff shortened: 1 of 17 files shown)


src/ApiService/ApiService/Program.cs
```diff
@@ -103,6 +103,7 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
             .AddScoped<IVmssOperations, VmssOperations>()
             .AddScoped<INodeTasksOperations, NodeTasksOperations>()
             .AddScoped<INodeMessageOperations, NodeMessageOperations>()
+            .AddScoped<IOnefuzzContext, OnefuzzContext>()
```
stishkin, line 106: as a person who'd be a new contributor - this is just confusing now, since I have to add a new Scoped to here and also to context. 

Can you remove all scoped additions from here (except for log) - and move them all to the context ? So we only have one place where we need to add stuff ?

stishkin (approved):

src/ApiService/ApiService/Program.cs
chkeita: The OnefuzzConext  is just a propertybag on top of the IserviceProvider. I does not handle initialization and scopes. To Remove these instrauctions i would have to reimplement the whole dependency injection engine.

tevoinea (approved):

PR #1893, opened 2022-05-04
Allow jobs with no log data to be schedule

### LINKED ISSUE #1892: Job was created without `config.logs`
## Information

- Onefuzz version: 5.3.0
- OS: Unknown

1. After we deployed 5.3.0, it was possible for someone to submit a job with no logs property in the config.
2. We throw if the logs property isn't present: https://github.com/microsoft/onefuzz/blob/eca88cb35f60c30fe7a6dbfbc436be0f7ddc36c9/src/api-service/__app__/onefuzzlib/tasks/config.py#L270-L273
3. Which breaks this task processing loop: https://github.com/microsoft/onefuzz/blob/eca88cb35f60c30fe7a6dbfbc436be0f7ddc36c9/src/api-service/__app__/onefuzzlib/tasks/scheduler.py#L235

### Expected result

Jobs should always be created with `logs` config property.

### Actual result

No tasks get processed.

chkeita:
closes #1892

### DIFF
--- a/src/pytypes/onefuzztypes/models.py
+++ b/src/pytypes/onefuzztypes/models.py
@@ -340,7 +340,7 @@ class AgentConfig(BaseModel):
 
 class TaskUnitConfig(BaseModel):
     instance_id: UUID
-    logs: str
+    logs: Optional[str]
     job_id: UUID
     task_id: UUID
     task_type: TaskType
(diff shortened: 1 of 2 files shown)


src/pytypes/onefuzztypes/models.py
```diff
@@ -340,7 +340,7 @@ class AgentConfig(BaseModel):
 
 class TaskUnitConfig(BaseModel):
     instance_id: UUID
-    logs: str
+    logs: Optional[str]
```
tevoinea, line 343: Could this cause issues on the agent side if the property isn't present?

chkeita: This field is optional on the client side already
https://github.com/microsoft/onefuzz/blob/457f03e4ae32c140c288327b2daf72db1c6b7b8b/src/agent/onefuzz-agent/src/tasks/config.rs#L49

tevoinea (approved):

PR #1897, opened 2022-05-04
Notification operations test hooks

stishkin:


chkeita (approved):

src/ApiService/ApiService/TestHooks/NotificationOperationsTestHooks.cs
```diff
@@ -0,0 +1,82 @@
...
+        private readonly INotificationOperations _notificationOps;
+
+        public NotificationOperationsTestHooks(ILogTracer log, IConfigOperations configOps, INotificationOperations notificationOps) {
+            _log = log.WithTag("TestHooks", nameof(NotificationOperationsTestHooks));
+            _configOps = configOps; ;
+            _notificationOps = notificationOps;
+        }
+
+        [Function("NewFilesTestHook")]
+        public async Task<HttpResponseData> NewFiles([HttpTrigger(AuthorizationLevel.Anonymous, "put", Route = "testhooks/notificationOperations/newFiles")] HttpRequestData req) {
```
tevoinea, line 25: This might be too meta but do you think it's worthwhile to use a source generator to auto create these?

tevoinea (approved):

src/ApiService/ApiService/TestHooks/NotificationOperationsTestHooks.cs
stishkin: you are right, i was thinking of dealing with that later. 

I was thinking of doing Azure API management and then work from there at some point. 

At this point I'm willing to type things

PR #1905, opened 2022-05-06
Set the table value to null when a property is null

### LINKED ISSUE #1901: [C#] ORM serializes not-set fields instead of ignoring them - which breaks currently running Python function
call
```
 onefuzz info get
```
and it succeeds.

Get latest Instance Config table entry JSON object, and  call "Save" on Instance Config operations in C# code with the object from above, it will serialize the full record to the table. Adding fields that were not present there before.

Thus if "network_config" was not present - it will add "network_config" as null. 
call 
```
:/# onefuzz info get
ERROR:cli:command failed: request did not succeed: HTTP 500 -*
```

chkeita:
fixes #1901

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
+++ b/src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
@@ -156,16 +156,20 @@ public class EntityConverter {
         if (type is null) {
             throw new NullReferenceException();
         }
-        var tableEntity = new TableEntity();
+
         var entityInfo = GetEntityInfo<T>();
-        foreach (var prop in entityInfo.properties.SelectMany(x => x)) {
-            //var prop = kvp.First();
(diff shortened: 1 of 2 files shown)


src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -156,16 +156,20 @@ public class EntityConverter {
...
             throw new NullReferenceException();
         }
-        var tableEntity = new TableEntity();
+
         var entityInfo = GetEntityInfo<T>();
-        foreach (var prop in entityInfo.properties.SelectMany(x => x)) {
-            //var prop = kvp.First();
+        Dictionary<string, object?> columnValues = entityInfo.properties.SelectMany(x => x).Select(prop => {
             var value = entityInfo.type.GetProperty(prop.name)?.GetValue(typedEntity);
+            if (value == null) {
```
chkeita, line 163: this is the actual fix. The null value was converted to the string "null" before this change

stishkin (approved):

PR #1918, opened 2022-05-09
Remove `client_secret` from config, add env var

ranweiler:
The client secret is only intended for confidential client use (e.g. ephemerally, in CI). If set via `onefuzz config`, it will be serialized as part of the config, for convenient reuse in that setting. When serialized, default filesystem permissions secure it against other local users, but this lacks defense in depth. On a long-lived host, this is brittle.

Make misuse impossible: remove the `client_secret` field from the config altogether. When initializing a `Onefuzz` API instance, check an env var for a client secret (to allow "setting" the client secret once across multiple commands in CI). This means every CLI invocation will _also_ perform this env lookup. Continue to allow passing / overriding the secret as a CLI argument, too.

Tested with a dev service principal and client secret:
- [x] No secret (falls back on public client flow)
- [x] Secret only in env var
- [x] Secret only as parameter
- [x] Secret as parameter overrides invalid env var
- [x] Secret misplaced in config always ignored


stishkin (approved):

src/cli/onefuzz/api.py
```diff
@@ -53,6 +53,8 @@
 
 UUID_RE = r"^[a-f0-9]{8}-?[a-f0-9]{4}-?[a-f0-9]{4}-?[a-f0-9]{4}-?[a-f0-9]{12}\Z"
 
+CLIENT_SECRET_ENV_VAR = "ONEFUZZ_CLIENT_SECRET"
```
chkeita: just a heads up. Check-pr will probably fail because the client-secret is not set.
You will need to update [this](https://github.com/microsoft/onefuzz/blob/a4357c8fb9b4eb2ceee1206c151e6c027f22a990/src/utils/check-pr/check-pr.py#L141)

ranweiler: Ha, yep, just fixed that!

ranweiler: Er, nope, something else. But I see what _you_ mean. Fixing.

ranweiler: Oh, no, nevermind. We still accept the `--client_secret` arg, so check-pr should still pass as-is.

chkeita:
This code that shows the command line to  setup the onefuzz cli should be updated to remove the [client_secret section](https://github.com/microsoft/onefuzz/blob/a4357c8fb9b4eb2ceee1206c151e6c027f22a990/src/deployment/deploy.py#L1087).
Note this is most likely a noop because we never set the cli client_secret in the deployment

chkeita (approved):

mgreisen (approved):

src/cli/examples/azure-functions-example/info/__init__.py
```diff
@@ -14,7 +14,6 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         endpoint=os.environ.get("ONEFUZZ_ENDPOINT"),
         authority=os.environ.get("ONEFUZZ_AUTHORITY"),
         client_id=os.environ.get("ONEFUZZ_CLIENT_ID"),
-        client_secret=os.environ.get("ONEFUZZ_CLIENT_SECRET"),
```
ranweiler, line 17: Now this automatically gets looked up by `Onefuzz.__init__()`.

PR #1925, opened 2022-05-10
pinning the version of dotnet to 6.0.202

chkeita:
A new version came out today and is causing the formatting break in the build release.
We decided to pin the release the version number.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,7 +260,7 @@ jobs:
     - name: Setup .NET Core SDK
       uses: actions/setup-dotnet@v2
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: '6.0.202'
     - name: Install dependencies
       run: |
         cd src/ApiService/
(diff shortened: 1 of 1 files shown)


mgreisen (approved):

.github/workflows/ci.yml
```diff
@@ -260,7 +260,7 @@ jobs:
     - name: Setup .NET Core SDK
       uses: actions/setup-dotnet@v2
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: '6.0.202'
```
stishkin, line 263: out of curiosity - this is required because ?

chkeita: a new version came out today and is causing the formatting break in the build release.
We decided to pin the release the version number.

mgreisen: b/c it changed today and is breaking the build. I suggested win Pin it until were ready to update it and reformat files

stishkin (approved):

Issue #1929, opened 2022-05-11
Trigger a build when a tag is created


mgreisen:
It's helpful to have a build created when a tag is pushed so that build artifacts for a tagged commit can be retrieved.

PR #1930, opened 2022-05-11
When a tag is created, trigger a build.

mgreisen:
When we tag a commit, it's useful to be able to get the build artifacts for our release process. 
#1929


nharper285 (commented):
Approved!

stishkin (approved):

PR #1931, opened 2022-05-11
Allow ignoring directories

### LINKED ISSUE #1856: `DirectoryMonitor` reports creation of both files and directories
As used, we only want to report new files. It might be useful to allow _optionally_ reporting new directories, too.

tevoinea:

_What is this about?_

Allow the DirectoryMonitor to skip reporting directories. The default behavior is to continue reporting them so we don't change the behavior unexpectedly.

## PR Checklist
* [ ] Closes #1856 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

Tested with the `dir-monitor` example.

### DIFF
--- a/src/agent/onefuzz/src/monitor.rs
+++ b/src/agent/onefuzz/src/monitor.rs
@@ -15,6 +15,7 @@ pub struct DirectoryMonitor {
     dir: PathBuf,
     notify_events: UnboundedReceiver<notify::Result<Event>>,
     watcher: notify::RecommendedWatcher,
+    report_directories: bool,
 }
 
 impl DirectoryMonitor {
@@ -43,11 +44,19 @@ impl DirectoryMonitor {
         let mut watcher = notify::recommended_watcher(event_handler)?;
         watcher.watch(&dir, RecursiveMode::NonRecursive)?;
 
+        let report_directories = true;
+
         Ok(Self {
             dir,
             notify_events,
             watcher,
+            report_directories,
         })
+    }
+
+    pub fn set_report_directories(mut self, report_directories: bool) -> Self {
+        self.report_directories = report_directories;
+        self
     }
 
     pub fn stop(&mut self) -> Result<()> {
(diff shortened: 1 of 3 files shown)


src/agent/onefuzz/src/monitor.rs
```diff
@@ -43,13 +44,21 @@ impl DirectoryMonitor {
...
+
         Ok(Self {
             dir,
             notify_events,
             watcher,
+            report_directories,
         })
     }
 
+    pub fn set_report_directories(mut self, report_directories: bool) -> Self {
```
ranweiler, line 57: API suggestion / random thoughts about setter signatures:

We don't have a true builder pattern in this situation, because there's just the one "built" type. Contrast with the API of `std::process::Command` (the builder) and `Child` (the... buildee? 😅).

As a result, the signature of this builder-style setter imposes the strongest-possible requirement on the receiver (we demand an owned `self`, when `&mut self` would do), and uses a quirky/surprising return value (_for a setter_), for (I claim!) no real convenience when we use it.

The reason I think there's no big usage benefit is that with this type, we're _always_ going to create a mutable instance, bound to a variable name. That's because it is basically a `Stream` (via `next_file()`) which we need to scrutinize in an async loop.

So the difference becomes:
```rust    
let mut monitor = DirectoryMonitor::new(opt.path)
        .await?
        .set_report_directories(true);
```
vs.
```rust    
let mut monitor = DirectoryMonitor::new(opt.path).await?;
monitor.set_report_directories(true);
```

Overall, this isn't a big deal, and is your call! But I like to use simple situations like this to reason about why/when to do one thing vs. another in designing APIs.

ranweiler (commented):
Looks great, just a few comments!

Also, can we add some unit tests?

src/agent/onefuzz/src/monitor.rs
tevoinea: Yup you're right, there isn't a Builder and a Buildee my original goal was to avoid having to pass an `Option<bool>` to `new` and the suggested way still accomplishes that so I'm happy!

PR #1934, opened 2022-05-12
Test timer repro

tevoinea:

_What is this about?_

Updates `TimerRepro` to use `IOnefuzzContext` and adds some unit tests.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/Tests/TimerReproTests.cs
+++ b/src/ApiService/Tests/TimerReproTests.cs
@@ -1,0 +1,116 @@
+using Microsoft.OneFuzz.Service;
+using Moq;
+using Xunit;
+using System.Linq;
+using System.Collections.Generic;
+using System;
+using Microsoft.Azure.Functions.Worker;
+
+namespace Tests;
+
+public class TimerReproTests {
+    private ILogTracer log;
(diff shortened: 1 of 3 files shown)


src/ApiService/Tests/TimerReproTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace Tests;
+
+public class TimerReproTests {
+    private ILogTracer log;
+    private Mock<IOnefuzzContext> mockCtx;
+    private Mock<IReproOperations> mockReproOperations;
+
+    public TimerReproTests() {
+        mockCtx = new Mock<IOnefuzzContext>();
```
chkeita, line 17: feel free to ignore. 
But 'IOnefuzzContext' can be created without needing a mock. It only depends on a service collection 
you can do something like this instead 
``` C#
var services = new ServiceCollection();
// adde your mocked dependncies here
//services.AddScoped<IReproOperations, Mock<IReproOperations>>();
var serviceProvider = services.BuildServiceProvider();
var context = new OnefuzzContext(serviceProvider);
```

PR #1960, opened 2022-05-17
Ensure `GlobalFlag` reg value is initialized for target

### LINKED ISSUE #1826: Missing shared library init doesn't work on a fresh Windows VM image
Attempting to write the gflags registry entry fails with:
```
additional error while checking for missing shared libraries: error accessing image global flags
```

ranweiler:
### Context

We ensured that the key exists for the target, but not the leaf `GlobalFlag` _value_. We tried to retrieve an existing value and update just [`FLG_SHOW_LDR_SNAPS`](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/show-loader-snaps), but somehow missed the base case in manual testing.

Fixes #1826.

### Changes

- Zero-initialize `GlobalFlag` registry values as needed
- Update an internal error type for easier debugging

### Testing

Manually tested by creating jobs using `windows-libfuzzer-linked-library` and `windows-libfuzzer-load-library` tasks, but with one DLL dependency removed from the setup dir. Both failed as expected with the descriptive error:
> `Error: fuzzer does not respond to '-help=1'. missing shared libraries: bad.dll.`

### DIFF
--- a/src/agent/dynamic-library/src/windows.rs
+++ b/src/agent/dynamic-library/src/windows.rs
@@ -110,8 +110,11 @@ pub enum ImageGlobalFlagsError {
     #[error("could not create registry key for `{image}`")]
     CreateKey { image: ImageFile, source: io::Error },
 
-    #[error("could not access `GlobalFlag` value of registry key for `{image}`")]
-    AccessValue { image: ImageFile, source: io::Error },
+    #[error("could not get `GlobalFlag` value of registry key for `{image}`")]
+    GetValue { image: ImageFile, source: io::Error },
+
+    #[error("could not set `GlobalFlag` value of registry key for `{image}`")]
+    SetValue { image: ImageFile, source: io::Error },
 }
 
 const GFLAGS_KEY_NAME: &str = "GlobalFlag"; // Singular
@@ -133,7 +136,7 @@ impl ImageGlobalFlags {
         let value = self
             .create_key()?
             .get_value(GFLAGS_KEY_NAME)
-            .map_err(|source| ImageGlobalFlagsError::AccessValue {
+            .map_err(|source| ImageGlobalFlagsError::GetValue {
                 source,
                 image: self.image.clone(),
             })?;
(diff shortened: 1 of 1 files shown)


src/agent/dynamic-library/src/windows.rs
```diff
@@ -193,15 +196,19 @@ impl ImageLoaderSnapsGuard {
     }
 
     fn enable(&self) -> Result<(), ImageGlobalFlagsError> {
-        let mut value = self.gflags.get_value()?;
+        // Value may not yet exist.
+        let mut value = self.gflags.get_value().unwrap_or(0x0);
```
ranweiler, line 200: Ideally `get_value()` would return an `Option`, but it doesn't, and the error is just a normal `io::Error`. If we really can't do registry I/O at all, we'll fail earlier (or in `set_value()`).

src/agent/dynamic-library/src/windows.rs
```diff
@@ -143,8 +146,8 @@ impl ImageGlobalFlags {
 
     pub fn set_value(&self, value: u32) -> Result<(), ImageGlobalFlagsError> {
         self.create_key()?
-            .set_value(GFLAGS_KEY_NAME, &value)
-            .map_err(|source| ImageGlobalFlagsError::AccessValue {
+            .set_value(GFLAGS_VALUE_NAME, &value)
```
chkeita, line 149: Why the naming change ? It looks like it is being used a key here

ranweiler: I'm glad I'm not the only one who finds that weird 😅. I'm aligning with the `winreg` API and official registry docs. In registry jargon, a "key" is a path-like object. The object that you and I would call a "key" is really called a "value name", and what we'd want to call the "value" is the "value data".

So to compare registry naming conventions vs. more common naming conventions:
| Registry | Common |
| ------- | -------|
| Key | Path |
| Value Name | Key |
| Value Data | Value |

See the worked example [here](https://docs.microsoft.com/en-us/windows/win32/sysinfo/structure-of-the-registry), as well as this comment:
> Each value consists of a value name and its associated data, if any.

chkeita (approved):

tevoinea (approved):

Issue #1939, opened 2022-05-13
Add timestamps to agent logs


tevoinea:
We now upload agent logs to a storage account. This is an example log: `INFO: starting regression task`.

We should prepend a timestamp (in UTC) to each logs message to make tracing easier.

PR #1972, opened 2022-05-24
Add timestamp to agent log

tevoinea:

_What is this about?_

Adds timestamp to agent tracing log output.

## PR Checklist
* [x] Applies to work item: #1939 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

* Rename `LogEvent` -> `LoggingEvent`
* Create `LogTrace` and `LogEvent` structs to organize necessary logging info https://github.com/microsoft/onefuzz/pull/1972/files#diff-52c53d4179505ba0dd3f46e2707afeb66f5208e6138b52159fd48134af41a73fR340-R356
* All agent logs now prepend the UTC timestamp
* Test to validate the behavior

## Validation Steps Performed

Ran check-pr, checked the blob logs and the AI logs.

### DIFF
--- a/src/agent/onefuzz-telemetry/src/lib.rs
+++ b/src/agent/onefuzz-telemetry/src/lib.rs
@@ -11,6 +11,7 @@ use uuid::Uuid;
 use z3_sys::ErrorCode as Z3ErrorCode;
 
 pub use appinsights::telemetry::SeverityLevel::{Critical, Error, Information, Verbose, Warning};
+pub use chrono::Utc;
 use tokio::sync::broadcast::{self, Receiver};
 #[macro_use]
 extern crate lazy_static;
@@ -631,7 +632,7 @@ macro_rules! log {
     ($level: expr, $($arg: tt)+) => {{
         let log_level = onefuzz_telemetry::to_log_level(&$level);
         if log_level <= log::max_level() {
-            let msg = format!("{}", format_args!($($arg)+));
+            let msg = format!("[{}] {}", onefuzz_telemetry::Utc::now(), format_args!($($arg)+));
             log::log!(log_level, "{}", msg);
             onefuzz_telemetry::try_broadcast_trace(msg.to_string(), log_level);
             onefuzz_telemetry::log_message($level, msg.to_string());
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -681,3 +682,21 @@ macro_rules! metric {
         client.track_metric($name.into(), $value);
     }};
 }
+
+#[cfg(test)]
+mod tests {
+    use crate as onefuzz_telemetry;
```
tevoinea, line 688: I didn't `use super::*;` here because the crate name needs to be `onefuzz_telemetry` in order for the macro to expand properly.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -631,7 +632,7 @@ macro_rules! log {
     ($level: expr, $($arg: tt)+) => {{
         let log_level = onefuzz_telemetry::to_log_level(&$level);
         if log_level <= log::max_level() {
-            let msg = format!("{}", format_args!($($arg)+));
+            let msg = format!("[{}] {}", onefuzz_telemetry::Utc::now(), format_args!($($arg)+));
             log::log!(log_level, "{}", msg);
```
chkeita, line 636: we also need a timestamp on the events. check `log_events`

chkeita (approved):

chkeita:
Adding the timestamp directly in the message might be redundant with app insight since everything gets a timestamp there. If the goal is to have it in the written logs we might want to format the message in the task logger instead

stishkin:
Any chance we don't write stuff to AppInisghts from the Agent ? Maybe only errors ? And the rest of the logs are kept in files ? 

There is so much noise in App Inisghts right now, is it required ?

tevoinea:
> Any chance we don't write stuff to AppInisghts from the Agent ? Maybe only errors ? And the rest of the logs are kept in files ?
> 
> There is so much noise in App Inisghts right now, is it required ?

We definitely could start removing stuff from AI after this change makes it in. I would propose moving traces to log file only and keep events and errors in AI as a first step. I'll create an issue

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -534,25 +551,33 @@ pub fn format_events(events: &[EventData]) -> String {
         .join(" ")
 }
 
-fn try_broadcast_event(event: &Event, properties: &[EventData]) -> bool {
+fn try_broadcast_event(timestamp: &DateTime<Utc>, event: &Event, properties: &[EventData]) -> bool {
```
chkeita, line 554: passing by reference here is not necessary since `DateTime` it implements the copy trait

chkeita (approved):

Issue #1948, opened 2022-05-16
Add command line parameter(s) to deploy.py to enable specific dotnet functions


AdamL-Microsoft:
OneFuzz deploy scripts setup both the python and dotnet Azure function apps, with all python functions enabled and dotnet all disabled currently.

To facilitate testing while we're transitioning to the dotnet refactored versions of these functions we'll want a way to selectively enable dotnet and disable the equivalent function in python at deployment time.

PR #1973, opened 2022-05-24
adding function app settings bicep template and updating deploy.py 

AdamL-Microsoft:

This adds app settings for all dotnet and python functions to automatically disable (AzureWebJobs.$FUNCTION.Disabled=1) all dotnet functions and enable all python functions in their corresponding function apps

To enable dotnet functions at deployment time, deploy.py was updated with an argument '--enable_dotnet' that accepts a space separated list of <del>dotnet</del> python functions to disable while enabling the dotnet function of the same name (except pascal cased names for dotnet vs python snake cased names)

example: 
```bash
./deploy.py --enable_dotnet tasks webhooks pool
```
This would enable the Tasks, Webhooks, and Pool functions in dotnet and turn these off in python

## PR Checklist
* [x] Applies to work item: #1948
* [x] Tests passed

#### added:
src/deployment/bicep-templates/function-settings-disabled-apps.bicep
#### updated:
src/deployment/azuredeploy.bicep
src/deployment/bicep-templates/function-settings.bicep
src/deployment/deploy.py

### DIFF
--- a/src/deployment/azuredeploy.bicep
+++ b/src/deployment/azuredeploy.bicep
@@ -24,6 +24,9 @@ param diagnosticsLogLevel string = 'Verbose'
 var log_retention = 30
 var tenantId = subscription().tenantId
 
+var python_functions_disabled = '0'
+var dotnet_functions_disabled = '1'
+
 var scaleset_identity = '${name}-scalesetid'
 
 var StorageBlobDataReader = '2a2b9908-6ea1-4ae2-8e65-a410df84e7d1'
@@ -261,6 +264,7 @@ module pythonFunctionSettings 'bicep-templates/function-settings.bicep' = {
     keyvault_name: keyVaultName
     monitor_account_name: operationalInsights.outputs.monitorAccountName
     multi_tenant_domain: multi_tenant_domain
+    functions_disabled: python_functions_disabled
   }
   dependsOn: [
     pythonFunction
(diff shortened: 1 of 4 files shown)


src/deployment/bicep-templates/function-settings-disabled-apps.bicep
```diff
@@ -0,0 +1,92 @@
+param functions_disabled_setting string
+
+var allFunctions = [
+  'agent_can_schedule'    //0
```
chkeita, line 4: make sure the to use the PascalCase version of these function for the dotnet code

src/deployment/azuredeploy.bicep
```diff
@@ -24,6 +24,9 @@ param diagnosticsLogLevel string = 'Verbose'
 var log_retention = 30
 var tenantId = subscription().tenantId
 
+var python_functions_disabled = '0'
+var dotnet_functions_disabled = '1'
```
stishkin, line 28: dotnet_functions_enabled ?

src/deployment/bicep-templates/function-settings-disabled-apps.bicep
AdamL-Microsoft: Good catch, fixed this in commit #24c9fd8

src/deployment/deploy.py
```diff
@@ -1238,7 +1304,14 @@ def main() -> None:
         nargs="*",
         help="Set additional AAD tenants beyond the tenant the app is deployed in",
     )
-
+    parser.add_argument(
+        "--enable_dotnet",
```
chkeita, line 1308: You should be able to parse the functions to a list instead of a string. check `--apply_migrations`.  But it might change the syntax. 
Feel free to ignore this suggestion.

chkeita (approved):

src/deployment/deploy.py
AdamL-Microsoft: Makes sense to keep the argument inputs uniform and this wouldn't require changing much of the previous code.

PR #2004, opened 2022-06-01
Properly compare job/task state enums in `JobHelper.check_current_job()`

DrChat:

`JobHelper.check_current_job()` currently always reports a successful status because it compares state enums with string literals, which will never succeed (confusingly).
In addition, by default `self.onefuzz.tasks.list` will only list active tasks (excluding ones that are in a shutting down state).

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

This simply removes string literal comparisons in favor of lists returned by the state object itself.

## Validation Steps Performed

Unsure - I don't know if there's any commands that can easily be ran to check this functionality.
This should be a simple enough code review.

### DIFF
--- a/src/cli/onefuzz/templates/__init__.py
+++ b/src/cli/onefuzz/templates/__init__.py
@@ -9,7 +9,7 @@ import zipfile
 from typing import Any, Dict, List, Optional, Tuple
 from uuid import uuid4
 
-from onefuzztypes.enums import OS, ContainerType, TaskState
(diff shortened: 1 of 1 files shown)


Porges (approved):

chkeita (approved):

Porges:
Will merge after validation run.

Porges:
Thanks for the contribution, @DrChat!

PR #2017, opened 2022-06-03
flushing the logs when the task is done

### LINKED ISSUE #2011: Flush log files in the agent before exiting the task
When the [Task exits](https://github.com/microsoft/onefuzz/blob/c24193f7e909800138cad90527badb78de37ab9d/src/agent/onefuzz-task/src/managed/cmd.rs#L44). Some of the logs accumulated by the task logger could be lost if they are not pushed to the server. 
This issue is about flushing those logs before we close the task

AB#39955753

chkeita:

adds a `flush_and_stop` function to the task logger to allow it write unprocessed logs before closing.

closes #2011


src/agent/onefuzz-task/src/tasks/task_logger.rs
```diff
@@ -254,127 +263,204 @@ impl TaskLogger {
...
+            } => {
+                match context
+                    .log_writer
+                    .write_logs(&context.pending_logs[start..start + count])
+                    .await?
+                {
+                    WriteLogResponse::Success => {
+                        if start + count >= context.pending_logs.len() {
+                            if flush {
+                                bail!("done");
```
tevoinea, line 283: Would `LoopState::Done` be more appropriate here? I know we break the loop for both Err and LoopState::Done:
https://github.com/microsoft/onefuzz/blob/853b8e99c599070479d87de31ca2f03f81f8280b/src/agent/onefuzz-task/src/tasks/task_logger.rs#L431-L443

But I think flushing the logs is a normal operation and not an error.

chkeita: good catch. This was supposed to be return the done state

tevoinea (approved):

PR #2027, opened 2022-06-08
Make Clippy run on all targets in CI

Porges:

While I'm browsing around the code clippy is shouting at me, so here's a janitorial exercise to clean up clippy lints and become a little more familiar with the codebase.

## Info on Pull Request

Fix all outstanding clippy lints. Fixes are almost mechanical, although I didn't automate them.


Porges:
@chkeita pointed out we are already gating on this in `ci/agent.sh` so it should be picking up any warnings. After investigation, `rust-analyzer` is checking all targets, so we will add `--all-targets` to the CI run.

src/ci/agent.sh
```diff
@@ -51,7 +51,7 @@ cargo fmt -- --check
 cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2020-0016 --ignore RUSTSEC-2020-0036 --ignore RUSTSEC-2019-0036 --ignore RUSTSEC-2021-0065 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0077
 cargo-license -j > data/licenses.json
 cargo build --release --locked
-cargo clippy --release -- -D warnings
+cargo clippy --release --all-targets -- -D warnings
```
chkeita, line 54: there is another clippy call in proxy.sh

chkeita (approved):

PR #2032, opened 2022-06-09
Convert `agent_events` to C# & add tests

### LINKED ISSUE #2031: C# conversion of `agent_events`
Convert the `agent_events` Azure Function to C#, and release to production.

This Function handles events emitted by the OneFuzz agent from nodes.

AB#39955736

Porges:
Depends on merging #2055 first.

Fixes #2031.

### Function implementation

Added implementation of `agent_events` function and some basic tests.

Fixed several issues encountered in `TestOperations`.

### Additional JsonConverter

The existing Python code for `agent_events` figures out which class to use only based upon the shape of the input, without a discriminator. I have added an additional `JsonConverter` named `SubclassConverter<T>` which will pick a subclass of `T` to deserialize into based upon what properties exist in the JSON.

### Enum helpers

Converted some enum helpers to extension methods to make them a bit more readable.

## PR Checklist
* [x] Tests added/passed


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -251,6 +249,13 @@ public record Task(
     List<TaskEventSummary> Events { get; set; } = new List<TaskEventSummary>();
     List<NodeAssignment> Nodes { get; set; } = new List<NodeAssignment>();
 }
+
+public record TaskEvent(
+    [PartitionKey] Guid TaskId,
```
tevoinea, line 254: This should be `[PartitonKey, RowKey]`, here's an example: https://github.com/microsoft/onefuzz/blob/cb1d89fce67f2b6f96c4d9c45b19d3f11213eee7/src/ApiService/ApiService/OneFuzzTypes/Model.cs#L309

src/ApiService/ApiService/ApiService.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.17.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.SignalRService" Version="1.7.0" />
+    <PackageReference Include="TaskTupleAwaiter" Version="2.0.0" />
```
tevoinea, line 37: I didn't see this used anywhere

src/ApiService/ApiService/OneFuzzTypes/Model.cs
tevoinea: The "ORM" requires both to be present to deserialize the entity properly from azure table. If you have a test deployment you can look at the `func{random_letters}` storage account in the `TaskEvent` table to see that the `PartitionKey` and `RowKey` columns have the exact same data.

Porges: Thanks!

src/ApiService/ApiService/ApiService.csproj
Porges: It adds into the `System.Threading.Tasks` namespace; allows the usage as seen in `OnWorkerEventDone` where you can do:

```csharp
var (task, node) = await (
    _context.TaskOperations.GetByTaskId(done.TaskId),
    _context.NodeOperations.GetByMachineId(machineId));
```

this will run the tasks concurrently so that they can hit Azure Storage in parallel.

src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
```diff
@@ -1,16 +1,45 @@
...
     public System.Threading.Tasks.Task Stopping(Repro repro);
 
     public IAsyncEnumerable<Repro> SearchStates(IEnumerable<VmState>? States);
+
+    public IAsyncEnumerable<Repro> SearchByTimeRange((DateTimeOffset min, DateTimeOffset max) timeRange);
+
+    public Async.Task<ResultVoid<(int, string)>> Replace(Repro updatedRepro);
+}
+
+public interface IReproOrm : IOrm<Repro> {
```
chkeita, line 18: I don't understand the need for this indirection. The IOrm and IStatefulOrm are meant to abstract the operations on the each specific entities. The `IStatefulOrm` is just only adds state management to the existing operations.

chkeita:
This PR is getting big. Can we split this up ?

tevoinea (approved):

PR #2033, opened 2022-06-09
Update `appinsights` to released version

Porges:
I noticed that there was a note to update `appinsights` to a released version once one was available, and now one is.

The `close` function was changed to be `async` in the latest version, so `try_flush_and_close` must also become async. Luckily, the three places we use this already have an async context available.

## PR Checklist
* [x] ~~Applies to work item: #xxx~~
* [x] Tests added/passed
* [x] ~~Requires documentation to be updated~~
* [x] ~~I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx~~

## Validation Steps Performed

Standard test run.

### DIFF
--- a/src/agent/onefuzz-telemetry/src/lib.rs
+++ b/src/agent/onefuzz-telemetry/src/lib.rs
@@ -66,7 +66,7 @@ impl fmt::Display for InstanceTelemetryKey {
     }
 }
 
-pub type TelemetryClient = appinsights::TelemetryClient;
+pub type TelemetryClient = appinsights::blocking::TelemetryClient;
 pub enum ClientType {
     Instance,
     Microsoft,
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -66,7 +66,7 @@ impl fmt::Display for InstanceTelemetryKey {
     }
 }
 
-pub type TelemetryClient = appinsights::TelemetryClient<appinsights::InMemoryChannel>;
+pub type TelemetryClient = appinsights::TelemetryClient;
```
Porges, line 69: Note that the channel is no longer explicit in the type. However, the default is the `InMemoryChannel`.

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -475,17 +475,17 @@ pub fn set_appinsights_clients(
...
+pub async fn try_flush_and_close() {
     let clients = global::take_clients();
 
     for client in clients {
-        client.flush_channel();
-        client.close_channel();
+        // close_channel performs an implicit flush
+        client.close_channel().await;
     }
 }
```
tevoinea, line 490: Maybe out of scope for this PR but now that this is async we could do this in parallel for some free perf!

```rust
tokio::join!(clients
        .iter()
        .map(|client| tokio::spawn(client.close_channel().await)));
```

tevoinea (approved):

src/agent/onefuzz-telemetry/src/lib.rs
chkeita: I don't think we gain a lot of perf with the move to async. This is meant to be a terminal call anyway. 
And it also introduce some weird asymetry. 
just like in the change in `src/agent/onefuzz-agent/src/main.rs`: the logging is creating without a runtime but we need a runtime to flush it. The non async version fit that scenario better. 
I propose we have both version async and non async and use them depending on if we are in an async context or not

Porges:
Validated with `check-pr`.

PR #2038, opened 2022-06-10
Add scripts to install dotnet on windows and ubuntu

### LINKED ISSUE #2005: Install `dotnet` and tools on VMs
Install the the .NET 6.0 runtime and CLI diagnostic global tools on fuzzing VMs. Do this for both Linux and Windows VMs.

Should include:
- `dotnet`
- [`dotnet-dump`](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dump)
- [`dotnet-coverage`](https://docs.microsoft.com/en-us/dotnet/core/additional-tools/dotnet-coverage)
- [`dotnet-sos`](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-sos) (might get via `dotnet-dump`?)

**TBD:** how should `dotnet` installation be bootstrapped? Once installed, should we install the diagnostic tools via `dotnet tool install --global`, or some other way? 

AB#39956292

tevoinea:

_What is this about?_

This PR adds scripts to install dotnet and related tooling on ubuntu and windows.

## PR Checklist
* [ ] Closes #2005 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

These scripts were written with the following goals:
* A small footprint regarding system configuration (environment variables, files in 'standard' locations)
* Ease of use by the agent

I made the tradeoff by setting the `DOTNET_ROOT` environment variable/adding /etc/dotnet/install_location on linux in order to make it easier to call the dotnet tools from the agent.

The alternative approach is that every time we call the dotnet tools from the agent, we must set the DOTNET_ROOT env var for that process every time.

I'm open to both approaches, the reason I chose the current implementation is because there's no way to indicate on the agent that the env var is required so it's easy to make mistakes.

## Validation Steps Performed

* I spun up a fresh ubuntu docker image and ran the script
* I spun up a fresh Win10 VM and ran the script

### DIFF
--- a/src/runtime-tools/linux/install-dotnet.sh
+++ b/src/runtime-tools/linux/install-dotnet.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+apt update
+apt install curl libicu-dev -y
+
+mkdir /onefuzz/tools/dotnet
+
+curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
+chmod u+x dotnet-install.sh
+. ./dotnet-install.sh --version 6.0.300 --install-dir /onefuzz/tools/dotnet
+
+mkdir /etc/dotnet
+touch /etc/dotnet/install_location
+echo "/onefuzz/tools/dotnet" >> /etc/dotnet/install_location
+
+dotnet tool install dotnet-dump --tool-path /onefuzz/tools
+dotnet tool install dotnet-coverage --tool-path /onefuzz/tools
+dotnet tool install dotnet-sos --tool-path /onefuzz/tools
--- a/src/runtime-tools/win64/install-dotnet.ps1
+++ b/src/runtime-tools/win64/install-dotnet.ps1
@@ -1,0 +1,19 @@
+# Install dotnet
+&powershell -NoProfile -ExecutionPolicy unrestricted -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; &([scriptblock]::Create((Invoke-WebRequest -UseBasicParsing 'https://dot.net/v1/dotnet-install.ps1'))) -Version 6.0.300 -InstallDir c:/onefuzz/tools/dotnet"
+
+# Set the variable for this current sessions
+$env:DOTNET_ROOT = "c:/onefuzz/tools/dotnet"
+# Set the variable for all future sessions
+[Environment]::SetEnvironmentVariable("DOTNET_ROOT", "c:/onefuzz/tools/dotnet", "Machine")
+
+# Install the necessary tools
+cd c:/onefuzz/tools/dotnet
+./dotnet.exe tool install dotnet-dump --tool-path c:/onefuzz/tools
+./dotnet.exe tool install dotnet-coverage --tool-path c:/onefuzz/tools
+./dotnet.exe tool install dotnet-sos --tool-path c:/onefuzz/tools
+
+# Verify they're working
+cd c:/onefuzz/tools
+./dotnet-dump.exe -h
+./dotnet-coverage.exe -h
+./dotnet-sos.exe -h


src/runtime-tools/win64/install-dotnet.ps1
```diff
@@ -0,0 +1,19 @@
+# Install dotnet
+&powershell -NoProfile -ExecutionPolicy unrestricted -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; &([scriptblock]::Create((Invoke-WebRequest -UseBasicParsing 'https://dot.net/v1/dotnet-install.ps1'))) -Version 6.0.300 -InstallDir c:/onefuzz/tools/dotnet"
+
+# Set the variable for this current sessions
+$env:DOTNET_ROOT = "c:/onefuzz/tools/dotnet"
+# Set the variable for all future sessions
+[Environment]::SetEnvironmentVariable("DOTNET_ROOT", "c:/onefuzz/tools/dotnet", "Machine")
```
chkeita, line 7: this environment variable should be set in onefuzz.ps1.

src/runtime-tools/linux/install-dotnet.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
chkeita, line 1: don't you need to call these from setup.sh and setup.ps1 ?

tevoinea: Ok! That's way simpler I'll make the updates

chkeita (approved):

src/runtime-tools/linux/setup.sh
```diff
@@ -122,6 +124,27 @@ if type apt > /dev/null 2> /dev/null; then
...
     fi
+
+    # Install dotnet
+    until sudo apt install -y curl libicu-dev; do
+        echo "apt failed, sleeping 10s then retrying"
+        sleep 10
+    done
+
+    echo "downloading dotnet install"
+    curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh > /dev/null
```
Porges, line 135: If we are retrying the `apt install`  we could put a `--retry X` here too

src/runtime-tools/linux/setup.sh
```diff
@@ -26,6 +27,7 @@ mkdir -p /onefuzz/setup
 mkdir -p /onefuzz/tools
 mkdir -p /onefuzz/etc
 mkdir -p /onefuzz/instance-specific-setup
+mkdir -p /onefuzz/tools/dotnet
```
Porges, line 30:
```suggestion
mkdir -p "$DOTNET_ROOT"
```

reuse existing variable

src/runtime-tools/linux/setup.sh
```diff
@@ -122,6 +124,27 @@ if type apt > /dev/null 2> /dev/null; then
...
+    echo "downloading dotnet install"
+    curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh > /dev/null
+    chmod +x dotnet-install.sh
+
+    echo "running dotnet install"
+    . ./dotnet-install.sh --version 6.0.300 --install-dir /onefuzz/tools/dotnet 2>&1 | logger -s -i -t 'onefuzz-dotnet-setup'
+    rm dotnet-install.sh
+
+    echo "install dotnet tools"
+    pushd /onefuzz/tools/dotnet
```
Porges, line 143:
```suggestion
    pushd "$DOTNET_ROOT"
```

src/runtime-tools/linux/setup.sh
```diff
@@ -122,6 +124,27 @@ if type apt > /dev/null 2> /dev/null; then
...
+        echo "apt failed, sleeping 10s then retrying"
+        sleep 10
+    done
+
+    echo "downloading dotnet install"
+    curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh > /dev/null
+    chmod +x dotnet-install.sh
+
+    echo "running dotnet install"
+    . ./dotnet-install.sh --version 6.0.300 --install-dir /onefuzz/tools/dotnet 2>&1 | logger -s -i -t 'onefuzz-dotnet-setup'
```
ranweiler, line 139: We could use a variable for the `dotnet` version.

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -6,6 +6,8 @@ $env:ONEFUZZ_ROOT = "C:\onefuzz"
 $env:ONEFUZZ_TOOLS = "C:\onefuzz\tools"
 $env:LLVM_SYMBOLIZER_PATH = "llvm-symbolizer"
 $env:RUST_LOG = "info"
+$env:DOTNET_ROOT = "c:\onefuzz\tools\dotnet"
+[Environment]::SetEnvironmentVariable("DOTNET_ROOT", "c:\onefuzz\tools\dotnet", "Machine")
```
ranweiler, line 10: I'm guessing it wasn't immediately obviously that it was necessary to set this at machine scope and also set it in the current session. Maybe worth a comment?

Also, consider reusing the session's env var when setting the machine-scoped one, just to emphasize that they're identical. I.e.
```suggestion
[Environment]::SetEnvironmentVariable("DOTNET_ROOT", $env:DOTNET_ROOT, "Machine")
```

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -168,3 +170,16 @@ function Install-VCRedist {
   Start-Process -FilePath C:\onefuzz\vcredist_x86.exe -ArgumentList "/install /q /norestart" -Wait -WindowStyle Hidden
   log "installing VC Redist: done"
 }
+
+function Install-Dotnet {
+  log "Installing dotnet"
+  &powershell -NoProfile -ExecutionPolicy unrestricted -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; &([scriptblock]::Create((Invoke-WebRequest -UseBasicParsing 'https://dot.net/v1/dotnet-install.ps1'))) -Version 6.0.300 -InstallDir c:\onefuzz\tools\dotnet"
```
ranweiler, line 176: We don't really need to use the one-liner version of this, right? That's probably more a convenience for users from the docs. It also has downsides: it's sort fussy to obvious parameters to this function (`$Version`, `$InstallDir`) without having to worry about string interpolation stuff.

Can we just explode this into multiple lines, and maybe parameterize the containing function so the end result gets used like:
```powershell
Install-Dotnet -Version $env:DOTNET_VERSION -InstallDir $env:DOTNET_ROOT
```

PR #2059, opened 2022-06-16
Use rust-cache action for Rust build cache

Porges:
This has some [smarts](https://github.com/Swatinem/rust-cache#cache-details) built in about what should and shouldn't be cached, which 1) removes our need to specify exactly what should be cached and 2) helps stop the cache size growing over time.


Porges:
Abandoning since I don't think this gains us much, and the new caching of binaries improves build times much more.

PR #2066, opened 2022-06-20
Adding unit tests for the scheduler

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/deployment/deploy.py
```diff
@@ -152,7 +152,7 @@ def __init__(
         subscription_id: Optional[str],
         admins: List[UUID],
         allowed_aad_tenants: List[UUID],
-        enable_dotnet: List[str],
+        enable_dotnet: List[str] = ["timer_tasks"],
```
chkeita, line 155: self reminder to remove this after testing

src/ApiService/Tests/SchedulerTests.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        foreach (var task in tasks) {
+            var seen = false;
+            foreach (var bucket in buckets) {
+                if (bucket.Contains(task)) {
+                    Assert.False(seen);
+                    seen = true;
+                }
+            }
+            Assert.True(seen);
```
Porges, line 137: use of `Assert.Single` can simplify this:

```suggestion
            Assert.Single(buckets, b => b.Contains(task));
```

src/ApiService/Tests/Integration/AzuriteStorage.cs
```diff
@@ -22,15 +22,11 @@ public Uri GetTableEndpoint(string _accountId)
     const string AccountName = "devstoreaccount1";
     const string AccountKey = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==";
 
-    public Task<(string?, string?)> GetStorageAccountNameAndKey(string _accountId)
-        => Async.Task.FromResult<(string?, string?)>((AccountName, AccountKey));
+    public Task<(string, string)> GetStorageAccountNameAndKey(string accountId)
+        => Async.Task.FromResult((accountId, AccountKey));
```
Porges, line 26: the test class here should ignore the `accountId`, it will always use the same account name

```suggestion
        => Async.Task.FromResult((AccountName, AccountKey));
```

src/ApiService/ApiService/onefuzzlib/JobOperations.cs
```diff
@@ -76,7 +88,12 @@ public class JobOperations : StatefulOrm<Job, JobState>, IJobOperations {
             await _context.Events.SendEvent(new EventJobStopped(job.JobId, job.Config, job.UserInfo, taskInfo));
         }
 
-        await Replace(job);
+        var result = await Replace(job);
 
+        if (result.IsOk) {
+            return job;
+        } else {
+            throw new Exception($"Failed to save job {job.JobId}");
```
tevoinea, line 96: I understand not wanting to directly expose the error from the ORM, but shouldn't we return our OnefuzzResult type instead of throwing?

tevoinea (approved):

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/JobOperations.cs
chkeita: i added the error from the orm to the exception message. 
In general i am not against the idea of returning a Result instead of of throwing but we would need to change all method to do the same to be consistent which is out of scope for the pr.

src/ApiService/ApiService/TimerTasks.cs
```diff
@@ -21,8 +21,8 @@ public class TimerTasks {
         _scheduler = scheduler;
     }
 
-    //[Function("TimerTasks")]
-    public async Async.Task Run([TimerTrigger("1.00:00:00")] TimerInfo myTimer) {
+    [Function("TimerTasks")]
```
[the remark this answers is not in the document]
chkeita: not necessary since the function is disabled by default

src/ApiService/ApiService/OneFuzzTypes/Validated.cs
```diff
@@ -63,10 +61,14 @@ public record PoolName : ValidatedString {
     }
 
     public static bool TryParse(string input, [NotNullWhen(returnValue: true)] out PoolName? result) {
-        if (!Check.IsAlnumDash(input)) {
-            result = default;
-            return false;
-        }
+
+        // bypassing the validation because this code has a stricter validation than the python equivalent
```
chkeita, line 65: @Porges  FYI, I had to disable this validation because is stricter that the python code. 
In python you can create a pool names "pool_linux" (with underscore). which should not work see #2080. But with this code, The pool name cannot be retrieved from the DB.

PR #2071, opened 2022-06-21
Updating AutoScale Feature to Sync Azure Table.

nharper285:

_What is this about?_
Currently, AutoScale table entries are not synced with the current state of the VMSS for a OneFuzz instance. This includes config updates and deletes. This PR addresses this. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
+++ b/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
@@ -87,7 +87,7 @@ def add_auto_scale_to_vmss(
     return None
 
 def get_auto_scale_profile(scaleset_id: UUID) -> AutoscaleProfile:
-    logging.info("Getting scaleset %s for existing auto scale resources" % vmss)
+    logging.info("Getting scaleset %s for existing auto scale resources" % scaleset_id)
     client = get_monitor_client()
     resource_group = get_base_resource_group()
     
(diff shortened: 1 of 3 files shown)


src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -86,6 +86,32 @@ def add_auto_scale_to_vmss(
...
+    except (ResourceNotFoundError, CloudError):
+        return Error(
+            code=ErrorCode.INVALID_CONFIGURATION,
+            errors=[
+                "Failed to query scaleset %s autoscale resource"
+                % vmss
+            ],
+        )
+          
+    return auto_scale_resource.AutoscaleProfile
```
tevoinea, line 113: `auto_scale_resource` can still be None here if the `list_by_resource_group` call succeeds but we simply don't find an auto scale resource associated with the scaleset.

src/api-service/__app__/onefuzzlib/workers/scalesets.py
```diff
@@ -875,6 +876,52 @@ def set_state(self, state: ScalesetState) -> None:
...
+        for rule in auto_scale_profile.rules:
+            scale_action = rule.scale_action
+            if scale_action.direction == "Increase":
+                scale_out_amount = scale_action.value
+                scale_out_cooldown = scale_action.cooldown
+            else if scale_action.direction == "Decrease":
+                scale_in_amount = scale_action.value
+                scale_in_cooldown = scale_action.cooldown
+            else:
+                pass
```
tevoinea, line 909: We have a pretty 'standardized' auto scaling profile across all of our scalesets so the assumptions you made here are true. But I think if something changes (intentionally or not) in our auto scale profiles and we don't convert as expected, we could benefit from having a warning here that we would hopefully pick up in telemetry.

Issue #2068, opened 2022-06-21
Refactor `node` to C#


Porges:
Implement the `node` Azure Function

AB#40113973

Porges:
Ready for prod, closing

PR #2072, opened 2022-06-22
Implement the `node` C# function

Porges:

1. Ports the `node` function from Python to C#.
2. Adds a missing authentication check.
3. Add validated string type `PoolName` for consistency with Python version.

## PR Checklist
* [x] Applies to work item: #2068 
* [x] Tests added/passed
* ~~Requires documentation to be updated~~
* ~~I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx~~

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -3,7 +3,6 @@ using System.Text.Json.Serialization;
 using Microsoft.OneFuzz.Service.OneFuzzLib.Orm;
 using Endpoint = System.String;
 using GroupId = System.Guid;
-using PoolName = System.String;
 using PrincipalId = System.Guid;
 using Region = System.String;
 
@@ -151,7 +150,6 @@ public record Proxy
 public record Error(ErrorCode Code, string[]? Errors = null);
 
 public record UserInfo(Guid? ApplicationId, Guid? ObjectId, String? Upn);
-
 
 public record TaskDetails(
     TaskType Type,
@@ -316,11 +314,11 @@ public record InstanceConfig
     NetworkSecurityGroupConfig ProxyNsgConfig,
     AzureVmExtensionConfig? Extensions,
     string ProxyVmSku,
-    IDictionary<Endpoint, ApiAccessRule>? ApiAccessRules,
-    IDictionary<PrincipalId, GroupId[]>? GroupMembership,
-
-    IDictionary<string, string>? VmTags,
-    IDictionary<string, string>? VmssTags
+    IDictionary<Endpoint, ApiAccessRule>? ApiAccessRules = null,
+    IDictionary<PrincipalId, GroupId[]>? GroupMembership = null,
+    IDictionary<string, string>? VmTags = null,
+    IDictionary<string, string>? VmssTags = null,
+    bool? RequireAdminPrivileges = null
 ) : EntityBase() {
     public InstanceConfig(string instanceName) : this(
         instanceName,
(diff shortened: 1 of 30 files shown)


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -316,11 +314,11 @@ public record InstanceConfig
...
-    IDictionary<Endpoint, ApiAccessRule>? ApiAccessRules,
-    IDictionary<PrincipalId, GroupId[]>? GroupMembership,
-
-    IDictionary<string, string>? VmTags,
-    IDictionary<string, string>? VmssTags
+    IDictionary<Endpoint, ApiAccessRule>? ApiAccessRules = null,
+    IDictionary<PrincipalId, GroupId[]>? GroupMembership = null,
+    IDictionary<string, string>? VmTags = null,
+    IDictionary<string, string>? VmssTags = null,
+    bool? RequireAdminPrivileges = null
```
Porges, line 321: This was left out of the original C# version. Are we looking to remove/deprecate this functionality?

tevoinea: It's possible it was added to the python side after the C# version was created

src/ApiService/ApiService/OneFuzzTypes/Responses.cs
```diff
@@ -15,6 +18,23 @@ public record PendingNodeCommand(
     NodeCommandEnvelope? Envelope
 ) : BaseResponse();
 
+// TODO: not sure how much of this is actually
+// needed in the search results, so at the moment
+// it is a copy of the whole Node type
```
tevoinea, line 23: Could we turn `BaseResponse` into an interface, add it on the existing `Node` record in `Model.cs`? The last thing to figure out would be the BaseResponse JsonConverter

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
```diff
@@ -11,6 +11,9 @@ public static class Query {
         public static string PartitionKey(string partitionKey)
             => TableClient.CreateQueryFilter($"PartitionKey eq {partitionKey}");
 
+        public static string PartitionKey(Guid partitionKey)
+            => TableClient.CreateQueryFilter($"PartitionKey eq {partitionKey}");
+
```
tevoinea, line 16: What about something like:
```csharp
public static string PartitionKey<T>(T partitionKey)
            where T : IFormattable
            => TableClient.CreateQueryFilter($"PartitionKey eq {partitionKey}");
```

That way we can also add it to `PoolName` for example.

tevoinea (approved):

src/ApiService/ApiService/OneFuzzTypes/Responses.cs
Porges: I think there's probably going to be things in `Node` (now or in future) that we don't want to reflect into the response here?

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
Porges: Also had this idea but hadn’t gotten around to it 👍

src/ApiService/ApiService/OneFuzzTypes/Responses.cs
Porges: I’ve also been bitten a lot of times in the past by sharing internal models as external ones, so I think I'll leave it like this for now.

PR #2078, opened 2022-06-23
Implement the `containers` function for C#

Porges:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/Containers.cs
```diff
@@ -196,4 +265,20 @@ public class Containers : IContainers {
         var client = await FindContainer(container, storageType) ?? throw new Exception($"unable to find container: {container.ContainerName} - {storageType}");
```
stishkin, line 265: should prefix be part of the log message  when logging container name ?

src/ApiService/ApiService/onefuzzlib/Containers.cs
```diff
@@ -185,7 +254,7 @@ public class Containers : IContainers {
 
         var sasBuilder = new BlobSasBuilder(permissions, endTime) {
             StartsOn = startTime,
-            BlobContainerName = container.ContainerName,
+            BlobContainerName = _config.OneFuzzStoragePrefix + container.ContainerName,
```
stishkin, line 257: this will be a source of hard to debug bugs forgetting to add prefix to container name. 

Can enforce that container name always requires prefix ?

Porges: At the moment we won't have any problems since production will always use an empty prefix. I'll think about a way to require that this is provided…

Porges: Discussed that we will address this later, as it won't affect production, only testing.

PR #2082, opened 2022-06-24
Fix tests

chkeita:


chkeita:
these changes are already merged as part of #2066

PR #2083, opened 2022-06-26
C# Integration tests project

Porges:
Move integration tests into their own project. This makes it easier to run unit tests if you don't want to (or don't have) `azurite` running, since you can `dotnet test` just the `Tests` directory.

Also add a check into the `AzuriteStorage` class that will abort the entire test run if `azurite` is not running, which is simpler than reading lots of socket exceptions.

Add a `README.md` with instructions for running the two types of tests.


Porges:
Coverage data is incorrect, needs fixing.

tevoinea (approved):

stishkin (approved):

PR #2090, opened 2022-06-27
Register '-net' application during deployment

Porges:

Adding the `-net` service to app registration so that OneFuzz CLI can access it by changing endpoint.

VS Code reformatted the file a bit.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -293,17 +293,14 @@ class Client:
             "cli_password", object_id, self.get_subscription_id()
         )
 
-    def get_instance_url(self) -> str:
+    def get_instance_urls(self) -> List[str]:
         # The url to access the instance
         # This also represents the legacy identifier_uris of the application
         # registration
         if self.multi_tenant_domain:
-            return "https://%s/%s" % (
-                self.multi_tenant_domain,
-                self.application_name,
-            )
+            return ["https://%s/%s" % (self.multi_tenant_domain, name) for name in [self.application_name, self.application_name + DOTNET_APPLICATION_SUFFIX]]
         else:
-            return "https://%s.azurewebsites.net" % self.application_name
+            return ["https://%s.azurewebsites.net" % name for name in [self.application_name, self.application_name + DOTNET_APPLICATION_SUFFIX]]
 
     def get_identifier_urls(self) -> List[str]:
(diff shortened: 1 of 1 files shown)


src/deployment/deploy.py
```diff
@@ -289,31 +293,25 @@ def create_password(self, object_id: UUID) -> Tuple[str, str]:
...
+    def get_instance_urls(self) -> List[str]:
         # The url to access the instance
         # This also represents the legacy identifier_uris of the application
         # registration
         if self.multi_tenant_domain:
-            return "https://%s/%s" % (
-                self.multi_tenant_domain,
-                self.application_name,
-            )
+            return ["https://%s/%s" % (self.multi_tenant_domain, name) for name in [self.application_name, self.application_name + DOTNET_APPLICATION_SUFFIX]]
```
chkeita, line 301: The "-net" functions are supposed to be temporary until we move everything to dotnet. 
with this changes the "-net" identifierUris and related info will become useless after the transfer.
Can you make this an option specified at deployment.

Porges: @chkeita I think we should be setting this as long as the `deploy.py` script is deploying the `-net` application separately. Would it be sufficient to add a note next to the application deployment to say to remove this setting once we remove the Python app?

chkeita: i am thinking about upgrade scenarios for existing deployments. 
If we keep it in the code as-is, we need to add a the code to remove it afterwards when we do the migration. 
We already have a setting in the deployment script that specify which -net function we want to activate. We could make that it dependent on that.

chkeita: The other options is to make sure we update the identifierUris (and related settings) at each deployment. Currently we only set it on the first time we create the registration.

Porges: @chkeita at the moment we do perform an update to add new identifier URIs ([see below](https://github.com/microsoft/onefuzz/pull/2090/files#diff-0489104f2cf3e8ff8968514b328b6cb496d9519ee74fa9f24b612ec060f2b4cfR452-R464)) but we never remove any. We could remove them in future?

Porges:
Closing in favour of combining this with #2084.

PR #2092, opened 2022-06-28
Add test for StatefulOrm static constructor

Porges:
`StatefulOrm` performs checks in the static constructor. Make sure that all these checks pass.

Currently `ReproOperations` and `ScalesetOperations` do not pass:

```
State transition method 'Stopping' in 'ReproOperations' does not have the correct signature.
Expected 'System.Threading.Tasks.Task`1[Microsoft.OneFuzz.Service.Repro] Invoke(Microsoft.OneFuzz.Service.Repro)' 
actual 'System.Threading.Tasks.Task Stopping(Microsoft.OneFuzz.Service.Repro)' 
```

```
State transition method 'Halt' in 'ScalesetOperations' does not have the correct signature.
Expected 'System.Threading.Tasks.Task`1[Microsoft.OneFuzz.Service.Scaleset] Invoke(Microsoft.OneFuzz.Service.Scaleset)'
actual 'System.Threading.Tasks.Task Halt(Microsoft.OneFuzz.Service.Scaleset)' 
```

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
@@ -123,7 +123,7 @@ public class ScalesetOperations : StatefulOrm<Scaleset, ScalesetState, ScalesetO
             }
         }
 
-        return scaleset with  { State = ScalesetState.Halt }; 
+        return scaleset;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -122,6 +122,8 @@ public ScalesetOperations(ILogTracer log, IOnefuzzContext context)
                 _log.WithHttpStatus(r.ErrorV).Error($"Failed to save scaleset record {scaleset.ScalesetId}");
             }
         }
+
+        return scaleset with  { State = ScalesetState.Halt };
```
chkeita, line 126: the state should already be halt at this point  so you should be returning the object as is

chkeita (approved):

tevoinea (approved):

PR #2105, opened 2022-06-29
Fix Racing condition in monitor.rs

chkeita:

This PR addresses the racing condition where a file could be discovered by the file monitor but disappear before the it reads its metadata.

### DIFF
--- a/src/agent/onefuzz/src/monitor.rs
+++ b/src/agent/onefuzz/src/monitor.rs
@@ -114,11 +114,12 @@ impl DirectoryMonitor {
                             continue;
                         }
                         Err(err) => {
-                            bail!(
+                            warn!(
                                 "error checking metadata for file. path = {}, error = {}",
                                 path.display(),
                                 err
                             );
+                            continue;
                         }
                     }
                 }
--- a/src/agent/onefuzz/src/monitor.rs
+++ b/src/agent/onefuzz/src/monitor.rs
@@ -1,7 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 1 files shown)


src/agent/onefuzz/src/monitor.rs
```diff
@@ -94,9 +94,17 @@ impl DirectoryMonitor {
...
-                    } else {
+                    if self.report_directories {
+                        return Ok(Some(path));
+                    }
+
+                    let is_dir = fs::metadata(&path)
+                        .await
+                        .ok()
+                        .map(|f| f.is_dir())
+                        .unwrap_or_default();
```
Porges, line 105: is this unwrap correct? if the path is missing we will unwrap the failed result to `false`, so the path will still be returned from the function

chkeita: good catch, i changed the logic to check if it is a file instead

ranweiler: Additionally, I think we probably want the following behavior:
1. Try to get the `Metadata`.
2. If `Ok(..)`, then report (conditionally if `report_directories` is `true`).
3. If `Err(err)` and `err.kind() == io::ErrorKind::NotFound`, we can assume we've lost a race with deletion so, ignore the file. Otherwise, something genuinely weird happened at the IO layer, so propagate the error (callers can always ignore it if desired).

chkeita: @ranweiler I opted to emit a warning instead of stopping the monitoring in case the error is intermittent

Porges (approved):

tevoinea (approved):

ranweiler (approved):

PR #2114, opened 2022-07-01
Cache radamsa builds in CI

Porges:
Some CI builds are getting spurious failures trying to clone the radamsa repository.

We don't need to build this every time so cache the built artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -349,7 +349,15 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/cache@v3
+      id: cache-radamsa-build-linux
+      with:
+        # key on the shell script only: this script fixes the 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -358,7 +366,15 @@ jobs:
     runs-on: windows-2019
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/cache@v3
+      id: cache-radamsa-build-windows
+      with:
+        # key on the shell script only: this script fixes the 
+        # version to a particular commit, so if it changes we need to rebuild
```
chkeita, line 373: We could pass the commit I’d to the script and use it in the key

chkeita (approved):

.github/workflows/ci.yml
Porges: I think it's good for the script to be self-contained

PR #2115, opened 2022-07-01
Reuse Agent artifacts if nothing in src/agent changes

Porges:
The agent build takes most of the CI runtime, so improve it by only rebuilding if the pre-reqs or something inside `src/agent` changes.

## PR Checklist
* [ ] Tests added/passed


Porges:
We should probably build the Rust version into the cache key for both this and the pre-reqs.

tevoinea:
If I remember correctly, the check-pr pipeline pulls down the github actions artifacts when testing against a PR. We should make sure this doesn't break that flow. I do see artifacts published at the end of the actions run so it might 'just work'.

tevoinea (approved):

Porges:
> If I remember correctly, the check-pr pipeline pulls down the github actions artifacts when testing against a PR. We should make sure this doesn't break that flow. I do see artifacts published at the end of the actions run so it might 'just work'.

Yes, the artifacts are still downloaded and packaged into the release zip, so it should work the same as before.

Porges:
> We should probably build the Rust version into the cache key for both this and the pre-reqs.

This is done now.

Porges:
Running one more `check-pr` to ensure this didn't break anything.

mgreisen (approved):

Porges:
One thing to think about is that the version number in the compiled binary might be wrong. Need to disable this cache on builds from the `main` branch.

Porges:
> One thing to think about is that the version number in the compiled binary might be wrong. Need to disable this cache on builds from a specific tag.

This is done now.

Porges:
Something is preventing `check-pr` from completing successfully. Need more investigation.

Porges:
Cleanly passed `check-pr`.

Porges:
Hold off on merging as I think we still have a problem with the version check in the API service, which will try to reimage the nodes if they don't have an exact version match.

Porges:
We now set `ONEFUZZ_ALLOW_OUTDATED_AGENT` during `check-pr` to allow the use of older versions.

PR #2116, opened 2022-07-01
Make the log sas url last as long as the job duration

chkeita:

Fixes an issue where the job logs sas token could expire before the job was scheduled

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/containers.py
+++ b/src/api-service/__app__/onefuzzlib/azure/containers.py
@@ -31,6 +31,8 @@ from .storage import (
     get_storage_account_name_key_by_name,
 )
 
+CONTAINER_SAS_DEFAULT_DURATION = datetime.timedelta(days=30)
+
 
 def get_url(account_name: str) -> str:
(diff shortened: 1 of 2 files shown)


src/api-service/__app__/onefuzzlib/azure/containers.py
```diff
@@ -31,6 +31,8 @@
     get_storage_account_name_key_by_name,
 )
 
+JOB_SCHEDULING_ESTIMATED_MAX_DELAY = datetime.timedelta(days=1)
```
stishkin, line 34: Actually that's worth a work item of it's own to figure out how long in worst case we wait for a job to become scheduled

stishkin (approved):

mgreisen (approved):

src/api-service/__app__/onefuzzlib/azure/containers.py
chkeita: I'll just match the existing default duration that we currently have for the jobs instead. 
The current code create sas of 30 days

stishkin (approved):

PR #2119, opened 2022-07-03
Enable .NET functions in `check-pr` for Agent-specific functions

Porges:
Enable the .NET functions for the agent by sending the agent the URI for the `-net` service.

Also fix some things causing failures when using the .NET functions (`CouldShrinkScaleset` was not implemented).

Improve error handling around table serialization/deserialization, fix an issue with int64/long mismatch between Python & C# code.

----

For `check-pr` testing:

1. There's a new parameter `enable_dotnet` which maps directly to the `--enable_dotnet` switch on `deploy.py`.
2. If you put `agent` there, all the `agent_*` functions will be enabled for .NET and disabled for Python.
3. If `agent_can_schedule` is disabled on the Python side, it will automatically tell the agent to use the .NET functions.

So to test the .NET agent functions, do a `check-pr` run with `enable_dotnet` set to `agent` and it should all work.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/creds.py
+++ b/src/api-service/__app__/onefuzzlib/azure/creds.py
@@ -82,6 +82,11 @@ def get_instance_url() -> str:
 
 
 @cached
+def get_agent_instance_url() -> str:
+    return "https://%s-net.azurewebsites.net" % get_instance_name()
+
+
+@cached
 def get_instance_id() -> UUID:
     from .containers import get_blob
     from .storage import StorageType
--- a/src/api-service/__app__/onefuzzlib/azure/creds.py
+++ b/src/api-service/__app__/onefuzzlib/azure/creds.py
@@ -85,7 +85,13 @@ def get_instance_url() -> str:
 def python_agent_functions_are_disabled() -> bool:
     # note that we only check one function here;
     # these should be enabled or disabled as a group
-    return os.environ["AzureWebJobs.agent_can_schedule.Disabled"] == "1"
+    try:
+        return os.environ["AzureWebJobs.agent_can_schedule.Disabled"] == "1"
+    except KeyError:
+        raise Exception(
+            "AzureWebJobs.agent_can_schedule.Disabled is not set. Environment: %s"
+            % os.environ
+        )
 
 
 @cached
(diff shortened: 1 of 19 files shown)


Porges:
Need to figure out how to enable these for `check-pr`.

tevoinea:
For the purposes of testing a refactor, I think it's totally OK to hardcode the change so that it'll run in check-pr (aka: the changes here are sufficient). If you wanted a way to test it in canary as well, an easy way would be to switch the URL based on an env var since we can modify those 'live'

Porges:
> For the purposes of testing a refactor, I think it's totally OK to hardcode the change so that it'll run in check-pr (aka: the changes here are sufficient). If you wanted a way to test it in canary as well, an easy way would be to switch the URL based on an env var since we can modify those 'live'

It's the Functions-level enabling that I'm missing. At the moment I'm going to run `check-pr` once, let it fail, then update the values in the deployment manually & re-run the tests.

Porges:
Have updated `check-pr` to enable Agent functions by default.

Porges:
`check-pr` now green with C# Agent functions enabled

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -81,6 +81,11 @@ def get_instance_url() -> str:
     return "https://%s.azurewebsites.net" % get_instance_name()
 
 
+@cached
+def get_agent_instance_url() -> str:
+    return "https://%s-net.azurewebsites.net" % get_instance_name()
```
Porges, line 86: we should enable this via config switch so we can turn it off in prod

src/api-service/__app__/onefuzzlib/azure/creds.py
```diff
@@ -81,6 +81,14 @@ def get_instance_url() -> str:
     return "https://%s.azurewebsites.net" % get_instance_name()
 
 
+@cached
+def get_agent_instance_url() -> str:
+    if os.environ.get("ONEFUZZ_USE_NET_AGENT_FUNCTIONS") == "true":
```
Porges, line 86: this should probably be `ONEFUZZ_USE_DOTNET_AGENT_FUNCTIONS`

Porges: we could actually have this automatically enabled if the Python functions are disabled, since that is done by environment variable

chkeita (approved):

PR #2126, opened 2022-07-06
Import `LibFuzzerDotnetLoader`, build .NET fuzzing tools

### LINKED ISSUE #2131: Install .NET fuzzing tools
Build and publish first and third-party fuzzing tools for .NET code: SharpFuzz, libfuzzer-dotnet, and a dynamic loader shim harness for libfuzzer-dotnet.

AB#40345137

ranweiler:
### Summary

To support .NET fuzzing, we want an end-user API that requires as little tool expertise as possible, but is effective and coverage-driven. These preliminary changes to the build are not yet used, but will be shortly by upcoming tasks. The new tools are:
- SharpFuzz's CLI tool for instrumenting user-submitted assemblies with coverage feedback. This does not appear to publishable as a self-contained executable at the moment. This is a CLI that simply accepts a path to an uninstrumented assembly. Its presence is not required to use `libfuzzer-dotnet`/`LibFuzzerDotnetLoader`, but will be applied to user assemblies in a setup stage.
- `libfuzzer-dotnet` (for both Linux and Windows), to enable using LibFuzzer with SharpFuzz instrumentation. This is actually a specially-parameterized native LibFuzzer binary which spawns `LibFuzzerDotnetLoader` as a child process, and is configured to use IPC to share data with the child's SharpFuzz runtime code.
- `LibFuzzerDotnetLoader`, which links to SharpFuzz, and uses dynamic loading so users don't have to link (or even be aware of) SharpFuzz or LibFuzzer. Users normally have to write and publish their own self-contained harness that is compatible with SharpFuzz. This removes that step / constraint.

Closes #2131.

### Changes

- Build pinned `SharpFuzz`, `libfuzzer-dotnet` tools for .NET fuzzing (Linux and Windows)
- Import and build `LibFuzzerDotnetLoader` project

### Testing

Manually checked that the expected artifacts get built and uploaded in CI. Downloaded the artifacts, and tested both fuzzing and crash repro against a sample .NET target library (on both Ubuntu 18.04 and Windows 10).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -283,7 +283,7 @@ jobs:
 
     # note that Test comes before Build:
     # Test will build things in Debug mode, Build will build them in Release
-    # Build comes second in case Test would do something (unspecified) that 
+    # Build comes second in case Test would do something (unspecified) that
     # would somehow alter the binaries that we want to package.
     - name: Test & Collect coverage info
       run: |
@@ -309,7 +309,7 @@ jobs:
         if [[ "$version" =~ '-' ]]; then
           # if it has a suffix, split it into two parts
           dotnet build -warnaserror --configuration Release /p:VersionPrefix=${version%-*} /p:VersionSuffix=${version#*-}
-        else 
+        else
           dotnet build -warnaserror --configuration Release /p:VersionPrefix=${version}
         fi
 
@@ -341,6 +341,26 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - run: src/ci/aflpp.sh
+    - uses: actions/upload-artifact@v2.2.2
+      with:
+        name: build-artifacts
+        path: artifacts
+  dotnet-fuzzing-tools-linux:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - run: src/ci/dotnet-fuzzing-tools.sh
+      shell: bash
+    - uses: actions/upload-artifact@v2.2.2
+      with:
+        name: build-artifacts
+        path: artifacts
+  dotnet-fuzzing-tools-windows:
+    runs-on: windows-2022
+    steps:
+    - uses: actions/checkout@v2
+    - run: src/ci/dotnet-fuzzing-tools.ps1
+      shell: pwsh
     - uses: actions/upload-artifact@v2.2.2
       with:
         name: build-artifacts
--- a/src/agent/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.csproj
+++ b/src/agent/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <LangVersion>10.0</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Requires commit ec28c46 or greater. -->
+    <ProjectReference Include="../sharpfuzz/src/SharpFuzz/SharpFuzz.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/agent/LibFuzzerDotnetLoader/Program.cs
+++ b/src/agent/LibFuzzerDotnetLoader/Program.cs
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Runtime;
+using System.Runtime.InteropServices;
+using System.Text;
+
+using TestOneSpan = SharpFuzz.ReadOnlySpanAction;
+delegate void TestOneArray(byte[] data);
+
+namespace LibFuzzerDotnetLoader {
+    public class Program {
+        public static void Main(string[] args) {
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -345,6 +345,26 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - run: src/ci/dotnet-fuzzing-tools.sh
+      shell: bash
+    - uses: actions/upload-artifact@v2.2.2
+      with:
+        name: build-artifacts
+        path: artifacts
+  dotnet-fuzzing-tools-windows:
+    runs-on: windows-2022
```
ranweiler, line 359: This is required to get .NET 6.0 on the build machine by default.

src/agent/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <OutputType>Exe</OutputType>
+    <LangVersion>10.0</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Requires commit ec28c46 or greater. -->
+    <ProjectReference Include="../sharpfuzz/src/SharpFuzz/SharpFuzz.csproj" />
```
ranweiler, line 13: This is gross, and rests atop load-bearing CI scripts. There may be a better way, but this is at least blunt. Regardless, we can change that after https://github.com/Metalnem/sharpfuzz/issues/26 is resolved.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+
+using TestOneSpan = SharpFuzz.ReadOnlySpanAction;
+delegate void TestOneArray(byte[] data);
+
+namespace LibFuzzerDotnetLoader {
+    public class Program {
+        public static void Main(string[] args) {
+            var target = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET");
```
ranweiler, line 15: We control this fully. I originally had it as 3 env vars, but that was sort of annoying in practice. We could always have it accept either this `:`-delimited string if present, and fall back on individual vars. The latter would mostly be nice for more agility when passing the assembly path.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (target == null) {
+                throw new Exception("`LIBFUZZER_DOTNET_TARGET` not set. " +
+                                    "Expected format: \"<assembly-path>:<class>:<static-method>\"");
+            }
+
+            var parts = target.Split(':');
+
+            var assemPath = parts[0];
```
ranweiler, line 24: We should probably fail helpfully and with context if any of these accesses are out of bounds, but we can revisit that.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (method == null) {
+                throw new Exception($"unable to resolve method: {methodName}");
+            }
+
+            if (TryTestOneSpan(method)) {
+                return;
+            }
+
+            if (TryTestOneArray(method)) {
```
ranweiler, line 46: We support multiple test function signatures, to support projects that target pre-`Span` SDKs.

src/ci/dotnet-fuzzing-tools.ps1
```diff
@@ -0,0 +1,43 @@
...
+dotnet publish src/SharpFuzz.CommandLine -f net6.0 -c Release -o $INSTALL_DIR/sharpfuzz --sc -r win10-x64
+popd
+popd
+
+# Build SharpFuzz and our dynamic loader harness for `libfuzzer-dotnet`.
+pushd src/agent/LibFuzzerDotnetLoader
+mkdir out
+dotnet publish . -c Release -o out --sc -r win10-x64 -p:PublishSingleFile=true
+cp out/LibFuzzerDotnetLoader.exe $INSTALL_DIR
+cp out/LibFuzzerDotnetLoader.pdb $INSTALL_DIR
```
ranweiler, line 33: There are also SharpFuzz PDBs in `out`, maybe we should copy those too. We should also probably stick `LibFuzzerDotnetLoader` in its own subdirectory entirely.

src/agent/LibFuzzerDotnetLoader/Program.cs
chkeita: I suggest printing out a warning message when we detect an old version. to let the user know that this might not be the most optimal route ie. there will be an extra allocation for each call of their test function

chkeita (approved):

tevoinea (approved):

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: Will add that in a follow-up PR.

ranweiler: Tracked in #2129.

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: Tracked in #2130.

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: Tracked in #2130.

PR #2136, opened 2022-07-06
Add multi-var target spec, improve errors

### LINKED ISSUE #2130: Improve `LibFuzzerDotnetLoader` env var ergonomics
1. Allow use of standalone (not `:`-delimited) env vars for passing assembly path, class, and method name
2. When splitting `LIBFUZZER_DOTNET_TARGET`, fail informatively if any component is missing (as a string).

ranweiler:
### Summary
Users can now specify a `libfuzzer-dotnet` target via specifying either:
1. `LIBFUZZER_DOTNET_TARGET` as `<assembly>:<class>:<method>`

or all of:

1. `LIBFUZZER_DOTNET_TARGET_ASSEMBLY`
1. `LIBFUZZER_DOTNET_TARGET_CLASS`
1. `LIBFUZZER_DOTNET_TARGET_METHOD`

Closes #2130.

### Changes

- Introduce the new `LibFuzzerDotnetTarget` class to encapsulate specification of a `libfuzzer-dotnet` target
- Hang static environment lookup methods on new class
- Add multiple-env var lookup option, and fall back on it if a `:`-delimited shorthand var is unset 
- Factor env lookup logic out of the main program

### DIFF
--- a/src/agent/LibFuzzerDotnetLoader/Program.cs
+++ b/src/agent/LibFuzzerDotnetLoader/Program.cs
@@ -12,31 +12,20 @@ delegate void TestOneArray(byte[] data);
 namespace LibFuzzerDotnetLoader {
     public class Program {
         public static void Main(string[] args) {
-            var target = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET");
+            var target = LibFuzzerDotnetTarget.FromEnvironment();
 
-            if (target == null) {
-                throw new Exception("`LIBFUZZER_DOTNET_TARGET` not set. " +
-                                    "Expected format: \"<assembly-path>:<class>:<static-method>\"");
-            }
+            var assem = Assembly.LoadFrom(target.AssemblyPath);
 
-            var parts = target.Split(':');
-
-            var assemPath = parts[0];
-            var typeName = parts[1];
-            var methodName = parts[2];
-
-            var assem = Assembly.LoadFrom(assemPath);
-
-            var ty = assem.GetType(typeName);
+            var ty = assem.GetType(target.ClassName);
 
             if (ty == null) {
-                throw new Exception($"unable to resolve type: {typeName}");
+                throw new Exception($"unable to resolve type: {target.ClassName}");
             }
 
-            var method = ty.GetMethod(methodName);
+            var method = ty.GetMethod(target.MethodName);
 
             if (method == null) {
-                throw new Exception($"unable to resolve method: {methodName}");
+                throw new Exception($"unable to resolve method: {target.MethodName}");
             }
 
             if (TryTestOneSpan(method)) {
@@ -47,7 +36,7 @@ namespace LibFuzzerDotnetLoader {
                 return;
             }
 
-            throw new Exception($"unable to bind method to a known delegate type: {methodName}");
+            throw new Exception($"unable to bind method to a known delegate type: {target.MethodName}");
         }
 
         // Returns `true` if delegate binding succeeded.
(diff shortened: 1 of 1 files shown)


src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +70,107 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
             return TryTestOne<TestOneArray>(method, t => span => t(span.ToArray()));
         }
     }
+
+    class LibFuzzerDotnetTarget
```
ranweiler, line 74: This class became surprisingly verbose out of a desire to throw context-rich exceptions. Happy to accept C# tips on improving brevity. To some extent I'm just not used to the amount of vertical space eaten up by this brace style 😅.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +70,107 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+        }
+
+        static LibFuzzerDotnetTarget FromEnvironmentVars()
+        {
+            var assemblyPath = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET_ASSEMBLY");
+
+            if (assemblyPath is null)
+            {
+                throw new Exception("`LIBFUZZER_DOTNET_TARGET_ASSEMBLY` not set");
+            }
```
chkeita, line 111:
```suggestion
            var assemblyPath = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET_ASSEMBLY")?? throw new Exception("`LIBFUZZER_DOTNET_TARGET_ASSEMBLY` not set");;
```
same for the others

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +70,107 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+            catch
+            {}
+
+            throw new Exception("No fuzzing target specified by environment variables");
+        }
+
+        static LibFuzzerDotnetTarget FromEnvironmentVars()
+        {
+            var assemblyPath = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET_ASSEMBLY");
```
ranweiler, line 106: Will add use constants for all of these env var keys.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +70,107 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+
+            var target = Environment.GetEnvironmentVariable("LIBFUZZER_DOTNET_TARGET");
+
+            if (target is null)
+            {
+                throw new Exception("`LIBFUZZER_DOTNET_TARGET` not set. " +
+                                    "Expected format: \"<assembly-path>:<class>:<static-method>\"");
+            }
+
+            var parts = target.Split(':');
```
chkeita, line 145: you can check that parts.Length == 3. that will save you the other checks.
You should also add StringSplitOptions.RemoveEmptyEntries

src/agent/LibFuzzerDotnetLoader/Program.cs
chkeita: you can use a [record ](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/records) instead if you want something smaller

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: This worked out really well. I kind of wanted the component-wise checks for specificity, but their value is questionable if the env var is wrong, especially given the use of split options.

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: I think the class is more well-sized now, thanks to your other suggestions.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,63 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
             return TryTestOne<TestOneArray>(method, t => span => t(span.ToArray()));
         }
     }
+
+    class LibFuzzerDotnetTarget
+    {
+        public string AssemblyPath { get; private set; }
+        public string ClassName { get; private set; }
+        public string MethodName { get; private set; }
```
chkeita, line 83: private set is not needed since you only assign those in the constructor
```suggestion
        public string AssemblyPath { get;}
        public string ClassName { get;}
        public string MethodName { get;}
```

chkeita (approved):

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: Ah, right, is really only necessary for non-`public` properties if we want to make it `protected`.

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,63 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+        }
+
+        static LibFuzzerDotnetTarget FromEnvironmentVars()
+        {
+            var assemblyPath = Environment.GetEnvironmentVariable(EnvVar.ASSEMBLY)??
+                throw new Exception($"`{EnvVar.ASSEMBLY}` not set");
+            var className = Environment.GetEnvironmentVariable(EnvVar.CLASS)??
+                throw new Exception($"`{EnvVar.CLASS}` not set");
+            var methodName = Environment.GetEnvironmentVariable(EnvVar.METHOD)??
+                throw new Exception($"`{EnvVar.METHOD}` not set");
```
Porges, line 116: I'd prefer to report all variables that aren't set, rather than just the first one that is not set

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,63 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+        static LibFuzzerDotnetTarget FromEnvironmentVarDelimited()
+        {
+            var target = Environment.GetEnvironmentVariable(EnvVar.TARGET)??
+                throw new Exception($"`{EnvVar.TARGET}` not set.");
+
+            var parts = target.Split(':', StringSplitOptions.RemoveEmptyEntries);
+
+            if (parts.Length < 3)
+            {
+                throw new Exception($"Value of `{EnvVar.TARGET}` is invalid");
```
Porges, line 131:
```suggestion
                throw new Exception($"Value of `{EnvVar.TARGET}` is invalid, should be of the form: '{{Assembly Path}}:{{Class Name}}:{{Method Name}}'");
```

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,63 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+
+
+        static LibFuzzerDotnetTarget FromEnvironmentVarDelimited()
+        {
+            var target = Environment.GetEnvironmentVariable(EnvVar.TARGET)??
+                throw new Exception($"`{EnvVar.TARGET}` not set.");
+
+            var parts = target.Split(':', StringSplitOptions.RemoveEmptyEntries);
+
+            if (parts.Length < 3)
```
Porges, line 129:
```suggestion
            if (parts.Length != 3)
```

Porges (commented):
A few suggested changes

src/agent/LibFuzzerDotnetLoader/Program.cs
ranweiler: Ha, this is basically what I had originally! I'll add it back in, even though we always catch this exception (for now).

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,63 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+            catch
+            {}
+
+            try {
+                return FromEnvironmentVars();
+            }
+            catch
+            {}
+
+            throw new Exception("No fuzzing target specified by environment variables");
```
tevoinea, line 106: It could be useful to include what we're expecting to see in the exception string. I think if the person hit this exception either they forgot the env vars are required or didn't know they were required in the first place. Something like:

`Expected either LIBFUZZER_DOTNET_TARGET=<assembly>:<class>:<method> or LIBFUZZER_DOTNET_TARGET_ASSEMBLY=<assembly>;LIBFUZZER_DOTNET_TARGET_CLASS=<class>;LIBFUZZER_DOTNET_TARGET_METHOD=<method> to be set`

Or even just a link to documentation explaining it.

tevoinea (approved):

src/agent/LibFuzzerDotnetLoader/Program.cs
```diff
@@ -81,4 +75,73 @@ static bool TryTestOne<T>(MethodInfo method, Func<T, SharpFuzz.ReadOnlySpanActio
...
+            var className = Environment.GetEnvironmentVariable(EnvVar.CLASS);
+            var methodName = Environment.GetEnvironmentVariable(EnvVar.METHOD);
+
+            var missing = new List<string>();
+
+            if (assemblyPath is null) { missing.Add(EnvVar.ASSEMBLY); }
+            if (className is null) { missing.Add(EnvVar.CLASS); }
+            if (methodName is null) { missing.Add(EnvVar.METHOD); }
+
+            if (assemblyPath is null || className is null || methodName is null)
```
ranweiler, line 121: I'm doing this so the compiler can understand that these variables are non-`null` on non-exceptional procedure exit. As you'd imagine, it can't infer that from a `missing.Count` check.

PR #2149, opened 2022-07-10
Depend on updated SharpFuzz package

### LINKED ISSUE #2143: Update builds / docs after next SharpFuzz release
After the changes in https://github.com/Metalnem/sharpfuzz/pull/24 are published to NuGet, update the following:
- Depend on a published SharpFuzz package in LibFuzzerDotnetLoader (instead of a local path)
- Update `dotnet-fuzzing-tools.{ps1,sh}` to match
- Remove pre-release instructions from the LibFuzzerDotnetLoader README

AB#40376873

ranweiler:
Closes #2143.


src/ci/dotnet-fuzzing-tools.sh
```diff
@@ -25,16 +25,12 @@ sudo apt-get install -y llvm-10 llvm-10-dev clang-10
 # Install `SharpFuzz` and `LibFuzzerDotnetLoader` pre-reqs.
 sudo apt-get install -y dotnet-sdk-6.0
 
-# Prepare SharpFuzz dependency for sibling project reference, build instrumentor.
-#
-# Redundant when commit ec28c46 is released.
-pushd src/agent
-git clone $SHARPFUZZ_REPO
+# Build SharpFuzz instrumentor.
+git clone $SHARPFUZZ_REPO sharpfuzz
```
ranweiler, line 29: The standalone CLI tool for instrumenting does not yet have its own NuGet release.

Metalnem: Do you really need the new version of the CLI tool for instrumenting? There were no changes in the instrumentation code, so current version of the tool should work fine with SharpFuzz 2.0.0. But if you would like the versions to match, I could publish a new version.

ranweiler: @Metalnem, you're correct: we could currently get by with the last-published version of `SharpFuzz.CommandLine`.

That said, an updated CLI package would be great! My main goal is eliminating any _future_ confusion or ABI mismatch by just depending on one version.

ranweiler: (I'm going to merge this for now, and will revisit if/when you cut a 2.0 release for the CLI!)

PR #2153, opened 2022-07-11
Pin Rust to 1.62

Porges:
By default we get whatever Rust version is installed in the Github Actions image. Pin it to latest Rust so it doesn't change unless we want it to.


Porges:
Waiting on #2152.

tevoinea (approved):

PR #2154, opened 2022-07-11
Path to source file in comment has changed, update it

Porges:
Non-code-change.


src/agent/libclusterfuzz/src/generated.rs
```diff
@@ -1,6 +1,5 @@
 // Rust generated by Microsoft, code originally from:
-//      https://github.com/google/clusterfuzz/blob/master/src/
-//          python/lib/clusterfuzz/stacktraces/constants.py
+//  https://github.com/google/clusterfuzz/blob/master/src/clusterfuzz/stacktraces/constants.py#:~:text=stack_frame_ignore_regexes
```
ranweiler, line 2: How about making this a permalink?

ranweiler (approved):

src/agent/libclusterfuzz/src/generated.rs
Porges: At the moment it would be a lie since the code is from some older version 🙂 

I need to double-check that we are up-to-date with latest.

Issue #2065, opened 2022-06-20
Crash deduping of non-deterministic targets misbehaves across related jobs


ranweiler:
The count of "unique" crashes for a given target are meant to represent _equivalence classes_ of _inputs_. Right now we compute this via equivocating crash _reports_, but the way we do this does is insensitive to nondeterministic crashes. In extreme cases, we can report more "unique" crashes than total crashes.

Right now, crash report task-level deduping is very naive: it uses the SHA-256 hash of the full call stack as a dedupe key. Notifications, on the other hand, are highly configurable. An overly rigid dedupe key function should only fail by producing partitions that are too small to be useful. However, when a target crashes in multiple ways for a single input, it can result in odd behavior, particularly given the current (template-defined) defaults for cross-job container sharing.

Suppose we have a fuzzing target, and (in _serial_) we create two template-based jobs from two builds, which share a `project` and target `name`. Say, project = `myapp`, target name = `myfuzzer`, and builds are `V1` and `V2`. Per our [conventional container naming rules](https://github.com/microsoft/onefuzz/blob/76d06efc8c924ab34edd31c910bd28644b623034/docs/containers.md#considerations-on-naming-containers), these jobs will be created with identical containers for `crashes`, `reports`, and `unique-reports`. **This part is critical to bug repro.**

Then the following can happen:
1. Create job `J1` for build `V1`. This causes the initial creation of the `crashes`, `reports`, and `unique-reports` containers.
2. Fuzzing task for `J1` finds an input `I`, writes it to the `crashes` container.
3. Crash report task for `J1` tests `I`, generates a crash report `C1` with call stack `S1`. A JSON blob with the name `sha256(I)` does not exist in the `reports` container, so it is written. A JSON blob with the name `sha256(C1)` (note the attempt at deduping) does not exist in the `unique-reports` container, so one is written there.
4. Now, create a new job `J2` for build `V2`. This does not create new containers for `crashes`, `reports`, or `unique-reports`.
5. Crash report task for `J2` initializes by batch processing any existing contents of the configured `crashes` container. This means it **re-tests** `I`, which was saved in the `crashes` container earlier, during `J1`.
[...]

PR #2155, opened 2022-07-11
Improvement to unique report generation

Porges:
This is a small change to improve the behaviour described in #2065.

Previously the crash reporter would do:

1. use input to generate failing call stack
2. save unique report keyed by `hash(call_stack)`
3. save report keyed by `hash(input)`

However (as describe in #2065), the output containers can be shared by multiple different versions of the same program, which are likely to generate different callstacks for the same input, which leads to `#unique_reports > #reports`.

This PR changes the flow to:

1. use input to generate a failing call stack
2. attempt to save report keyed by `hash(input)`
3. if a fresh report was _not_ saved, save unique report keyed by `hash(call_stack)`

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/report/crash_report.rs
+++ b/src/agent/onefuzz-task/src/tasks/report/crash_report.rs
@@ -139,6 +139,19 @@ impl CrashTestResult {
     ) -> Result<()> {
         match self {
             Self::CrashReport(report) => {
+                // try to save the report first — if it already exists, that
+                // means this is an already-tested input, so we shouldn’t save
+                // a unique report for it:
+                if let Some(reports) = reports {
+                    let name = report.blob_name();
+                    if upload_or_save_local(&report, &name, reports).await? {
+                        event!(new_report; EventData::Path = name);
+                    } else {
+                        // report already exists; don’t save a unique report
+                        return Ok(());
+                    }
+                }
+
                 // Use SHA-256 of call stack as dedupe key.
                 if let Some(unique_reports) = unique_reports {
                     let name = report.unique_blob_name();
@@ -147,12 +160,7 @@ impl CrashTestResult {
                     }
                 }
 
-                if let Some(reports) = reports {
-                    let name = report.blob_name();
-                    if upload_or_save_local(&report, &name, reports).await? {
-                        event!(new_report; EventData::Path = name);
-                    }
-                }
+                Ok(())
             }
 
             Self::NoRepro(report) => {
@@ -162,9 +170,10 @@ impl CrashTestResult {
                         event!(new_unable_to_reproduce; EventData::Path = name);
                     }
                 }
+
+                Ok(())
             }
         }
-        Ok(())
     }
 }
 


ranweiler (approved):

Porges:
On second thought this is not a very good fix, since:

1. assume input `X` would cause crashes `C1`, `C2`
2. we find crash `C1`, generate a `report` (keyed on `X`) and `unique_report` (keyed on `C1`)
3. if at any time later (even after `C1` is fixed) we find `C2`, we will never generate a new `unique_report` for it

ranweiler:
> On second thought this is not a very good fix, since:
> 
> 1. assume input `X` would cause crashes `C1`, `C2` non-deterministically
> 2. we find crash `C1`, generate a `report` (keyed on `X`) and `unique_report` (keyed on `C1`)
> 3. if at any time later (even after `C1` is fixed) we find `C2`, we will never generate a new `unique_report` for it

I _think_ this fix is still a strict improvement. The reason for this hinges on the presence (and correct functioning) of the `regression` tasks.

The scenario you describe is reachable with our current template-determined container sharing conventions. The impact of this PR is that we wouldn't get custom `new_report`/`new_unique_report` **events** fired during the job, and in some scenarios, we will no longer add to the `unique_reports` container when we previously would have. _Without_ these changes, we wouldn't get the `new_report` event, but we _would_ get the `new_unique_report` event (which is confusing).

However, our job templates also create `regression` tasks. These have a `regression_reports` container whose named is `build`-sensitive (unlike the `crashes`/`reports`/`unique_reports` containers). The regression task tests each input in the `inputs` corpus, _and_ each historically-known crashing input in the `crashes` set of inputs. If any of these input test cases cause a crash (whether `C1` or `C2`, or `Cn`), it creates a regression report (in the `regression_reports` container) and a `regression_report` event. This is true even with this PR. So even if `C1` is fixed, `C2` will be observed by the `regerssion` task.

Tying that all together, I think that:
1. This PR makes the `new_{unique_}report` eventing more consistent, and easier to reason about. We emit a `new_report` event (and write to the `reports` container) exactly once ever, per input: the first time we find the report. If the observed crash has a novel call stack at that time, then we will also emit a `new_unique_report` event / add to the `unique_reports` container. The number of `unique_reports`/events is always at most the number of `reports`/`new_report` events.
2. The regression task will continue to catch new / previously masked / previously non-deterministically triggered crashes that are reachable in a new `build`, although there will never be any `crash_report` task-emitted eventing / updates to the `reports`/`unique_reports` container for them. In the status quo, it's more confusing, because we will _never_ update the `reports` container, but we _could_ add to the `unique_reports` container. With this PR, it is _solely_ the job of the `regression` task to observe regressions / previously-hidden crashes. Without this PR, the `crash_report` task sometimes will observe them (in a numerically-confusing way).

chkeita:
> I _think_ this fix is still a strict improvement. The reason for this hinges on the presence (and correct functioning) of the `regression` tasks.
>  (...)

it seems to me that we are defining unique_report to be the First crash report we find. And not the uniquely identifiable crash report (by stack trace) .

I agree with @ranweiler  with this definition being most compatible with the expectations of the other task s

However @Porges  raised a good point in the PR. we use to store all the crash report we found for the job in the unique report. Now with this change if we detect a new type of crash we loose it. I propose that we keep adding any new crash report we find to the report container by changing our naming scheme to `<input-name>-hash(stack)` instead of just <input-name> to avoid any collisions.

Issue #2156, opened 2022-07-12
Refactor `jobs` Function to C#


Porges:
Convert `jobs` from Python to C#.

AB#40414202

Porges:
Ready for prod, closing

PR #2157, opened 2022-07-12
Convert `jobs` function for C#

Porges:
#2156 - convert the `jobs` function from Python to C#


src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -72,18 +78,18 @@ public class Orm<T> : IOrm<T> where T : EntityBase {
         }
 
         public async Task<ResultVoid<(int, string)>> Update(T entity) {
+            if (entity.ETag is null) {
+                throw new ArgumentException("ETag must be set when updating an entity", nameof(entity));
```
Porges, line 82: Converted this to an exception as it is a coding logic bug if it happens.

chkeita (approved):

PR #2112, opened 2022-07-01
Scaleset/autoscale cli default min size to zero

AdamL-Microsoft:
## Summary of the Pull Request
 
```bash
> onefuzz scalesets create $NAME --image $IMAGE --region $REGION
# now accepts & implies the following:
> onefuzz scalesets create $NAME --image $IMAGE --region $REGION --min_instances 0
```

Allows setting the scaleset + autoscale minimum instances count to zero at creation time and also sets zero as the `onefuzz scaleset create` command default if `--min_instances` argument isn't specified.

## PR Checklist
* [x] Applies to work item: #2144
* [x] Passed Integration Tests. 

## Info on Pull Request

updated:
src/cli/onefuzz/api.py
src/pytypes/onefuzztypes/requests.py

PR #2162, opened 2022-07-13
Allow autoscale with a minimum of 0

Porges:
`check-pr` failed during setup with:

```
1 validation error for AutoScale min ensure this value is greater than or equal to 1
```

Possibly due to #2112; needed a server-side change to match.


Porges:
Re-running `check-pr` to validate.

PR #2165, opened 2022-07-14
Use .get instead of lookup on `os.environ`

Porges:
Python service would crash (`KeyError`) if `ONEFUZZ_ALLOW_OUTDATED_AGENT` was not set.

### DIFF
--- a/src/api-service/__app__/onefuzzlib/workers/nodes.py
+++ b/src/api-service/__app__/onefuzzlib/workers/nodes.py
@@ -133,7 +133,7 @@ class Node(BASE_NODE, ORMMixin):
     @classmethod
     def mark_outdated_nodes(cls) -> None:
         # if outdated agents are allowed, do not attempt to update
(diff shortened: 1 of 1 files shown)


chkeita (approved):

src/api-service/__app__/onefuzzlib/workers/nodes.py
```diff
@@ -133,7 +133,7 @@ def search_outdated(
     @classmethod
     def mark_outdated_nodes(cls) -> None:
         # if outdated agents are allowed, do not attempt to update
-        if os.environ["ONEFUZZ_ALLOW_OUTDATED_AGENT"] == "true":
+        if os.environ.get("ONEFUZZ_ALLOW_OUTDATED_AGENT") == "true":
```
chkeita, line 136: don't forget to port the logic to the C# code

Porges: The C# version already does the right thing here.

Issue #2160, opened 2022-07-13
Non fatal errors in azcopy are causing a failure of the tasks


chkeita:
Here is an error in az copy that's causing a failure of the tasks
The failure is due to the fact that `6 of 2162` files did not sync.
However, that logs does not seem to include the any string identifying a [serious error](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-configure#log-and-plan-files). 

```cmd 
2162 Total, 2-sec Throughput (Mb/s): 0.0018
91.2 %, 1313 Done, 3 Failed, 846 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0     
91.2 %, 1313 Done, 3 Failed, 846 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0
91.2 %, 1313 Done, 3 Failed, 846 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0
91.2 %, 1315 Done, 3 Failed, 844 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0.0012
91.2 %, 1315 Done, 3 Failed, 844 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0     
91.2 %, 1315 Done, 3 Failed, 844 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0
91.2 %, 1318 Done, 3 Failed, 841 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0.0018
91.2 %, 1318 Done, 3 Failed, 841 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0     
91.2 %, 1319 Done, 3 Failed, 840 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0.0006
91.2 %, 1319 Done, 3 Failed, 840 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0     
91.2 %, 1320 Done, 3 Failed, 839 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0.0006
91.2 %, 1320 Done, 3 Failed, 839 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0     
91.2 %, 1320 Done, 3 Failed, 839 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0
91.3 %, 1322 Done, 3 Failed, 837 Pending, 2162 Total, 2-sec Throughput (Mb/s): 0.0012
[...]

PR #2166, opened 2022-07-14
Increase the size of the output  collected from the worker  

chkeita:
This is not a fix but a way to help debugging the linked issue. 
The logs are truncated so in this PR we are increasing the size of the worker output

#2160


src/agent/onefuzz-agent/src/worker/tests.rs
```diff
@@ -253,10 +253,16 @@ sys.stderr.write('C' * 65536 + 'D' * 4000)";
     redirected.child.wait().unwrap();
     let captured = redirected.streams.unwrap().join().unwrap();
 
-    let stdout: String = repeat('A').take(96).chain(repeat('B').take(4000)).collect();
+    let stdout: String = repeat('A')
+        .take(MAX_TAIL_LEN - 4000)
+        .chain(repeat('B').take(4000))
```
mgreisen, line 258: Do these values need to be higher as well? It looks like the code was just reformatted.

chkeita: i made the test dependent on the size of the buffer. The previous version assumed (MAX_TAIL_LEN == 4096)

mgreisen (approved):

Issue #1837, opened 2022-04-25
[Refactor] timer_repro to C#


tevoinea:

mgreisen:
@tevoinea please add a description

Porges:
Released into prod, closing

PR #2168, opened 2022-07-15
Laying groundwork for TimerRepro implementation

tevoinea:

_What is this about?_

* Fixes a TimerRepro state
* Improve error struct output by overriding ToString. Before, this would only print the `string[]` object name but not the contents of the strings
* Factor `ReportOrRegression` into `IReport`
* Add `ParseResourceId` and `GetData` to `Creds` which will be used in a follow up PR for TimerRepro

## PR Checklist


src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs
```diff
@@ -96,15 +98,15 @@ public NotificationOperations(ILogTracer log, IOnefuzzContext context)
         return null;
     }
 
-    private void GithubIssue(GithubIssuesTemplate config, Container container, string filename, RegressionReportOrReport? report) {
+    private void GithubIssue(GithubIssuesTemplate config, Container container, string filename, IReport? report) {
```
chkeita, line 101: can we make the report parameter non nullable. this will save us a null check in the implementation since we check for null before the call. (same goes for NotifyTeams)

chkeita (approved):

PR #2202, opened 2022-07-28
Add missing job state transition

chkeita:
The state transition Init was missing from the port of JobOperations


src/ApiService/ApiService/onefuzzlib/JobOperations.cs
```diff
@@ -73,6 +74,13 @@ public class JobOperations : StatefulOrm<Job, JobState, JobOperations>, IJobOper
         }
     }
 
+    public async Async.Task<Job> Init(Job job) {
+        _logTracer.Info($"init job: {job.JobId}");
+        var enabled = job with { State = JobState.Enabled };
+        var result = await Replace(enabled);
```
stishkin, line 80: log in case replace fails ?

stishkin (approved):

PR #2211, opened 2022-08-03
migrate repro_vmss

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/Functions/ReproVmss.cs
+++ b/src/ApiService/ApiService/Functions/ReproVmss.cs
@@ -1,0 +1,115 @@
+﻿using System.Net;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+
+namespace Microsoft.OneFuzz.Service.Functions;
+
+public class ReproVmss {
+    private readonly ILogTracer _log;
+    private readonly IEndpointAuthorization _auth;
+    private readonly IOnefuzzContext _context;
+
+    public ReproVmss(ILogTracer log, IEndpointAuthorization auth, IOnefuzzContext context) {
+        _log = log;
+        _auth = auth;
+        _context = context;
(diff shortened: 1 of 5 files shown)


src/ApiService/ApiService/Functions/ReproVmss.cs
```diff
@@ -0,0 +1,115 @@
...
+            if (vm == null) {
+                return await _context.RequestHandling.NotOk(req, new Error(ErrorCode.INVALID_REQUEST, new[] { "no such VM" }), $"{request.OkV.VmId}");
+            }
+
+            var response = req.CreateResponse(HttpStatusCode.OK);
+            await response.WriteAsJsonAsync(vm);
+            return response;
+        }
+
+        var vms = await _context.ReproOperations.SearchStates(VmStateHelper.Available).Select(vm => vm with { Auth = null }).ToListAsync();
```
tevoinea, line 46: Out of scope for this PR, but what do you think of creating an attribute called `Sensitive` (or similar) that clears out the information for that field when serializing? That way we wouldn't need to remember to do `vm with { Auth = null })` every time.

tevoinea: Or I guess we could just change the `Authentication` record to 
```csharp
public record Authentication
(
    SecretData<string> Password,
    string PublicKey,
    SecretData<string> PrivateKey
);
```

tevoinea (approved):

src/ApiService/ApiService/Functions/ReproVmss.cs
chkeita: agreed. SecretData is the way to got for this. we have a work item (in ado) related to this. i plan on doing this after the migration.

PR #2214, opened 2022-08-03
Factor out generic libfuzzer task

ranweiler:
Abstract the existing `libfuzzer_fuzz` task implementation. This is done in a way that will support introducing variant `libfuzzer_fuzz` task types that have extra config and do special process setup for certain kinds of LibFuzzer targets (i.e. `libfuzzer-dotnet`). 

Without changing task behavior, use the abstracted API to express the _existing_ task via the `GenericLibFuzzer`, which has no special additional config.

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
+++ b/src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
@@ -40,8 +40,14 @@ pub fn default_workers() -> usize {
     usize::max(1, cpus - 1)
 }
 
+pub trait LibFuzzerType {
+    type Config;
+
+    fn from_config(config: &Config<Self>) -> LibFuzzer;
+}
+
 #[derive(Debug, Deserialize, Clone)]
-pub struct Config {
+pub struct Config<L: LibFuzzerType + ?Sized> {
     pub inputs: SyncedDir,
     pub readonly_inputs: Option<Vec<SyncedDir>>,
     pub crashes: SyncedDir,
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
```diff
@@ -40,8 +40,14 @@ pub fn default_workers() -> usize {
     usize::max(1, cpus - 1)
 }
 
+pub trait LibFuzzerType {
```
ranweiler, line 43: Will add doc comments.

I do think this is a bit awkward, but I'm trying to keep the changes here as minimal as possible. The `libfuzzer-dotnet` fuzzing task will really be extremely similar to the generic one, except for how we construct the fuzzer process environment and arguments.

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
```diff
@@ -61,14 +67,23 @@ pub struct Config {
 
     #[serde(flatten)]
     pub common: CommonConfig,
+
+    #[serde(flatten)]
+    pub extra: L::Config,
```
ranweiler, line 72: In the `libfuzzer-dotnet` task type, this will be a struct along the lines of:
```rust
struct LibFuzzerDotnetConfig {
    target_assembly: PathBuf,
    target_class: String,
    target_method: String,
}
```

tevoinea (approved):

PR #2216, opened 2022-08-04
Add networking related operations for C# refactor

tevoinea:

_What is this about?_

Related to migrating `timer_repro`, in order to keep the PR sizes manageable this is a part of the new API surface required to migrate the whole functionality.

## PR Checklist

### DIFF
--- a/src/ApiService/ApiService/TestHooks/IpOperationsTestHooks.cs
+++ b/src/ApiService/ApiService/TestHooks/IpOperationsTestHooks.cs
@@ -30,7 +30,7 @@ namespace ApiService.TestHooks {
             var nic = await _ipOps.GetPublicNic(rg, name);
             if (nic != null) {
                 var resp = req.CreateResponse(HttpStatusCode.OK);
-                await resp.WriteStringAsync(nic.Get().Value.Data.Name);
+                await resp.WriteStringAsync((await nic.GetAsync()).Value.Data.Name);
                 return resp;
             } else {
                 return req.CreateResponse(HttpStatusCode.NotFound);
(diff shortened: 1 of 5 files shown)


src/ApiService/ApiService/onefuzzlib/NsgOperations.cs
```diff
@@ -19,10 +28,13 @@ public class NsgOperations : INsgOperations {
         private readonly ICreds _creds;
         private readonly ILogTracer _logTracer;
 
+        private readonly IOnefuzzContext _context;
+
 
-        public NsgOperations(ICreds creds, ILogTracer logTracer) {
+        public NsgOperations(ICreds creds, ILogTracer logTracer, IOnefuzzContext context) {
```
chkeita, line 34: ICred is in the context too

src/ApiService/ApiService/TestHooks/IpOperationsTestHooks.cs
```diff
@@ -28,10 +28,13 @@ public class IpOperationsTestHooks {
             var name = query["name"];
 
             var nic = await _ipOps.GetPublicNic(rg, name);
-
-            var resp = req.CreateResponse(HttpStatusCode.OK);
-            await resp.WriteStringAsync(nic.Get().Value.Data.Name);
-            return resp;
+            if (nic != null) {
+                var resp = req.CreateResponse(HttpStatusCode.OK);
+                await resp.WriteStringAsync(nic.Get().Value.Data.Name);
```
chkeita, line 33: i think calling nic.get here is querying the server twice. 
The data should already be available in Value.Data.

Same for ip.Get below

chkeita: also if you have to use the get method use the async version

chkeita (approved):

PR #2218, opened 2022-08-04
Add extensions functionality for C# refactor

tevoinea:

_What is this about?_

Related to migrating `timer_repro`.

Highlights of changes:
* Added `ImageOperations` which is used for working with VM images.
* Made generic wrapper type over `VirtualMachineScaleSetExtensionData` and `VirtualMachineExtensionData`. We use the same extensions (and same extension properties) for both and since I can't add a shared interface between the types ( and instead of duplicating the extension configuration functions), I created `VMExtensionWrapper`.
* Fixed some bugs in the existing extensions functions (ex: making sure our types are serialized as Azure expects, a couple of missing `await`s)

## PR Checklist

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmExtensionWrapper.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmExtensionWrapper.cs
@@ -1,4 +1,4 @@
-using Azure.Core;
+﻿using Azure.Core;
 using Azure.ResourceManager.Compute;
 
 namespace Microsoft.OneFuzz.Service {
--- a/src/ApiService/ApiService/onefuzzlib/VmExtensionWrapper.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmExtensionWrapper.cs
@@ -1,0 +1,62 @@
+using Azure.Core;
+using Azure.ResourceManager.Compute;
+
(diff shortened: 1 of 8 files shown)


src/ApiService/ApiService/onefuzzlib/VmExtensionWrapper.cs
```diff
@@ -0,0 +1,62 @@
...
+        public string? TypePropertiesType { get; init; }
+        public string? Publisher { get; init; }
+        public string? TypeHandlerVersion { get; init; }
+        public string? ForceUpdateTag { get; init; }
+        public bool? AutoUpgradeMinorVersion { get; init; }
+        public bool? EnableAutomaticUpgrade { get; init; }
+        public BinaryData? Settings { get; init; }
+        public BinaryData? ProtectedSettings { get; init; }
+
+        public (string, VirtualMachineExtensionData) GetAsVirtualMachineExtension() {
```
chkeita, line 17: You can use an implicit [conversion instead](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators).

just suggestion, feel free to ignore

chkeita (approved):

PR #2221, opened 2022-08-04
Add `libfuzzer_dotnet_fuzz` task to agent

ranweiler:
Add a new `libfuzzer_dotnet_fuzz` task to the agent.

This task has a config which is an extension to the `libfuzzer_fuzz` with 3 extra parameters: `target_assembly`, `target_class`, and `target_method`. At runtime, it uses the new `LibFuzzerType::extra_setup()` hook to to instrument the `target_assembly` for coverage feedback using SharpFuzz, and generates `libfuzzer-dotnet`-based fuzzer invocations of `LibFuzzer`. These use `LibFuzzerDotnetLoader` as the `--target_path` arg for `libfuzzer-dotnet`, and sets the env vars that tell `LibFuzzerDotnetLoader` what to fuzz.

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/dotnet.rs
+++ b/src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/dotnet.rs
@@ -1,7 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use anyhow::Result;
+use async_trait::async_trait;
 use onefuzz::libfuzzer::LibFuzzer;
+use tokio::process::Command;
 
 use crate::tasks::fuzz::libfuzzer::common;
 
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/dotnet.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::tasks::fuzz::libfuzzer::common;
+
+#[cfg(target_os = "linux")]
+const LIBFUZZER_DOTNET_PATH: &str =
+    "/onefuzz/third-party/dotnet-fuzzing-linux/libfuzzer-dotnet/libfuzzer-dotnet";
+
+#[cfg(target_os = "windows")]
+const LIBFUZZER_DOTNET_PATH: &str =
+    "/onefuzz/third-party/dotnet-fuzzing-windows/libfuzzer-dotnet/libfuzzer-dotnet.exe";
```
tevoinea, line 17: Shouldn't this be `C:/onefuzz/...`?

tevoinea (approved):

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/dotnet.rs
ranweiler: As-is, the path is relative to the current drive (the drive of the current working directory). In this context, that means the drive of the current working dir of the `onefuzz-task` process. We can either assume that these drives are the same, or that it should be `C:`. IMO the former is a safer assumption. In reality, it doesn't matter, and if we ever needed to stop hardcoding this value, we'd have to make it so implementers of `LibFuzzerTaskType` can have their own data fields. WDYT is clearest?

tevoinea: Oh I didn't know about that behavior, I thought it would've been an invalid path. I think leaving it as it is is good then!

Issue #2225, opened 2022-08-05
Repro code fails if the task os isn't in the image list


mgreisen:
- Onefuzz version: 5.9

In repro.py the following check is made on line 62. 

            if task.os not in DEFAULT_OS:
                raise NotImplementedError("unsupported OS for repro %s" % task.os)

This prevents us from moving to a new OS without a code change. This test should be removed or at the very least only be a warning.
This code has also been migrated to C# so the change needs to be made in both places. 

AB#40800406

PR #2226, opened 2022-08-05
Update Default Windows Image

nharper285:

_What is this about?_
Windows-10 20H2 is no longer serviced. We need to upgrade to 21H2. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
```diff
@@ -16,7 +16,7 @@ public class ReproOperations : StatefulOrm<Repro, VmState, ReproOperations>, IRe
     private static readonly Dictionary<Os, string> DEFAULT_OS = new Dictionary<Os, string>
     {
         {Os.Linux, "Canonical:UbuntuServer:18.04-LTS:latest"},
-        {Os.Windows, "MicrosoftWindowsDesktop:Windows-10:20h2-pro:latest"}
```
mgreisen, line 19: There is another issue #2225 where I'm suggesting that we remove the test, which makes these lists redundant. Further, after deployment if we still have some nodes running the older version this would cause the test to fail after this change. 

If we consider this list to be the list of fully tested images for the repro operation, then we should just add to the list.

nharper285: Understood. I can take out the change to the repro-related files.

nharper285: Honestly, this wouldn't be a bad instance_config value - i.e., make default image config values for windows and linux in the instance_config.

tevoinea: +1 to migrating it to `instance_config`. Presumably, if someone chooses to use the default linux/windows they're deferring to us to make a good decision about what the image should be which includes keeping the image up to date. By adding it to the instance_config, we don't need to make code changes to update those images. When we do update them, auto scale will naturally phase the old ones out and the ones ones in.

nharper285: New issue #2229

nharper285: @mgreisen / @tevoinea - I've made an issue for a long term solution, but we need this in the short term. Please approve when you see fit.

tevoinea (approved):

Issue #2203, opened 2022-07-29
New nodes are created when the scalesets state is shutdown


chkeita:
## Information

- Onefuzz version: 
- OS:

## Provide detailed reproduction steps (if any)

The situation was observed in canary in the scaleset: a0e09e42-da6a-450f-b2a8-656897707508
at the time of writing this here is the result of quering the state of the scaleset twice ( about 10 min apart)
```json 
{
    "client_object_id": "1ab466cd-c6c3-4687-a6ad-00d452f152db",
    "ephemeral_os_disks": false,
    "image": "Canonical:0001-com-ubuntu-server-focal:20_04-lts:latest",
    "needs_config_update": false,
    "nodes": [
        {
            "instance_id": "0",
            "machine_id": "e87924b6-400f-4a44-946e-74c9e2219f2e",
            "state": "busy"
        },
        {
            "instance_id": "1",
            "machine_id": "e5393b4e-b61f-45c6-b1ba-c76b80249553",
            "state": "busy"
        },
        {
            "instance_id": "105",
            "machine_id": "ab51fb3e-b7bf-4c7c-be2c-0def88a23d5b",
            "state": "init"
        }
    ],
    "pool_name": "test_pool_6e133403",
    "region": "westus3",
    "scaleset_id": "a0e09e42-da6a-450f-b2a8-656897707508",
    "size": 5,
    "spot_instances": false,
    "state": "shutdown",
    "tags": {},
    "vm_sku": "Standard_D2s_v3"
}
```

```json 
{
    "client_object_id": "1ab466cd-c6c3-4687-a6ad-00d452f152db",
    "ephemeral_os_disks": false,
    "image": "Canonical:0001-com-ubuntu-server-focal:20_04-lts:latest",
    "needs_config_update": false,
    "nodes": [
        {
            "instance_id": "0",
            "machine_id": "e87924b6-400f-4a44-946e-74c9e2219f2e",
            "state": "busy"
        },
        {
            "instance_id": "1",
            "machine_id": "e5393b4e-b61f-45c6-b1ba-c76b80249553",
            "state": "busy"
        },
        {
            "instance_id": "106",
            "machine_id": "091539a4-9b3b-453e-8173-e77e08a1c5fc"
        }
    ],
    "pool_name": "test_pool_6e133403",
    "region": "westus3",
    "scaleset_id": "a0e09e42-da6a-450f-b2a8-656897707508",
    "size": 5,
    "spot_instances": false,
[...]

PR #2232, opened 2022-08-09
Adjust auto scale to scale down nodes on shutdown

tevoinea:

_What is this about?_

Auto scale can prevent scalesets from shutting down by continuing to spin up nodes even though the scale set is in shutdown state.

## PR Checklist
* [ ] Applies to work item: #2203
* [*] Tests added/passed

## Info on Pull Request

_What does this include?_

When a scale set enters the shutdown state, onefuzz will do the following:
* Replace all existing auto scale rules with a single scale in rule that will always evaluate to `True` and will scale in 1 node at a time
* If no existing auto scale resources are configured for the scaleset, skip it

### 1. No new nodes in the scale set

Since we remove all scale out rules, new nodes will never be created regardless of whether there continues to be work in the pool's queue (effectively isolating the scale set from the pool).

### 2. No node will acquire scale in protection

Since the scale set is in shutdown state, nodes can no longer be assigned work. Nodes that are not assigned work are never marked with `scale_in_protection`.

_Note: Nodes not marked with scale in protection can be scaled in by the auto scaling service._
_Note: Once a node completes a WorkSet, it releases its scale in protection._

**Based on 1. and 2. we can say that the number of nodes in the scale set must _strictly_ decrease over time (as the nodes complete their WorkSets).**

## Validation Steps Performed

1. Create a new scale set
4. Submit a job (choose a 'short' duration like 1 hour)
5. Wait for the nodes the pick up work
6. Shut down the scale set using `onefuzz scalesets shutdown {id}`
7. Validate the job continues to run
8. Wait for the job to complete
9. Watch the scale set scale down to 0 nodes
10. Validate the scale set resource is deleted

### DIFF
--- a/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
+++ b/src/api-service/__app__/onefuzzlib/azure/auto_scale.py
@@ -55,6 +55,7 @@ def get_auto_scale_settings(
         )
         for auto_scale in auto_scale_collections:
             if str(auto_scale.target_resource_uri).endswith(str(vmss)):
+                logging.info("Found auto scale settings for %s" % vmss)
                 return auto_scale
 
     except (ResourceNotFoundError, CloudError):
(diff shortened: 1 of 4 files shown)


src/api-service/__app__/onefuzzlib/azure/auto_scale.py
```diff
@@ -42,23 +42,22 @@
 
 
 @retry_on_auth_failure()
-def add_auto_scale_to_vmss(
-    vmss: UUID, auto_scale_profile: AutoscaleProfile
-) -> Optional[Error]:
-    logging.info("Checking scaleset %s for existing auto scale resources" % vmss)
+def get_auto_scale_settings(
```
chkeita, line 45: don't forget to port the code to C# later

chkeita (approved):

mgreisen:
Is there a C# implementation of this code?

tevoinea:
> Is there a C# implementation of this code?

Not yet, I wanted to get the python in to unblock first. Will follow up with C#

PR #2235, opened 2022-08-09
Port TimerWorkers to C#

stishkin:


chkeita (approved):

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -168,14 +175,104 @@ IOnefuzzContext context
...
+    Guid? scalesetId = null,
+    IEnumerable<NodeState>? states = null,
+    PoolName? poolName = null,
+    bool excludeUpdateScheduled = false,
+    int? numResults = null) {
+
+        List<string> queryParts = new();
+
+        if (poolId is not null) {
+            queryParts.Add($"(pool_id eq '{poolId}')");
```
Porges, line 220: should use `TableClient.CreateQueryFilter` so escaping happens properly: see `Queries.cs` for examples

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -168,14 +175,104 @@ IOnefuzzContext context
...
+            IEnumerable<NodeState>? states = null,
+            PoolName? poolName = null,
+            bool excludeUpdateScheduled = false,
+            int? numResults = null) {
+
+        var query = SearchOutdatedQuery(_context.ServiceConfiguration.OneFuzzVersion, poolId, scalesetId, states, poolName, excludeUpdateScheduled, numResults);
+        if (numResults is null) {
+            return QueryAsync(query);
+        } else {
+            return QueryAsync(query).TakeWhile((_, i) => i < numResults);
```
Porges, line 261:
```suggestion
            return QueryAsync(query).Take(numResults);
```

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -168,14 +175,104 @@ IOnefuzzContext context
...
+
+    public async Async.Task CleanupBusyNodesWithoutWork() {
+        //# There is a potential race condition if multiple `Node.stop_task` calls
+        //# are made concurrently.  By performing this check regularly, any nodes
+        //# that hit this race condition will get cleaned up.
+        var nodes = _context.NodeOperations.SearchStates(states: NodeStateHelper.BusyStates);
+
+        await foreach (var node in nodes) {
+            await StopIfComplete(node, true);
+        }
```
Porges, line 273: We can issue all of these in parallel (there are more examples below as well):
```suggestion
        await Task.WhenAll(nodes.Select(n => StopIfComplete(n, true));
```

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -258,6 +355,19 @@ IOnefuzzContext context
         await SendStopIfFree(updatedNode);
     }
 
+    public async Async.Task SetShutdown(Node node) {
+        //don't give out more work to the node, but let it finish existing work
+        _logTracer.Info($"setting delte_requested: {node.MachineId}");
+        node = node with { DeleteRequested = true };
+        var r = await Replace(node);
```
Porges, line 362: prefer `Update` to `Replace`

```suggestion
        var r = await Update(node);
```

Porges (approved):

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -3,15 +3,24 @@
 
 namespace Microsoft.OneFuzz.Service;
 
-public interface IScalesetOperations : IOrm<Scaleset> {
+public interface IScalesetOperations : IStatefulOrm<Scaleset, ScalesetState> {
     IAsyncEnumerable<Scaleset> Search();
 
-    public IAsyncEnumerable<Scaleset> SearchByPool(PoolName poolName);
+    IAsyncEnumerable<Scaleset?> SearchByPool(PoolName poolName);
```
tevoinea, line 9: Could the nullable modifier on `Scaleset` be removed since the implementation is returning non-nullable?

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
stishkin: I see what you mean. 

100% onboard with doing it better than Python implementation (since Python just replaces the record)

stishkin: We need to revisit all this later on, since originally we just did "Replace" to follow the Python implementation.

So there are lots of places where Update should be done instead of Replace

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
stishkin: will address in follow up PR

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
stishkin: will address in follow up pr

PR #2241, opened 2022-08-10
Check if state transitions exist at test time

Porges:
Don't ignore missing state transitions in `StatefulOrm` static constructor test, instead error out. 

`Scaleset` was missing the `Resize` transition.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/NodeMessageOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/NodeMessageOperations.cs
@@ -1,0 +1,46 @@
+﻿using ApiService.OneFuzzLib.Orm;
+using Microsoft.OneFuzz.Service.OneFuzzLib.Orm;
+
+namespace Microsoft.OneFuzz.Service;
+
+//# this isn't anticipated to be needed by the client, hence it not
+//# being in onefuzztypes
+public record NodeMessage(
+    [PartitionKey] Guid MachineId,
+    [RowKey] string MessageId,
+    NodeCommand Message
+) : EntityBase {
+    public NodeMessage(Guid machineId, NodeCommand message) : this(machineId, NewSortedKey, message) { }
+};
+
+public interface INodeMessageOperations : IOrm<NodeMessage> {
+    IAsyncEnumerable<NodeMessage> GetMessage(Guid machineId);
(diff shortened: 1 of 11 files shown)


src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -178,9 +180,12 @@ public class StatefulOrm<T, TState, TSelf> : Orm<T>, IStatefulOrm<T, TState> whe
                     continue;
                 }
 
-                throw new Exception($"State transition method '{state}' in '{thisType?.Name}' does not have the correct signature. Expected '{delegateSignature}'  actual '{methodInfo}' ");
+                throw new InvalidOperationException($"State transition method '{state}' in '{thisType.Name}' does not have the correct signature. Expected '{delegateSignature}'  actual '{methodInfo}' ");
             };
 
+            if (missing.Any()) {
```
chkeita, line 186: Not all state have a corresponding transition method.
ex: JobState.Enabled and JobState.Stopped do not have a transition method in jobs.py

Porges: > Not all state have a corresponding transition method. ex: JobState.Enabled and JobState.Stopped do not have a transition method in jobs.py

I think that it's probably a good idea that we explicitly have these as no-op transitions, so we can check them all?

chkeita: sounds good

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/NodeMessageOperations.cs
```diff
@@ -0,0 +1,46 @@
...
+//# being in onefuzztypes
+public record NodeMessage(
+    [PartitionKey] Guid MachineId,
+    [RowKey] string MessageId,
+    NodeCommand Message
+) : EntityBase {
+    public NodeMessage(Guid machineId, NodeCommand message) : this(machineId, NewSortedKey, message) { }
+};
+
+public interface INodeMessageOperations : IOrm<NodeMessage> {
```
Porges, line 16: Moved this type and NodeTasksOperations into their own files.

stishkin (approved):

PR #2252, opened 2022-08-15
Retry on "No connection could be made because target machine actively…

### LINKED ISSUE #2217: Create_and_connect failures
## Information

- Onefuzz version: 5.9
- OS: Windows

1. onefuzz --endpoint <endpoint> repro create_and_connect oft-unique-reports-<unique-filename>.json 

### Expected result

That I would be dropped into the cdb debugger

### Actual result
Often the result is a failed connection with this error message. 

DebugConnect failed, Win32 error 0n10061
No connection could be made because the target machine actively refused it.
The server was reached but it refused the connection attempt. Ensure that the server specified is correct and that it is listening at the specified port.

After some testing it appears that if I use the connect command immediately after the failure the connection works. And if I use the separate commands create, then connect it also works. So it seems that there is a random window of time after the VM is created where it's "running" but still refuses connections. 

Wrapping the dbg_windows and dbg_linux functions in try/catch here https://github.com/microsoft/onefuzz/blob/992751044fd190b1eef10367303cbfd58120d48a/src/cli/onefuzz/api.py#L641 would likely resolve this issue. 

AB#40780347

stishkin:
Added retries when CLI fails to connect to CDB on a repro VM.

CLI will retry every 10 seconds on failure in a forever loop until it succeeds or user presses CTRL-C to kill the CLI process.

Closes #2217

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -9,6 +9,7 @@ import os
 import pkgutil
 import re
 import subprocess  # nosec
+import time
 import uuid
 from enum import Enum
 from shutil import which
@@ -553,7 +554,9 @@ class Repro(Endpoint):
         return None
 
     def _dbg_windows(
-        self, repro: models.Repro, debug_command: Optional[str]
+        self,
+        repro: models.Repro,
+        debug_command: Optional[str],
     ) -> Optional[str]:
         """Setup an SSH tunnel, then connect via CDB over SSH tunnel"""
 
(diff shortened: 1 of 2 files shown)


src/cli/onefuzz/api.py
```diff
@@ -568,29 +571,48 @@ def _dbg_windows(
         proxy = "*:" + REPRO_SSH_FORWARD if bind_all else REPRO_SSH_FORWARD
         with ssh_connect(repro.ip, repro.auth.private_key, proxy=proxy):
             dbg = ["cdb.exe", "-remote", "tcp:port=1337,server=localhost"]
-            if debug_command:
-                dbg_script = [debug_command, "qq"]
-                with temp_file("db.script", "\r\n".join(dbg_script)) as dbg_script_path:
-                    dbg += ["-cf", _wsl_path(dbg_script_path)]
-
+            while True:
```
tevoinea, line 574: Maybe we should stop after N tries, if it hasn't worked after a few minutes it _probably_ won't and that way people don't waste time thinking it might eventually work. Maybe N=10/20/30?

tevoinea (approved):

src/cli/onefuzz/api.py
stishkin: I'll set it to 10

PR #2263, opened 2022-08-17
Enabling csharp webhook and proxy heartbeat functions

nharper285:

_What is this about?_
The csharp functions I worked on some time ago had not been enabled. Doing in this PR. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/deployment/deploy.py
+++ b/src/deployment/deploy.py
@@ -1147,6 +1147,10 @@ class Client:
                             ],
                             env=dict(os.environ, CLI_DEBUG="1"),
                         )
+
+                        if function_name == "queue_proxy_update":
+                            dotnet_name = "QueueProxyHeartbeat"
+
                         # enable dotnet function
                         logger.info(f"enabling DOTNET function: {dotnet_name}")
                         subprocess.check_output(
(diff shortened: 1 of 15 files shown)


chkeita (approved):

src/deployment/deploy.py
```diff
@@ -1147,6 +1147,10 @@ def expand_agent(f: str) -> List[str]:
                             ],
                             env=dict(os.environ, CLI_DEBUG="1"),
                         )
+
+                        if function_name == "queue_proxy_update":
```
chkeita, line 1151: why the change of name ?

nharper285: I changed it awhile back because I thought it was a more accurate function name. The function _only_ handles the heartbeat updates.

chkeita: i propose we keep it as is and change it after the migration is done.

chkeita: this will minimize the custom code we need to support the rename

nharper285: Done.

PR #2266, opened 2022-08-17
Update Rust azure_* crates

Porges:
Dependabot tried and failed to do this.

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/task_logger.rs
+++ b/src/agent/onefuzz-task/src/tasks/task_logger.rs
@@ -3,12 +3,15 @@
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
-use azure_core::HttpError;
-use azure_storage::core::prelude::*;
+use azure_core::error::HttpError;
+use azure_core::StatusCode;
+use azure_storage::prelude::StorageClient;
 use azure_storage_blobs::prelude::*;
+use azure_storage_blobs::{container::operations::ListBlobsResponse, prelude::AsContainerClient};
+use futures::TryStreamExt;
 use onefuzz_telemetry::{LogTrace, LoggingEvent};
-use reqwest::{StatusCode, Url};
-use std::{path::PathBuf, sync::Arc, time::Duration};
+use reqwest::Url;
+use std::{path::PathBuf, time::Duration};
 use uuid::Uuid;
 
 use tokio::sync::broadcast::{error::TryRecvError, Receiver};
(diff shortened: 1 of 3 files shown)


tevoinea (approved):

Porges:
Manually verified that task logs appear while running, and passed `check-pr`.

src/agent/onefuzz-task/src/tasks/task_logger.rs
```diff
@@ -150,25 +154,24 @@ impl LogWriter<BlobLogWriter> for BlobLogWriter {
...
-                    if s == &StatusCode::PRECONDITION_FAILED
-                        && b.contains("MaxBlobSizeConditionNotMet")
+                Some(herr) => match herr.status() {
+                    StatusCode::PreconditionFailed
+                        if herr.error_code() == Some("MaxBlobSizeConditionNotMet") =>
                     {
                         Ok(WriteLogResponse::MaxSizeReached)
-                    } else if s == &StatusCode::CONFLICT && b.contains("BlockCountExceedsLimit") {
+                    }
+                    StatusCode::Conflict if herr.error_code() == Some("BlockCountExceedsLimit") => {
```
ranweiler, line 169: This (and the above) are different checks, but probably preferable. Maybe worth double-checking that by using an equality check (instead of substring match), we aren't excluding any error codes that we were previously accepting.

src/agent/onefuzz-task/src/tasks/task_logger.rs
```diff
@@ -45,7 +48,7 @@ trait LogWriter<T>: Send + Sync {
 
 /// Writes logs on azure blobs
 pub struct BlobLogWriter {
-    container_client: Arc<ContainerClient>,
```
ranweiler, line 48: Are there perf implications here, wrt an underlying HTTP client?

Porges: From what I can tell the client is still shared. `ContainerClient` has a `StorageClient` has a `Pipeline` which is a `Vec<Arc<dyn Policy>>`, of which the last entry is the HTTP transport (phew!).

src/agent/onefuzz-task/src/tasks/task_logger.rs
Porges: Going by the [Azure Blob Storage error codes](https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes) article, these are not substrings of any other error codes.

ranweiler (approved):

PR #2267, opened 2022-08-18
Make EntityConverter singleton

chkeita:
Make EntityConverter singleton to reduce the allocations


src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -88,22 +88,24 @@ class OnefuzzNamingPolicy : JsonNamingPolicy {
     }
 }
 public class EntityConverter {
```
Porges, line 90: we might as well make this a `static` class

Porges (approved):

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
chkeita: yeah that was my initial intent but keep it a class allow us to inject a request specific logger.

Issue #2272, opened 2022-08-18
`libfuzzer dotnet` job template


ranweiler:
AB#40985415

PR #2273, opened 2022-08-18
Enable dotnet fuzzing

### LINKED ISSUE #2271: Update server-side data model for dotnet fuzzing
AB#40985412

ranweiler:
Add a new CLI job template, `libfuzzer dotnet_dll`, and supporting server-side definitions

- Add `libfuzzer dotnet_dll` job template, which uses SharpFuzz, `libfuzzer-dotnet`, and `LibFuzzerDotnetLoader` to fuzz, record coverage, and create crash reports
- Update server data model to support task creation and config generation for the `libfuzzer_dotnet_fuzz`, `dotnet_coverage`, and `dotnet_crash_report` tasks

Closes #2271, #2272.


src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -17,6 +17,8 @@
 from . import JobHelper
 
 LIBFUZZER_MAGIC_STRING = b"ERROR: libFuzzer"
+LIBFUZZER_DOTNET_LOADER_PATH_LINUX = "/onefuzz/third-party/dotnet-fuzzing-linux/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.dll"
+LIBFUZZER_DOTNET_LOADER_PATH_WINDOWS = "/onefuzz/third-party/dotnet-fuzzing-windows/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.dll"
```
ranweiler, line 21: Yes, this DLL is cross-platform. However, `libfuzzer-dotnet` needs a platform-specific wrapper to invoke the managed code under test as a child process. As a result, we (as of this PR) will deploy two copies of the DLL on VMs, but accompanied by a bunch of other platform native files to support the fuzzing task. The `dotnet_coverage` and `_crash_report` tasks just happen to only want/need the underlying DLL, because we're invoking its `main()` directly.

src/ci/dotnet-fuzzing-tools.ps1
```diff
@@ -25,7 +25,7 @@ popd
 
 # Build SharpFuzz and our dynamic loader harness for `libfuzzer-dotnet`.
 pushd src/agent/LibFuzzerDotnetLoader
-dotnet publish . -c Release -o $ARTIFACTS/LibFuzzerDotnetLoader --sc -r win10-x64 -p:PublishSingleFile=true
+dotnet publish . -c Release -o $ARTIFACTS/LibFuzzerDotnetLoader --sc -r win10-x64
```
ranweiler, line 28: We were only setting `PublishSingleFile` to keep file counts down, and keep the third-party tools folder easier to navigate. However, the non-fuzzing tasks actually need access to the underlying DLL so we can `dotnet exec` it.

src/cli/onefuzz/templates/libfuzzer.py
chkeita: can you please add this as a comment it will be useful to know when maintaining the code

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/Defs.cs
```diff
@@ -39,8 +39,131 @@ public static class Defs {
                 )},
             MonitorQueue: ContainerType.ReadonlyInputs)
     },
+    {
+        TaskType.DotnetCoverage, new TaskDefinition(
```
tevoinea, line 43: `DotnetCoverage` task definition was already present:
https://github.com/microsoft/onefuzz/blob/ac54c576aef5c444cb85d7f6420769c35905b24d/src/ApiService/ApiService/onefuzzlib/Defs.cs#L558

tevoinea: Same for python:
https://github.com/microsoft/onefuzz/blob/ac54c576aef5c444cb85d7f6420769c35905b24d/src/api-service/__app__/onefuzzlib/tasks/defs.py#L561

ranweiler: Nice catch, thanks! I did originally _move_ it, but lost that commit at some point 😅.

ranweiler: Looks like I _had_ gotten the Python dupe, but not the C# one.

src/ci/dotnet-fuzzing-tools.sh
```diff
@@ -34,7 +34,7 @@ popd
 
 # Build SharpFuzz and our dynamic loader harness for `libfuzzer-dotnet`.
 pushd src/agent/LibFuzzerDotnetLoader
-dotnet publish . -c Release -o $ARTIFACTS/LibFuzzerDotnetLoader --sc -r linux-x64 -p:PublishSingleFile=true
+dotnet publish . -c Release -o $ARTIFACTS/LibFuzzerDotnetLoader --sc -r linux-x64
```
Porges, line 37: Suggest using the full name of this flag as isn't as clear as the arguments which take a value are:
```suggestion
dotnet publish . -c Release -o $ARTIFACTS/LibFuzzerDotnetLoader --self-contained -r linux-x64
```

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -18,6 +18,24 @@
 
 LIBFUZZER_MAGIC_STRING = b"ERROR: libFuzzer"
 
+# These path constants point to identically-named managed DLLs, so it may seem like one
+# of them is redundant. This is false: the DLLs are _not_ identical, and each one links
+# platform-native code.
```
Porges, line 23: Do we want to put the platform name into the DLL filename to make this clearer?

Porges (approved):

tevoinea (approved):

src/cli/onefuzz/templates/libfuzzer.py
ranweiler: This is naturally namespaced by container name (which does include the platform). Also, it turns out we don't need to make the distinction in this path constant, so we only need one. I've updated the comment accordingly.

PR #2276, opened 2022-08-19
Ignore info advisory about unmaintained `ansi_term` crate

ranweiler:
The `ansi_term` crate is unmaintained, as described by the INFO-level advisory [RUSTSEC-2021-0139](https://rustsec.org/advisories/RUSTSEC-2021-0139.html).

Allow continued use of it for now, and ignore the advisory when invoking `cargo audit`.


ranweiler:
Added https://github.com/microsoft/onefuzz/issues/2277 to capture moving to an alternative library.

mgreisen (approved):

chkeita (approved):

PR #2279, opened 2022-08-22
Devcontainer updates

Porges:
Addresses two things with the devcontainer:

- We want Rust 1.63, since we are pinned to that version now. I couldn't get this version to work with the devcontainer "features" so install it using rustup instead.
- The `dotnet` install was in a broken state due to a mix of Canonical/MS packages (see [here](https://github.com/dotnet/sdk/issues/27129)). The `dotnet6` package is included with Ubuntu now, so install it directly using `apt` instead of via the feature which fetched it from the MS package repository.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -3,17 +3,14 @@
 ARG VARIANT="ubuntu-22.04"
 FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
 
-# Setup MS package signing key:
-# NB: needs changing if VARIANT changes
-RUN wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
-    dpkg -i packages-microsoft-prod.deb && \
-    rm packages-microsoft-prod.deb
-
-# Install APT HTTPS transport required for MS packages:
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && apt-get install -y apt-transport-https
-
 # Install packages required for build:
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends \
     libunwind-dev clang build-essential libssl-dev pkg-config lldb \
-    bash-completion npm python-is-python3
+    bash-completion npm python-is-python3 \
+    dotnet6
+
+# Install Rust (installing via devcontainer feature failing for 1.63)
+USER vscode
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.63 -y
+USER root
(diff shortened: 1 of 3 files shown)


chkeita (approved):

.devcontainer/Dockerfile
```diff
@@ -3,17 +3,14 @@
 ARG VARIANT="ubuntu-22.04"
 FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
 
-# Setup MS package signing key:
-# NB: needs changing if VARIANT changes
-RUN wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
```
ranweiler, line 8: Double-checking: the mismatch here between `20.04` and the `VARIANT` was _not_ the only problem, correct? I'm assuming that trying to update this is how you surfaced the problem.

Then a question: aren't we installing the Microsoft repo elsewhere, e.g. on VMs? Would it be better to try to make sure always install that on the devcontainer, use the same sources, &c? My main worry is that one of the packages might be split into more granular packages, and so they stop being 1-1 replacements.

ranweiler (approved):

.devcontainer/Dockerfile
```diff
@@ -3,17 +3,14 @@
...
-
 # Install packages required for build:
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends \
     libunwind-dev clang build-essential libssl-dev pkg-config lldb \
-    bash-completion npm python-is-python3
+    bash-completion npm python-is-python3 \
+    dotnet6
+
+# Install Rust (installing via devcontainer feature failing for 1.63)
```
ranweiler, line 13: Is there a bug worth linking?

.devcontainer/Dockerfile
Porges: Second one:

I don't think we are using the MS packages anywhere else. I think we have two scenarios:

- For the Azure Functions service we rely on the preinstalled dotnet binaries (we build a in what is called a "framework-dependent" mode), so we already relinquish control over that.

- For the code that runs on the VMs (the new dotnet tasks), you've set them to build in self-contained mode, so the runtime is bundled with the binaries and we will not use any external dotnet runtime if it exists on the VMs.

Also worth noting is that the devcontainer does not affect our build at all (at the moment), it is purely for development purposes. What version of dotnet we are building against is inherited from the `ubuntu-18.04` image that our Github Actions is based upon. Getting this all aligned so we are using the exact same version everywhere would be worth doing but I'm not attempting that at the moment.

.devcontainer/Dockerfile
Porges: I didn't look too deeply into it; the script failed with an "exit 1" with no further error…

.devcontainer/Dockerfile
Porges: We do have some external tools being added (like SharpFuzz) which will use the platform-installed dotnet tooling, but I don't think we have anything on the VMs which we build ourselves. I'll ask around :)

tevoinea (approved):

.devcontainer/Dockerfile
```diff
@@ -3,17 +3,14 @@
...
+USER vscode
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.63 -y
```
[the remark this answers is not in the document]
Porges: Done, and added a note about what else needs updating if it changes.

PR #2282, opened 2022-08-22
Fix function name

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/Functions/ReproVmss.cs
+++ b/src/ApiService/ApiService/Functions/ReproVmss.cs
@@ -15,8 +15,8 @@ public class ReproVmss {
         _context = context;
     }
 
-    [Function("repro_vmss")]
-    public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "GET", "PATCH", "POST", "DELETE")] HttpRequestData req) {
+    [Function("repro_vms")]
+    public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "GET", "POST", "DELETE")] HttpRequestData req) {
         return _auth.CallIfUser(req, r => r.Method switch {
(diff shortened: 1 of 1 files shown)


stishkin (approved):

src/ApiService/ApiService/Functions/ReproVmss.cs
```diff
@@ -43,7 +43,7 @@ public class ReproVmss {
             return response;
         }
 
-        var vms = await _context.ReproOperations.SearchStates(VmStateHelper.Available).Select(vm => vm with { Auth = null }).ToListAsync();
+        var vms = _context.ReproOperations.SearchStates(VmStateHelper.Available).Select(vm => vm with { Auth = null });
         var response2 = req.CreateResponse(HttpStatusCode.OK);
         await response2.WriteAsJsonAsync(vms);
```
Porges, line 48: I think that this isn't going to write the correct result when it's given an `IAsyncEnumerable`

Porges (changes requested):

src/ApiService/ApiService/Functions/ReproVmss.cs
chkeita: It works with `IAsyncEnumerable` because System.Text.Json [supports it](https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/6.0/iasyncenumerable-serialization#new-behavior)

Porges: > It works with `IAsyncEnumerable` because System.Text.Json [supports it](https://docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/6.0/iasyncenumerable-serialization#new-behavior)

nice!

Porges (approved):

PR #2285, opened 2022-08-23
Update `chrono`, remove `time` to fix audit failures

### LINKED ISSUE #1266: move to using async appinsights
When we first started using `appinsights-rs`, only a blocking mode was available.

With the 0.2 release, an async version is available.  Moving to the async model rather than a separate thread would be beneficial from a maintenance stance. 

AB#39994197

### LINKED ISSUE #1366: Upgrade `time` and `chrono` crates
The `chrono` and `time` crates share a possible segfault related to thread-unsafe use of `getenv()` and `setenv()`.

- [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html): potential segfault in `time`
- [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071.html): potential segfault in `chrono`

Ensure these are both fixed upstream.

Note: a patch is available for `time`, which we depend on via `chrono` (it suffers twice from the same API usage error). However, `chrono` itself does not yet have a public patch. 

AB#39994083

Porges:
Closes #1366 and incidentally closes #1266.

Requires using a not-yet-merged modification to the latest version of `appinsights-rs` to remove the `time` feature from the `chrono` dependency (https://github.com/dmolokanov/appinsights-rs/pull/280). 

There are changes to use the new (tokio-based) version of `appinsights-rs`, e.g., made `set_appinsights_clients` async to ensure it is always called from an async context, since the constructor for appinsights now invokes `Tokio::spawn`.


Porges:
Something isn't working right with appinsights; VMs never come up properly.

Porges:
Failure is `0xc0000409` (`STATUS_STACK_BUFFER_OVERRUN`).

Porges:
Passed `check-pr` with latest changes.

Porges:
`check-pr` failing during the log check phase due to:

```
ERROR:integration:error log: error running task: QueueClient.pop Caused by:     0: storage queue pop failed     1: request attempt 6 failed     2: HTTP status client error (404 Not Found) for url (https://fuzzmaowljd7y6ejc.queue.core.windows.net/acb73cf3-0fb4-495e-a02a-864f8c307ab8/messages?se=2022-10-01T22%3A32%3A08Z&sp=raup&sv=2018-03-28&sig=dB9XtLT1xVzZ37hcIz1dpPw9Xz4yBr/yI%2BnXm/NAVAQ%3D) 
```

Porges:
Now green in `check-pr`.

src/agent/Cargo.toml
```diff
@@ -19,3 +19,6 @@ members = [
 
 [profile.release]
 lto = "thin"
+
+[patch.crates-io]
+appinsights = { git = "https://github.com/Porges/appinsights-rs", branch="audit-fixes" }
```
tevoinea, line 24: You should link your PR so we know that when it's released we can go back to the 'official' crate

Porges: It's in the commit message but I'll add it here too.

src/agent/onefuzz-agent/src/main.rs
```diff
@@ -173,15 +173,15 @@ fn run(opt: RunOpt) -> Result<()> {
     }
 
     // We can't send telemetry if this fails.
-    let config = load_config(opt);
+    let rt = tokio::runtime::Runtime::new()?;
+    let config = rt.block_on(load_config(opt));
 
     if let Err(err) = &config {
```
ranweiler, line 179: Maybe we can add something like the following (which we probably should've had already):
```suggestion
    // We can't send telemetry, because we couldn't get a telemetry key from the config.
    // Instead, log to an assumed-redirected stdout for the sake of debugging.
    if let Err(err) = &config {
```

src/agent/coverage/Cargo.toml
```diff
@@ -35,7 +35,7 @@ winapi = "0.3"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 pete = "0.9"
-procfs = "0.12"
+procfs = { version = "0.12", default-features = false, features=["flate2"] }
```
ranweiler, line 38: Can we please document what we are opting out of with this specification?

src/agent/onefuzz-agent/src/main.rs
```diff
@@ -191,20 +191,20 @@ fn run(opt: RunOpt) -> Result<()> {
         }
     }
 
-    telemetry::try_flush_and_close();
+    rt.block_on(telemetry::try_flush_and_close());
```
ranweiler, line 194: Reuse of the runtime seems reasonable, but I just want to call out [this bit of the docs](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#multi-thread-scheduler):

> Any spawned tasks will continue running after block_on returns.

Should we maybe play it safe and call [`Runtime::shutdown_background()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background) after our main call to `block_on()`?

src/ci/agent.sh
```diff
@@ -46,16 +46,9 @@ if [ X${CARGO_INCREMENTAL} == X ]; then
...
-# RUSTSEC-2020-0036: a dependency `failure` (pulled from proc-maps) is deprecated
-# RUSTSEC-2019-0036: a dependency `failure` (pulled from proc-maps) has type confusion vulnerability
-# RUSTSEC-2021-0065: a dependency `anymap` is no longer maintained
-# RUSTSEC-2020-0077: `memmap` dependency unmaintained, via `symbolic` (see: `getsentry/symbolic#304`)
-# RUSTSEC-2020-0159: potential segfault in `time`, not yet patched (#1366)
-# RUSTSEC-2020-0071: potential segfault in `chrono`, not yet patched (#1366)
 # RUSTSEC-2022-0048: xml-rs is unmaintained
 # RUSTSEC-2021-0139: ansi_term is unmaintained
-cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2020-0016 --ignore RUSTSEC-2020-0036 --ignore RUSTSEC-2019-0036 --ignore RUSTSEC-2021-0065 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0077 --ignore RUSTSEC-2022-0048 --ignore RUSTSEC-2021-0139
+cargo audit --deny warnings --deny unmaintained --deny unsound --deny yanked --ignore RUSTSEC-2022-0048 --ignore RUSTSEC-2021-0139
```
ranweiler, line 51: 🎉

src/agent/onefuzz-telemetry/src/lib.rs
```diff
@@ -450,7 +450,10 @@ mod global {
 }
 
 const REDACTED: &str = "Redacted";
-pub fn set_appinsights_clients(
+// This function doesn't do anything async, but TelemetryClient::new must be invoked
```
ranweiler, line 453: Should this be also be enforced upstream by making `TelemetryClient::new()` an `async` method in the `audit-fixes` fork?

ranweiler (approved):

src/agent/coverage/Cargo.toml
Porges: Done.

src/agent/onefuzz-agent/src/main.rs
Porges: The telemetry data is sent on a background task, so we want it to continue running up until we flush the telemetry client.

src/agent/onefuzz-telemetry/src/lib.rs
Porges: Taking this to the Rustaceans channel for discussion 🙂

src/agent/onefuzz-agent/src/main.rs
ranweiler: Ah, good point!

src/agent/Cargo.toml
```diff
@@ -19,3 +19,7 @@ members = [
 
 [profile.release]
 lto = "thin"
+
+[patch.crates-io]
+# TODO: remove once PR is merged upstream https://github.com/dmolokanov/appinsights-rs/pull/280
+appinsights = { git = "https://github.com/Porges/appinsights-rs", branch="audit-fixes" }
```
ranweiler, line 25: Oh, I missed this: here and wherever we depend on this branch, I suggest we actually pin to the appropriate `rev`.

If we don't do this, and we merged this PR as-is, then the build could break if you pushed a breaking change to the branch (e.g. making `TelemetryClient::new()` async).

src/agent/onefuzz-telemetry/src/lib.rs
ranweiler: FWIW: I think this PR is good-to-go regardless. If you implement this idea, it'd force the `async`-ness change anyway. Though it _would_ break the build, which prompted me to add one more suggestion re our the dependency spec for your fork.

Porges: @ranweiler Yes I am inclined to leave it as-is for now since that would be a source-breaking change whereas the current PR should (ha!) not affect downstream users at all.

PR #2292, opened 2022-08-24
Add `--use_dotnet_agent_functions` to `deploy.py` and Python service

Porges:
Add an explicit `USE_DOTNET_AGENT_FUNCTIONS` setting to Python service, instead of relying on whether or not functions are enabled/disabled.

We now run the Python & C# functions side-by-side so this wasn’t taking effect.

### DIFF
--- a/src/deployment/bicep-templates/function-settings.bicep
+++ b/src/deployment/bicep-templates/function-settings.bicep
@@ -71,6 +71,6 @@ resource functionSettings 'Microsoft.Web/sites/config@2021-03-01' = {
       ONEFUZZ_KEYVAULT: keyvault_name
       ONEFUZZ_OWNER: owner
       ONEFUZZ_CLIENT_SECRET: client_secret
-      ONEFUZZ_USE_DOTNET_AGENT_FUNCTIONS: use_dotnet_agent_functions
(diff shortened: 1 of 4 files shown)


src/deployment/bicep-templates/function-settings.bicep
```diff
@@ -50,26 +51,26 @@ resource functionSettings 'Microsoft.Web/sites/config@2021-03-01' = {
...
+      FUNCTIONS_WORKER_PROCESS_COUNT: '1'
+      APPINSIGHTS_INSTRUMENTATIONKEY: app_insights_key
+      APPINSIGHTS_APPID: app_insights_app_id
+      ONEFUZZ_TELEMETRY: telemetry
+      AzureWebJobsStorage: func_sas_url
+      MULTI_TENANT_DOMAIN: multi_tenant_domain
+      AzureWebJobsDisableHomepage: 'true'
+      AzureSignalRConnectionString: signal_r_connection_string
+      AzureSignalRServiceTransportType: 'Transient'
+      ONEFUZZ_INSTANCE_NAME: instance_name
```
Porges, line 65: I was getting a bicep lint about not using quotes here

AdamL-Microsoft (approved):
Looks good

PR #2293, opened 2022-08-24
Changing CLI Scaleset `size` parameter to `max_size`.

### LINKED ISSUE #2227: Refactor Scaleset Scaling Params - Change Size to MaxSize
Simple request to update the current CLI commands and parameters for scaleset create. Currently `size` is set to both `default` and `MaxSize`. We propose adding a `default` parameter and setting the meaning of `size` to `MaxSize`. 

**Current:** 
   auto_scale = requests.AutoScaleOptions(
            min=min_instances,
            max=size,
            default=size,
            scale_out_amount=scale_out_amount,
            scale_out_cooldown=scale_out_cooldown,
            scale_in_amount=scale_in_amount,
            scale_in_cooldown=scale_in_cooldown,
        )

**Proposed**: 
   auto_scale = requests.AutoScaleOptions(
            min=min_instances,
            max=size,
            default=default_instances **(new flag),**
            scale_out_amount=scale_out_amount,
            scale_out_cooldown=scale_out_cooldown,
            scale_in_amount=scale_in_amount,
            scale_in_cooldown=scale_in_cooldown,
        )

nharper285:

_What is this about?_
Updating the scaleset `size` parameter to be called `max_size` and set the initial size of the scaleset to be `min_instances` - aka 0. This means that the scaleset wont' be initialized with it's max number of nodes and instead grow to meet demand. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

Closes #2227


tevoinea (approved):
Ignore me :)

src/cli/onefuzz/api.py
```diff
@@ -1439,7 +1439,7 @@ def create(
                 vm_sku=vm_sku,
                 image=image,
                 region=region,
-                size=size,
+                size=1,
```
mgreisen, line 1442: It would be good to explain _why_ we set the size to 1. Probably not going to remember later.

mgreisen (approved):

src/cli/onefuzz/api.py
nharper285: I'll add a comment.

src/cli/onefuzz/api.py
```diff
@@ -1393,8 +1393,9 @@ def _expand_scaleset_machine(
     def create(
         self,
         pool_name: primitives.PoolName,
-        size: int,
+        max_size: int,
         *,
+        intialized_size: Optional[int] = 1,
```
mgreisen, line 1398: I think `initial_size` makes a bit more sense because the scaleset has not been created yet, so a past tense parameter is a little confusing.  We want to communicate that this is the size we will initialize the scale set to.

nharper285: Oh I much prefer that. Good call.

PR #2297, opened 2022-08-25
Devcontainer Dockerfile fixes

Porges:
ARG needs to be after the FROM or it isn't "visible" later on.

Also, the `rust-analyzer` extension has a new ID.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,10 +2,10 @@
 
 ARG VARIANT="ubuntu-22.04"
 
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
 # note: keep this in sync with .github/workflows/ci.yml
 ARG RUSTVERSION="1.63"
-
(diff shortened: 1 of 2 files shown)


tevoinea:
I also had an issue with dotnet where it only found the runtime and not the sdk. I was working up a fix in this branch (https://github.com/tevoinea/onefuzz/tree/tevoinea/MoveToPrecreateScript) but I was also going to include moving some of our dependency installation to the Dockerfile to take advantage of the prebuilt containers that we have now.

The only parts you would need to add to fix the sdk issue are:
1. Add lines 14-22 https://github.com/microsoft/onefuzz/compare/main...tevoinea:tevoinea/MoveToPrecreateScript?expand=1#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6R14-R22
2. Change the `apt install ... dotnet6` -> `apt install ... dotnet-sdk-6.0`

tevoinea (approved):

PR #2309, opened 2022-08-28
Create tables on startup

Porges:
Attempt to create storage tables upon startup rather than once per table client.


stishkin (approved):

Porges:
Manually validated since `check-pr` is out of quota.

PR #2315, opened 2022-08-29
Enable AppInsights dependency tracking

Porges:
At the moment we do not have this enabled, but it will be useful to diagnose issues with our use of Azure Storage.


chkeita (approved):

Porges:
Confirmed working manually.

PR #2316, opened 2022-08-29
Fix ServiceConfiguration.OneFuzzVersion

Porges:
.NET code was not getting the correct version from `ServiceConfiguration.OneFuzzVersion` as that expected to read it from the config. It should be read from the assembly attributes instead, but I still allow it to be overridden here.

### DIFF
--- a/src/ApiService/ApiService/ServiceConfiguration.cs
+++ b/src/ApiService/ApiService/ServiceConfiguration.cs
@@ -99,7 +99,7 @@ public class ServiceConfiguration : IServiceConfig {
             // version can be overridden by config:
             return Environment.GetEnvironmentVariable("ONEFUZZ_VERSION")
                 ?? _oneFuzzVersion
-                ?? "0.0.0";
+                ?? throw new InvalidOperationException("Unable to read OneFuzz version from assembly");
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/ServiceConfiguration.cs
```diff
@@ -86,7 +93,16 @@ public class ServiceConfiguration : IServiceConfig {
...
     public string? OneFuzzResourceGroup { get => Environment.GetEnvironmentVariable("ONEFUZZ_RESOURCE_GROUP"); }
     public string? OneFuzzTelemetry { get => Environment.GetEnvironmentVariable("ONEFUZZ_TELEMETRY"); }
-    public string OneFuzzVersion { get => Environment.GetEnvironmentVariable("ONEFUZZ_VERSION") ?? "0.0.0"; }
+
+    public string OneFuzzVersion {
+        get {
+            // version can be overridden by config:
+            return Environment.GetEnvironmentVariable("ONEFUZZ_VERSION")
+                ?? _oneFuzzVersion
+                ?? "0.0.0";
```
stishkin, line 102: prefer "0.0.0" over throwing ?

stishkin (approved):

chkeita (approved):

PR #2325, opened 2022-08-30
Update deploy to set all settings at once

Porges:
Setting them one-by-one is slow.


nharper285 (approved):

Porges:
Confirmed via `check-pr`.

PR #2328, opened 2022-08-31
Enforce no extra properties

### LINKED ISSUE #2314: C# port should enforce missing fields in Json deserialization
POSTing following JSON to Pool APIs in C#
``` json
{
  "name": "FT-DELETE-4812d4da-a9ab-40e1-88f1-2d9e148254fa",
  "oss": "linux",
  "arch": "x86_64",
  "managed": true
}
```
with typo for "OS" will succeed with setting OS to Windows

Meanwhile Python endpoint fails creation with error:
```json
{"code": 450, "errors": ["os: field required"]}
```

ORM must enforce that all fields are present

AB#41119374

Porges:
Closes #2314.


src/ApiService/ApiService/OneFuzzTypes/Requests.cs
```diff
@@ -192,57 +211,68 @@ public record ScalesetSearch(
...
+) : BaseRequest;
 
-public record TaskGet(Guid TaskId);
+public record TaskGet(Guid TaskId) : BaseRequest;
+
+public record TaskCreate(
+   Guid JobId,
+   List<Guid>? PrereqTasks,
+   TaskDetails Task,
+   [property: Required]
```
Porges, line 233: in ASP.NET Core they make `[Required]` the default for any non-null properties, we should try to do the same

tevoinea (approved):

tevoinea:
We should be sure to let this change bake a little longer than usual because it has the potential to expose a lot of small serialization bugs we missed.

Porges:
Found an issue already in `check-pr`.

Porges:
Found during testing that the CLI does send `null` for some no-longer supported properties, so I’ve updated the PR to allow extraneous properties to be set iff they are `null`.

tevoinea (approved):

Porges:
Passed `check-pr` with full set of changes.

PR #2333, opened 2022-08-31
Functional tests

stishkin:


src/ApiService/FunctionalTests/1f-api/Helpers.cs
```diff
@@ -0,0 +1,28 @@
...
+    public class Helpers {
+
+        public static async Task<(Pool, Scaleset)> CreatePoolAndScaleset(PoolApi poolApi, ScalesetApi scalesetApi, string os = "linux", string? region = null, int numNodes = 2) {
+            var image = (os == "linux") ? ScalesetApi.Image_Ubuntu_20_04 : ScalesetApi.ImageWindows;
+
+            var newPoolId = Guid.NewGuid().ToString();
+            var newPoolName = PoolApi.TestPoolPrefix + newPoolId;
+            var newPool = await poolApi.Create(newPoolName, os);
+
+            Assert.True(newPool.IsOk, $"failed to create new pool: {newPool.ErrorV}");
```
tevoinea, line 18: In a recent PR, I added the `FluentAssertions` library so we can do things like

`newPool.IsOk.Should().BeTrue()`

I'm not saying you should change it here, just spreading the message 😛

tevoinea (approved):

src/ApiService/FunctionalTests/1f-api/Helpers.cs
stishkin: I'll start doing this. There is more tests to write, so I'll redo these ones as well

PR #2339, opened 2022-09-02
[WIP] unmanaged node prototype

chkeita:

- [ ] endpoint to retrieve the configuration of a pool for an unmanaged node
- [ ] enable role verification in the authentication
- [ ] added Os to the registration request from the agent 
- [ ] new "unmanagedNode" app role
- [ ] registration command to associate authorize an existing appid to be used to connect an unmanaged node


stishkin:
Can you write a one page architecture spec that can be reviewed first before proceeding with implementation ?

stishkin (changes requested):
- Is it dedicated pool for unmanaged nodes or is pool can be mixed with managed and unmanaged nodes ? 
	- What if pool is Linux but unmanaged node is Windows, how is it enforced (same 32 bit vs 64 bit) Same for OS image - Ubuntu vs Redhat ?
	- How is unmanaged node name is assigned ? what if there is a duplicate in the pool ? 
	- what happens if unmanaged node goes rogue ? Do we try to prevent it from taking work ? (Since OneFuzz was designed with "tear it down" in mind, and unmanaged node we cannot tear it down)
	- Which queue unmanaged node gets it's work from ? 
	- Currently managed nodes auth using token that they get from inside vnet. Unmanaged node cannot do that - what's the auth, and can it be mixed with "user" auth when node makes a request ? 
	- Can I have 2 agents running on one unmanged node ? Are they going to interfere with each other ?
	- Which app insights unmanaged node logs to ? how is that setup ?
	- What if pool does not exist for which unmanaged node claims to be registered with
	
	- What happens if unmanaged node is outdated ? What's the update process ? 
	- Does it have to be part of a scaleset ? or is it standalone VM ? 
	
	- What's the testing strategy ?
        - What's test automation strategy ?

chkeita:
> Can you write a one page architecture spec that can be reviewed first before proceeding with implementation ?

This is a prototype in a draft PR. I am not committing to an specific implementation. The point of this prototype is to understand what is needed.

chkeita:
> - Is it dedicated pool for unmanaged nodes or is pool can be mixed with managed and unmanaged nodes ? 

dedicated. The pool [already has a field](https://github.com/chkeita/onefuzz/blob/c54db04083ad21cfaabe6075de269eb799e6f773/src/ApiService/ApiService/OneFuzzTypes/Model.cs#L600) to that says if it is managed or not.

> - What if pool is Linux but unmanaged node is Windows, how is it enforced (same 32 bit vs 64 bit) Same for OS image - Ubuntu vs Redhat ?

we have validtion the os matches at the registration. But keep in mind that in this scenrio we do not control the nodes. So we rely on whatever the gent gives which makes it hard to enforce. 

> - How is unmanaged node name is assigned ? what if there is a duplicate in the pool ? 

The node id will be provided by the agent. At this point in the prototype, a duplicate will overwrit the previous registration

> - what happens if unmanaged node goes rogue ? Do we try to prevent it from taking work ? (Since OneFuzz was designed with "tear it down" in mind, and unmanaged node we cannot tear it down)

No. we don't since we do not control it. it is up to the node owner to tear it down

> - Which queue unmanaged node gets it's work from ? 

The queue of the pool it is registered to. Same as the current nodes

> - Currently managed nodes auth using token that they get from inside vnet. Unmanaged node cannot do that - what's the auth, and can it be mixed with "user" auth when node makes a request ? 

The client support [2 types of authentication](https://github.com/chkeita/onefuzz/blob/94e29047d7954035012898b4f0797452682d7f46/src/agent/onefuzz/src/auth.rs#L59)
we currently use `ManagedIdentity`. The unmanaged node will use `Client`. The credential will created by the unmnaged node owner by creating a new Application registration in azure and assigning it to the onefuzz instance. There will be helpers for that

> - Can I have 2 agents running on one unmanged node ? Are they going to interfere with each other ?

Currently both agent can run but they will interfere with each other because they will share the same machine id

> - Which app insights unmanaged node logs to ? how is that setup ?

The appinsight config is part of the job config and is independent of the node. the setup for is the same as the current implementation

> - What if pool does not exist for which unmanaged node claims to be registered with

the agent registration will fail

> - What happens if unmanaged node is outdated ? What's the update process ? 

it is up to the owner of the unmanaged node

> - Does it have to be part of a scaleset ? or is it standalone VM ? 

it could be part of a scaleset or a standalone VM . That is up to the unmanaged node owner. The service side will only see the node.

> - What's the testing strategy ?    
> - What's test automation strategy ?

create an unmanaged pools,
create the auth data
create vms. 
deploy the latest agent on that vm and run it unmnged on thosse VM
run check PR on those pools

stishkin:
Can you write a one page spec with scenarios that implementation supposed to handle and the whole team can review and ask questions and provide feedback ? 

All those questions that I wrote were just of the top of my head. And it would be great to have a some sort of a plan to read through to understand what are you trying to implement. 

And also have a paragraph explaining test strategy.

chkeita:
> Can you write a one page spec with scenarios that implementation supposed to handle and the whole team can review and ask questions and provide feedback ?
> 
> All those questions that I wrote were just of the top of my head. And it would be great to have a some sort of a plan to read through to understand what are you trying to implement.
> 
> And also have a paragraph explaining test strategy.


When i am done with prototyping

PR #2349, opened 2022-09-06
Use 64-bit worker for function

Porges:
We want the .NET code running as 64-bit. This shouldn't affect the Python code.


stishkin (approved):

Porges:
Confirmed working.

Issue #2350, opened 2022-09-07
srcview crate: SrcView should support try_insert to prevent multiple parses of the same PDB


suttonbradley:
Currently the `SrcView` struct in the srcview crate only has an `insert` function, with no `contains` or `try_insert` function, so there is no way to ensure that a PDB at the given path has been parsed and had its metadata stored within the `SrcView` without doing a potentially unnecessary `insert` call. This makes it so users of the srcview crate have to either a) maintain their own metadata on which PDBs they have inserted into the `SrcView` or b) accept the slowdown of potential duplicate PDB parsing, rather than having the `SrcView` do a lookup in its tree.

In order to mitigate this, there should be a `try_insert` and/or a `contains` function for SrcView.

PR #2351, opened 2022-09-07
Add SrcView::try_insert function

suttonbradley:

_What is this about?_

## PR Checklist
* [x] Applies to work item: #2350 
* [x] Tests added/passed
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: chat
* [x] Requires documentation to be updated: doc comments

## Info on Pull Request

Adds `try_insert` function to the `SrcView` struct of the `srcview` crate so that a user can ensure a PDB associated with the given module name has been parsed without re-parsing the PDB.

## Validation Steps Performed

TODO: currently working on integrating with the existing tests

### DIFF
--- a/src/agent/srcview/src/srcview.rs
+++ b/src/agent/srcview/src/srcview.rs
@@ -52,30 +52,30 @@ impl SrcView {
     /// Insert a new pdb into the SrcView only if the `pdb` path is not in the SrcView already,
     /// returning a  [Result] indicating the success of the insert, if any was necessary.
     /// If the [Result] is [Ok], the contained bool indicates whether a value was inserted.
-    /// 
+    ///
     /// # Arguments
-    /// 
+    ///
     /// * `module` - Module name to store the PDB info as
     /// * `pdb` - Path to PDB
-    /// 
+    ///
     /// # Errors
-    /// 
+    ///
     /// If the PDB at the provided path has not been inserted already **and**
     /// cannot be parsed or contains otherwise unexpected data.
-    /// 
+    ///
     /// # Example
-    /// 
+    ///
     /// ```no_run
     /// use srcview::SrcView;
-    /// 
+    ///
     /// let mut sv = SrcView::new();
     /// let modoffs = get_coverage();
-    /// 
+    ///
     /// // Map each modoff to a PDB name/path and make sure it's in the SrcView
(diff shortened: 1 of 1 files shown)


src/agent/srcview/src/srcview.rs
```diff
@@ -49,6 +49,42 @@ impl SrcView {
...
+    /// let mut sv = SrcView::new();
+    /// let modoffs = get_coverage();
+    /// 
+    /// // Map each modoff to a PDB name/path and make sure it's in the SrcView
+    /// for modoff in modoffs {
+    ///     let module_name =  mod_name_from_modoff(modoff);
+    ///     sv.try_insert(module_name, format!("~/pdbs/{module_name}.pdb"));
+    /// }
+    /// ```
+    pub fn try_insert<P: AsRef<Path>>(&mut self, module: &str, pdb: P) -> Result<()> {
```
Porges, line 80: It's often useful for the caller to know if an insertion happened or not, suggest returning `Result<bool>`?

suttonbradley: Great suggestion. Fixed.

Porges:
@suttonbradley note that the files will need to be `rustfmt`’d to pass CI

src/agent/srcview/src/srcview.rs
```diff
@@ -49,6 +49,46 @@ impl SrcView {
...
+    ///     if let Ok(inserted) = res {
+    ///         println!("PDB was inserted: {inserted}");
+    ///     }
+    /// }
+    /// ```
+    pub fn try_insert<P: AsRef<Path>>(&mut self, module: &str, pdb: P) -> Result<bool> {
+        if self.0.contains_key(&module.to_owned()) {
+            Ok(false)
+        } else {
+            self.insert(module, pdb).map(|_| true)
```
comcma, line 88: I think the value to be inserted should be a PdbCache 
`let cache = PdbCache::new(pdb)?;`

comcma: Ah, thanks. Bad eyes on my part.

Porges (approved):

suttonbradley:
@bot agree

PR #2358, opened 2022-09-08
Fix: Inconsistencies in VMSS creation between C#/Python

Porges:
For spot instances:
* `EvictionPolicy` was `Delete` in Python, `Deallocate` in C#
* `BillingMaxPrice` was `-1.0` in Python, `1.0` in C#

At top level:

* `DoNotRunExtensionsOnOverprovisionedVms` was `true` in Python, `false` in C# (however, at the moment we disable overprovisioning)

Reformatted code to make it similar to Python and easier to compare.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -259,18 +259,7 @@ public class VmssOperations : IVmssOperations {
             return getOsResult.ErrorV;
         }
 
-        var vmssData = new VirtualMachineScaleSetData(location) {
-            DoNotRunExtensionsOnOverprovisionedVms = false,
-            Sku = new ComputeSku() { Name = vmSku, Capacity = vmCount },
-            Overprovision = false,
-            SinglePlacementGroup = false,
-            UpgradePolicy = new UpgradePolicy() { Mode = UpgradeMode.Manual },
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -325,9 +352,9 @@ public class VmssOperations : IVmssOperations {
             //
             // https://docs.microsoft.com/en-us/azure/
             //   virtual-machine-scale-sets/use-spot#resource-manager-templates
-            vmssData.VirtualMachineProfile.EvictionPolicy = VirtualMachineEvictionPolicyTypes.Deallocate;
+            vmssData.VirtualMachineProfile.EvictionPolicy = VirtualMachineEvictionPolicyTypes.Delete;
             vmssData.VirtualMachineProfile.Priority = VirtualMachinePriorityTypes.Spot;
-            vmssData.VirtualMachineProfile.BillingMaxPrice = 1.0;
+            vmssData.VirtualMachineProfile.BillingMaxPrice = -1.0;
```
stishkin, line 357: -1.0 ?

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
Porges: See comment above: "Setting max price to -1 means it won't be evicted because of price."

tevoinea (approved):

Issue #1754, opened 2022-04-06
[Refactor] queue_file_changes Azure App Function


tevoinea:
Refactor the QueueFileChanges function and required sub classes to C#. Validate by disabling the python function and enabling the .net function in a testing environment before deploying the change to production.

**Blocked on templating**:
* https://github.com/microsoft/onefuzz/blob/447f1813b1c3fac0e932c575f51bbe1538194936/src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs#L37
* https://github.com/microsoft/onefuzz/blob/447f1813b1c3fac0e932c575f51bbe1538194936/src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs#L45
* https://github.com/microsoft/onefuzz/blob/447f1813b1c3fac0e932c575f51bbe1538194936/src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs#L49

AB#39993385

AB#40263775

chkeita:
@tevoinea Can we keep the issue open until we have  full check pr pass with the changes integrated.

PR #2363, opened 2022-09-08
Refactor notification support

tevoinea:

_What is this about?_

Makes progress on #1754. After this PR we need to add a script that can migrate the templates supported by python to ones supported by C#. Then, it'll be ready to be enabled in testing environments/prod.

## Info on Pull Request

_What does this include?_

* Capability to notify teams/ado/github
* Render templates

## Validation Steps Performed

**Only tested with teams so far. No expecting any huge changes after validating ado/gh**

1. Create a new notification config for teams following the docs in `teams.md`
2. Spin up a new libfuzzer job
3. Watch the notifications come in
4. Validate the produced link works


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -85,7 +85,11 @@ public record ProxyHeartbeat
     Guid ProxyId,
     List<Forward> Forwards,
     DateTimeOffset TimeStamp
-);
+) {
+    public override string ToString() {
+        return JsonSerializer.Serialize(this);
+    }
```
tevoinea, line 91: `ProxyHeartbeat.ToString()` was being used in some log messages but the `List<Forwards>` was being serialized as ```System.Collections.Generic.List`1[Microsoft.OneFuzz.Service.Forward]```

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -471,19 +475,33 @@ public class NotificationTemplateConverter : JsonConverter<NotificationTemplate>
     public override NotificationTemplate? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) {
         using var templateJson = JsonDocument.ParseValue(ref reader);
         try {
-            return templateJson.Deserialize<AdoTemplate>(options);
-        } catch (JsonException) {
+            return ValidateDeserialization(templateJson.Deserialize<AdoTemplate>(options));
+        } catch (Exception ex) when (
+            ex is JsonException
```
stishkin, line 480: log exceptions ?

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -471,19 +475,33 @@ public class NotificationTemplateConverter : JsonConverter<NotificationTemplate>
...
 
         try {
-            return templateJson.Deserialize<TeamsTemplate>(options);
-        } catch (JsonException) {
+            return ValidateDeserialization(templateJson.Deserialize<TeamsTemplate>(options));
+        } catch (Exception ex) when (
+            ex is JsonException
+            || ex is ArgumentNullException
+            || ex is ArgumentOutOfRangeException
+        ) {
```
stishkin, line 493: log exceptions

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -471,19 +475,33 @@ public class NotificationTemplateConverter : JsonConverter<NotificationTemplate>
...
-        } catch (JsonException) {
+            return ValidateDeserialization(templateJson.Deserialize<GithubIssuesTemplate>(options));
+        } catch (Exception ex) when (
+            ex is JsonException
+            || ex is ArgumentNullException
+            || ex is ArgumentOutOfRangeException
+        ) {
+
         }
         throw new JsonException("Unsupported notification template");
```
stishkin, line 506: so if github template throws JsonException, then we'll get here and throw "Unsupported notification template"

is that correct logic ?

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -0,0 +1,283 @@
...
+container:{container} filename:{filename}";
+
+        _logTracer.Info($"notify ado: {notificationInfo}");
+
+        try {
+            var ado = await AdoConnector.AdoConnectorCreator(_context, container, filename, config, report, _logTracer);
+            await ado.Process(notificationInfo);
+        } catch (Exception e) {
+            /*
+            TODO: Catch these
```
stishkin, line 36: TODO for later ?

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -0,0 +1,283 @@
...
+                AzureDevOpsAuthenticationError,
+                AzureDevOpsClientError,
+                AzureDevOpsServiceError,
+                AzureDevOpsClientRequestError,
+                ValueError,
+            */
+            if (!failTaskonTransientError && IsTransient(e)) {
```
stishkin, line 43: NIT: failTaskOnTransientError 

Capitalize "On"

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
tevoinea: Yup, it's going to be part of testing ado

src/ApiService/ApiService/OneFuzzTypes/Model.cs
tevoinea: That's right, there's only those 3 templates that we support.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
tevoinea: The exceptions are just a mechanism to "fall-through" until we deserialize to the right type, that's why they're always caught. The only exception that's important is the last one because it indicates we weren't able to deserialize to any of our known types. Also, logging isn't available here, it would have to be done after serialization (which we already do).

I improved the exception message to be clearer in the case where none of the type deserialize.

PR #2373, opened 2022-09-11
Validate changes to devcontainer files

Porges:
When a PR is sent that changes devcontainer configuration files (under `.devcontainer`), ensure that the devcontainer can be built and run.

### DIFF
--- a/.github/workflows/validate-devcontainer.yml
+++ b/.github/workflows/validate-devcontainer.yml
@@ -1,0 +1,21 @@
+name: Validate Devcontainer configuration
+on:
+  # run on PRs to main branch, if any 
+  # devcontainer files have changed
+  pull_request:
+    branches:
+    - main
+    paths:
+    - .devcontainer/**
+
+jobs:
+  use-devcontainer:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-devcontainer.yml
```diff
@@ -0,0 +1,21 @@
...
+  use-devcontainer:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    # note that all we do here is ensure the devcontainer can
+    # setup and run one command, ensuring that it works
+    - name: Build and use devcontainer
+      uses: devcontainers/ci@v0.2
+      with:
+        runCmd: cargo --version
```
tevoinea, line 21: We could also assert that all of the tools we expect to exist are there:

`cargo`, `dotnet`, `func`, `azcopy`, `az`, `python`

tevoinea (approved):

nharper285 (approved):

.github/workflows/validate-devcontainer.yml
Porges: I think the next step would be to use the devcontainer as the build image, which would check this implicitly (as well as provide other benefits). I won't do that in this PR.

PR #2379, opened 2022-09-12
tests and bug fixes

stishkin:
tests and bug fixes

there is more tests to write...


tevoinea (approved):

src/ApiService/FunctionalTests/1f-api/Info.cs
```diff
@@ -0,0 +1,48 @@
...
+        public InfoVersion(JsonElement e) => _e = e;
+        public InfoVersion Convert(JsonElement e) => new InfoVersion(e);
+
+        public string Git => _e.GetProperty("git").GetString()!;
+        public string Build => _e.GetProperty("build").GetString()!;
+        public string Version => _e.GetProperty("version").GetString()!;
+    }
+
+
+    public class InfoResponse : IFromJsonElement<InfoResponse> {
```
Porges, line 20: Question: what is the reason we aren't reusing the service request/response types in these tests?

stishkin: Long time ago a Microsoft test manager told me that you never should use dev code to test dev code. Since that will miss bunch of bugs.

I don't think we should be doing it everywhere, but at least in one place it might be a good idea.

Porges: Maybe it will be a good idea to use Swagger-generated definitions here, when we have them.

PR #2380, opened 2022-09-12
implement not-implemented: GetInputContainerQueues

stishkin:


src/ApiService/ApiService/onefuzzlib/TaskOperations.cs
```diff
@@ -15,7 +15,7 @@ public interface ITaskOperations : IStatefulOrm<Task, TaskState> {
 
     IAsyncEnumerable<Task> SearchStates(Guid? jobId = null, IEnumerable<TaskState>? states = null);
 
-    IEnumerable<Container>? GetInputContainerQueues(TaskConfig config);
```
tevoinea, line 18: Did you mean to use `OneFuzzResult`? I don't think we use `Result<T, E>` anywhere outside of `FunctionalTests`

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/TaskOperations.cs
Porges: We should probably just rename `OneFuzzResult` to `Result`…

stishkin: `OneFuzzResult` Error is of OneFuzz Error type `{int code, string[] errors}`

Anywhere in the code where `TaskConfigError` is returned it is returned as `ResultVoid<TaskConfigError>.Error(new TaskConfigError(err));`

In this specific case `ResultVoid` won't do

Next code iteration - maybe might make sense to get rid of `ResultVoid<TaskConfigError>` as well as fix up this code and use `OneFuzzResult` across the board

Porges (approved):

src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs
```diff
@@ -84,7 +84,8 @@ public NotificationOperations(ILogTracer log, IOnefuzzContext context)
         // Nullability mismatch: We filter tuples where the containers are null
         return _context.TaskOperations.SearchStates(states: TaskStateHelper.AvailableStates)
             .Select(task => (task, _context.TaskOperations.GetInputContainerQueues(task.Config)))
-            .Where(taskTuple => taskTuple.Item2 != null)!;
+            .Where(taskTuple => taskTuple.Item2.IsOk && taskTuple.Item2.OkV != null)
+            .Select(x => (x.Item1, x.Item2.OkV))!;
```
Porges, line 88: We should return a named tuple here to make the code easier to understand.

PR #2386, opened 2022-09-13
Require client-side prefixing of task config params

ranweiler:
Remove ad hoc hardcoded server-side prefixing of task config parameters. This allows us to use them to their full potential, such as invoking a task where the task-relative `target_exe` is actually shipped as part of the `tools_dir` and not the `setup_dir`.

Ideally, we'd also use `{setup_dir}` prefixing instead of `setup/`, but this doesn't yet work as expected due to the implementation of eager path canonicalization in `Expand`.


src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -550,7 +550,7 @@ def dotnet(
         if target_options is None:
             target_options = []
         target_options = [
-            "--target_path={setup_dir}/" f"{target_harness}"
+            "--target_path={setup_dir}/" + "{target_harness}"
```
ranweiler, line 553: Opportunistic fix, decided this was too brittle (works for constants, but not variables).

ranweiler:
Declining these changes in favor of removing prefixing requirements completely (for now).

PR #2391, opened 2022-09-13
Use Github-hosted Ubuntu until issue with self-hosted pool is resolved

Porges:


mgreisen (approved):

.github/workflows/ci.yml
```diff
@@ -21,7 +21,9 @@ jobs:
       matrix:
         os:
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-windows-2019" ]
-        - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-18.04" ] 
+        # disabled while issue is affecting instances
+        #- [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-18.04" ] 
+        - ubuntu-18.04
```
ranweiler, line 26: 1. I'm guessing not, but can this be controlled with input parameters, or some similar mechanism? Something like adding a `use_github_hosted` variable that defaults to `false`.
2. Even if "yes" to (1), could we make the branch protection checks agnostic to the pool? If not, then the idea in (1) doesn't really help us anyway.

Porges: Just going to resolve this so I can get it merged, but will investigate these. 

The branch-protection checks already don't care about the pool, since we require `package` to pass, which depends on the agent builds.

PR #2403, opened 2022-09-15
Split build test

Porges:
Testing splitting agent build into build/test portions.


Porges:
Quite fiddly. Won't do it for now.

Issue #2399, opened 2022-09-14
Revisit API and implementation of `target_exe`


ranweiler:
Right now, for any given task, the following are true:
- The `target_exe` parameter (`TaskDetails.target_exe`) must be a `setup` blob container-relative virtual path
- The `target_exe` parameter cannot contain any `Expand` placeholders (see #2387)

Server-side, when generating a `TaskUnitConfig`', we hard-prepend `setup/` to `TaskUnitConfig.target_exe` (which is what the task executor on the VM eventually receives). As a result, the implementations of the tasks assume that `target_exe` is always directly addressable relative to the current working directory of the executor, which contains the `setup` dir under the concrete name `setup`.

From the perspective of the `onefuzz-task` executor, if we invoked `tree .`, we'd see something like:
```
.
├── crashes_1
├── inputs_2
├── readonly_inputs_3
├── setup
│   └── fuzzer.exe
└── tmp.5cj9Wh2pzl
``` 
...with the specifics depending on the task being executed. However, we will _always_ have a directory `./setup`, which will always contain the task `target_exe`. The server-side tables will have `TaskDetails.target_exe == fuzzer.exe`, but the task config JSON on the VM (via the server-generated `TaskUnitConfig`) will have `TaskUnitConfig.target_exe == setup/fuzzer.exe`. Similarly, for crash repro, we generate a GDB script ([here](https://github.com/microsoft/onefuzz/blob/2fe73ab79cfd77caecba82b39f8d3f6ae3db2260/src/api-service/__app__/onefuzzlib/repro.py#L218) or [here](https://github.com/microsoft/onefuzz/blob/2fe73ab79cfd77caecba82b39f8d3f6ae3db2260/src/ApiService/ApiService/onefuzzlib/ReproOperations.cs#L251)) that does a similar hardcoding, directly from the `TaskDetails`.

This all has a few quirks and/or problems:
- `target_exe` is special and different from other `.._exe` config parameters, which _must_ use an expansion placeholder variable to relativize to `{tools_dir}` or `{setup_dir}`. In contrast, `target_exe` _cannot_ be relativized, either conceptually or in implementation. The other `.._exe` params can be _any_ path (including absolute paths)
- At least in templates, some special conditions apply to the relationship between `target_exe` and `setup`
[...]

PR #2405, opened 2022-09-15
Prefix `target_exe` with `setup` dir at use sites

ranweiler:
Stop server-side prefixing of task `target_exe` parameters with a hardcoded `setup/` relative path. Instead, prefix `target_exe` in each task implementation, using the absolute path of the local mirror of the `setup` container.

Note that this preserves the existing end-user API and internal data model, but internally enforces the following invariant: the `target_exe` field that occurs in `TaskDetails` and `TaskUnitConfig` now match, and are always a `setup`-relative virtual path (blob name). Users of `target_exe` (like the `onefuzz-task` runner) must combine this relativized subpath with a path to a local `setup` directory to concretely locate the `target_exe`.

This change will effect the following breakage on upgrade: old `TaskUnitConfig`'s sitting in a work queue as part of a work set will have a `target_exe` that has been pre-prefixed with `setup/`. If an upgraded agent picks up the old task, it will result in double-prefixing the `target_exe` at runtime, causing the task to fail.

See larger long-term design discussion in #2399.


src/ApiService/ApiService/onefuzzlib/Config.cs
```diff
@@ -257,9 +257,8 @@ await _containers.GetContainerSasUrl(x.Item2.Name, StorageType.Corpus, ConvertPe
         }
 
         if (definition.Features.Contains(TaskFeature.CoverageFilter)) {
-            var coverageFilter = task.Config.Task.CoverageFilter;
             if (coverageFilter != null) {
-                config.CoverageFilter = $"setup/{coverageFilter}";
+                config.CoverageFilter = task.Config.Task.CoverageFilter;
```
ranweiler, line 261: This change has not been accounted for the VM agent, addressing now.

src/agent/onefuzz-task/src/tasks/analysis/generic.rs
```diff
@@ -194,11 +194,13 @@ pub async fn run_tool(
     config: &Config,
     reports_dir: &Option<PathBuf>,
 ) -> Result<()> {
+    let target_exe = config.common.setup_dir.join(&config.target_exe);
```
ranweiler, line 197: To avoid the upgrade breakage described in the PR message, we could temporarily guard this `.join()` on a call to `.starts_with("setup/")`. This does have some edge cases, though, e.g. if a user had a `setup` container which had a nested `target_exe` that (abstractly) resided in `{setup_dir}/setup/fuzz.exe`.

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -550,7 +550,7 @@ def dotnet(
         if target_options is None:
             target_options = []
         target_options = [
-            "--target_path={setup_dir}/" f"{target_harness}"
```
ranweiler, line 553: Opportunistic fix from other PR, which I closed without merging.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -138,9 +138,34 @@ impl CoverageTask {
...
 
+        let setup_dir = &self.config.common.setup_dir;
+
         // Ensure users can locate the filter relative to the setup container.
-        let expand = Expand::new().setup_dir(&self.config.common.setup_dir);
-        let filter_path = expand.evaluate_value(raw_filter_path)?;
+        let expand = Expand::new().setup_dir(setup_dir);
+        let expanded_filter_path = expand.evaluate_value(raw_filter_path)?;
+
+        // Determine if `coverage_filter` is already prefixed with a absolute or relative
```
ranweiler, line 147: Hmm, we can do this more robustly by introducing a `find_setup_relative_file()` function which could use file existence checks and attempted expansions to handle the cases of no prefixing (a fully relativized subpath), `setup/`-prefixing, and `{setup_dir}`-prefixing. We could use that same function wherever we are prepending `{setup_dir}` to `target_exe`. Let's try that instead.

src/agent/onefuzz-task/src/tasks/analysis/generic.rs
tevoinea: I think the guard makes sense -- we go from guaranteed breakage of all inflight 
 WorkUnits  to possible/unlikely breakage. The only thing is, after this change gets deployed, we should remove the guard in the following deploy so we don't keep this edge case around.

tevoinea (approved):

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -138,9 +139,17 @@ impl CoverageTask {
...
+        let resolved =
+            resolve_setup_relative_path(&self.config.common.setup_dir, raw_filter_path).await?;
+        let filter_path = if let Some(path) = resolved {
+            path
+        } else {
+            error!(
+                "unable to resolve setup-relative coverage filter path: {}",
+                raw_filter_path
+            );
+            return Ok(CmdFilter::default());
```
ranweiler, line 151: Note the behavior here: if a `coverage_filter` was specified, but we couldn't find it, fall back to the default (empty) filter.

Alternatively, we may just fail the task. Currently, this is the only way to proactively inform the user that their `coverage_filter` subpath was invalid (present, but not resolvable under `setup_dir`).

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
```diff
@@ -236,12 +236,7 @@ where
                 .for_each(|d| inputs.push(&d.local_path));
         }
 
-        let fuzzer = LibFuzzer::new(
-            &self.config.target_exe,
-            self.config.target_options.clone(),
-            self.config.target_env.clone(),
-            &self.config.common.setup_dir,
-        );
+        let fuzzer = L::from_config(&self.config).await?;
```
ranweiler, line 239: Opportunistic fix. We should only ever be creating `LibFuzzer` values via `L::from_config()`.

src/agent/onefuzz-task/src/tasks/utils.rs
```diff
@@ -96,3 +96,157 @@ pub fn parse_key_value(value: String) -> Result<(String, String)> {
...
+            anyhow::format_err!(
+                "unable to resolve subpath `{}` under setup dir `{}`",
+                subpath.display(),
+                setup_dir.display()
+            )
+        })
+}
+
+/// Try to resolve an ambiguous setup-relative subpath, returning `None` if not found.
+pub async fn resolve_setup_relative_path(
```
ranweiler, line 120: @tevoinea, let me know if you have any questions about this behavior.

One thing I just noticed: I think a (valid, existing) absolute path will result in `Ok(None)`. We should probably check for that.

ranweiler: (that is: we should either reject any absolute `subpath`, or alternatively, if `subpath` is absolute, and `setup_dir` is already a prefix of `subpath`, then this should arguably succeed)

src/agent/onefuzz-task/src/tasks/utils.rs
```diff
@@ -96,3 +96,157 @@ pub fn parse_key_value(value: String) -> Result<(String, String)> {
...
+
+        test_case(RELATIVE, "x/y/fuzz.exe").await?;
+        test_case(RELATIVE, "setup/x/y/fuzz.exe").await?;
+        test_case(RELATIVE, "{setup_dir}/x/y/fuzz.exe").await?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_resolve_setup_relative_path_nested_setup() -> Result<()> {
```
ranweiler, line 243: This tests the unlikely edge case that we have something like a _new_, non-legacy `target_exe` that is prefixed by `setup/` because the target genuinely exists in a path like `{setup_dir}/setup/fuzz.exe`.

src/agent/onefuzz-task/src/tasks/utils.rs
tevoinea: What do you think of logging a warning if more than one of these resolve? I imagine it's _probably_ a misconfiguration and the user would want to know.

ranweiler: I'm going to stick with "return the first resolved" for now, because I don't actually want to actually attempt every single case. The main goal here is to have have a comprehensible resolution rule/order to aid debugging.

tevoinea (approved):

PR #2406, opened 2022-09-15
fix some bugs

stishkin:
- fix queries in timer retention
- do not discard proxy record after proxy state is processed, since that record needs to persist

### DIFF
--- a/src/ApiService/ApiService/Functions/TimerProxy.cs
+++ b/src/ApiService/ApiService/Functions/TimerProxy.cs
@@ -20,7 +20,9 @@ public class TimerProxy {
 
         var proxies = await proxyOperations.QueryAsync().ToListAsync();
 
-        foreach (var proxy in proxies) {
+        foreach (var p in proxies) {
+            var proxy = p;
+
             if (VmStateHelper.Available.Contains(proxy.State)) {
                 // Note, outdated checked at the start, but set at the end of this loop.
                 // As this function is called via a timer, this works around a user
@@ -40,11 +42,14 @@ public class TimerProxy {
 
             if (VmStateHelper.NeedsWork.Contains(proxy.State)) {
                 _logger.Info($"scaleset-proxy: update state. proxy:{proxy.Region} state:{proxy.State}");
-                await proxyOperations.ProcessStateUpdate(proxy);
+                proxy = await proxyOperations.ProcessStateUpdate(proxy);
             }
 
-            if (proxy.State != VmState.Stopped && proxyOperations.IsOutdated(proxy)) {
-                await proxyOperations.Replace(proxy with { Outdated = true });
+            if (proxy is not null && (proxy.State != VmState.Stopped && proxyOperations.IsOutdated(proxy))) {
+                var r = await proxyOperations.Replace(proxy with { Outdated = true });
+                if (!r.IsOk) {
+                    _logger.Error($"Failed to replace proxy recordy for proxy {proxy.ProxyId} due to {r.ErrorV}");
+                }
             }
         }
 
@@ -54,8 +59,8 @@ public class TimerProxy {
         foreach (var region in regions) {
             var allOutdated = proxies.Where(x => x.Region == region).All(p => p.Outdated);
             if (allOutdated) {
-                await proxyOperations.GetOrCreate(region);
-                _logger.Info($"Creating new proxy in region {region}");
+                var proxy = await proxyOperations.GetOrCreate(region);
+                _logger.Info($"Creating new proxy with id {proxy.ProxyId} in region {region}");
             }
 
             // this is required in order to support upgrade from non-nsg to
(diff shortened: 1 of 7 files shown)


src/ApiService/ApiService/Functions/TimerProxy.cs
```diff
@@ -40,11 +42,14 @@ public class TimerProxy {
 
             if (VmStateHelper.NeedsWork.Contains(proxy.State)) {
                 _logger.Info($"scaleset-proxy: update state. proxy:{proxy.Region} state:{proxy.State}");
-                await proxyOperations.ProcessStateUpdate(proxy);
+                proxy = await proxyOperations.ProcessStateUpdate(proxy);
             }
 
-            if (proxy.State != VmState.Stopped && proxyOperations.IsOutdated(proxy)) {
-                await proxyOperations.Replace(proxy with { Outdated = true });
```
nharper285, line 47: This was the cause of the bad record replacements?

nharper285 (approved):

src/ApiService/ApiService/Functions/TimerProxy.cs
stishkin: we are using a potentially outdated record here, so we could overwrite an updated one with older version.

src/ApiService/ApiService/onefuzzlib/VmOperations.cs
```diff
@@ -218,7 +218,9 @@ await _context.Creds.GetResourceGroupResource()
                 extensionName,
                 extension
             );
-        } catch (RequestFailedException ex) when (ex.Status == 409 && ex.Message.Contains("VM is marked for deletion")) {
+        } catch (RequestFailedException ex) when
+            (ex.Status == 409 &&
+                (ex.Message.Contains("VM is marked for deletion") || ex.Message.Contains("The request failed due to conflict with a concurrent request."))) {
             _logTracer.Info(ex.Message);
```
tevoinea, line 224:
```suggestion
            _logTracer.Info($"Tried to create extension but failed due to {ex.Message}");
```

So we know what the message is about

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
```diff
@@ -46,12 +46,25 @@ public static string And(IEnumerable<string> queries)
...
+        }
+
         public static string TimeRange(DateTimeOffset min, DateTimeOffset max) {
             // NB: this uses the auto-populated Timestamp property, and will result in a table scan
             // TODO: should this be inclusive at the endpoints?
             return TableClient.CreateQueryFilter($"Timestamp lt {max} and Timestamp gt {min}");
         }
 
+        public static string NewerThan(DateTimeOffset t) {
+            return TableClient.CreateQueryFilter($"Timestamp gt {t}");
```
tevoinea, line 61:
```suggestion
            return TableClient.CreateQueryFilter($"Timestamp gt {t.ToString("o")}");
```

To keep the same behavior from `TimerRetention`

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
```diff
@@ -46,12 +46,25 @@ public static string And(IEnumerable<string> queries)
...
+        }
+
+        public static string OlderThan(DateTimeOffset t) {
+            return TableClient.CreateQueryFilter($"Timestamp lt {t}");
```
tevoinea, line 65:
```suggestion
            return TableClient.CreateQueryFilter($"Timestamp lt {t.ToString("o")}");
```

Same here.

src/ApiService/ApiService/Functions/TimerProxy.cs
```diff
@@ -40,11 +42,14 @@ public class TimerProxy {
...
+            if (proxy is not null && (proxy.State != VmState.Stopped && proxyOperations.IsOutdated(proxy))) {
```
tevoinea, line 48:
```suggestion
            if (proxy is not null && proxy.State != VmState.Stopped && proxyOperations.IsOutdated(proxy)) {
```

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
stishkin: TableClient.CreateQueryFilter <- does the conversion based on type. So it's supposed to do t.ToString("o") implicitly

src/ApiService/ApiService/onefuzzlib/orm/Queries.cs
stishkin: same as above

src/ApiService/ApiService/Functions/TimerProxy.cs
stishkin: ok

PR #2409, opened 2022-09-16
mark tasks as failed if a work unit cannot be created for the task

stishkin:


src/ApiService/ApiService/onefuzzlib/TaskOperations.cs
```diff
@@ -122,7 +120,7 @@ public TaskOperations(ILogTracer log, IOnefuzzContext context)
             return;
         }
 
-        _logTracer.Error($"task failed {task.JobId}:{task.TaskId} - {error}");
+        _logTracer.Info($"task failed {task.JobId}:{task.TaskId} - {error}");
```
tevoinea, line 123: I guess you're doing this because we're intentionally marking a task as failed?

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/TaskOperations.cs
stishkin: yes, it's not a bug, it's normal operation

PR #2415, opened 2022-09-20
fix linux repro extensions

stishkin:
Windows and Linux C# check-pr started to pass

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Auth.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Auth.cs
@@ -15,17 +15,19 @@ public class Auth {
             FileName = keyGen,
             CreateNoWindow = false,
             UseShellExecute = false,
-            RedirectStandardOutput = true,
-            RedirectStandardError = true
+            RedirectStandardOutput = false,
+            RedirectStandardError = true,
+            ArgumentList = {
+                "-t",
+                "rsa",
+                "-f",
+                tempFile,
+                "-P",
+                "",
+                "-b",
+                "2048"
+            }
         };
-        p.ArgumentList.Add("-t");
-        p.ArgumentList.Add("rsa");
-        p.ArgumentList.Add("-f");
-        p.ArgumentList.Add(tempFile);
-        p.ArgumentList.Add("-P");
-        p.ArgumentList.Add("");
-        p.ArgumentList.Add("-b");
-        p.ArgumentList.Add("2048");
         return p;
     }
 
@@ -35,14 +37,12 @@ public class Auth {
         File.Delete(tmpFile);
         tmpFile = tmpFile + ".key";
         var startInfo = SshKeyGenProcConfig(tmpFile);
-        var proc = new Process() { StartInfo = startInfo };
-        if (proc is null) {
-            throw new Exception($"ssh-keygen process could start (got null)");
-        }
+        using var proc = new Process() { StartInfo = startInfo };
         if (proc.Start()) {
+            var stdErr = await proc.StandardError.ReadToEndAsync();
             await proc.WaitForExitAsync();
             if (proc.ExitCode != 0) {
-                throw new Exception($"ssh-keygen failed due to {await proc.StandardError.ReadToEndAsync()}");
+                throw new Exception($"ssh-keygen failed due to {stdErr}");
             }
 
             var priv = File.ReadAllText(tmpFile);
--- a/src/ApiService/ApiService/onefuzzlib/Auth.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Auth.cs
@@ -13,7 +13,7 @@ public class Auth {
         }
         return new ProcessStartInfo() {
             FileName = keyGen,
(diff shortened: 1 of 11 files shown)


src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
+            RedirectStandardError = true
+        };
+        p.ArgumentList.Add("-t");
+        p.ArgumentList.Add("rsa");
+        p.ArgumentList.Add("-f");
+        p.ArgumentList.Add(tempFile);
+        p.ArgumentList.Add("-P");
+        p.ArgumentList.Add("");
+        p.ArgumentList.Add("-b");
+        p.ArgumentList.Add("2048");
```
Porges, line 28: You can initialize these in the object like so:

```suggestion
            RedirectStandardError = true,
            ArgumentList = {
                "-t",
                "rsa",
                "-f",
                tempFile,
                "-P",
                "",
                "-b",
                "2048"
            },
        };
```

src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
-        spanResult = WriteLengthPrefixedBytes(parameters.Exponent, spanResult);
-        spanResult = WriteLengthPrefixedBytes(parameters.Modulus, spanResult);
-        Debug.Assert(spanResult.Length == 0);
+    // This works both on Windows and Linux azure function hosts
+    private static async Async.Task<(string, string)> GenerateKeyValuePair() {
+        var tmpFile = Path.GetTempFileName();
+        File.Delete(tmpFile);
+        tmpFile = tmpFile + ".key";
+        var startInfo = SshKeyGenProcConfig(tmpFile);
+        var proc = new Process() { StartInfo = startInfo };
```
Porges, line 38: I think this should have `using` on it

Porges (approved):

src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
+        if (proc is null) {
```
Porges, line 39: This will never be true

src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
+            throw new Exception($"ssh-keygen process could start (got null)");
+        }
```
Porges, line 41: This can't happen, remove
```suggestion
```

src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
+        if (proc.Start()) {
+            await proc.WaitForExitAsync();
+            if (proc.ExitCode != 0) {
+                throw new Exception($"ssh-keygen failed due to {await proc.StandardError.ReadToEndAsync()}");
```
Porges, line 45: We need to start reading redirected outputs before waiting for exit or the program can get stuck and deadlock (output pipe fills up).

Porges: Set `RedirectStandardOutput = false` since we never read it, then put this read before the `WaitForExitAsync`.

src/ApiService/ApiService/onefuzzlib/Auth.cs
```diff
@@ -1,45 +1,65 @@
...
-        var result = new byte[sizeof(int) + SSHRSABytes.Length + sizeof(int) + parameters.Modulus!.Length + sizeof(int) + parameters.Exponent!.Length];
-        var spanResult = result.AsSpan();
-        spanResult = WriteLengthPrefixedBytes(SSHRSABytes, spanResult);
-        spanResult = WriteLengthPrefixedBytes(parameters.Exponent, spanResult);
-        spanResult = WriteLengthPrefixedBytes(parameters.Modulus, spanResult);
-        Debug.Assert(spanResult.Length == 0);
+    // This works both on Windows and Linux azure function hosts
+    private static async Async.Task<(string, string)> GenerateKeyValuePair() {
+        var tmpFile = Path.GetTempFileName();
+        File.Delete(tmpFile);
```
tevoinea, line 35: Maybe try/catch this since there's many things that can go wrong in file io?

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/Auth.cs
stishkin: I did this initially and it did not work since ArgumentList does not have "public set"

src/ApiService/ApiService/onefuzzlib/Auth.cs
stishkin: if we catch an exception - what do we do ? 
if auth creation fails - then we fail the whole operation.
And try again later.

PR #2423, opened 2022-09-20
Remove use of unapproved crypto libraries

Porges:
To comply with Microsoft security policies, we must use only approved crypto libraries.

Notably, `ring` is not on the approved libraries list. We should use the platform crypto libraries instead. The official guidance is:

- on Windows, schannel and friends
- on Linux, the OpenSSL 1.1.1 series if available, otherwise OpenSSL 3.0. OpenSSL _must_ be dynamically and not statically linked so that distro-provided updates can be used.

Following these guidelines strictly would mean that we have to build distro/version specific binaries for Linux. Instead we have an exception which allows us to statically link against the OpenSSL 1.1.1 series, allowing us to have a (somewhat) portable binary.

OpenSSL 1.1.1. can be statically linked by switching to the `native-tls-vendored` feature of `reqwest`.  (Verified by inspection that `ring` and `rustls` are no longer in `Cargo.lock` files.)

So the result is:

- on Windows, we use schannel
- on Linux we statically link against 1.1.1

It is up to us to ensure we keep up-to-date with OpenSSL issues and releases. The version can be checked by looking at the version of `openssl-src` that we depend upon. Currently this is `111.22.0+1.1.1q`. This is the latest version; also check the [Vulnerabilities page](https://www.openssl.org/news/vulnerabilities-1.1.1.html).

--- 

Verified by inspection that `rustls` and `ring` are no longer present in `Cargo.lock`.

Current dynamic library dependencies are:

```console
$ ldd ./onefuzz-agent 
        linux-vdso.so.1 (0x00007ffd9ba60000)
        libunwind.so.8 => /lib/x86_64-linux-gnu/libunwind.so.8 (0x00007ff5ab5e5000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007ff5ac1d1000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff5ac1cc000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff5ab4fe000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff5ac1c7000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff5ab2d4000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff5ac1dc000)
        liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007ff5ab2a9000)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,10 +34,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install specific Rust version
+      uses: actions-rs/toolchain@v1
       # note: keep this in sync with .devcontainer/Dockerfile
-      run: |
-        rustup default 1.64 
-        rustup component add clippy rustfmt llvm-tools-preview
+      with:
+        toolchain: 1.64
+        default: true
+        components: clippy, rustfmt, llvm-tools-preview
     - name: Get Rust version & build version
       shell: bash
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -492,7 +492,7 @@ jobs:
           cp artifacts/agent-MINGW*/onefuzz_task.pdb src/deployment/tools/win64/
 
           for linux_dir in artifacts/agent-Linux/*; do 
-            dest="src/deployment/tools/$(basename "$linux_dir")"
+            dest="src/deployment/tools/linux/$(basename "$linux_dir")"
             mkdir -p "$dest"
             cp "$linux_dir/onefuzz-agent" "$dest"
             cp "$linux_dir/onefuzz-task" "$dest"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
         os:
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-windows-2019" ]
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-18.04" ] 
+        - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-20.04" ] 
     runs-on: "${{ matrix.os }}"
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -56,6 +60,11 @@ jobs:
         path: artifacts
         key: agent-artifacts|${{ join(matrix.os, ':') }}|${{steps.rust-version.outputs.RUST_VERSION}}|${{ env.ACTIONS_CACHE_KEY_DATE }}|${{ hashFiles('src/agent/**/*') }}|${{hashFiles('src/ci/agent.sh')}}
         # note: also including the ACTIONS_CACHE_KEY_DATE to rebuild if the Prereq Cache is invalidated
+    - name: Linux Prereqs
```
Porges, line 63: This was reordered because we need the Linux pre-reqs installed before we run `rust-prereqs.sh` in the Rust pre-reqs step (on newer versions of Ubuntu).

.github/workflows/ci.yml
```diff
@@ -28,14 +28,18 @@ jobs:
         os:
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-windows-2019" ]
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-18.04" ] 
+        #- [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-20.04" ] 
+        #- [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-22.04" ] 
     runs-on: "${{ matrix.os }}"
     steps:
     - uses: actions/checkout@v2
     - name: Install specific Rust version
+      uses: actions-rs/toolchain@v1
```
Porges, line 37: Newer images do not have `rustup` by default.

.github/workflows/ci.yml
```diff
@@ -28,14 +28,18 @@ jobs:
         os:
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-windows-2019" ]
         - [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-18.04" ] 
+        #- [ self-hosted, "1ES.Pool=onefuzz-ci", "1ES.ImageOverride=github-runner-image-ubuntu-20.04" ]
```
Porges, line 31: I have been testing this with newer versions as well but keeping them disabled for now.

chkeita (approved):

mgreisen (approved):

PR #2433, opened 2022-09-21
cleanup queues for non-existent pools and non-existent tasks

### LINKED ISSUE #2432: Add task queue garbage collection
Add queue task garbage collection in case tasks records are deleted from task table manually. Then delete task queues when tasks do not exist

stishkin:
closes #2432

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Queue.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Queue.cs
@@ -2,6 +2,7 @@
 using System.Threading.Tasks;
 using Azure.Core;
 using Azure.Storage.Queues;
+using Azure.Storage.Queues.Models;
 using Azure.Storage.Sas;
 using Microsoft.OneFuzz.Service.OneFuzzLib.Orm;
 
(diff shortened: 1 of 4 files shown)


Porges (approved):

src/ApiService/ApiService/onefuzzlib/Queue.cs
```diff
@@ -30,6 +34,12 @@ public class Queue : IQueue {
         _log = log;
     }
 
+    public async IAsyncEnumerable<QueueItem> ListQueues(StorageType storageType) {
+        var queueServiceClient = await GetQueueClientService(storageType);
+        await foreach (var q in queueServiceClient.GetQueuesAsync()) {
```
Porges, line 39: this can return `queueServiceClient.GetQueuesAsync()` directly rather than yielding each item

stishkin: like this ?
image

stishkin: I think if you suggestion works - we can do this in ORM as well

stishkin (commented):
..

PR #2438, opened 2022-09-22
Add more logs

stishkin:
Logs, logs, logs

### DIFF
--- a/src/ApiService/ApiService/Functions/AgentRegistration.cs
+++ b/src/ApiService/ApiService/Functions/AgentRegistration.cs
@@ -136,7 +136,7 @@ public class AgentRegistration {
 
         var r = await _context.NodeOperations.Replace(node);
         if (!r.IsOk) {
-            _log.WithHttpStatus(r.ErrorV).Error($"failed to replace node operations for node {node.MachineId}");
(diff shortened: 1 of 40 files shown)


src/ApiService/ApiService/Functions/AgentRegistration.cs
```diff
@@ -134,7 +134,10 @@ public class AgentRegistration {
             ScalesetId: scalesetId,
             Version: version);
 
-        await _context.NodeOperations.Replace(node);
+        var r = await _context.NodeOperations.Replace(node);
+        if (!r.IsOk) {
+            _log.WithHttpStatus(r.ErrorV).Error($"failed to replace node operations for node {node.MachineId}");
+        }
```
mgreisen, line 140: does the machineId show up as a field in the customDimension?

stishkin: no it's part of the message

I will redo logging later on to process formatted strings and then it will show up in custom dimensions (in this specific case I'll do it manually)

mgreisen: Yeah, it's just really handy to have those kinds of fields factored out.

Porges (approved):

PR #2443, opened 2022-09-23
Add Roslyn analyzer to check results are used

### LINKED ISSUE #2442: Make ORM use more foolproof
At the moment it's easy to drop an updated entity and fail to use it in subsequent operations. This is because we have immutable record objects.

I suggest we update methods of the form `T UpdateEntity(T entity, …)` to take a `ref` parameter instead `UpdateEntity(ref T entity, …)`, so that the caller cannot fail to use the updated entity.

This will also make explicit operations which update the entity state.

Porges:
As seen in #2441, it is easy to drop return values of updated entities accidentally.

This PR adds a Roslyn Analyzer which will detect when return values are unused. To explicitly ignore a value you can drop it with `_ = …;`

Closes #2442.

### DIFF
--- a/src/ApiService/ApiService/Functions/AgentEvents.cs
+++ b/src/ApiService/ApiService/Functions/AgentEvents.cs
@@ -129,7 +129,7 @@ public class AgentEvents {
                     // Other states we would want to preserve are excluded by the
                     // outermost conditional check.
                     if (task.State != TaskState.Running && task.State != TaskState.SettingUp) {
-                        await _context.TaskOperations.SetState(task, TaskState.SettingUp);
+                        task = await _context.TaskOperations.SetState(task, TaskState.SettingUp);
                     }
 
                     var nodeTask = new NodeTasks(
@@ -155,7 +155,8 @@ public class AgentEvents {
             // if tasks are running on the node when it reports as Done
             // those are stopped early
             await _context.NodeOperations.MarkTasksStoppedEarly(node, error);
-            await _context.NodeOperations.ToReimage(node, done: true);
+            // result ignored: not used after this point
+            _ = await _context.NodeOperations.ToReimage(node, done: true);
         }
 
         return null;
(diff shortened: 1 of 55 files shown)


src/ApiService/ApiService/Functions/TimerRepro.cs
```diff
@@ -14,20 +14,25 @@ public class TimerRepro {
...
+                _ = expiredVmIds.Add(repro.VmId);
+                // ignoring result: value not used later
+                _ = await _onefuzzContext.ReproOperations.Stopping(repro);
+            }
         }
 
-        var expiredVmIds = expired.Select(repro => repro?.VmId);
-
         await foreach (var repro in _onefuzzContext.ReproOperations.SearchStates(VmStateHelper.NeedsWork)) {
-            if (await expiredVmIds.ContainsAsync(repro.VmId)) {
```
Porges, line 26: Also fixed this: this would perform the query again each time and be very slow.

stishkin (approved):

src/ApiService/ApiService/Functions/AgentEvents.cs
```diff
@@ -155,7 +155,8 @@ public class AgentEvents {
             // if tasks are running on the node when it reports as Done
             // those are stopped early
             await _context.NodeOperations.MarkTasksStoppedEarly(node, error);
-            await _context.NodeOperations.ToReimage(node, done: true);
+            // result ignored: not used after this point
+            _ = await _context.NodeOperations.ToReimage(node, done: true);
```
chkeita, line 159: I propose we encapsulate this pattern in an extension method called ignore 
soemthing like 
``` C#
static class TaskExtensions
{
	public static async Task Ignore<T>(this Task<T> task) {
		_ = await task;
	}
}
```

and this call can become. The method name shows the intent so no need to add the comment 
``` suggestion
            await _context.NodeOperations.ToReimage(node, done: true).Ignore();
```

chkeita (approved):

src/ApiService/ApiService/Functions/AgentEvents.cs
Porges: Good idea. This will make it easier to find and fix things in future as well.

Porges: At the moment I'm only using this helper in cases where we need to address/fix this in future.

PR #2450, opened 2022-09-27
use InterpolatedStringHandler to move values to CustomDimensions Tags instead of keeping them in the error message

stishkin:


src/ApiService/ApiService/Log.cs
```diff
@@ -1,14 +1,57 @@
...
+    private StringBuilder _builder;
+    private Dictionary<string, string>? _tags;
+
+    public LogStringHandler(int literalLength, int formattedCount) {
+        _builder = new StringBuilder(literalLength);
+        _tags = null;
+    }
+
+    public void AppendLiteral(string message) {
+        _ = _builder.Append(message);
```
Porges, line 24: Instead of having to discard, you can add StringBuilder to the CSharpExtensions config file

Porges (approved):

PR #2460, opened 2022-09-28
fix null ref exception

stishkin:


src/ApiService/ApiService/Log.cs
```diff
@@ -25,7 +25,9 @@ public struct LogStringHandler {
     }
 
     public void AppendFormatted<T>(T message) {
-        _builder.Append(message?.ToString());
+        if (message is not null) {
+            _builder.Append(message.ToString());
```
Porges, line 29: do we want to put a placeholder in like `<null>`?

src/ApiService/ApiService/Log.cs
```diff
@@ -105,14 +109,12 @@ class AppInsights : ILog {
...
 
-
         Dictionary<string, double>? copyMetrics = null;
         if (metrics is not null) {
             copyMetrics = new(metrics);
         }
         _telemetryClient.TrackException(ex, copyTags, copyMetrics);
-
-        Log(correlationId, $"{message} : {ex.Message}", SeverityLevel.Error, tags, caller);
+        Log(correlationId, $"[{message}] {ex.Message}", SeverityLevel.Error, tags, caller);
```
Porges, line 117: the callstack made me wonder if it was `ex` that was `null` here

stishkin: it's string builder inside of message

I reproed

stishkin: (and then I'll be proven wrong ....)

Porges (approved):

PR #2461, opened 2022-09-28
[Draft] Allow multiple analysis outputs

### LINKED ISSUE #2123: Arbitrary artifact creation from crash analysis task
One feature that would be useful for our project is to generate a single file that can be rerun to precisely reproduce a crash found during fuzzing.
While we could accomplish the same thing with the input file and snapshot (both part of the inputs container), combining these into a single file streamlines the workflow our users have to go through in order to reproduce a crash.

In order to do this, the most appropriate method would be to use the crash analyzer task to generate the minimized reproduction file, save it to a OneFuzz container, and include a download link in the JSON template returned by the analyzer.

However, it appears that analysis tasks are only permitted up to one output container at a time, and the crash analysis task outputs single-file JSON reports that OneFuzz interprets (as the output container is designated as an input for notifications).
https://github.com/microsoft/onefuzz/blob/52ccf05a29658ddc0036eb9721cbf128a80b81be/src/api-service/__app__/onefuzzlib/tasks/defs.py#L53-L93

Would bumping the number of output containers from `== 1` to `>= 1` be a good solution? Or maybe something else?

Porges:

Closes #2123.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Defs.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Defs.cs
@@ -231,7 +231,7 @@ public static class Defs {
                 ),
                 new ContainerDefinition(
                     Type:ContainerType.Analysis,
-                    Compare: Compare.Equal,
+                    Compare: Compare.AtLeast,
                     Value:1,
                     Permissions: ContainerPermission.Write | ContainerPermission.Read | ContainerPermission.List
                 ),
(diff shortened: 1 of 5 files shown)


Porges:
Working on a different solution for this.

PR #2464, opened 2022-09-29
Cache VMSS VM InstanceID lookups

Porges:
This is a potentially expensive operation as we need to fetch every VM in the Scaleset to find the VMID we are looking for.

This should be safe to cache "forever" since the mapping will not change.

It seems like we should be storing the `InstanceID` instead of (as well as?) the VMID, but that is hard to change at the moment. Perhaps after C# port is completed.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -193,7 +193,6 @@ public class VmssOperations : IVmssOperations {
     private record InstanceIdKey(Guid Scaleset, Guid VmId);
     private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
         => _cache.GetOrCreateAsync(new InstanceIdKey(scaleset, vmId), async entry => {
-            entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(10);
             var scalesetResource = GetVmssResource(scaleset);
             var vmIdString = vmId.ToString();
             await foreach (var vm in scalesetResource.GetVirtualMachineScaleSetVms().AsAsyncEnumerable()) {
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -180,58 +190,74 @@ public class VmssOperations : IVmssOperations {
         return results;
     }
 
+    private record InstanceIdKey(Guid Scaleset, Guid VmId);
+    private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
+        => _cache.GetOrCreateAsync(new InstanceIdKey(scaleset, vmId), async _ => {
```
stishkin, line 195: should there be expiration on the entries ?

stishkin (approved):

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
tevoinea: It can be configured but I don't think we set it anywhere. It could be nice to have different eviction policies for our different caches.

Also, we should be keeping an eye on memory usage as we're adding caches.

chkeita: we can reuse the TTL value from the [python code](
https://github.com/chkeita/onefuzz/blob/4de2a0cda65de900ee1840e0c6eb219732b52c18/src/api-service/__app__/onefuzzlib/azure/vmss.py#L148) 1min

chkeita: @Porges as discussed in the meeting today i will be taking over this PR and i will add the TTL to unblock prod with hotfix

nharper285 (approved):

mgreisen (approved):

PR #2468, opened 2022-09-29
CLI: Retry on connection reset

Porges:
In our integration test run we are seeing some connection-reset errors which causes the CLI operation to fail.

To fix this:
1. Set TCP-KeepAlive to keep Azure load balancer connections alive longer than the default timeout (4 minutes).
2. Treat ConnectionResetError as retryable.


stishkin (approved):

Porges:
Doing a `check-pr` run first to validate.

Porges:
Didn't cause any problems, at least.

PR #2469, opened 2022-09-29
Allow worker loops to continue after errors

Porges:
During `TimerWorkers` if updating one entity fails and throws an exception we will abandon the whole update. Instead log the error and continue to attempt to process the remaining entities. This will allow us to make progress even if one entity is stuck.

### DIFF
--- a/src/ApiService/ApiService/Functions/TimerWorkers.cs
+++ b/src/ApiService/ApiService/Functions/TimerWorkers.cs
@@ -48,7 +48,7 @@ public class TimerWorkers {
             try {
                 _log.Info($"updating pool: {pool.PoolId:Tag:PoolId} ({pool.Name:Tag:PoolName}) - state: {pool.State:Tag:PoolState}");
                 var newPool = await _poolOps.ProcessStateUpdate(pool);
-                _log.Info($"completed updating pool - now in state {newPool.State:Tag:PoolState}");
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/Functions/TimerWorkers.cs
```diff
@@ -62,13 +65,24 @@ public class TimerWorkers {
...
 
         var scalesets = _scaleSetOps.SearchAll();
         await foreach (var scaleset in scalesets) {
-            await ProcessScalesets(scaleset);
+            try {
+                _log.Info($"updating scaleset: {scaleset.ScalesetId:Tag:ScalesetId} - state: {scaleset.State:Tag:ScalesetState}");
+                var newScaleset = await ProcessScalesets(scaleset);
+                _log.Info($"completed updating scaleset - now in state {newScaleset.State:Tag:ScalesetState}");
+            } catch (Exception ex) {
+                _log.Exception(ex, $"failed to process scaleset");
```
stishkin, line 84: log names ? but I guess it's ok to dismiss those too....

stishkin (approved):

src/ApiService/ApiService/Functions/TimerWorkers.cs
Porges: We can get the sequence of logs by correlation ID, so I don't want to repeat too much? But I can add it in.

PR #2476, opened 2022-10-02
Debug failing check pr

stishkin:
See comments for explanations

### DIFF
--- a/src/ApiService/ApiService/Functions/AgentEvents.cs
+++ b/src/ApiService/ApiService/Functions/AgentEvents.cs
@@ -103,7 +103,6 @@ public class AgentEvents {
             return null;
         }
 
-        _log.Info($"node state update: {machineId:Tag:MachineId} from {node.State:Tag:FromState} to {ev.State:Tag:ToState}");
         node = await _context.NodeOperations.SetState(node, ev.State);
 
         if (ev.State == NodeState.Free) {
--- a/src/ApiService/ApiService/Functions/AgentEvents.cs
+++ b/src/ApiService/ApiService/Functions/AgentEvents.cs
@@ -97,6 +97,8 @@ public class AgentEvents {
             // Don’t check reimage_requested, as nodes only send 'init' state once.  If
             // they send 'init' with reimage_requested, it's because the node was reimaged
             // successfully.
+            // Node has to be in state other than Init, otherwise SetState ignores the change, and therefore
+            // InitializedAt field is never set
             node = node with { ReimageRequested = false, InitializedAt = DateTimeOffset.UtcNow };
             // discard result: node not used after this point
             _ = await _context.NodeOperations.SetState(node, ev.State);
--- a/src/ApiService/ApiService/Functions/AgentEvents.cs
+++ b/src/ApiService/ApiService/Functions/AgentEvents.cs
@@ -97,8 +97,6 @@ public class AgentEvents {
             // Don’t check reimage_requested, as nodes only send 'init' state once.  If
             // they send 'init' with reimage_requested, it's because the node was reimaged
             // successfully.
-            // Node has to be in state other than Init, otherwise SetState ignores the change, and therefore
-            // InitializedAt field is never set
             node = node with { ReimageRequested = false, InitializedAt = DateTimeOffset.UtcNow };
             // discard result: node not used after this point
             _ = await _context.NodeOperations.SetState(node, ev.State);
--- a/src/ApiService/ApiService/OneFuzzTypes/Enums.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Enums.cs
@@ -294,7 +294,6 @@ public enum AgentMode {
 }
 
 public enum NodeState {
-    New,
     Init,
     Free,
     SettingUp,
--- a/src/ApiService/ApiService/OneFuzzTypes/Enums.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Enums.cs
@@ -294,6 +294,7 @@ public enum AgentMode {
 }
 
 public enum NodeState {
+    New,
     Init,
     Free,
     SettingUp,
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -98,7 +98,7 @@ public record Node
     string Version,
     DateTimeOffset? Heartbeat = null,
     DateTimeOffset? InitializedAt = null,
-    NodeState State = NodeState.Init,
+    NodeState State = NodeState.New,
 
     Guid? ScalesetId = null,
     bool ReimageRequested = false,
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -98,7 +98,7 @@ public record Node
     string Version,
     DateTimeOffset? Heartbeat = null,
     DateTimeOffset? InitializedAt = null,
-    NodeState State = NodeState.New,
+    NodeState State = NodeState.Init,
 
     Guid? ScalesetId = null,
     bool ReimageRequested = false,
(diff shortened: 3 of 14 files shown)


src/ApiService/ApiService/OneFuzzTypes/Enums.cs
```diff
@@ -294,6 +294,7 @@ public enum AgentMode {
 }
 
 public enum NodeState {
+    New,
```
stishkin, line 297: Need to add new State for the node (this bug is also in Python code) See "Explanation" for why

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -98,7 +98,7 @@ public record Node
     string Version,
     DateTimeOffset? Heartbeat = null,
     DateTimeOffset? InitializedAt = null,
-    NodeState State = NodeState.Init,
+    NodeState State = NodeState.New,
```
stishkin, line 101: Node has to begin in a state that is not "Init", hence adding new state New

src/ApiService/ApiService/OneFuzzTypes/Enums.cs
stishkin: We convert enums to strings, so order of the Enum members does not matter

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -345,7 +346,17 @@ public NodeOperations(ILogTracer log, IOnefuzzContext context)
             } catch (RequestFailedException ex) when (
                 ex.Status == (int)HttpStatusCode.Conflict ||
                 ex.ErrorCode == "EntityAlreadyExists") {
-                return null;
+
+                var existingNode = await QueryAsync(Query.SingleEntity(poolName.ToString(), machineId.ToString())).FirstOrDefaultAsync();
```
stishkin, line 350: if insert failed, then at get entry from the table and do sanity check

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -497,17 +507,6 @@ public NodeOperations(ILogTracer log, IOnefuzzContext context)
             queryParts.Add($"({q})");
         }
 
-        if (excludeUpdateScheduled) {
```
stishkin, line 500: this was first bug - SearchStatesQuery would ignore current version node. We do not want that. (And Python code does not ignore it ether, so this fix is just aligning implementation with Python code)

src/ApiService/ApiService/Functions/AgentEvents.cs
```diff
@@ -97,13 +97,14 @@ public class AgentEvents {
             // Don’t check reimage_requested, as nodes only send 'init' state once.  If
             // they send 'init' with reimage_requested, it's because the node was reimaged
             // successfully.
+            // Node has to be in state other than Init, otherwise SetState ignore the change, and therefore
```
stishkin, line 100: Explanation on why we need a new state added to nodes. And this is bug number two, that would force nodes to go from Free to Done state right away, since InitializedAt is not set, and so the node is considered to be outdated right after it is created. Since we perform the check on timer and LInux is faster to provision than Windows - Linux nodes succeed more often than Windows

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -470,12 +482,10 @@ public NodeOperations(ILogTracer log, IOnefuzzContext context)
     }
 
     public static string SearchStatesQuery(
```
stishkin, line 484: this was first bug - SearchStatesQuery would ignore current version node. We do not want that. (And Python code does not ignore it ether, so this fix is just aligning implementation with Python code)

src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
```diff
@@ -90,15 +90,15 @@ public ReproOperations(ILogTracer log, IOnefuzzContext context)
         var vm = await GetVm(repro, config);
         var vmOperations = _context.VmOperations;
         if (!await vmOperations.IsDeleted(vm)) {
-            _logTracer.Info($"vm stopping: {repro.VmId}");
+            _logTracer.Info($"vm stopping: {repro.VmId:Tag:VmId} {vm.Name:Tag:VmName}");
             var rr = await vmOperations.Delete(vm);
-            if (!rr) {
-                _logTracer.Error($"failed to delete repro {repro.VmId:Tag:VmId}");
+            if (rr) {
+                _logTracer.Info($"repro vm fully deleted {repro.VmId:Tag:VmId} {vm.Name:Tag:VmName}");
```
tevoinea, line 96: Out of scope for this PR, but what do you think of enforcing some sort of pre+result logging in each of the `*Operations` functions?

That way we don't need to remember to wrap the call site in logs for starting/succeeding/failing an operation and we get the benefit of traceability everywhere else for free.

By that I mean...

**VmOperations.cs**:
```csharp
Delete(Vm vm) {
    _logTracer.Info($"vm stopping: {repro.VmId:Tag:VmId} {vm.Name:Tag:VmName}");
    ....
    // log if failed or succeded
    return result;
}
```

**ReproOperations.cs**
```csharp
Stopping(Repro repro) {
    // no need to log starting operation here
    await vmOperations.Delete(vm).IgnoreResult();
    // no need to log succeeded/failed, only use the result
}
```

tevoinea: I'll start a discussion in the repo to get feedback from the team.

src/ApiService/IntegrationTests/AgentEventsTests.cs
```diff
@@ -279,7 +279,7 @@ public AgentEventsTestsBase(ITestOutputHelper output, IStorage storage)
             Async.Task.Run(async () => {
                 // should still be in init state:
```
tevoinea, line 280: NIT:
```suggestion
                // should still be in new state:
```

tevoinea (approved):

src/ApiService/ApiService/Functions/AgentEvents.cs
nharper285: This makes a lot of sense.

src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
stishkin: good point

src/ApiService/ApiService/OneFuzzTypes/Enums.cs
chkeita, line 297: I don't think we need the new state. 
The implementation of NodeOperations.SetState does not match the python code. 
- [in python](https://github.com/chkeita/onefuzz/blob/3347e7e67b3c51dddf47eafc9320cfbc7bd609ae/src/api-service/__app__/onefuzzlib/workers/nodes.py#L486) the state is saved even if the state is the same (we don't send the event in that case).
- [in C#](https://github.com/chkeita/onefuzz/blob/4662df3e397c6bb1406647515a47b4fa868c4f26/src/ApiService/ApiService/onefuzzlib/NodeOperations.cs#L450) we don't save the node if the state is the same.

stishkin: yeah you are right.

stishkin: New state could be good - since it makes SetState implementation consistent across all components.

chkeita (approved):

src/ApiService/ApiService/OneFuzzTypes/Enums.cs
chkeita: yes adding  new state might be beneficial. But i think we should add it after the migration is done. 
Keeping things the same  simplify the comparison with the behavior of the python code.

PR #2481, opened 2022-10-03
Persist instance ids

stishkin:


src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -195,26 +198,40 @@ private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
...
+                await foreach (var vm in scalesetResource.GetVirtualMachineScaleSetVms().AsAsyncEnumerable()) {
+                    var response = await vm.GetAsync();
+                    var instanceId = response.Value.Data.InstanceId;
+
+                    _ = await _nodeOps.SetInstanceId(scaleset, vmId, instanceId);
+
+                    if (response.Value.Data.VmId == vmIdString) {
+                        // we found the VM we are looking for
+                        entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(10);
+                        _ = entry.SetValue(instanceId);
```
stishkin, line 218: I added "entry.SetValue" - since we should cache instance id of the found node. Am I wrong ?

chkeita: The return value is used to set the value. of the entry when we call GetOrCreateAsync so no need to use SetValue here. 
This can actually be dangerous if the implementation does not allow reentrancy on the same entry

chkeita:
Saving the instance id makes the caching obsolete. We should only have one of them. I vote for keeping the caching for now and add the intance_id once we have validated the migration is complete.

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
chkeita: Note that we use setValue below for other entries in the cache not the one that is currently open.

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -195,26 +198,40 @@ private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
         => _cache.GetOrCreateAsync(new InstanceIdKey(scaleset, vmId), async entry => {
```
chkeita, line 198: Saving the instance id makes the caching obsolete. We should only have one of them. I vote for keeping the caching for now and add the intance_id once we have validated the migration is complete.

stishkin: We just have 2 layers of caching.

Having "local" caching is good - makes things faster, and less requests to Azure storage required.

If my change is very risky - makes sense to revert. If change is low risk - then it has benefit

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
stishkin: reverted

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
chkeita: I think if we don't have an immediate needs we should not add it now. However,  if we have to have 2 layers of caching let's not intermingle them in the same function it makes the logic complex.

This function should only be used for caching the azure request. I suggest saving the instance  to the database in the `GetInstanceId` method instead

stishkin: If we move it to GetInstanceId - then it defeats the point of storing it in the table at all. We want to store instance ids when we iterate through scaleset so we avoid Azure requests later on.

Moving to GetInstanceId will only cache one entry of the VM that was found.

PR #2486, opened 2022-10-04
Migrating notification templates

tevoinea:

_What is this about?_

Adds support for migrating our jinja templates to scriban on the fly.

## PR Checklist
* [ ] Applies to work item: #1754 

## Info on Pull Request

_What does this include?_

* Detect whether a template is a jinja template
* Convert our jinja templates to scriban template
* Tests
* Small docs improvement for `NodeDisposalStrategy`

## Validation Steps Performed

Test ADO/github/teams notification creation with an old jinja template (so long as the template is not very complicated).

The tests that are added in this PR are testing the jinja template features we use in prod (`if`, `for`, and a combination of the 2).


stishkin (approved):

src/ApiService/ApiService/onefuzzlib/notifications/JinjaTemplateAdapter.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.OneFuzz.Service;
+
+public class JinjaTemplateAdapter {
```
chkeita, line 3: Will this convert every jinja template to scriban or is it a best effort implementation ?

tevoinea: In the scope of all jinja templates, it's definitely best effort. In the scope of the jinja templates we use, it will convert all of them.

src/ApiService/ApiService/OneFuzzTypes/Enums.cs
```diff
@@ -331,8 +331,14 @@ public static class NodeStateHelper {
 }
 
 
+/// Select how nodes should be disposed of after they complete a WorkSet
 public enum NodeDisposalStrategy {
+    /// Re-images the node (which resets its state), then either it can pick up more work
```
Porges, line 336: nice, we should document like this on other state enums as well

Porges (approved):

PR #2506, opened 2022-10-11
Only fetch VM InstanceView data when required

Porges:
Fetching the `InstanceView` data for a VM is considered a “high-cost” operation by VMSS and contributes to throttling. In most cases we do not need this data, so split the `GetVm` method into two versions, one that fetches the `InstanceView` data and one that doesn’t.

We only need `InstanceView` data during VM provisioning failures, to see what went wrong.


chkeita (approved):

Porges:
Validating with `check-pr`.

Porges:
Had some problems with the run but I don't believe they are due to this change.

PR #2515, opened 2022-10-13
Optimize the output side of `srcview`

DrChat:

This PR contains performance improvements for the `srcview` crate.

## PR Checklist
* [x] Tests added/passed

## Info on Pull Request

1) Comparing `PathBuf` is more expensive than comparing raw strings, so we should strive to use strings while building the `PdbCache`.
2) It's more expensive to allocate temporary backing memory for the XML cobertura file than it is to simply write directly to a handle provided by the caller.
3) Swap to `quick-xml` and try to avoid allocating any memory in the hot path when outputting cobertura XML output.

## Validation Steps Performed

`srcview` binary with a large PDB (ntoskrnl) and modoff file.


src/agent/srcview/src/pdbcache.rs
```diff
@@ -49,7 +52,11 @@ impl PdbCache {
 
             while let Some(symbol) = symbols.next()? {
                 if let Ok(SymbolData::Procedure(proc)) = symbol.parse() {
+                    let proc_name = proc.name.to_string();
                     let mut lines = program.lines_at_offset(proc.offset);
+
+                    let symbol_to_lines = symbol_to_lines.entry(proc_name.to_string()).or_default();
```
Porges, line 58: Stylistic but `clone` instead of `to_string` to be explicit about what's happening, since it's already a `String`? And another instance of this below…

src/agent/srcview/src/report.rs
```diff
@@ -366,30 +377,35 @@ impl Report {
...
+                ("branch-rate", "0"),
+                ("timestamp", format!("{}", unixtime).as_str()),
+                ("complexity", "0"),
+                ("version", "0.1"),
+            ]),
+        ))?;
+
+        ew.create_element("sources").write_inner_content(|ew| {
+            ew.create_element("source")
+                .write_text_content(BytesText::new(""))?;
```
Porges, line 403: Is it required to write an empty text node?

Porges (approved):

src/agent/srcview/src/report.rs
DrChat: I'm not sure if this is required vs. just writing an `Empty` node, though this should be the same as what the previous code was doing.

src/agent/srcview/src/pdbcache.rs
Porges: @DrChat that's fine then, it won't be confusing in the IDE with type information

src/agent/srcview/src/report.rs
Porges: @DrChat yes, and I don't care enough to find out 🙂

DrChat:
Of particular note: When writing a file, callers will _need_ to wrap the file with a `BufWriter` before passing it into `Report::cobertura` or their performance will drastically suffer.
While I don't think it should be the job of `srcview` to try and buffer any writes, it might be appropriate to update the example with one that does this.

DrChat:
Haha, I mean we might as well as just leave the `&mut impl Write` function and nudge callers to use a `BufWriter` if they're writing a file.
Maybe if it becomes a problem, we could do something like just wrap it internally?

Porges:
> @DrChat Haha, I mean we might as well as just leave the `&mut impl Write` function and nudge callers to use a `BufWriter` if they're writing a file. Maybe if it becomes a problem, we could do something like just wrap it internally?

For now I'll just update our example executable, and the NOTE should hopefully point the way.

tevoinea (approved):

DrChat:
Also fyi - I'll probably follow up with another few PRs with further refactoring.
There's definitely a lot that can be done to reduce memory use and improve performance.

ranweiler:
In general, for incremental perf optimization PRs, we should include some before/after benchmarks of the relevant metrics (time, memory). This is helpful for understanding what one gets in exchange for using (sometimes) less obvious / more complex implementation strategies. Will also help diagnose regressions and guide future optimization attempts.

ranweiler:
@drchat, I don't think we need to get too strict with it or anything, especially with more obvious big wins. If you're using WPA, even just noting the changes you're seeing at a high level would be useful. That'd help share effect magnitude, and ideally give others a sense of how you were measuring it / how to repeat on their own.

DrChat:
Sounds good!
Before (symbols are a bit mangled, but you can see `SrcView` and `Report` functions in the 5 entries past the highlighted frame):
wpa_4SgF6m3fcF

After:
wpa_WgKwYIiLiG

Overall, from 13.637s to 8.937s (roughly 1.5x speedup). For our use-case, much higher because we call the output side more than `SrcView::insert`.

PR #2526, opened 2022-10-17
[Draft] Chase `Duplicate Of` links when filing ADO bugs

Porges:
Chase `Duplicate Of` links when updating bugs, so we always update the main bug in question.

⚠️ This code is hypothetical and has not been tested or even run.


stishkin (approved):

Porges:
@tevoinea pointed out that this link name is not guaranteed across different ADO instances. We could probably push this into the config so it can be overridden if needed.

PR #2529, opened 2022-10-18
Produce an error if coverage recording failed due to timeout

### LINKED ISSUE #2520: Race in coverage recording?
I noticed while trying to repro #1781 locally that sometimes the recording would not work properly:

Commandline:

```console
$ RUST_LOG=trace cargo run --example coverage -- /workspaces/onefuzz/src/integration-tests/libfuzzer/fuzz.exe @@ --inputs  /workspaces/onefuzz/src/integration-tests/libfuzzer/input.txt
```

Coverage as expected:
```
[2022-10-16T22:16:40Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/librt-2.27.so
[2022-10-16T22:16:40Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/libm-2.27.so
[2022-10-16T22:16:40Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/libdl-2.27.so
[2022-10-16T22:16:40Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/libgcc_s.so.1
[2022-10-16T22:16:40Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/libc-2.27.so
[2022-10-16T22:16:41Z INFO  coverage::block::linux] new thread: 28928
[2022-10-16T22:16:41Z INFO  coverage] recorded in 2.480324274s
[2022-10-16T22:16:41Z INFO  coverage] input = /workspaces/onefuzz/src/integration-tests/libfuzzer/input.txt
[2022-10-16T22:16:41Z INFO  coverage] ld-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libc-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libdl-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libgcc_s.so.1 = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libm-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libpthread-2.27.so = 185 / 3192 (5.80%)
[2022-10-16T22:16:41Z INFO  coverage] librt-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libstdc++.so.6.0.29 = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] fuzz.exe = 5 / 22 (22.73%)
[2022-10-16T22:16:41Z INFO  coverage] ld-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libc-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libdl-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libgcc_s.so.1 = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libm-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libpthread-2.27.so = 185 / 3192 (5.80%)
[2022-10-16T22:16:41Z INFO  coverage] librt-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] libstdc++.so.6.0.29 = 0 / 0 (NaN%)
[2022-10-16T22:16:41Z INFO  coverage] fuzz.exe = 5 / 22 (22.73%)
```

Coverage not as expected:
```
[2022-10-16T22:19:43Z INFO  coverage::block::linux] module load: /lib/x86_64-linux-gnu/ld-2.27.so
[2022-10-16T22:19:43Z INFO  coverage] recorded in 11.445283278s
[2022-10-16T22:19:43Z INFO  coverage] input = /workspaces/onefuzz/src/integration-tests/libfuzzer/input.txt
[2022-10-16T22:19:43Z INFO  coverage] ld-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:19:43Z INFO  coverage] fuzz.exe = 0 / 22 (0.00%)
[2022-10-16T22:19:43Z INFO  coverage] ld-2.27.so = 0 / 0 (NaN%)
[2022-10-16T22:19:43Z INFO  coverage] fuzz.exe = 0 / 22 (0.00%)
```

In the latter case it only receives "module load" events for `fuzz.exe` & `ld`. It seems that `pete`/`Ptracer` is missing some events?

Traces:
- [ok.txt](https://github.com/microsoft/onefuzz/files/9795874/ok.txt)
- [bad.txt](https://github.com/microsoft/onefuzz/files/9795873/bad.txt)

AB#41811543

Porges:
Closes #2520.

After investigation the underlying cause here is that the process is getting killed due to timeout, but we don't report the timeout or produce an error. Modify the coverage code so that it fails if timeout is hit.

## Open questions

Do we want to be able to support a partial coverage instead?

### DIFF
--- a/src/agent/coverage/src/block/linux.rs
+++ b/src/agent/coverage/src/block/linux.rs
@@ -5,7 +5,8 @@ use std::collections::BTreeMap;
 use std::convert::TryInto;
 use std::ffi::OsStr;
 use std::process::Command;
-use std::sync::mpsc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{mpsc, Arc};
 use std::thread;
 use std::time::{Duration, Instant};
 
@@ -40,7 +41,14 @@ impl<'c> Recorder<'c> {
         let mut tracer = Ptracer::new();
         let mut child = tracer.spawn(cmd)?;
 
-        let _timer = Timer::new(timeout, move || child.kill());
+        let timed_out = Arc::new(AtomicBool::new(false));
+
+        let timer_flag = timed_out.clone();
+        let timer = Timer::new(timeout, move || {
+            if child.kill().is_ok() {
+                timer_flag.as_ref().store(true, Ordering::Relaxed);
+            }
+        });
 
         let recorder = Recorder {
             breakpoints: Breakpoints::default(),
@@ -53,8 +61,16 @@ impl<'c> Recorder<'c> {
         };
 
         let coverage = recorder.wait()?;
+        timer.cancel();
 
-        Ok(coverage)
+        if timed_out.load(Ordering::Relaxed) {
+            Err(anyhow::format_err!(
+                "timed out creating recording after {}s",
+                timeout.as_secs_f64()
+            ))
+        } else {
+            Ok(coverage)
+        }
     }
 
     fn wait(mut self) -> Result<CommandBlockCov> {
--- a/src/agent/coverage/src/block/linux.rs
+++ b/src/agent/coverage/src/block/linux.rs
@@ -41,14 +41,7 @@ impl<'c> Recorder<'c> {
         let mut tracer = Ptracer::new();
         let mut child = tracer.spawn(cmd)?;
 
-        let timed_out = Arc::new(AtomicBool::new(false));
-
-        let timer_flag = timed_out.clone();
-        let timer = Timer::new(timeout, move || {
-            if child.kill().is_ok() {
-                timer_flag.store(true, Ordering::Relaxed);
-            }
-        });
+        let timer = Timer::new(timeout, move || child.kill());
 
         let recorder = Recorder {
             breakpoints: Breakpoints::default(),
@@ -61,9 +54,8 @@ impl<'c> Recorder<'c> {
         };
 
         let coverage = recorder.wait()?;
-        timer.cancel();
 
-        if timed_out.load(Ordering::Relaxed) {
+        if timer.timed_out() {
             Err(anyhow::format_err!(
                 "timed out creating recording after {}s",
                 timeout.as_secs_f64()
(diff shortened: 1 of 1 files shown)


src/agent/coverage/src/block/linux.rs
```diff
@@ -40,7 +41,14 @@ impl<'c> Recorder<'c> {
         let mut tracer = Ptracer::new();
         let mut child = tracer.spawn(cmd)?;
 
-        let _timer = Timer::new(timeout, move || child.kill());
+        let timed_out = Arc::new(AtomicBool::new(false));
+
+        let timer_flag = timed_out.clone();
```
ranweiler, line 46: Maybe this could be internal to the `Timer` struct, and set right before we call [`on_timeout()`](https://github.com/microsoft/onefuzz/blob/b3fd6d5ed0fcc851478a02f169e0f5ae70f114a6/src/agent/coverage/src/block/linux.rs#L458-L459)?

Then we could add a `timed_out(&self) -> bool` method to `Timer` to further hide the atomics.

ranweiler (approved):

stishkin (approved):

src/agent/coverage/src/block/linux.rs
Porges: At the moment I have it as conditional on whether or not the kill actually "worked" so that we don't report an error if we timeout but the process had already stopped, so the `timeout` function will need to return some kind of result to move that logic into `Timer`.

ranweiler: Alternately: if the process _would_ have exited on its own, there's always still a race with our call to `kill()`. Checking the `kill()` result for an `InvalidInput` case doesn't squash the race.

So, I think we could fairly just say that if we made it to the `on_timeout()` callback at all, the timeout state was semantically reached (even if we could have in principle observed the exit).

PR #2536, opened 2022-10-20
Ignoring the scanning log file when reporting an issue with azcopy

chkeita:

The current code assumes that only one log file is produced by azcopy. But since [10.9.0](https://github.com/Azure/azure-storage-azcopy/blob/main/ChangeLog.md#version-1090) a scanning log file has been introduced.
This PR ignores the scanning log file.


mgreisen (approved):

chkeita:
related to #2160

PR #2539, opened 2022-10-20
Reduce number of ARM calls in ListVmss

Porges:
1. `ListVmss` was invoking `GetVirtualMachineScaleSetAsync` and then `GetVirtualMachineScaleSetVms`; instead we can get the VMSS resource directly and invoke `GetVirtualMachineScaleSetVms` on it.

2. I also removed some unneeded invocations of `AsAsyncEnumerable` which might be turning async enumerables into sync enumerables.

3. Do the same `HasData` check in the instance ID lookup that we were already doing in `ListVmss`.

After this change (#3) we no longer do any `GET`s against individual VMSS VMs during the `check-pr` run: 
image

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -41,7 +41,7 @@ public interface IVmssOperations {
         string sshPublicKey,
         IDictionary<string, string> tags);
 
-    Async.Task<List<string>?> ListVmss(Guid name, Func<VirtualMachineScaleSetVmResource, bool>? filter);
+    IAsyncEnumerable<string> ListVmss(Guid name, Func<VirtualMachineScaleSetVmResource, bool>? filter);
     Async.Task<OneFuzzResultVoid> ReimageNodes(Guid scalesetId, IReadOnlySet<Guid> machineIds);
     Async.Task DeleteNodes(Guid scalesetId, IReadOnlySet<Guid> machineIds);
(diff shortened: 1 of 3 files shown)


tevoinea (approved):

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -411,19 +410,21 @@ private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
         }
     }
 
-    public async Task<List<string>?> ListVmss(Guid name, Func<VirtualMachineScaleSetVmResource, bool>? filter) {
+    public async Async.Task<List<string>?> ListVmss(Guid name, Func<VirtualMachineScaleSetVmResource, bool>? filter) {
         try {
-            var vmss = await _creds.GetResourceGroupResource().GetVirtualMachineScaleSetAsync(name.ToString());
-            return vmss.Value.GetVirtualMachineScaleSetVms()
+            // have to assign to a typed variable so that the Where call can be disambiguated
+            IAsyncEnumerable<VirtualMachineScaleSetVmResource> vms =
```
chkeita, line 416: how about we return the asyncenumerable list and let the caller do the filtering. 
We won't need to consume the full list here.

Porges: done! There's only one caller so it was an easy change.

Issue #2547, opened 2022-10-21
OneFuzz ScaleSet VM management gets out of sync (??) 


stishkin:
We are hitting following issue. Looks like OneFuzz request info to azure is failing 

```
The provided instanceId 2143 is not an active Virtual Machine Scale Set VM instanceId. Status: 400 (Bad Request) ErrorCode: InvalidParameter

Content: { "error": { "code": "InvalidParameter", "message": "The provided instanceId 2143 is not an active Virtual Machine Scale Set VM instanceId.", "target": "instanceIds" } }
```

Call Stack:
```
Azure.RequestFailedException:
   at Azure.ResourceManager.Compute.VirtualMachineScaleSetVMsRestOperations+<UpdateAsync>d__15.MoveNext (Azure.ResourceManager.Compute, Version=1.0.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Azure.ResourceManager.Compute.VirtualMachineScaleSetVmCollection+<CreateOrUpdateAsync>d__5.MoveNext (Azure.ResourceManager.Compute, Version=1.0.0.0, Culture=neutral, PublicKeyToken=92742159e12e44c8)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
[...]

PR #2549, opened 2022-10-21
Downgrade instance not found exception

tevoinea:

_What is this about?_

#2547 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -262,7 +262,7 @@ public class VmssOperations : IVmssOperations {
             return OneFuzzResultVoid.Ok;
         } catch (RequestFailedException ex) when (ex.Status == 409 && ex.Message.StartsWith("The request failed due to conflict with a concurrent request")) {
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"protection policy update is already in progress: {vmId}:{instanceId} in vmss {scaleset.ScalesetId}");
-        } catch (RequestFailedException ex) when(ex.Status == 400 && ex.Message.Contains("The provided instanceId") && ex.Message.Contains("not an active Virtual Machine Scale Set VM instanceId.")) {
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -262,6 +262,8 @@ private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
             return OneFuzzResultVoid.Ok;
         } catch (RequestFailedException ex) when (ex.Status == 409 && ex.Message.StartsWith("The request failed due to conflict with a concurrent request")) {
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"protection policy update is already in progress: {vmId}:{instanceId} in vmss {scaleset.ScalesetId}");
+        } catch (RequestFailedException ex) when (ex.Status == 400 && ex.Message.Contains("The provided instanceId") && ex.Message.Contains("not an active Virtual Machine Scale Set VM instanceId.")) {
```
chkeita, line 265: This will still get logged as an error right ? i think you have to update the call to this in `AcquireScaleInProtection` and `ReleaseScaleInProtection`

tevoinea: It's going to be logged as a trace with severity level: Error aka: it won't show up in the Failures tab of app insights.

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
tevoinea: Resolved offline. It makes sense that we don't want to log it as an exception because it's an expected error but, it's still an error that shouldn't be happening even though we're ok with it happening given our current system architecture. Removing the exception logging but keeping the trace severity level as error seems to be an OK balance. Otherwise, we can add a specific case for this exception to only log as info.

Issue #2553, opened 2022-10-24
Add exponential back off for notification creation


tevoinea:
Today, when we fail to create a notification, we retry 5 times before the message ends up in the poison queue. We found that many times we can requeue the poison queue messages at a later date (a few days later) and the notifications have no issue being created.

We should automatically re-attempt notifications (like we already do) but now each time waiting longer before re-attempting to create the notification.

AB#41923566

PR #2555, opened 2022-10-24
Exponential backoff for failed notifications

tevoinea:

_What is this about?_

#2553 

Today, our file-changes-poison queue does not get consumed and we manually re-queue those messages. This PR automatically re-processes those messages using an exponential backoff. The backoff times are:
```
// 1 - 5 mins
// 2 - 25 mins
// 3 - 2 hours 5 mins
// 4 - 10 hours 25 mins
// 5 - 2 days 4 hours 5 mins
// >6 - 7 days +/- 1 day of random variance so all the messages don't dequeue at the same time
```

Once a message hits the 1 week backoff point (or 6 attempts), we log a warning (though we should consider adding a webhook as well).

## Info on Pull Request

_What does this include?_

* Consuming the `file-changes-poison` queue
* Exponential backoff retry for poison queue messages
* Logging when poison queue messages are _really_ failing
* Rename `failTaskOnTransientError` to `isLastRetryAttempt` to more accurately reflect the behavior of that flag (it was changed in a previous PR)

## Validation Steps Performed

* Enqueue a poison queue message
* Run the C# code such that it will fail when processing the message (aka: always throw an exception)
* Validate the message was consumed from the queue
* Validate the message was put back on the queue but hidden (azure portal shows 0 of 1 messages)
* Re-run the C# code but don't throw any exceptions
* Validate the message was consumed from the queue after the visibility period and make sure it was created successfully and not put back in the poison queue


src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -47,7 +53,39 @@ public class QueueFileChanges {
...
+        if (!fileChangeEvent.RootElement.TryGetProperty(eventType, out var eventTypeElement)
+            || eventTypeElement.GetString() != "Microsoft.Storage.BlobCreated") {
+            return;
+        }
+
+        const string topic = "topic";
+        if (!fileChangeEvent.RootElement.TryGetProperty(topic, out var topicElement)
+            || !_storage.CorpusAccounts().Contains(new ResourceIdentifier(topicElement.GetString()!))) {
+            return;
+        }
```
chkeita, line 74: let's put all this processing inside the try catch. A malformed message could cause an exception here and we will endup with a `QueueFileChangesPoisonPoison` queue

chkeita: Better yet, Can we  invoke `QueueFileChanges.Run` directly so we don't repeat the logic.

src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -47,7 +53,39 @@ public class QueueFileChanges {
...
+
+        try {
+            // Setting isLastRetryAttempt to false will rethrow any exceptions
+            // With the intention that the azure functions runtime will handle requeing
+            // the message for us. The difference is for the poison queue, we're handling the
+            // requeuing ourselves because azure functions doesn't support retry policies
+            // for queue based functions.
+            await FileAdded(_log, fileChangeEvent, false);
+        } catch {
+            await RequeueMessage(msg);
```
chkeita, line 84: I think we can add this logic to the existing `QueueFileChanges` function and keep the poison queue for message that need manual intervention. we will avoid splitting the processing code in two places.

src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -47,7 +53,39 @@ public class QueueFileChanges {
...
+            return;
+        }
+
+        try {
+            // Setting isLastRetryAttempt to false will rethrow any exceptions
+            // With the intention that the azure functions runtime will handle requeing
+            // the message for us. The difference is for the poison queue, we're handling the
+            // requeuing ourselves because azure functions doesn't support retry policies
+            // for queue based functions.
+            await FileAdded(_log, fileChangeEvent, false);
```
Porges, line 82: Stylistic suggestion: I prefer naming bare arguments like this so it is clear what they are doing:
```suggestion
            await FileAdded(_log, fileChangeEvent, isLastRetryAttempt: false);
```

src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -56,6 +94,51 @@ public class QueueFileChanges {
...
+    // >6 - 7 days +/- 1 day of random variance so all the messages don't dequeue at the same time
+    private static TimeSpan CalculateExponentialBackoff(int dequeueCount) {
+        var numMinutes = Math.Pow(5, dequeueCount);
+        var backoff = TimeSpan.FromMinutes(numMinutes);
+        var maxBackoff = MaxTtl();
+        return backoff < maxBackoff ? backoff : maxBackoff;
+    }
+
+    private static TimeSpan MaxTtl() {
+        var variance = new Random().Next(0, 48);
```
Porges, line 140: Use the 'shared' random number generator instead of instantiating a new one:

```suggestion
        var variance = Random.Shared.Next(0, 48);
```

Porges (approved):

PR #2556, opened 2022-10-24
Extend default coverage timeout to 120s

Porges:
As of #2529 we now fail if the coverage recording times out. The default timeout (5s) does not appear to be long enough for even simple tasks to run, so extend it to 120s.

### DIFF
--- a/src/agent/coverage/examples/src-cov.rs
+++ b/src/agent/coverage/examples/src-cov.rs
@@ -22,7 +22,7 @@ struct Opt {
     #[structopt(min_values = 2)]
     cmd: Vec<String>,
 
-    #[structopt(short, long, long_help = "Timeout in ms", default_value = "5000")]
+    #[structopt(short, long, long_help = "Timeout in ms", default_value = "120000")]
     timeout: u64,
 
     #[structopt(short = "x", long)]
--- a/src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
+++ b/src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
@@ -30,7 +30,7 @@ use crate::tasks::{
 };
 
 const MAX_COVERAGE_RECORDING_ATTEMPTS: usize = 2;
-const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(600);
+const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(120);
 
 #[derive(Debug, Deserialize)]
 pub struct Config {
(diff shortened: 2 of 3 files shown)


AdamL-Microsoft (approved):

src/agent/coverage/examples/src-cov.rs
```diff
@@ -22,7 +22,7 @@ struct Opt {
     #[structopt(min_values = 2)]
     cmd: Vec<String>,
 
-    #[structopt(short, long, long_help = "Timeout in ms", default_value = "5000")]
+    #[structopt(short, long, long_help = "Timeout in ms", default_value = "120000")]
```
ranweiler, line 25: Relevant task executor constant can be found [here](https://github.com/microsoft/onefuzz/blob/cafba18c8500bce241a7480880ae6f2e9717a31a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs#L39). May want to check for any places where it gets set server-side or in the CLI.

ranweiler: (the attr here never used within OneFuzz, it's just used by the example binary for debugging)

Porges: doh!

src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
```diff
@@ -30,7 +30,7 @@ use crate::tasks::{
 };
 
 const MAX_COVERAGE_RECORDING_ATTEMPTS: usize = 2;
-const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(120);
```
ranweiler, line 33: FWIW, this is fine, but doesn't ever apply to the code path that is failing tasks. .NET coverage tasks wholly rely on the external `dotnet-coverage` tool.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -36,7 +36,7 @@ const COVERAGE_FILE: &str = "coverage.json";
 const SOURCE_COVERAGE_FILE: &str = "source-coverage.json";
 const MODULE_CACHE_FILE: &str = "module-cache.json";
 
-const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_TARGET_TIMEOUT: Duration = Duration::from_secs(120);
```
ranweiler, line 39: If a user sets their own `target_timeout` that is less than the time required for the initial coverage recording (and associated cold module / symbol cache misses), then this generous window will not help.

Consider, at this constant's [use site](https://github.com/microsoft/onefuzz/blob/cafba18c8500bce241a7480880ae6f2e9717a31a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs#L59-L63), either adding a `TIMEOUT_BUFFER` value, setting a minimum, or applying a multiplier (perhaps with a max?).

Porges: I'll won't address it here but I'll open an issue to track.

Issue #2550, opened 2022-10-21
Post VM instance ID from the Agent as part of the "AgentCanSchedule" request


stishkin:
Implementing this work item will fix Azure throttling issues when OneFuzz iterates over scalesets in order to find matching Instance Id for a VM, based on a VM ID. Also implementing this removes need for caching Instance IDs in memory cache altogether.

When OneFuzz agent starts up it can get it's own Instance ID by performing following calculation:

```
Each VM in a scale set also gets a computer name assigned to it. This computer name is the hostname of the VM in the [Azure-provided DNS name resolution within the virtual network](https://learn.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances). This computer name is of the form "{computer-name-prefix}{base-36-instance-id}".

The {base-36-instance-id} is in [base 36](https://en.wikipedia.org/wiki/Base36) and is always six digits in length. If the base 36 representation of the number takes fewer than six digits, the {base-36-instance-id} is padded with zeros to make it six digits in length. For example, an instance with {computer-name-prefix} "nsgvmss" and instance ID 85 will have computer name "nsgvmss00002D".
```
source: https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids

When calling AgentCanSchedule, agent includes instnace ID as part of the message - OneFuzz service saves Instance ID as part of the Node record in the Node Table.

Upgrade path:
- When instance ID is needed to apply scale in protection (or to remove it) - first check the Node table for the instance id. If not there then it is older version of the node, and use current logic to iterate over the scaleset to find the instance id.
- In about 8 days - it is safe to remove "current logic" i.e. caching and scaleset iteration to find the instance id, since all nodes will be upgraded by that time, and will be reporting instance IDs that will be persisted in Node table

AB#41893228

Porges:
Sent a PR which implements this but including it as part of the AgentRegistration message instead, and the decoding of the computer name to instance ID happens in the server since it's a VMSS-specific thing, so I feel like the knowledge should be there instead of in the agent itself.

PR #2557, opened 2022-10-24
Include computer name in AgentRegistration & decode Instance ID from it

Porges:
Addressing #2550 but in a slightly different way:

1. Always include the machine name in the agent registration message. From the machine name we can extract the instance ID.
2. Use the stored value in preference to fetching it every time.
3. To back-fill existing nodes that do not have the value stored, update it in AcquireScaleInProtection if it is not already present.

### DIFF
--- a/src/agent/onefuzz/src/machine_id.rs
+++ b/src/agent/onefuzz/src/machine_id.rs
@@ -43,6 +43,14 @@ pub async fn get_ims_id() -> Result<Uuid> {
     Ok(value)
 }
 
+// Here we use the "computer name" terminology from the VMSS documentation: https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name
+// this is the hostname of the compute node.
+pub fn get_computer_name() -> Result<String> {
+    let hostname = hostname::get().context("getting hostname")?;
+    // we expect that Azure VM names should always be ASCII-only
+    Ok(hostname.to_string_lossy().into_owned())
+}
+
 pub async fn get_machine_name() -> Result<String> {
     let path = onefuzz_etc()?.join("machine_name");
     let body = match fs::read_to_string(&path).await {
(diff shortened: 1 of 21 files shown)


mgreisen:
@Porges I just remembered the work that @chkeita is doing with unmanaged nodes. Unmanaged nodes are likely to have some other naming scheme. I'm not sure how we will distinguish when registering.

Porges:
> @Porges I just remembered the work that @chkeita is doing with unmanaged nodes. Unmanaged nodes are likely to have some other naming scheme. I'm not sure how we will distinguish when registering.

Yes, we will have to have different handling for them. If they send `null` for this then we would fall back to the existing code path. We are going to have to have different handling for unmanaged nodes in any case, so I don’t think this change will adversely affect that work.

src/agent/onefuzz/src/machine_id.rs
```diff
@@ -43,6 +43,14 @@ pub async fn get_ims_id() -> Result<Uuid> {
     Ok(value)
 }
 
+// Here we use the "computer name" terminology from the VMSS documentation: https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name
+// this is the hostname of the compute node.
+pub fn get_computer_name() -> Result<String> {
```
chkeita, line 48: just curious. Is computer_name different from machine_name below ?

chkeita (approved):

src/agent/onefuzz/src/machine_id.rs
Porges: Yes, `machine_name` is `{scale-set-name}_{instance-id}`. But now that you point it out, maybe we should be using this instead of `computer_name` since we will not need to decode it!

Porges: I’m going to do this instead of the Base-36 stuff.

tevoinea (approved):

chkeita (approved):

PR #2561, opened 2022-10-25
Use the vitual env folder to store the config if it exists

chkeita:

Save the onefuzz config in the virtual environment folder. This simplifies the scenarios of connecting to multiple environments.


tevoinea (approved):

src/cli/onefuzz/backend.py
```diff
@@ -41,7 +41,8 @@
 
 _ACCESSTOKENCACHE_UMASK = 0o077
 
-ONEFUZZ_BASE_PATH = os.path.join("~", ".cache", "onefuzz")
+HOME_PATH = os.environ.get("VIRTUAL_ENV") if os.environ.get("VIRTUAL_ENV") else "~"
+ONEFUZZ_BASE_PATH = os.path.join(HOME_PATH, ".cache", "onefuzz")
```
ranweiler, line 45: By modifying `ONEFUZZ_BASE_PATH`, we also modify`DEFAULT_TOKEN_PATH`. In the virtualenv case, this removes the filesystem access controls which restrict access to the token cache. We should ensure that the base path for `DEFAULT_TOKEN_PATH` stays the same (is always `$HOME`).

ranweiler (changes requested):

ranweiler (approved):

PR #2563, opened 2022-10-25
Don't log an error if we delete a Repro and it is already missing

Porges:
If we delete a Repro and it is already missing, this shouldn't be considered an error.

Also update the Insert/Update/Delete/Replace methods to return an `HttpStatusCode` instead of a raw `int`.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
@@ -1,4 +1,5 @@
 ﻿using System.Globalization;
+using System.Net;
 using System.Threading.Tasks;
 using ApiService.OneFuzzLib.Orm;
 using Azure.ResourceManager.Compute.Models;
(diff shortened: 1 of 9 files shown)


src/ApiService/ApiService/onefuzzlib/ReproOperations.cs
```diff
@@ -111,9 +112,10 @@ public ReproOperations(ILogTracer log, IOnefuzzContext context)
         // BUG?: why are we updating repro and then deleting it and returning a new value
         repro = repro with { State = VmState.Stopped };
         var r = await Delete(repro);
-        if (!r.IsOk) {
+        if (!r.IsOk && r.ErrorV.Status != HttpStatusCode.NotFound) {
```
Porges, line 115: Actual fix here.

stishkin (approved):

tevoinea (approved):

PR #2564, opened 2022-10-25
Override libfuzzer -runs parameter during coverage recording

Porges:
When we run the coverage task we pass all `target_options` on to the fuzzer being used for coverage.

For libfuzzer, if this includes the `-runs=X` parameter, we will run the recording step that many times; however, for coverage we only want to run it once.

This PR changes the coverage task so that if we see a `runs` parameter, we override it to `-runs=1`. Note that this is specific to libfuzzer, even though the coverage task is meant to be “generic”. Thus, the code in question is marked `//HACK`.


Porges:
Belatedly found out that we have `--fuzzing_target_options`.

PR #2565, opened 2022-10-25
Update the libfuzzer-with-options test to use fuzzing_target_options

Porges:
This currently causes the coverage task to time out, ensure that the `runs` parameter is only passed to the fuzzing task and not the coverage task.

### DIFF
--- a/src/integration-tests/integration-test.py
+++ b/src/integration-tests/integration-test.py
@@ -76,6 +76,7 @@ class Integration(BaseModel):
     reboot_after_setup: Optional[bool] = Field(default=False)
     test_repro: Optional[bool] = Field(default=True)
     target_options: Optional[List[str]]
+    fuzzing_target_options: Optional[List[str]]
     inject_fake_regression: bool = Field(default=False)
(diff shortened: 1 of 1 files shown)


src/integration-tests/integration-test.py
```diff
@@ -114,7 +115,7 @@ class Integration(BaseModel):
             ContainerType.inputs: 2,
         },
         reboot_after_setup=True,
-        target_options=["-runs=10000000"],
+        fuzzing_target_options=["-runs=10000000"],
```
ranweiler, line 118: It'd be good to have some other use of `target_options`, _ideally_ in a way that the target actually depends on. Something like an `--enable-crash` flag that gets passed on to the target. That'd require a new target to be written, so not for this PR.

ranweiler (approved):

src/integration-tests/integration-test.py
Porges: Yes, we were discussing earlier how the test doesn’t actually test that the `target_options` is used.

PR #2575, opened 2022-10-27
Enable MSVC security features

Porges:
Enabling some native code features which are SDL requirements. (Specifically, the requirement "Executable binary files must be hardened to leverage available platform security mitigations".)

I attempted to use BinSkim to enforce the requirements but it has trouble with Rust binaries at present.

### DIFF
--- a/src/agent/.cargo/config
+++ b/src/agent/.cargo/config
@@ -1,2 +1,0 @@
-[target.x86_64-pc-windows-msvc]
-rustflags = ["-C", "target-feature=+crt-static"]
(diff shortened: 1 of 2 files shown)


tevoinea (approved):

Porges:
It would probably make sense to do this in the packaging step to ensure we are scanning all binaries.

Porges:
Ideally we would generate and upload the SARIF file from BinSkim but this doesn't work with GitHub, see: https://github.com/microsoft/binskim/issues/509

PR #2535, opened 2022-10-20
Update References to Ubuntu 18.04 to 20.04

nharper285:
## Summary of the Pull Request

_What is this about?_
We will be deprecating OneFuzz's use of 18.04 Ubuntu Images. This is in accordance with Github's similar deprecation plan - linked [here](https://github.com/actions/runner-images/issues/6002).
This upgrade will prevent OneFuzz build failures when Github's plan goes into effect.

## PR Checklist
* [ ] Applies to work item: #2532 
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign the CLI.
* [ ] Tests added/passed
* [ ] Requires documentation to be updated
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Info on Pull Request

_What does this include?_

**Automation**
* [ ] ci.yml - build automation updates
* [ ] deploy-onefuzz.yml 

**Documentation**
* [ ] custom-images.md
* [ ] getting-started.md
* [ ] fuzzing-dotnet-with-libfuzzer.md
* [ ] support-platforms.md
* [ ] webhooks_events.md
* [ ] generate_docs.py

**Service Code**
* [ ] models.cs
* [ ] models.py
* [ ] test_models.py

## Validation Steps Performed

_How does someone test & validate?_
* [ ] Ensure Build Works
* [ ] Ensure Deployment Works
* [ ] Integration Tests

PR #2576, opened 2022-10-27
Fix Github actions warnings

Porges:
Node.js 12 actions are deprecated, as well as `::set-output` 

- Update `actions/checkout` to v3
- Update `actions/upload-artifact` to v3
- Update `actions/setup-python` to v4
- Update `actions/cache` to v3
- `set-output` on stdout is deprecated, update to `$GITHOUT_OUTPUT` method
- Change from `actions-rs/toolchain` (unsupported) to `dtolnay/rust-toolchain`

The only warning remaining after this is the one about Ubuntu 18.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,11 +34,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Install specific Rust version
-      uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


chkeita (approved):

AdamL-Microsoft:
> The only warning remaining after this is the one about Ubuntu 18.04.

I think Noah's PR #2535 should clear that as well.

PR #2577, opened 2022-10-27
Reduce fetches to VMSS

Porges:
Remove another `GET` from `ListInstanceIds`, and only invoke `ListInstanceIds` if we have to (i.e. one of the nodes is missing `InstanceId`).


src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -632,39 +636,37 @@ where x.State.ReadyForReset()
...
                 _log.Info($"decommissioning nodes");
-                await Async.Task.WhenAll(nodes
-                    .Where(node => machineIds.Contains(node.MachineId))
+                await Async.Task.WhenAll(nodesToReimage
                     .Select(async node => {
                         await _context.NodeOperations.ReleaseScaleInProtection(node).IgnoreResult();
                     }));
                 return;
 
             case NodeDisposalStrategy.ScaleIn:
```
Porges, line 666: I'm a little unclear on the logic of what we are doing here:

- reimage the VMSS node
- release scale-in protection on the VMSS node
- delete our internal knowledge of the node?

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -693,33 +695,32 @@ where x.State.ReadyForReset()
...
-                await _context.VmssOperations.DeleteNodes(scaleset.ScalesetId, machineIds);
-                await Async.Task.WhenAll(nodes
-                    .Where(node => machineIds.Contains(node.MachineId))
+                _log.Info($"deleting nodes {scaleset.ScalesetId:Tag:ScalesetId} {string.Join(", ", nodesToDelete.Select(n => n.MachineId)):Tag:MachineIds}");
+                await _context.VmssOperations.DeleteNodes(scaleset.ScalesetId, nodesToDelete);
+                await Async.Task.WhenAll(nodesToDelete
                     .Select(async node => {
                         await _context.NodeOperations.Delete(node);
-                        await _context.NodeOperations.ReleaseScaleInProtection(node).IgnoreResult();
+                        // REMOVE?: don't need to release scale-in protection if we have deleted the node
```
Porges, line 722: this removal to be checked

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -693,33 +695,32 @@ where x.State.ReadyForReset()
...
             case NodeDisposalStrategy.ScaleIn:
-                _log.Info($"deleting nodes {scaleset.ScalesetId:Tag:ScalesetId} {string.Join(", ", machineIds):Tag:MachineIds}");
-                await _context.VmssOperations.DeleteNodes(scaleset.ScalesetId, machineIds);
-                await Async.Task.WhenAll(nodes
-                    .Where(node => machineIds.Contains(node.MachineId))
+                _log.Info($"deleting nodes {scaleset.ScalesetId:Tag:ScalesetId} {string.Join(", ", nodesToDelete.Select(n => n.MachineId)):Tag:MachineIds}");
+                await _context.VmssOperations.DeleteNodes(scaleset.ScalesetId, nodesToDelete);
+                await Async.Task.WhenAll(nodesToDelete
                     .Select(async node => {
                         await _context.NodeOperations.Delete(node);
```
chkeita, line 721: we now have a [DeleteAll](https://github.com/chkeita/onefuzz/blob/49ca5b524802cb370394b654956504d91fc3ebc7/src/ApiService/ApiService/onefuzzlib/orm/Orm.cs#L159) operation in the ORM

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -444,22 +428,43 @@ public Async.Task<IReadOnlyList<string>> ListAvailableSkus(Region region)
...
-        foreach (var machineId in machineIds) {
-            if (machineToInstance.TryGetValue(machineId, out var instanceId)) {
-                _ = instanceIds.Add(instanceId);
+        foreach (var node in nodes) {
+            if (node.InstanceId is not null) {
+                _ = instanceIds.Add(node.InstanceId);
+                continue;
+            }
+
+            var lookup = await machineToInstanceLazy.Value;
```
chkeita, line 450: suggestion:
we could use this logic to update the instanceId in the node entry in the table. 
I would do it in separate method but that is not a strong preference.

Porges: This is just here for back-compat. After the next release we should never see missing InstanceID here.

Porges: I've made a change in the commit e41a2e9e47cdc6c50b761ea8b72d41ea6e752d2f to insert InstanceID during the CleanupNodes phase, so now there's never a case where we insert a Node without an InstanceID.

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
Porges: I'll update to this separately, I think

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
Porges: @tevoinea I think you’re the most familiar with this code/sequence

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
Porges: I think this is also the cause of the "not an active Instance ID" exceptions we are seeing

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -514,6 +514,11 @@ public Async.Task<Scaleset> SetShutdown(Scaleset scaleset, bool now)
 
         //ground truth of existing nodes
         var azureNodes = await _context.VmssOperations.ListInstanceIds(scaleSet.ScalesetId);
+        if (azureNodes is null) {
+            // didn't find scaleset
+            return (false, scaleSet);
+        }
+
```
tevoinea, line 521: This return is too early, lines 525-530 will clean up our database of nodes that are now deleted in azure.

tevoinea, line 521: I don't think ListInstanceIds can return null, only an empty dictionary.

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
tevoinea: Yup, that's right. The `Decommission` strategy is what we want to move to (I'm wrapping up validating it in canary). It's going to skip reimaging and just let azure reap our nodes via scale-in. Then instead of deleting the node in our database here, we delete it in `CleanupNodes` (lines 525-530 that I mentioned in the earlier comment).

src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
tevoinea: I think you're right, we could probably skip releasing scale in protection here.

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -161,37 +158,24 @@ public class VmssOperations : IVmssOperations {
         }
     }
 
-    public async Async.Task<IDictionary<Guid, string>> ListInstanceIds(Guid name) {
+    public async Async.Task<IDictionary<Guid, string>?> ListInstanceIds(Guid name) {
```
tevoinea, line 161: NIT: What do you think of this pattern?

```csharp
public async Async.Task<OneFuzzResult<IDictionary<Guid, string>>> ListInstanceId(...) {
    try {
        var results = new Dictionary<Guid, string>();
        ...
        return OneFuzzResult<Dictionary<Guid, string>>.Ok(results);
    } catch (RequestFailedException ex) when (ex.Status == 404) {
        _log.Verbose($"scaleset does not exist {name:Tag:VmssName}");
        return OneFuzzResult<Dictionary<Guid, string>>.Error(ErrorCode.UNABLE_TO_FIND, "Scaleset not found");
    } catch (Exception e) {
        _log.Exception(e);
        return OneFuzzResult<Dictionary<Guid, string>>.Error(ErrorCode.UNABLE_TO_FIND, "failed to list instance ids");
    }
}
```

We get better handling/logging for exceptions and the error/success case is explicit instead of implicitly through null.

tevoinea (approved):

PR #2584, opened 2022-10-28
Update the Authorization layer to allow unmanaged nodes

### LINKED ISSUE #2524: Update the Authorization layer to allow unmanaged nodes

chkeita:

In the authorization layer we now look for the claim `UnmamagedNode` or `ManagedNode` to allow a request from an agent.
closes #2524


src/ApiService/ApiService/UserCredentials.cs
```diff
@@ -9,16 +10,20 @@ namespace Microsoft.OneFuzz.Service;
 public interface IUserCredentials {
     public string? GetBearerToken(HttpRequestData req);
     public string? GetAuthToken(HttpRequestData req);
-    public Task<OneFuzzResult<UserInfo>> ParseJwtToken(HttpRequestData req);
+    public Task<OneFuzzResult<UserAuthInfo>> ParseJwtToken(HttpRequestData req);
 }
 
+public record UserAuthInfo(UserInfo UserInfo, List<string> Roles);
```
tevoinea, line 16: What if the roles are stored as an enum instead of a string? We could still (de)serialize into a string in requests/responses.

tevoinea (approved):

src/ApiService/ApiService/UserCredentials.cs
tevoinea: Non blocking. just floating the idea

chkeita: The claims are not all defined by us. some of the claims are added by the azure authentication layer. the string type allows us to collect them all

PR #2585, opened 2022-10-30
Add version to Application Insights

Porges:
This will help us identify if a particular version causes a problem.


Porges:
Can't find this info anywhere in AI logs…

tevoinea:
I'm not sure why it wasn't working for you (seems right to me?) but we could also add a tag with the OneFuzz version here: https://github.com/microsoft/onefuzz/blob/8be6e3c946965a30b902ec5f76000ddfc47ba3c8/src/ApiService/ApiService/Log.cs#L74

PR #2586, opened 2022-10-31
Delete nodes on decommission instead of just releasing scale in protection

tevoinea:

_What is this about?_

We want to discard nodes as soon as they're done completing their assigned tasks. This is different from the current behavior where we try to re-use nodes by reimaging them. Originally, this discard process was accomplished by removing scale-in protection and letting auto scale take care of everything. The issue with that approach is that scale-in actions only happen when the work queue for the pool is empty so letting auto scale delete the nodes won't work for a system under load.

This PR changes our approach to deleting nodes right away instead of reimaging or waiting for scale-in actions. The behavior continues to be behind the "NodeDisposalStrategies" env var until we decide to enable it by default.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
@@ -615,7 +615,7 @@ public class ScalesetOperations : StatefulOrm<Scaleset, ScalesetState, ScalesetO
         };
 
         await ReimageNodes(scaleSet, toReimage.Values, strategy);
-        await DeleteNodes(scaleSet, toDelete.Values, strategy);
+        await DeleteNodes(scaleSet, toDelete.Values);
 
         return (toReimage.Count > 0 || toDelete.Count > 0, scaleSet);
     }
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -658,10 +658,11 @@ where x.State.ReadyForReset()
 
         switch (disposalStrategy) {
             case NodeDisposalStrategy.Decommission:
-                _log.Info($"decommissioning nodes");
+                _log.Info($"Skipping reimage, deleting nodes: {string.Join(", ", nodesToReimage.Select(n => n.MachineId)):Tag:MachineIds}");
+                await _context.VmssOperations.DeleteNodes(scaleset.ScalesetId, nodesToReimage);
                 await Async.Task.WhenAll(nodesToReimage
                     .Select(async node => {
-                        await _context.NodeOperations.ReleaseScaleInProtection(node).IgnoreResult();
+                        await _context.NodeOperations.Delete(node);
```
tevoinea, line 665: This is the core of the change.

stishkin (approved):

chkeita (approved):

Porges (approved):

PR #2593, opened 2022-11-01
Ensure `wheel` updated when `pipenv check` invoked

ranweiler:
When `pipenv check` is invoked, it will always fail due to the `wheel` version installed in our current runner environments (which has a known DoS vuln).

In the one CI task where we invoke `pipenv check`, ensure that we update the system's ambient `wheel` beforehand.

Also, remove the ignore directive added in #2594, as a satisfying `wheel` package has now published (and not yanked).


AdamL-Microsoft (approved):

ranweiler:
`pipenv check` uses `safety` to query [`safety-db`](https://github.com/pyupio/safety-db). This has [`wheel <0.38.0` listed as insecure](https://github.com/pyupio/safety-db/blob/d440ab59ed8e4be3beacd4165ba3eb12ef3dc527/data/insecure.json#L8730-L8732), but a satisfying update has not yet been published to PyPI (0.38.0 was published, but [yanked](https://pypi.org/project/wheel/0.38.0/)).

ranweiler:
Before merging, and once a new version of `wheel` is published, re-enable the check that was disabled in #2594.

contrib/deploy-onefuzz-via-azure-devops/deploy-onefuzz.yml
```diff
@@ -48,6 +48,7 @@ stages:
             workingDirectory: "$(ONEFUZZ_DEPLOY_LOC)"
             script: |
               set -ex
+              python -m pip install -U wheel
```
ranweiler, line 51: I think this one may be unnecessary, because it is definitely insufficient. What really matters is the `wheel` version used in the `tox` environment (where `pipenv check` gets invoked). Let me try removing it.

PR #2607, opened 2022-11-08
Catching Model Exceptions During VMSS Update

nharper285:

_What is this about?_
Adding a more granular catching mechanism for the model error during calls of `UpdateAsync`.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -256,6 +256,8 @@ public class VmssOperations : IVmssOperations {
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"protection policy update is already in progress: {instanceId} in vmss {scaleset.ScalesetId}");
         } catch (RequestFailedException ex) when (ex.Status == 400 && ex.Message.Contains("The provided instanceId") && ex.Message.Contains("not an active Virtual Machine Scale Set VM instanceId.")) {
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"The node with instanceId {instanceId} no longer exists in scaleset {scaleset.ScalesetId}");
+        } catch (RequestFailedException ex) when (ex.Status == 400 && ex.Message.Contains(" reached its limit") && ex.Message.Contains("Upgrade the VMs to the latest model")) {
+            return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"The node with instanceId {instanceId} no longer exists in scaleset {scaleset.ScalesetId}");
         } catch (Exception ex) {
             _log.Exception(ex, $"unable to set protection policy on: {instanceId:Tag:InstanceId} in vmss {scaleset.ScalesetId:Tag:ScalesetId}");
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"unable to set protection policy on: {instanceId} in vmss {scaleset.ScalesetId}");
--- a/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
@@ -500,7 +500,7 @@ public class VmssOperations : IVmssOperations {
                 WaitUntil.Started,
                 new VirtualMachineScaleSetVmInstanceRequiredIds(instanceIds));
             if (r.GetRawResponse().IsError) {
-                _log.Error($"failed to start update instance for scaleset {scalesetId:Tag:ScalesetId} due to {r.GetRawResponse().ReasonPhrase:Tag:Error}");
+                _log.Error($"failed to start upgrade instance for scaleset {scalesetId:Tag:ScalesetId} due to {r.GetRawResponse().ReasonPhrase:Tag:Error}");
             }
         } catch (RequestFailedException ex) {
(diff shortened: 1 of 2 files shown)


src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -485,26 +494,33 @@ public Async.Task<IReadOnlyList<string>> ListAvailableSkus(Region region)
...
+        try {
+            _log.Info($"upgrading VMSS nodes - name: {scalesetId:Tag:ScalesetId} ids: {string.Join(", ", instanceIds):Tag:InstanceIds}");
+            var r = await vmssResource.UpdateInstancesAsync(
+                WaitUntil.Started,
+                new VirtualMachineScaleSetVmInstanceRequiredIds(instanceIds));
+            if (r.GetRawResponse().IsError) {
+                _log.Error($"failed to start update instance for scaleset {scalesetId:Tag:ScalesetId} due to {r.GetRawResponse().ReasonPhrase:Tag:Error}");
+            }
+        } catch (RequestFailedException ex) {
+            _log.Exception(ex, $"failed to upgrade scaleset instances");
```
mgreisen, line 506: let's use consistent terms in our log messages. choose `update` or `upgrade`.

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
```diff
@@ -85,12 +85,17 @@ public class VmssOperations : IVmssOperations {
...
             var scalesetResource = GetVmssResource(name);
             var patch = new VirtualMachineScaleSetPatch();
             patch.Sku.Capacity = capacity;
-            _ = await scalesetResource.UpdateAsync(WaitUntil.Started, patch);
-            return OneFuzzResultVoid.Ok;
+            try {
+                _log.Info($"updating VM count {name:Tag:VmssName} - {capacity:Tag:Count}");
+                _ = await scalesetResource.UpdateAsync(WaitUntil.Started, patch);
+                return OneFuzzResultVoid.Ok;
+            } catch (RequestFailedException ex) {
```
tevoinea, line 95: Do we want to catch all exceptions or specifically the model exceptions?

We could follow a model similar to here:
https://github.com/microsoft/onefuzz/blob/f20c482052178ea20d9adf13d063bb979e900607/src/ApiService/ApiService/onefuzzlib/NodeOperations.cs#L400-L402

And maybe add ex.Message.contains("has reached its limit of 10 models that may be referenced by one or more VMs belonging to the Virtual Machine Scale Set. Upgrade the VMs to the latest model of the Virtual Machine Scale Set before trying again."

tevoinea: Or even better, factor out that logic into a separate function so it's repeatable in the rest of the file! 

`catch (RequestFailedException ex) when (IsModelException(ex) {`

then

```csharp
static bool IsModelException(RequestFailedException ex) {
    return ex.Status == 400 
        && ex.ErrorCode == "BadRequest" 
        && ex.Message.Contains("has reached its limit of 10 models that may be referenced by one or more VMs belonging to the Virtual Machine Scale Set. Upgrade the VMs to the latest model of the Virtual Machine Scale Set before trying again.");
}

Comment (nharper285) [reply src/ApiService/ApiService/onefuzzlib/VmssOperations.cs:95]: I don't neccessarily see an issue with catching all exceptions considering how much we monitor them. If any of these calls fail, then we get into the pattern of not processing scalesets fully and having issues like node's state hanging.

File: src/ApiService/ApiService/onefuzzlib/VmssOperations.cs [hunk 3]
```diff
@@ -247,6 +256,8 @@ private Task<string> GetInstanceIdForVmId(Guid scaleset, Guid vmId)
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"protection policy update is already in progress: {instanceId} in vmss {scaleset.ScalesetId}");
         } catch (RequestFailedException ex) when (ex.Status == 400 && ex.Message.Contains("The provided instanceId") && ex.Message.Contains("not an active Virtual Machine Scale Set VM instanceId.")) {
             return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"The node with instanceId {instanceId} no longer exists in scaleset {scaleset.ScalesetId}");
+        } catch (RequestFailedException ex) when (ex.Status == 400 && ex.Message.Contains(" reached its limit") && ex.Message.Contains("Upgrade the VMs to the latest model")) {
+            return OneFuzzResultVoid.Error(ErrorCode.UNABLE_TO_UPDATE, $"The node with instanceId {instanceId} no longer exists in scaleset {scaleset.ScalesetId}");
```
tevoinea, line 260: Error string should be along the lines of: "the scaleset has reached the model limit"

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
tevoinea: Fair enough, we're logging the exception, so we know at least it won't go unnoticed.

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/VmssOperations.cs
nharper285: Whoops

mgreisen (approved):

PR #2612, opened 2022-11-10
Use broker or browser login instead of device flow

Porges:
Update CLI to attempt broker or browser-based authentication first; if you `Ctrl-C` to cancel it, you can fall back to device code login.

Also updated the MSAL dependency to latest version and pass `allow_broker=True` which will allow the use of Web Account Manager (WAM), if it is available.

Using browser auth requires the `http://localhost` redirect URI, and using the broker requires a special custom URI including the app ID (see code).

### DIFF
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -236,7 +236,6 @@ class Backend:
                 authority=self.config.authority,
                 token_cache=self.token_cache,
                 allow_broker=True,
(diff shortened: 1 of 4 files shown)


Porges:
Verified; ready for review.

mgreisen (approved):

chkeita (approved):

stishkin (approved):

PR #2615, opened 2022-11-11
Update to .NET 7

Porges:
Update service code to .NET 7.0.

Also update the version installed onto agent nodes via setup scripts to 7.0.100.


src/agent/LibFuzzerDotnetLoader/LibFuzzerDotnetLoader.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <LangVersion>10.0</LangVersion>
```
tevoinea, line 5: Might as well bump to C# 11 too right?

chkeita (approved):

PR #2617, opened 2022-11-13
Update LLVM from v10 to v12

Porges:
While I was investigating `llvm-symbolizer` issues, I found that we are installing LLVM-10 by default.

Now that we are using the Ubuntu 20.04 image as the default, use the latest LLVM version available for that release. (In addition, v12 is also available on 22.04, whereas v10 is not.)


Porges:
Passes `check-pr`.

tevoinea (approved):

PR #2620, opened 2022-11-14
Add support for feature flags

tevoinea:

_What is this about?_

Adds a new App Configuration resource to the onefuzz deploy.

## Info on Pull Request

_What does this include?_

* Deploy new App Config resource
* Wire up the azure functions to allow using it
* Demo 1 case where it can be used

## Validation Steps Performed

* Ran deploy locally
* Enabled feature flag in azure
* Validated it was enabled
* Disabled feature flag in azure
* Validated it was disabled

Ran check-pr


src/deployment/bicep-templates/feature-flags.bicep
```diff
@@ -0,0 +1,20 @@
+param location string
+
+var suffix = uniqueString(resourceGroup().id)
+var appConfigName = 'app-config-${suffix}'
+
+resource featureFlags 'Microsoft.AppConfiguration/configurationStores@2022-05-01' = {
+  name: appConfigName
+  location: location
+  sku:{
```
stishkin, line 9: Managed identity is better than connection string:

  identity: {
    type: 'SystemAssigned'
  }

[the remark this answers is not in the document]
stishkin: resource func 'Microsoft.Web/sites@2021-03-01' existing = {
    name: <AZURE FUNCTION NAME>
}
var readerRoleId = '516239f1-63e1-4d78-a4de-a74fb236a071'

resource assignReaderRole 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
  name: guid('${func.identity.principalId}-role-assignment-reader')
  scope: featureFlags
  properties: {
    principalId: func.identity.principalId
    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', readerRoleId)
  }
}

stishkin (changes requested):

src/ApiService/ApiService/ServiceConfiguration.cs
```diff
@@ -82,6 +83,8 @@ public class ServiceConfiguration : IServiceConfig {
     public string? ApplicationInsightsAppId => GetEnv("APPINSIGHTS_APPID");
     public string? ApplicationInsightsInstrumentationKey => GetEnv("APPINSIGHTS_INSTRUMENTATIONKEY");
 
+    public string? AppConfigurationConnectionString => GetEnv("APPCONFIGURATION_CONNECTION_STRING");
```
stishkin, line 86: And so this is not needed when using rbac with managed identity

stishkin: but you still need to pass feature flags info in order to instantiate the object and then use 

new Azure.Identity.DefaultAzureCredential(false)

for auth

stishkin:
Deploying new azure resource just to store one value.... seems like too much ? 
Are there any other configuration options we can move there as well ?

src/ApiService/ApiService/Program.cs
```diff
@@ -47,13 +48,22 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+            })
             .ConfigureFunctionsWorkerDefaults(builder => {
                 builder.UseMiddleware<LoggingMiddleware>();
                 builder.AddApplicationInsights(options => {
                     options.ConnectionString = $"InstrumentationKey={configuration.ApplicationInsightsInstrumentationKey}";
                 });
+                builder.UseAzureAppConfiguration();
             })
             .ConfigureServices((context, services) => {
+                services.AddAzureAppConfiguration();
```
Porges, line 66: Not sure if you found this while enabling AppConfiguration but with the AppConfiguration stuff there's also [explicit feature flag support](https://learn.microsoft.com/en-us/azure/azure-app-configuration/quickstart-feature-flag-azure-functions-csharp?tabs=in-process) via `AddFeatureManagement` which gives you an `IFeatureManagerSnapshot` injected to use instead of having to parse the configuration settings manually out of e.g. `context.AppConfiguration[FeatureFlagConstants.EnableJinjaConvert]` (instead it would be something like `_features.IsEnabledAsync(FeatureFlagConstants.EnableJinjaConvert)`).

src/ApiService/ApiService/Program.cs
```diff
@@ -47,13 +49,22 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+            })
             .ConfigureFunctionsWorkerDefaults(builder => {
                 builder.UseMiddleware<LoggingMiddleware>();
                 builder.AddApplicationInsights(options => {
                     options.ConnectionString = $"InstrumentationKey={configuration.ApplicationInsightsInstrumentationKey}";
                 });
             })
             .ConfigureServices((context, services) => {
+                services.AddAzureAppConfiguration();
+                var _ = services.AddFeatureManagement();
```
Porges, line 67:
```suggestion
                _ = services.AddFeatureManagement();
```

Porges: can use discard

src/ApiService/ApiService/ServiceConfiguration.cs
```diff
@@ -82,6 +83,8 @@ public class ServiceConfiguration : IServiceConfig {
     public string? ApplicationInsightsAppId => GetEnv("APPINSIGHTS_APPID");
     public string? ApplicationInsightsInstrumentationKey => GetEnv("APPINSIGHTS_INSTRUMENTATIONKEY");
 
+    public string? AppConfigurationEndpoint => GetEnv("APPCONFIGURATION_ENDPOINT");
```
Porges, line 86: If this is always required we should make it return `string?` and do an up-front check.

Ideally for the ServiceConfiguration I'd like us to have a method that loads everything from the environment up front so we catch any mis-configurations as soon as the program loads. But that's a future change :)

src/ApiService/ApiService/onefuzzlib/notifications/NotificationsBase.cs
```diff
@@ -93,13 +99,15 @@ protected class Renderer {
             _reportUrl = reportUrl;
             _targetUrl = targetUrl;
             _inputUrl = inputUrl;
+            _shouldConvertJinja = shouldConvertJinja || true;
```
Porges, line 102: `_shouldConvertJinja` will always be `true`?

tevoinea: The default for the FF if it doesn't exist is false and I wanted it to be enabled when we deployed.

I renamed the FF to `scribanOnly` so that it's compatible with the default false.

PR #2632, opened 2022-11-16
Bug fixes related to the unmanaged nodes

chkeita:
- fix the token request in the client
- adding a managed field to the agentConfig
- fix typo in the appId claim type
- fix typo in the UnmanagedNode claim value
- fix query PoolOperation.GetByClientId

### DIFF
--- a/src/ApiService/ApiService/Functions/Pool.cs
+++ b/src/ApiService/ApiService/Functions/Pool.cs
@@ -134,7 +134,7 @@ public class Pool {
                     InstanceId: instanceId,
                     ClientCredentials: null,
                     MultiTenantDomain: _context.ServiceConfiguration.MultiTenantDomain,
-                    IsUnmanaged: !p.Managed)
+                    Managed: p.Managed)
         };
     }
 }
--- a/src/ApiService/ApiService/Functions/Pool.cs
+++ b/src/ApiService/ApiService/Functions/Pool.cs
@@ -133,7 +133,8 @@ public class Pool {
                     HeartbeatQueue: queueSas,
                     InstanceId: instanceId,
                     ClientCredentials: null,
-                    MultiTenantDomain: _context.ServiceConfiguration.MultiTenantDomain)
+                    MultiTenantDomain: _context.ServiceConfiguration.MultiTenantDomain,
+                    IsUnmanaged: !p.Managed)
         };
     }
 }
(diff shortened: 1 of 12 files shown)


src/ApiService/ApiService/onefuzzlib/EndpointAuthorization.cs
```diff
@@ -46,8 +45,8 @@ public class EndpointAuthorization : IEndpointAuthorization {
             return await _context.RequestHandling.NotOk(req, tokenResult.ErrorV, "token verification", HttpStatusCode.Unauthorized);
         }
 
-        var token = tokenResult.OkV;
-        if (await IsUser(token)) {
+        var token = tokenResult.OkV.UserInfo;
+        if (await IsUser(tokenResult.OkV)) {
             if (!allowUser) {
                 return await Reject(req, tokenResult.OkV.UserInfo);
```
tevoinea, line 51:
```suggestion
                return await Reject(req, token);
```

src/ApiService/ApiService/onefuzzlib/EndpointAuthorization.cs
```diff
@@ -58,7 +57,7 @@ public class EndpointAuthorization : IEndpointAuthorization {
             }
         }
 
-        if (await IsAgent(token) && !allowAgent) {
+        if (await IsAgent(tokenResult.OkV) && !allowAgent) {
             return await Reject(req, tokenResult.OkV.UserInfo);
```
tevoinea, line 61:
```suggestion
            return await Reject(req, token);
```

src/ApiService/ApiService/Functions/Pool.cs
```diff
@@ -133,7 +133,8 @@ private static PoolGetResult PoolToPoolResponse(Service.Pool p)
                     HeartbeatQueue: queueSas,
                     InstanceId: instanceId,
                     ClientCredentials: null,
-                    MultiTenantDomain: _context.ServiceConfiguration.MultiTenantDomain)
+                    MultiTenantDomain: _context.ServiceConfiguration.MultiTenantDomain,
+                    IsUnmanaged: !p.Managed)
```
tevoinea, line 137: I know managed is our current default which is why you added IsUnmanaged but I think it'll be easier to reason about in the future if we do `IsManaged`. Specifically, to avoid the double negative with `!IsUnmanaged`.

tevoinea (approved):

PR #2637, opened 2022-11-20
Bump azure_* crates

Porges:
Bump the crates which dependabot is trying to bump.

Remove invocations of `into_future` as `.await` now automatically uses the [`IntoFuture` trait](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) as of 1.64.


tevoinea (approved):
Are you using a formatter for the Cargo.toml? If so, do you want to add the extension/configuration to the devcontainer so we can all use the same one? Or open an issue to add it?

Porges:
> Are you using a formatter for the Cargo.toml? If so, do you want to add the extension/configuration to the devcontainer so we can all use the same one? Or open an issue to add it?

Good idea, I'll send it as another PR.

PR #2642, opened 2022-11-21
Fixing .NET crash report no-repro

Porges:
~~Based upon `check-pr` results we still need .NET 6 in order to run the existing SharpFuzz binary; modify install scripts to install multiple supported versions.~~

The minidump file being generated did not have enough information for SOS to correctly locate the exception object. Switch to “full” minidumps instead.

- Fix a small bug with coverage reporting task ID as job ID.

- I had thought we needed .NET 6 installed as well to run SharpFuzz, but now I'm not so sure. In any case, installing both versions will not hurt (and may help users).

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Extension.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Extension.cs
@@ -303,7 +303,7 @@ public class Extensions : IExtensions {
             urlsUpdated.Add(toolsAzCopy);
             urlsUpdated.Add(toolsSetup);
 
-            var toExecuteCmd = $"sh setup.sh {mode.ToString().ToLowerInvariant()}";
+            var toExecuteCmd = $"bash setup.sh {mode.ToString().ToLowerInvariant()}";
             var extensionSettings = JsonSerializer.Serialize(new { CommandToExecute = toExecuteCmd, FileUris = urlsUpdated }, _extensionSerializerOptions);
 
             var extension = new VMExtensionWrapper {
--- a/src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
+++ b/src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
@@ -49,7 +49,7 @@ const MINIDUMP_TYPE_VAR: &str = "COMPlus_DbgMiniDumpType";
 const MINIDUMP_NAME_VAR: &str = "COMPlus_DbgMiniDumpName";
 
 const MINIDUMP_ENABLE: &str = "1";
-const MINIDUMP_TYPE_HEAP: &str = "2";
+const MINIDUMP_TYPE_FULL: &str = "4";
 
 // Invoke target with .NET runtime environment vars set to create minidumps.
 //
@@ -70,7 +70,7 @@ async fn collect_dump(
 
     // Set `dotnet` environment vars to enable saving minidumps on crash.
     cmd.env(ENABLE_MINIDUMP_VAR, MINIDUMP_ENABLE);
-    cmd.env(MINIDUMP_TYPE_VAR, MINIDUMP_TYPE_HEAP);
+    cmd.env(MINIDUMP_TYPE_VAR, MINIDUMP_TYPE_FULL);
     cmd.env(MINIDUMP_NAME_VAR, dump_path);
 
     let mut child = cmd.spawn()?;
--- a/src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
+++ b/src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
@@ -27,12 +27,18 @@ pub async fn collect_exception_info(
         }
     };
 
-    let exception = dump.exception().await?;
+    let exception = dump.exception().await;
 
     // Remove temp dir cooperatively.
     spawn_blocking(move || tmp_dir).await?;
 
-    Ok(exception)
+    match exception {
+        Ok(r) => Ok(Some(r)),
+        Err(e) => {
+            error!("unable to extract exception info: {}", e);
+            Ok(None)
+        }
+    }
 }
 
 const DUMP_FILE_NAME: &str = "tmp.dmp";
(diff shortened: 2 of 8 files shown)


src/agent/onefuzz-task/src/tasks/report/dotnet/generic.rs
```diff
@@ -171,7 +171,7 @@ impl AsanProcessor {
             format_err!("unable to sha256 digest input file: {}", input.display())
         })?;
 
-        let job_id = self.config.common.task_id;
+        let job_id = self.config.common.job_id;
```
Porges, line 174: found another bug while investigating

src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
```diff
@@ -208,7 +212,8 @@ pub fn parse_sos_print_exception_output(text: &str) -> Result<DotnetExceptionInf
         }
     }
 
-    let exception = exception.ok_or_else(|| format_err!("missing exception type"))?;
+    let exception =
+        exception.ok_or_else(|| format_err!("missing exception type, output was:\n{}", text))?;
```
Porges, line 216: really we should collect up all the missing info and report it all at once (is there a `Validation` type?)

src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
```diff
@@ -43,7 +49,7 @@ const MINIDUMP_TYPE_VAR: &str = "COMPlus_DbgMiniDumpType";
 const MINIDUMP_NAME_VAR: &str = "COMPlus_DbgMiniDumpName";
 
 const MINIDUMP_ENABLE: &str = "1";
-const MINIDUMP_TYPE_HEAP: &str = "2";
+const MINIDUMP_TYPE_FULL: &str = "4";
```
Porges, line 52: the actual fix is here

src/runtime-tools/win64/onefuzz.ps1
```diff
@@ -6,7 +6,7 @@ $env:ONEFUZZ_ROOT = "C:\onefuzz"
 $env:ONEFUZZ_TOOLS = "C:\onefuzz\tools"
 $env:LLVM_SYMBOLIZER_PATH = "llvm-symbolizer"
 $env:RUST_LOG = "info"
-$env:DOTNET_VERSION = "7.0.100"
+$env:DOTNET_VERSIONS = "7.0.100;6.0.403"
```
Porges, line 9: At first I thought this was needed for SharpFuzz, now I'm not so sure…

src/agent/onefuzz-task/src/tasks/report/dotnet/common.rs
```diff
@@ -97,12 +103,10 @@ impl DotnetDumpFile {
...
     }
 
-    pub async fn exception(&self) -> Result<Option<DotnetExceptionInfo>> {
+    pub async fn exception(&self) -> Result<DotnetExceptionInfo> {
         let output = self.exec_sos_command(SOS_PRINT_EXCEPTION).await?;
         let text = String::from_utf8_lossy(&output.stdout);
-        let exception = parse_sos_print_exception_output(&text).ok();
-
-        Ok(exception)
+        parse_sos_print_exception_output(&text)
```
Porges, line 109: Fixed to not throw away error info

src/runtime-tools/win64/onefuzz.ps1
Porges: …it won’t hurt to leave this here, though.

src/ApiService/ApiService/onefuzzlib/Extension.cs
```diff
@@ -303,7 +303,7 @@ private class Settings {
             urlsUpdated.Add(toolsAzCopy);
             urlsUpdated.Add(toolsSetup);
 
-            var toExecuteCmd = $"sh setup.sh {mode.ToString().ToLowerInvariant()}";
+            var toExecuteCmd = $"bash setup.sh {mode.ToString().ToLowerInvariant()}";
```
Porges, line 306: Script is `#!/bin/bash` so run it with Bash, not Dash.

src/runtime-tools/linux/setup.sh
```diff
@@ -144,16 +144,18 @@ if type apt > /dev/null 2> /dev/null; then
     curl --retry 10 -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh 2>&1 | logger -s -i -t 'onefuzz-curl-dotnet-install'
     chmod +x dotnet-install.sh
 
-    logger "running dotnet install"
-    /bin/bash ./dotnet-install.sh --version "$DOTNET_VERSION" --install-dir "$DOTNET_ROOT" 2>&1 | logger -s -i -t 'onefuzz-dotnet-setup'
+    for version in "${DOTNET_VERSIONS[@]}"; do
+        logger "running dotnet install $version"
+        /bin/bash ./dotnet-install.sh --version "$version" --install-dir "$DOTNET_ROOT" 2>&1 | logger -s -i -t 'onefuzz-dotnet-setup'
```
tevoinea, line 149: What's the interaction here when we install 2 different versions to the same `install-dir`? For example, in line 156 (`"$DOTNET_ROOT"/dotnet tool install dotnet-dump --version 6.0.351802 --tool-path /onefuzz/tools 2>&1 | logger -s -i -t 'onefuzz-dotnet-tools'`), will it be using dotnet 6 or 7? Do we need to install the same tool twice but with each version of dotnet? How do we decide which version is being used?

chkeita (approved):

src/runtime-tools/linux/setup.sh
Porges: I think it will install it for the latest framework you have installed that the tool supports. In this case [`dotnet-dump` supports net6.0](https://www.nuget.org/packages/dotnet-dump#supportedframeworks-body-tab). 

Note that for these tools we don't need to have them on the same framework as our code, since we don't run any .NET on the nodes, and the tool does not attach to any running .NET code, it is used to analyze dumps after they are created.

Porges: In the case you invoke it with `dotnet` it will use the framework that the executable is targetted at, so we have both v6 and v7 installed at the moment.

PR #2643, opened 2022-11-22
Include a reason when telling a node it is not allowed to schedule work

chkeita:
Updated the response given by the server when denying work to a node. 
We now include a reason field to containing an explanation

### DIFF
--- a/src/ApiService/ApiService/Functions/AgentCanSchedule.cs
+++ b/src/ApiService/ApiService/Functions/AgentCanSchedule.cs
@@ -42,29 +42,30 @@ public class AgentCanSchedule {
                 canScheduleRequest.MachineId.ToString());
         }
 
-        var allowed = true;
-        if (!await _context.NodeOperations.CanProcessNewWork(node)) {
-            allowed = false;
-        }
+
+        var canProcessNewWork = await _context.NodeOperations.CanProcessNewWork(node);
+        var allowed = canProcessNewWork.IsAllowed;
+        var reason = canProcessNewWork.Reason;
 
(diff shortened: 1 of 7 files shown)


src/ApiService/ApiService/Functions/AgentCanSchedule.cs
```diff
@@ -42,29 +42,30 @@ public class AgentCanSchedule {
...
-        if (allowed) {
-            var scp = await _context.NodeOperations.AcquireScaleInProtection(node);
-            if (!scp.IsOk) {
-                _log.Warning($"Failed to acquire scale in protection for: {node.MachineId:Tag:MachineId} in: {node.PoolName:Tag:PoolName} due to {scp.ErrorV:Tag:Error}");
-            }
-
-            _ = scp.OkV; // node could be updated but we don't use it after this
-
-            allowed = scp.IsOk;
+        var scp = await _context.NodeOperations.AcquireScaleInProtection(node);
```
tevoinea, line 63: We might be acquiring scaling protection for a node where work is stopped. Did you mean to add an early return in the `if (workStopped)` block?

tevoinea (approved):

src/deny.toml
```diff
@@ -17,6 +17,8 @@ yanked = "deny"
 ignore = [
     "RUSTSEC-2022-0048", # xml-rs is unmaintained
     "RUSTSEC-2021-0139", # ansi_term is unmaintained
+    "RUSTSEC-2021-0145", # atty bug: we are unaffected (no custom allocator)
+]
```
chkeita, line 21:
```suggestion 
```

Issue #2644, opened 2022-11-22
Deployment Failure when Resource group name does not match the onefuzz instance name


Preeti123456:
## Information

- Onefuzz version: 5.20.0
- OS: Windows

## Provide detailed reproduction steps (if any)

1. Download the version 
2. Deploy when the instance name and resource group name is different.

### Expected result

_What is the expected result of the above steps?_
The deployment should be successful.

### Actual result

_What is the actual result of the above steps?_

```INFO:deploy:enabling DOTNET function: 'WebhookPing'
INFO:deploy:updating Python settings
**ERROR**: (ResourceNotFound) The Resource 'Microsoft.Web/sites/DevExOneFuzzApp' under resource group 'DevExOneFuzzApp' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix
Code: ResourceNotFound
Message: The Resource 'Microsoft.Web/sites/DevExOneFuzzApp' under resource group 'DevExOneFuzzApp' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix```


### Error - RCA  

Script -  deploy.py  
Function - enable_dotnet_func
![image](https://user-images.githubusercontent.com/22413185/203222192-190ba5f8-0345-49e2-826b-9e37ecb48bbf.png)

You are using the same value of name and resource group.


AB#42331416

PR #2645, opened 2022-11-22
Fix invocation of functionapp in the deployment script

chkeita:
The deployment script was using the application name as the resource group when updating the pp settings

#2644


mgreisen (approved):
:shipit:

Issue #2501, opened 2022-10-11
Support Scriban template verification


tevoinea:
From the discussion #2491, users should be able to validate their scriban notification templates.

- [x] Create an endpoint which can render a notification template using scriban - #2655 
- [ ] Create a CLI command for the above endpoint (ex: `onefuzz debug notification_template @./myconfig.json`

AB#41749595

PR #2655, opened 2022-11-25
Validate notification template

tevoinea:

_What is this about?_

Adds a new endpoint `/ValidateScriban` which can be used to test scriban templates.

## PR Checklist
* [ ] Applies to work item: #2501
* [x] Tests added/passed

## Info on Pull Request

_What does this include?_

* **A new endpoint `/ValidateScriban`.** Which will attempt to render a template using scriban, return the rendered template and the rendering context used.
* **New model: `TemplateRenderContext`.** This simply gives a name to the anonymous type we were already using before for rendering templates.
* **New request/response objects.** Used for receiving and returning rendering data.
* **Allow disabling jinja migration in the render function.** The `ValidateScriban` endpoint should not be attempting to support jinja, therefore we need a way to disable the functionality. This will be modifiable by feature flag as well once #2620 merges.

## Validation Steps Performed

1. Start azure functions
2. Do a post request with the following body:
```json
{
  "template": "{{ report.crash_site }} - {{ report.executable }}",
}
```
3. Validate it was rendered correctly.
4. Do a post request with the following body:
```json
{
  "template": "{{ report.NOTAREALPROPERTY }} - {{ report.executable }}",
}
```
5. Validate an error was thrown.
6. Do a post request with the following body:
```json
{
  "template": "{{ report.crash_site }} - {{ report.executable }}",
  "context": {
    "report": {
      "crash_site": "ABC - 123  - THIS IS AN AWESOME CRASH SITE"
    }
  }
}
```
7. Validate that the new crash site string was used.

In all of the above cases, validate that the returned `TemplateRenderContext` was correct for the input as well.


stishkin (approved):

src/ApiService/ApiService/Functions/ValidateScriban.cs
```diff
@@ -0,0 +1,150 @@
...
+        var (renderer, templateRenderContext) = await GenerateTemplateRenderContext(request.OkV.Context);
+
+        var renderedTemaplate = await renderer.Render(request.OkV.Template, new Uri(instanceUrl), enableJinjaAdapter: false);
+
+        var response = new TemplateValidationResponse(
+            renderedTemaplate,
+            templateRenderContext
+        );
+
+        return await RequestHandling.Ok(req, response);
```
mgreisen, line 31: What kind of error messages will people see if the rendering fails?

tevoinea: In the current state of the PR, the invalid templates simply won't render and there won't be an error message.

For example, given the input:
`This exists: {{ report.target_executable }} but this doesn't: {{ report.NOT_A_REAL_PROPERTY }}`

The result:
`This exists: target.exe but this doesn't: `

From reading the scriban docs, there is a way to enable 'strict' variable resolution which would cause the above example to throw an exception which we could catch and return in the response. I'm investigating this option to see if the exception information is valuable or not.

tevoinea: Discussed offline, we'll be validating the template using strict mode before inserting it into the database. This way we don't need to worry about notification creation time because only strictly validated templates should be in the database. This means we'll also use strict mode in the validation endpoint.

PR #2662, opened 2022-11-30
pass the machine_identity from the agent to the task

chkeita:
The default value of the MachineIdentity was being passed to the task instead of the actual machine identity. This would cause all agent to have the id `00000000-0000-0000-0000-000000000000`


src/agent/onefuzz-agent/src/worker.rs
```diff
@@ -221,7 +229,7 @@ impl IWorkerRunner for WorkerRunner {
 
         config.insert(
             "machine_identity".to_string(),
-            serde_json::to_value(MachineIdentity::default())?,
+            serde_json::to_value(&self.machine_identity)?,
```
chkeita, line 232: this is the fix

tevoinea (approved):

src/agent/onefuzz-task/src/local/common.rs
```diff
@@ -248,7 +251,12 @@ pub fn build_local_context(
         task_id,
         instance_id,
         setup_dir,
-        ..Default::default()
+        machine_identity: MachineIdentity::from_metadata().await?,
+        instance_telemetry_key: None,
+        heartbeat_queue: None,
+        microsoft_telemetry_key: None,
+        logs: None,
+        min_available_memory_mb: u64::default(),
```
Porges, line 259: we should just say `0` 😆

Porges (approved):

src/agent/onefuzz-task/src/local/common.rs
chkeita:
```suggestion
        min_available_memory_mb: 0,
```

PR #2663, opened 2022-12-01
Better rejection message when parsing validated strings

Porges:
When parsing a `ValidatedString` from JSON and it fails, include a message about the expected format of the string.

Reworked the classes using C#7 features to reduce the amount of boilerplate needed to add a new validated string type.


src/ApiService/ApiService/OneFuzzTypes/Validated.cs
```diff
@@ -1,161 +1,107 @@
...
         if (value is null) {
             throw new JsonException("expected a string");
         }
 
-        if (TryParse(value, out var result)) {
+        if (ValidatedStringBase<T>.TryParse(value, out var result)) {
             return result;
         } else {
-            throw new JsonException($"unable to parse input as a {typeof(T).Name}");
+            throw new JsonException($"unable to parse input as a {typeof(T).Name}: {T.Requirements}");
```
Porges, line 65: Here is where we now included more information

src/ApiService/ApiService/OneFuzzTypes/Validated.cs
```diff
@@ -1,161 +1,107 @@
...
 using System.Text.RegularExpressions;
 using Azure.Core;
 
 namespace Microsoft.OneFuzz.Service;
 
-static class Check {
-    private static readonly Regex _isAlnum = new(@"\A[a-zA-Z0-9]+\z", RegexOptions.Compiled);
-    public static bool IsAlnum(string input) => _isAlnum.IsMatch(input);
+static partial class Check {
+    [GeneratedRegex("\\A[a-zA-Z0-9]+\\z")]
```
tevoinea, line 10: Interesting read about `GeneratedRegex` for anyone that's reviewing this:

https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/#source-generation

src/ApiService/ApiService/OneFuzzTypes/Validated.cs
```diff
@@ -1,161 +1,107 @@
...
+[JsonConverter(typeof(ValidatedStringConverter<Container>))]
+public sealed record Container : ValidatedStringBase<Container>, IValidatedString<Container> {
+    private Container(string value) : base(value) { }
+    public static Container Parse(string input) => new(input);
+
     // See: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage
     // - 3-63
     // - Lowercase letters, numbers, and hyphens.
     // - Start with lowercase letter or number. Can't use consecutive hyphens.
     private static readonly Regex _containerRegex = new(@"\A(?!-)(?!.*--)[a-z0-9\-]{3,63}\z", RegexOptions.Compiled);
```
tevoinea, line 104: Should we add this regex to `Check` too?

src/ApiService/ApiService/OneFuzzTypes/Validated.cs
Porges: The C# linter suggested doing this so I let it make the change at the same time I was reworking the code.

src/ApiService/ApiService/OneFuzzTypes/Validated.cs
Porges: Will do

stishkin (approved):

PR #2665, opened 2022-12-01
Except on uppercase application names in deployments

### LINKED ISSUE #2652: App name with uppercasing gives error
## Information
App name with anything apart from lowercase gives error.

- Onefuzz version: 5.20.0
- OS: Windows

Deploy a cluster with app name with uppercase letters

### Expected result

Nodes should get configured and jobs should be deployed.

### Actual result

The nodes are stuck in init state

AdamL-Microsoft:

Fixes #2652 

OneFuzz deployments will have issues with names created containing uppercase letters. This PR adds an exception and stops deploying if the ~~resource group or~~ application name are not all lowercase.

## Info on Pull Request

updated:
* src/deployment/deploy.py


src/deployment/deploy.py
```diff
@@ -1236,8 +1249,8 @@ def main() -> None:
     formatter = argparse.ArgumentDefaultsHelpFormatter
     parser = argparse.ArgumentParser(formatter_class=formatter)
     parser.add_argument("location")
-    parser.add_argument("resource_group")
-    parser.add_argument("application_name")
+    parser.add_argument("resource_group", type=lower_case)
```
chkeita, line 1252: the resource group does not need the restriction only the application_name
```suggestion 
    parser.add_argument("resource_group")
```

chkeita (approved):

PR #2673, opened 2022-12-06
rename client_id in pool to object_id

chkeita:

Currently the Pool has an optional field for the `client_id` meant to be used in the unmanaged node scenario to identify the agent. It is associated with an application registration.

This PR replaces it with the `object id` of the service principal associated with that application registration. 
This gives us more flexibility giving us the ability to use a managed identity as the authentication method in the unmanaged case.

This PR also updates the authentication logic to also include the reason for rejecting a request. 
The agent logic is also updated to give the content of the response body when a request fails.

Note: The client_id can safely be removed from the pool record because it is not currently used


src/ApiService/ApiService/Functions/AgentRegistration.cs
```diff
@@ -70,9 +70,10 @@ public class AgentRegistration {
     }
 
     private async Async.Task<AgentRegistrationResponse> CreateRegistrationResponse(Service.Pool pool) {
-        var baseAddress = _context.Creds.GetInstanceUrl();
-        var eventsUrl = new Uri(baseAddress, "/api/agents/events");
-        var commandsUrl = new Uri(baseAddress, "/api/agents/commands");
+
+
+        var eventsUrl = new Uri(_context.ServiceConfiguration.OneFuzzBaseAddress, "/api/agents/events");
+        var commandsUrl = new Uri(_context.ServiceConfiguration.OneFuzzBaseAddress, "/api/agents/commands");
```
chkeita, line 76:
```suggestion 
```suggestion 
    var baseAddress = _context.Creds.GetInstanceUrl();
    var eventsUrl = new Uri(baseAddress, "/api/agents/events");
    var commandsUrl = new Uri(baseAddress, "/api/agents/commands");
```
```

File: src/ApiService/ApiService/onefuzzlib/EndpointAuthorization.cs [hunk 2]
```diff
@@ -57,20 +60,18 @@ public class EndpointAuthorization : IEndpointAuthorization {
...
 
-    public async Async.Task<bool> IsUser(UserAuthInfo tokenData) {
-        return !await IsAgent(tokenData);
-    }
 
-    public async Async.Task<HttpResponseData> Reject(HttpRequestData req, UserInfo token) {
+    public async Async.Task<HttpResponseData> Reject(HttpRequestData req, UserInfo token, String? reason = null) {
         var body = await req.ReadAsStringAsync();
-        _log.Error($"reject token. url:{req.Url:Tag:Url} token:{token:Tag:Token} body:{body:Tag:Body}");
+        _log.Error($"reject token. reason:{reason} url:{req.Url:Tag:Url} token:{token:Tag:Token} body:{body:Tag:Body}");
```
tevoinea, line 74: Would it be useful to include the reason why we're rejecting the token in the response?

tevoinea: Or is that too much information to expose?

src/ApiService/ApiService/onefuzzlib/PoolOperations.cs
```diff
@@ -87,8 +87,8 @@ public PoolOperations(ILogTracer log, IOnefuzzContext context)
         return await _context.Queue.QueueObject(GetPoolQueue(pool.PoolId), workSet, StorageType.Corpus);
     }
 
-    public IAsyncEnumerable<Pool> GetByClientId(Guid clientId) {
-        return QueryAsync(filter: $"client_id eq '{clientId}'");
+    public IAsyncEnumerable<Pool> GetByObjectId(Guid objectId) {
+        return QueryAsync(filter: $"object_id eq '{objectId}'");
```
tevoinea, line 91: Won't this need a column rename?

tevoinea (approved):
Just the column rename question

src/ApiService/ApiService/onefuzzlib/PoolOperations.cs
chkeita: no need, the `client_id` was not used before. in the worst case, we will have entities with both the client_id and the object_id

PR #2677, opened 2022-12-08
post JobCreated when job is created

stishkin:


Porges (approved):

chkeita (approved):

src/ApiService/ApiService/Functions/Jobs.cs
```diff
@@ -72,6 +72,8 @@ public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymou
         if (!r.IsOk) {
             _logTracer.WithTag("HttpRequest", "POST").WithHttpStatus(r.ErrorV).Error($"failed to insert job {job.JobId:Tag:JobId}");
```
Porges, line 73: We should return failure when this happens
```suggestion
            _logTracer.WithTag("HttpRequest", "POST").WithHttpStatus(r.ErrorV).Error($"failed to insert job {job.JobId:Tag:JobId}");
            return r.ErrorV;
```

src/ApiService/ApiService/Functions/Jobs.cs
```diff
@@ -72,6 +72,8 @@ public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymou
...
         }
+        await _context.Events.SendEvent(new EventJobCreated(job.JobId, job.Config, job.UserInfo));
```
chkeita, line 75: this should be sent only if the job was created

src/ApiService/ApiService/Functions/Jobs.cs
```diff
@@ -71,7 +71,16 @@ public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymou
         var r = await _context.JobOperations.Insert(job);
         if (!r.IsOk) {
             _logTracer.WithTag("HttpRequest", "POST").WithHttpStatus(r.ErrorV).Error($"failed to insert job {job.JobId:Tag:JobId}");
+            return await _context.RequestHandling.NotOk(
+                req,
+                new Error(
```
stishkin, line 76: @Porges  you mean like that ?
@chkeita  - yes it fails

Porges: @stishkin I think probably just `UNABLE_TO_CREATE`; there is no container setup here?

stishkin: @chkeita  - what do you mean ? job created should be sent when it failed to insert ?

chkeita: oh never mind i misunderstood the call

PR #2681, opened 2022-12-08
Remove unused coverage recorder from `input-tester`

### LINKED ISSUE #2680: Removed unused coverage dependency
On Windows, the `input-tester` crate supports recording coverage while testing inputs. This is unused, and will not be used in the future. Remove it to support improving coverage with fewer constraints.

AB#42550757

ranweiler:
Closes #2680.

### DIFF
--- a/src/agent/input-tester/src/crash_detector.rs
+++ b/src/agent/input-tester/src/crash_detector.rs
@@ -14,11 +14,7 @@ use std::{
 };
 
 use anyhow::Result;
-use coverage::{
-    block::{windows::Recorder as BlockCoverageRecorder, CommandBlockCov},
-    cache::ModuleCache,
-};
-use debugger::{BreakpointId, DebugEventHandler, Debugger, ModuleLoadInfo};
+use debugger::{DebugEventHandler, Debugger};
 use log::{debug, error, trace};
 use win_util::{
     pipe_handle::{pipe, PipeReaderNonBlocking},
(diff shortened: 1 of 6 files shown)


src/agent/input-tester/src/crash_detector.rs
```diff
@@ -272,44 +262,16 @@ impl<'a> DebugEventHandler for CrashDetectorEventHandler<'a> {
             debugger.quit_debugging();
         }
     }
-
-    fn on_create_process(&mut self, dbg: &mut Debugger, module: &ModuleLoadInfo) {
```
ranweiler, line 276: These callback hooks have no-op default impls.

src/agent/input-tester/src/crash_detector.rs
```diff
@@ -328,16 +290,13 @@ pub fn test_process<'a>(
         command.env(k, v);
     }
 
-    let filter = coverage::code::CmdFilter::default();
-    let recorder = cache.map(|c| BlockCoverageRecorder::new(c, filter));
```
ranweiler, line 332: We only created a coverage recorder if the `cache` was `None`.

src/agent/onefuzz/src/input_tester.rs
```diff
@@ -147,7 +147,6 @@ impl<'a> Tester<'a> {
             &env,
             self.timeout,
             IGNORE_FIRST_CHANCE_EXCEPTIONS,
-            None,
```
ranweiler, line 150: The task impls use the `input-tester` crate via this call site, which hardcodes the module cache to `None` (meaning we never used it to record coverage).

chkeita (approved):

Issue #2699, opened 2022-12-14
Improve source coverage of HTML reports


ranweiler:
HTML coverage reports are confusingly sparse. While we will always expect some reporting noise for optimized targets, this is exacerbated by our use of Sancov instrumentation sites as the only source of binary coverage. This is because in general, the instrumented sites will not include blocks that are redundant. To recover such block, we would need something like a postdominator analysis of the CFG. Even if no blocks were pruned, we have the same problem for instructions _within_ a block, which may include multiple source lines of code.

To fix this, we will only instrument based on our own CFG analysis. Then, when generating source coverage, we will "reinflate" blocks and look up source line coverage for every instruction in the block (not just the block entrypoint).

To enable this:
- [x] Write a new crate for easier, unified analysis of debuginfo and block recovery
- [x] Rewrite coverage analysis to use this crate
- [x] Factor out existing coverage-related file formats into new crate, adding versioning and simpler, normalized formats
- [x] Use the new coverage recording in the coverage tasks (with backwards-compatible report file deserialization)

AB#42609819

PR #2700, opened 2022-12-14
Add `debuggable-module`

ranweiler:
Part of #2699.


chkeita (approved):

src/agent/debuggable-module/src/block.rs
```diff
@@ -0,0 +1,286 @@
...
+        visited.insert(entry);
+
+        // Reset decoder for `entry`.
+        let position = (entry - offset.0) as usize;
+        decoder.set_position(position)?;
+        decoder.set_ip(entry);
+
+        // Decode instructions (starting from `entry`) until we reach a block
+        // terminator or run out of valid data.
+        while decoder.can_decode() {
```
ranweiler, line 76: We identify basic blocks through a two-pass sweep of the function. I could've abstracted each loop and expressed the passes with callbacks, but I think it'd be a lot harder to read.

PR #2701, opened 2022-12-14
Record coverage using `debuggable-module`

ranweiler:
Part of #2699.


src/agent/coverage/src/allowlist.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+            (true, true) => {
+                // Allow all.
+                true
+            }
+        }
+    }
+
+    fn has_allow_match(&self, path: &str) -> bool {
+        for re in &self.allow {
```
Porges, line 58: Suggest using [`RegexSet`](https://docs.rs/regex/latest/regex/struct.RegexSet.html) instead?

src/agent/coverage/src/binary.rs
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+            bail!("unknown coverage offset: {offset:x}");
+        };
+
+        Ok(())
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Count(pub u32);
```
Porges, line 36: Is this the _concept_ you mentioned? 🙂

ranweiler: One of them! 😄

src/agent/coverage/src/allowlist.rs
ranweiler: The only reason I didn't is because it felt wasteful to discard the `Regex` values we construct when parsing `AllowListLine`'s, and I wanted to avoid fallibility in the `RegexSet` ctor.

But on a second look, we _can_ easily access the raw regex with `as_str()`, and obtaining those strings (only) from a parsed `Regex` lets us safely `unwrap()` the ctor call.

src/agent/coverage/src/binary.rs
```diff
@@ -0,0 +1,96 @@
...
+    let exe = Object::parse(module.executable_data())?;
+    converter.process_object(&exe)?;
+    let di = Object::parse(module.debuginfo_data())?;
+    converter.process_object(&di)?;
+    converter.serialize(&mut std::io::Cursor::new(&mut symcache))?;
+    let symcache = SymCache::parse(&symcache)?;
+
+    let mut offsets = BTreeSet::new();
+
+    // If we wanted to apply an allowlist to function names, this is where we'd do it.
```
Porges, line 65: this comment implies (to me) that the functionality doesn't exist, but it does?

ranweiler: Ha, stale comment, good catch!

ranweiler: https://github.com/microsoft/onefuzz/pull/2701/commits/cbd6f117c2bb6926f14b97af280ccf44e150a576

src/agent/coverage/src/allowlist.rs
ranweiler: https://github.com/microsoft/onefuzz/pull/2701/commits/71bef6c1cea3650e33a8af83f86f460729205442

ranweiler: (adding tests for this)

src/agent/coverage/src/allowlist.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct AllowList {
+    allow: Option<RegexSet>,
+    deny: Option<RegexSet>,
+}
+
+impl AllowList {
+    pub fn new(allow: impl Into<Option<RegexSet>>, deny: impl Into<Option<RegexSet>>) -> Self {
```
tevoinea, line 35: Is there a difference between the impl trait argument and defining a generic T in the function?

I'm asking because the types for allow and deny are the same so they could be factored out together:

```rust
pub fn new<T: Into<Option<RegexSet>>>(allow: T, deny: T) -> Self {
```

ranweiler: There is a difference when there are multiple `impl Trait` arguments. Each `impl Trait` introduces a fresh type variable. So your example is not equivalent, because it requires that the arguments have the same concrete type at each call site.

This would prevent what we're trying to achieve, which is simulating optional arguments. As written, we could call `AllowList::new(regex_set, None)`, where one argument has the non-optional type `RegexSet`, but where the second argument has type `Option<RegexSet>`.

chkeita (approved):

src/agent/coverage/src/allowlist.rs
ranweiler: I ended up changing the signature anyway 😄.

Issue #2702, opened 2022-12-14
onefuzz-agent failing


mgreisen:
## Information

- Onefuzz version: 6.0.0
- OS: Windows

## Provide detailed reproduction steps (if any)

Tasks failing with the following error. 
------------------------------------------

task failed. exit_status:ExitStatus { Code = 1, Signal = , Success = False }
onefuzz-agent failed: C:\onefuzz\logs\04995d23-84cd-498d-804d-833df7467738-stdout.txt
[2022-12-10T13:24:34Z INFO  onefuzz_agent] loading supervisor agent config
[2022-12-10T13:24:34Z INFO  onefuzz_agent::config] loaded dynamic-config: C:\onefuzz\etc\dynamic-config.json
[2022-12-10T13:24:34Z INFO  onefuzz_agent::config] renewing registration
[2022-12-10T13:24:34Z INFO  onefuzz_agent::config] saved dynamic-config: C:\onefuzz\etc\dynamic-config.json
[2022-12-10T13:24:34Z INFO  onefuzz_agent::reboot] checking for saved reboot context: C:\onefuzz\reboot_context.json
[2022-12-10T13:24:34Z INFO  onefuzz_agent::reboot] loaded reboot context
[2022-12-10T13:24:34Z INFO  onefuzz_agent] running agent
[2022-12-10T13:24:35Z INFO  onefuzz_agent::worker] spawning ``onefuzz-task``; cwd = C:\onefuzz\63a2f8ea-b506-456f-97aa-3387b6482617, job_id = d97d637c-8bfc-4de0-b277-873d7e32a59a, task_id = 63a2f8ea-b506-456f-97aa-3387b6482617
[2022-12-10T13:24:36Z INFO  onefuzz_agent::worker] spawning ``onefuzz-task``; cwd = C:\onefuzz\77009b8d-36dd-4fdb-88ea-f335164332cc, job_id = d97d637c-8bfc-4de0-b277-873d7e32a59a, task_id = 77009b8d-36dd-4fdb-88ea-f335164332cc
[2022-12-10T13:24:36Z INFO  onefuzz_agent::worker] spawning ``onefuzz-task``; cwd = C:\onefuzz\7b9ecfae-2a1b-4fea-bdc7-75dcd8b0efe4, job_id = d97d637c-8bfc-4de0-b277-873d7e32a59a, task_id = 7b9ecfae-2a1b-4fea-bdc7-75dcd8b0efe4
[...]

PR #2710, opened 2022-12-16
Reduce mutation in the agent state machine

chkeita:
Removing the shared mutable state (&mut self) from the agent state machine. The intent is to make the code easier to reason with.

### DIFF
--- a/src/agent/onefuzz-agent/src/coordinator.rs
+++ b/src/agent/onefuzz-agent/src/coordinator.rs
@@ -166,15 +166,15 @@ impl_downcast!(ICoordinator);
 #[async_trait]
 impl ICoordinator for Coordinator {
     async fn poll_commands(&mut self) -> Result<Option<NodeCommand>, PollCommandError> {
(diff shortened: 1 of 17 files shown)


tevoinea (approved):

src/agent/onefuzz-agent/src/coordinator.rs
```diff
@@ -153,25 +156,25 @@ pub struct TaskInfo {
...
+    async fn can_schedule(&self, work: &WorkSet) -> Result<CanSchedule>;
 }
 
 impl_downcast!(ICoordinator);
 
 #[async_trait]
 impl ICoordinator for Coordinator {
     async fn poll_commands(&mut self) -> Result<Option<NodeCommand>, PollCommandError> {
-        self.poll_commands().await
+        Coordinator::poll_commands(self).await
```
chkeita, line 169: This might be related to #2702 the existing code seemed to create an infinite loop in some case

PR #2715, opened 2022-12-20
Add key for block coverage, tests

ranweiler:
Add a `blocks` subkey to the binary block coverage JSON schema, and test conversions. This will make it easier to add new per-module metadata in the future, if needed.

**Old:**
```json
{
    "/setup/main.exe": {
        "1": 0,
        "12c": 1,
        "1388": 0
    },
    "/setup/lib/some.dll": {
        "7b": 0,
        "1c8": 10
    }
}

```

**New:**
```json
{
    "/setup/main.exe": {
        "blocks": {
            "1": 0,
            "12c": 1,
            "1388": 0
        }
    },
    "/setup/lib/some.dll": {
        "blocks": {
            "7b": 0,
            "1c8": 10
        }
    }
}

```

### DIFF
--- a/src/agent/onefuzz-file-format/tests/integration.rs
+++ b/src/agent/onefuzz-file-format/tests/integration.rs
@@ -32,12 +32,12 @@ fn test_binary_coverage_formats() -> Result<()> {
     let expected = expected_binary_coverage()?;
 
     let v0_text = include_str!("files/binary-coverage.v0.json");
-    let v0_json: v0::BinaryCoverageJson = serde_json::from_str(&v0_text)?;
+    let v0_json: v0::BinaryCoverageJson = serde_json::from_str(v0_text)?;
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz-file-format/tests/integration.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    let mut binary = BinaryCoverage::default();
+    binary.modules.insert(some_dll_path, some_dll);
+    binary.modules.insert(main_exe_path, main_exe);
+
+    Ok(binary)
+}
+
+#[test]
+fn test_binary_coverage_formats() -> Result<()> {
```
Porges, line 31: This would be a good opportunity to use golden-file testing (e.g. `goldenfile`), but that can be done at a later stage.

Porges (approved):

PR #2718, opened 2022-12-21
Add `lines` key to v1 source coverage JSON

ranweiler:
- Mirror binary format by adding a `lines` subkey, enabling easier file metadata
- Serialize line numbers as decimal strings
- Integration test source coverage conversions


src/agent/onefuzz-file-format/tests/files/source-coverage.v1.json
```diff
@@ -0,0 +1,15 @@
+{
+    "src/bin/main.c": {
+        "lines": {
+            "4": 1,
+            "9": 0,
```
tevoinea, line 5: I like having the line numbers as keys since it guarantees only one entry per line. But I think having the value be an object allows more flexibility down the line if we want to bring column back.

What I'm suggesting:

```json
"lines": {
    "4": {
        "count": 1,
    }
}
```

I guess we also have some degree of control because these formats are already versioned but just a thought.

tevoinea (approved):

src/agent/onefuzz-file-format/tests/files/source-coverage.v1.json
ranweiler: Good to be thinking about. I was imagining that if we reintroduced columns in the future, it'd probably turn into something further nested for the sake of normalization, like `line -> column -> count`:
```json
"locations": {
  "4": {
    "1": 1,
    "3": 0,
    "6": 1
  }
}
```
or more descriptively:
```json
"lines": {
  "4": {
    "columns": {
      "1": 1,
      "3": 0,
      "6": 1
    }
  }
}
```

And yeah, in the presence of versioning, I suggest we just wait until we can actually provide columns often enough for it to matter.

PR #2719, opened 2022-12-21
Add deserialization compat functions

ranweiler:
- Use semver strings for coverage file versioning
- Use adjacent tagging for coverage file formats (`{ "version": "...", "coverage": { ... } }`)
- Add compat `deserialize()` functions to handle unversioned `v0`/0.1 files

### DIFF
--- a/src/agent/onefuzz-file-format/src/coverage/source.rs
+++ b/src/agent/onefuzz-file-format/src/coverage/source.rs
@@ -8,13 +8,27 @@ pub mod v0;
 pub mod v1;
 
 #[derive(Serialize, Deserialize)]
-#[serde(tag = "version")]
+#[serde(tag = "version", content = "coverage")]
 pub enum SourceCoverageJson {
-    #[serde(rename = "0")]
+    #[serde(rename = "0.1")]
     V0(v0::SourceCoverageJson),
 
-    #[serde(rename = "1")]
+    #[serde(rename = "1.0")]
     V1(v1::SourceCoverageJson),
+}
+
+impl SourceCoverageJson {
+    pub fn deserialize(text: &str) -> Result<Self> {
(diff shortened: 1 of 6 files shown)


src/agent/onefuzz-file-format/tests/files/binary-coverage.v1.json
```diff
@@ -1,15 +1,18 @@
...
-            "1": 0,
-            "12c": 1,
-            "1388": 0
-        }
-    },
-    "/setup/lib/some.dll": {
-        "blocks": {
-            "7b": 0,
-            "1c8": 10
+    "version": "1.0",
```
ranweiler, line 2: We intentionally do _not_ add outer `version`/`coverage` keys to the `v0`/legacy test files, because we will only ever deserialize old files (which are always untagged).

In contrast, all non-legacy schemas will be deserialized and deserialized via the appropriate `..CoverageJson` enum, so the test files should be tagged.

src/agent/onefuzz-file-format/src/coverage/source.rs
```diff
@@ -8,15 +8,29 @@ pub mod v0;
...
+    #[serde(rename = "0.1")]
     V0(v0::SourceCoverageJson),
 
-    #[serde(rename = "1")]
+    #[serde(rename = "1.0")]
     V1(v1::SourceCoverageJson),
 }
 
+impl SourceCoverageJson {
+    pub fn deserialize(text: &str) -> Result<Self> {
```
ranweiler, line 21: This is what we will use to load old and new files for further processing.

We will round-trip from old versions to new via the fallible conversions to the appropriate non-`Json` coverage type.

tevoinea (approved):

PR #2723, opened 2022-12-22
Use a custom `Output` type when recording coverage

ranweiler:
We can't actually always observe a recorded process exit using `Child::wait()`, so the `ExitStatus` of our `Output` needs to be optional. Create a less-presumptuous `Output` type of our own that lets us express these situations, and can be constructed in a platform-appropriate way.

### DIFF
--- a/src/agent/coverage/src/record/linux/debugger.rs
+++ b/src/agent/coverage/src/record/linux/debugger.rs
@@ -2,13 +2,16 @@
 // Licensed under the MIT License.
 
 use std::collections::BTreeMap;
-use std::process::{Command, Output};
+use std::io::Read;
+use std::process::Command;
 
 use anyhow::{bail, format_err, Result};
 use debuggable_module::path::FilePath;
 use debuggable_module::Address;
 use pete::{Ptracer, Restart, Signal, Stop, Tracee};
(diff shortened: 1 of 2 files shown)


src/agent/coverage/src/record/linux/debugger.rs
```diff
@@ -46,7 +49,30 @@ impl<'eh> Debugger<'eh> {
             return Err(err);
         }
 
-        let output = child.wait_with_output()?;
+        // Currently unavailable on Linux.
+        let status = None;
```
ranweiler, line 53: If we had the `wait(2)` status integer that `pete` internally parses, then we could construct a standard library `ExitStatus` using [`ExitStatusExt::from_raw()`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.from_raw).

This raw `wait(2)` status isn't currently exposed in the `Stop` enum. Rather than try to be clever and resynthesize one from the exited/signaled/core dumped info we _do_ get, just omit it for now.

tevoinea: We should create an issue in pete and link to it from here so we don't lose track of this

tevoinea (approved):

PR #2740, opened 2023-01-12
Add readonly_inputs to dotnet & dotnet_dll templates

### LINKED ISSUE #2739: dotnet_dll libfuzzer task does not contain readonly_inputs parameter
seems to be like this was missed accidentally.

Need to add support for readonly_inputs parameter to the CLI template

AB#42935474

Porges:
Closes #2739

### DIFF
--- a/src/cli/onefuzz/templates/libfuzzer.py
+++ b/src/cli/onefuzz/templates/libfuzzer.py
@@ -376,13 +376,11 @@ class Libfuzzer(Command):
         )
 
         if existing_inputs:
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -331,13 +331,11 @@ def basic(
         )
 
         if existing_inputs:
-            self.onefuzz.containers.get(existing_inputs)
```
Porges, line 334: These are already invoked above

PR #2741, opened 2023-01-12
Switch over to new `coverage` task

### LINKED ISSUE #2699: Improve source coverage of HTML reports

ranweiler:
- Switch to new `coverage` task, remove `coverage-legacy`
- Update backend and CLI to support new new `.._allowlist` task fields

These changes are backwards-compatible with the old actual coverage JSON file formats. The new `coverage` task knows the old (unversioned) format, can deserialize to modern in-memory values, and can serialize into the new versioned format. This means that existing coverage blob storage containers with old coverage JSON files won't break new tasks.

**Warning:** this PR introduces a _breaking change_ for users of the (optional) `coverage.coverage_filter` field. Instead, we have 3 separate optional allowlists (`function_allowlist`, `module_allowlist`, and `source_allowlist`) that govern coverage recording. However, the deprecated (and no longer user-exposed) server-side field definitions are retained for the sake of being able to parse old task details.

Closes #2699.


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -209,6 +209,9 @@ public record TaskDetails(
     List<string>? ReportList = null,
     long? MinimizedStackDepth = null,
     string? CoverageFilter = null,
```
ranweiler, line 211: Now deprecated and only for server-side back-compat. Will add a comment.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -932,6 +935,9 @@ Uri HeartbeatQueue
     public List<string>? ReportList { get; set; }
     public long? MinimizedStackDepth { get; set; }
     public string? CoverageFilter { get; set; }
```
ranweiler, line 937: Now deprecated and only for server-side back-compat. Will add a comment.

src/ApiService/ApiService/onefuzzlib/Defs.cs
```diff
@@ -10,8 +10,10 @@ public static class Defs {
                     TaskFeature.TargetEnv,
                     TaskFeature.TargetOptions,
                     TaskFeature.TargetTimeout,
-                    TaskFeature.CoverageFilter,
```
ranweiler, line 13: This should be kept and commented as deprecated.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -45,7 +47,9 @@ pub struct Config {
     pub target_options: Vec<String>,
     pub target_timeout: Option<u64>,
 
-    pub coverage_filter: Option<String>,
```
ranweiler, line 48: This deletion means that, on instance update, if there are old coverage tasks that are enqueued for a pool, they will fail when agents pick them up (because the agent isn't aware of this type).

We could add it as an unused, optional field to let the task run (but any existing `coverage_filter` will be ignored).

I believe that old tasks that do NOT have a coverage filter will be unimpacted.

chkeita: @ranweiler  Is this field currently used. 
I suggest that we keep it for this release and fail the task with a message saying that option is no longer supported. 
We can remove that option on the next release. By that time, any old tasks using this would have been drained from the queue.

ranweiler: Got it, so you're basically suggesting just failing more verbosely and specifically on the first release that contains the new task?

Then the first line of the task would be something like:
```rust
if config.coverage_filter.is_some() {
    bail!("the `coverage_filter` option for the `coverage` task is no longer supported");
}
```

We could also choose to ignore that field, run the task, and just _warn_ the user, but I'm don't think we can alert them reasonably.

ranweiler: I assume that we agree that it'd be a breaking change / major version bump either way; this is just to make it easier to understand why the old tasks don't get run.

chkeita: I agree. This will save the investigation time before we just tell the use to just resubmit their job.

chkeita (approved):

tevoinea (approved):

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -245,25 +235,30 @@ impl<'a> TaskContext<'a> {
...
+            CoverageRecorder::new(cmd)
+                .allowlist(allowlist)
+                .timeout(timeout)
+                .record()
         })
         .await??;
 
-        Ok(coverage)
+        if let Some(status) = recorded.output.status {
+            if !status.success() {
```
ranweiler, line 256: Offline suggestion from @Porges: check for error code that matches an unhandled debug break on Windows, yield a more-specific error to aid debugging.

ranweiler: This extra scrutiny is no longer necessary: such runtime faults are directly converted to to recording errors in #2772.

src/agent/onefuzz-task/src/tasks/coverage/generic/windows-interceptor.list
```diff
@@ -0,0 +1,16 @@
+# Required to avoid recording errors.
+! *\llvm-project\compiler-rt\*
+! *\vctools\crt\*
+! *\Windows Kits\10\include\*\ucrt\*
+! *\Windows Kits\10\Include\*\ucrt\*
```
ranweiler, line 5: Oops, repeat.

src/ApiService/ApiService/onefuzzlib/Config.cs
```diff
@@ -262,6 +262,24 @@ await _containers.GetContainerSasUrl(x.Item2.Name, StorageType.Corpus, ConvertPe
             }
         }
 
+        if (definition.Features.Contains(TaskFeature.FunctionAllowlist)) {
+            if (task.Config.Task.FunctionAllowlist != null) {
+                config.FunctionAllowlist = task.Config.Task.FunctionAllowlist;
+            }
```
Porges, line 268: tip: that this can be written

```suggestion
            if (task.Config.Task.FunctionAllowlist is string functionAllowList) {
                config.FunctionAllowlist = functionAllowlist;
            }
```

but don't bother updating all of these unless you really want 🙂

Porges (approved):

src/ApiService/ApiService/onefuzzlib/Config.cs
ranweiler: Will do a follow-up pass!

PR #2745, opened 2023-01-17
Wrap fallible ORM functions in try/catch

### LINKED ISSUE #2709: Wrap fallible ORM functions in `Try`/`Catch`
Some ORM operations are fallible. We have existing code that attempts to handle the failure modes using `.IsError` but it turns out that they can also throw exceptions. Here's an example:

https://github.com/microsoft/onefuzz/blob/349604f03adb9821add877cd0880798fcf328e18/src/ApiService/ApiService/onefuzzlib/orm/Orm.cs#L81-L92

That example can also throw `Azure.RequestFailedException`.

### Feature request

For all fallible ORM functions, wrap them in `try`/`catch` and return the appropriate `OneFuzzResult` type.

AB#42650423

tevoinea:

_What is this about?_

Closes #2709 

## Info on Pull Request

_What does this include?_

* Catch `RequestFailedException` in fallible ORM functions
* Update `BatchOperation` signature to better indicate the 'all-or-nothing' behavior - [more info](https://learn.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions#batch-response-syntax)

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/orm/Orm.cs
```diff
@@ -150,13 +162,32 @@ public IAsyncEnumerable<T> SearchByRowKeys(IEnumerable<string> rowKeys)
...
+                var tableClient = await GetTableClient(typeof(T).Name);
+                var transactions = await entities.Select(e => new TableTransactionAction(actionType, _entityConverter.ToTableEntity(e))).ToListAsync();
+                var responses = await tableClient.SubmitTransactionAsync(transactions);
+                var wrappingResponse = responses.GetRawResponse();
+                if (wrappingResponse.IsError) {
+                    return ResultVoid<(int, string, int?)>.Error((wrappingResponse.Status, wrappingResponse.ReasonPhrase, null));
+                }
+
+                var subTransactionFailures = responses.Value.Where(response => response.IsError);
+                if (subTransactionFailures.Any()) {
```
chkeita, line 176: isn't this redundant. My understanding is if the `wrappingResponse` is successful then all `subtransactions` are successful. 
I think this logic should be moved inside the check for `wrappingResponse.IsError` to give more detail about what failed in that batch

tevoinea: From the docs, it looks like the `wrappingResponse` can be successful while having `subtransactions` that are unsuccessful.

> Assuming that the batch request has been properly authorized and has been successfully received by the Table service, the batch request returns status code 202 (Accepted), even if one of the operations in the change set fails.

https://learn.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions#batch-response-syntax

So there's 3 failure modes that are being handled:

1. The `wrappingResponse` fails and throws a `RequestFailedException`
2. The `wrappingResponse` fails but doesn't throw
3. The `wrappingResponse` succeeds but one of the `subtransactions` fails

chkeita (approved):

PR #2746, opened 2023-01-17
Remove wcwidth hook

Porges:
This is breaking the build with the most recent version; the `version.json` file was removed upstream.


AdamL-Microsoft (approved):

Porges:
Now bundled into #2744

PR #2770, opened 2023-01-24
Include `AssignedTo` when failing to create a work item due to an authentication exception

tevoinea:

_What is this about?_

PATs in notification configs expire, when they do the operation throws a VssAuthenticationException. We then need to manually track down the owner of the config to update the PAT. This PR includes the `AssignedTo` field of the config if available, to help us find the owner of the config faster.


src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs
```diff
@@ -142,4 +143,8 @@ public NotificationOperations(ILogTracer log, IOnefuzzContext context)
         _logTracer.Error($"unable to find crash_report or no repro entry for report: {JsonSerializer.Serialize(report)}");
         return null;
     }
+
+    public async Async.Task<Notification> GetNotification(Guid notifificationId) {
```
chkeita, line 147: is this used ?

tevoinea: Forgot to include the 2nd part. Added an endpoint for querying a single notification by id.

Will follow up with CLI change so we can do `onefuzz notifications get <notification_id>`

chkeita (approved):

PR #2771, opened 2023-01-24
Config Refactor Round 2.

nharper285:

_What is this about?_
Currently, a number of parameters and arguments must be specified when a user configures their endpoint to an instance. 
ex. 
`onefuzz config --endpoint https://[endpoint].azurewebsites.net/ --authority https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47 --client_id 72f1562a-8c0c-41ea-beb9-fa2b71c80134`

These parameters - `authority`, `client_id`, `tenant_domain`, etc. - are used to authenticate to the instance when a user runs a CLI command. They are stored on the local machine, as well.

**Proposal**
Refactor the CLI config command such that it does not require these parameters and only asks the user to specify the `endpoint`. Instead, store the additional config params on the instance - to be set at deployment time - and pull them down every time a user makes a request.  

## PR Checklist
* [ ] Applies to work item: #2773, #2774, #2775

## Validation Steps Performed

* [x] Single-tenant deployment w/ ms tenant
* [x] Single-tenant deployment w/ secure tenant
* [x] Multi-tenant deployment 
* [x] Run command against ms tenant instance
* [x] Run command against secure tenant instance
* [x] Run command against multi-tenant instance
* [x] Run command w/ service prinicpal 
* [x] Deploy upgrade using existing service principal


src/ApiService/ApiService/onefuzzlib/ConfigOperations.cs
```diff
@@ -26,7 +26,7 @@ private sealed record InstanceConfigCacheKey();
     private static readonly InstanceConfigCacheKey _key = new(); // singleton key
     public Task<InstanceConfig> Fetch()
         => _cache.GetOrCreateAsync(_key, async entry => {
-            entry = entry.SetAbsoluteExpiration(TimeSpan.FromMinutes(10)); // cached for 10 minutes
+            entry = entry.SetAbsoluteExpiration(TimeSpan.FromMinutes(1)); // cached for 1 minute
```
stishkin, line 29: 10 minutes was wrong ? or you forgot to change back after testing ?

src/deployment/config.json
```diff
@@ -1,6 +1,15 @@
 {
+    "tenant_id": "72f988bf-86f1-41af-91ab-2d7cd011db47",
+    "tenant_domain": "azurewebsites.net",
+    "multi_tenant_domain": "",
+    "cli_client_id": "72f1562a-8c0c-41ea-beb9-fa2b71c80134",
     "proxy_nsg_config": {
-        "allowed_ips": ["*"],
+        "allowed_ips": [
```
stishkin, line 7: duplicate intended ?

src/deployment/deploy.py
```diff
@@ -305,7 +304,7 @@ def get_instance_url(self) -> str:
         # The url to access the instance
         # This also represents the legacy identifier_uris of the application
         # registration
-        if self.multi_tenant_domain:
+        if self.multi_tenant_domain != "":
```
stishkin, line 307: if not self.multi_tenant_domain:  <<-- will be true if not null and not empty

https://docs.python.org/3/library/stdtypes.html#truth-value-testing

stishkin, line 307: Previous code should be fine, or you hit an issue with it ? 
https://docs.python.org/3/library/stdtypes.html#truth-value-testing

src/deployment/deploylib/configuration.py
```diff
@@ -107,6 +112,59 @@ def parse_nsg_json(self, config: Any) -> None:
...
+    def parse_endpoint_json(self, config: Any) -> None:
+
+        if "cli_client_id" not in config:
+            raise Exception(
+                "CLI client_id not provided as valid key. Please Provide Valid Config."
+            )
+
+        if (
+            not isinstance(config["cli_client_id"], str)
+            or config["cli_client_id"] == ""
```
stishkin, line 124: probably client ID then should be also a valid GUID if you want to do validation here

src/deployment/deploylib/configuration.py
```diff
@@ -225,3 +283,38 @@ def update_nsg(
...
+
+
+def update_endpoint_params(
+    config_client: InstanceConfigClient,
+    authority: str,
+    client_id: str,
+    tenant_domain: str,
+) -> None:
+
+    config_client.table_service.insert_or_merge_entity(
```
stishkin, line 295: Does this go into Azure and updates config table directly in the storage account ?

src/ApiService/ApiService/onefuzzlib/ConfigOperations.cs
nharper285: No I'm changing it permanently. With the instance_config, we want the changes to be immediate. It doesn't make sense to update the instance_config and then wait for ten minutes to see the changes.

src/deployment/config.json
nharper285: Nope, good catch.

src/deployment/deploy.py
nharper285: No I want this param to be required in the config.json. This way, if an instance isn't multi-tenant, they can just leave it blank. So in this case self.multi_tenant_domain is always set to at least "". The previous check is therefore not sufficient.

src/deployment/deploylib/configuration.py
nharper285: Done.

src/deployment/deploylib/configuration.py
nharper285: Yes!

stishkin: I think this is a bad idea. I'd argue that  all of the storage updates should go through service. Never direct....

nharper285: Similar behavior has been in place since updating the NSG work - we upload the NSG rules directly. 

I don't disagree that there might be a better methodology, but I'm not sure what that'd be and it'd require changes for both this work and the NSG deployment code. Could be a follow-up PR to figure out another way to upload.

stishkin: I guess since there is existing behavior ....

stishkin: I'll open a work item

stishkin (approved):

src/ApiService/ApiService/Functions/Config.cs
```diff
@@ -0,0 +1,35 @@
...
+    private readonly IOnefuzzContext _context;
+
+    public Config(ILogTracer log, IOnefuzzContext context) {
+        _log = log;
+        _context = context;
+    }
+
+    [Function("Config")]
+    public Async.Task<HttpResponseData> Run(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "GET", Route = "config")] HttpRequestData req) {
```
mgreisen, line 18: >Route = "config" [](http://example.com/codeflow?start=58&length=16)

I don't think this is strictly needed as it's the same as the name of the function.

src/ApiService/ApiService/onefuzzlib/ConfigOperations.cs
mgreisen: By the comments I understand this to be the cached value. So we don't wait to see the changes, but we're deciding how long the cache is good for before invalidating it and going back to storage to fetch again.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -347,7 +347,10 @@ public record InstanceConfig
     IDictionary<Endpoint, ApiAccessRule>? ApiAccessRules = null,
     IDictionary<PrincipalId, GroupId[]>? GroupMembership = null,
     IDictionary<string, string>? VmTags = null,
-    IDictionary<string, string>? VmssTags = null
+    IDictionary<string, string>? VmssTags = null,
+    string? Authority = "",
```
mgreisen, line 351: > string? [](http://example.com/codeflow?start=3&length=8)

Do we need nullable types here? Don't we have required parameters?

src/deployment/config.json
```diff
@@ -1,6 +1,12 @@
 {
+    "tenant_id": "72f988bf-86f1-41af-91ab-2d7cd011db47",
```
mgreisen, line 2: I thought we were getting out of the business of hard coding these values.

nharper285: As a first step, I was just moving it out of the deploy script. This acts as an example tenant_id.

src/ApiService/ApiService/ServiceConfiguration.cs
```diff
@@ -26,7 +26,9 @@ public interface IServiceConfig {
 
     public string? DiagnosticsAzureBlobContainerSasUrl { get; }
     public string? DiagnosticsAzureBlobRetentionDays { get; }
-
+    public string? CliAppId { get; }
```
Porges, line 29: Let's use the full form here (and update the config setting name)

```suggestion
    public string? ClientId { get; }
```

src/cli/onefuzz/api.py
```diff
@@ -153,6 +158,42 @@ def _req_model_list(
...
+        self,
+    ) -> None:
+
+        endpoint_params = responses.Config(
+            authority="",
+            client_id="",
+            tenant_domain="",
+        )
+
+        if self.onefuzz._backend.config.endpoint is not None:
```
Porges, line 171: Just a stylistic thing but I'd invert the if to do the check and throw first, to reduce overall indentation

src/cli/onefuzz/api.py
```diff
@@ -153,6 +158,42 @@ def _req_model_list(
 
         return [model.parse_obj(x) for x in response]
 
+    def _req_config_params(
+        self,
+    ) -> None:
+
+        endpoint_params = responses.Config(
```
Porges, line 165: No need for this I think since we are overwriting this variable later?

src/cli/onefuzz/backend.py
```diff
@@ -321,6 +321,7 @@ def request(
         if not endpoint:
             raise Exception("endpoint not configured")
 
+        LOGGER.info(self.headers)
```
Porges, line 324: remove before merge?

src/cli/onefuzz/backend.py
```diff
@@ -181,7 +181,7 @@ def get_access_token(self) -> Any:
         if not self.config.endpoint:
             raise Exception("endpoint not configured")
 
-        if self.config.tenant_domain:
+        if "common" in self.config.authority:
```
Porges, line 184: I think we should probably check this precisely. I think we're testing for `https://login.microsoftonline.com/common` here?

Porges (approved):

src/ApiService/ApiService/ServiceConfiguration.cs
mgreisen, line 29: Do these need to be nullable?

src/ApiService/ApiService/Functions/Config.cs
nharper285: Removed.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
nharper285: Removed.

src/deployment/config.json
nharper285: I'd like to leave this in the first pass.

src/ApiService/ApiService/ServiceConfiguration.cs
nharper285: I'd like to keep the naming as is. In the deploy script, we already distinguish between a `client_id` and `cli_app_id`, who's uses are distinct.

src/cli/onefuzz/api.py
nharper285: Removed.

src/cli/onefuzz/api.py
nharper285: Done.

src/cli/onefuzz/backend.py
nharper285: Done.

src/ApiService/ApiService/ServiceConfiguration.cs
nharper285: Yup, just based-off how the service config is set up. All of the values are nullable, despite the fact that they are mostly all present all of the time.

src/cli/onefuzz/backend.py
nharper285: Done.

chkeita (approved):

src/ApiService/ApiService/UserCredentials.cs
```diff
@@ -92,7 +92,7 @@ public class UserCredentials : IUserCredentials {
                 } else {
                     var tenantsStr = allowedTenants.OkV is null ? "null" : String.Join(';', allowedTenants.OkV!);
                     _log.Error($"issuer not from allowed tenant. issuer: {token.Issuer:Tag:Issuer} - tenants: {tenantsStr:Tag:Tenants}");
-                    return OneFuzzResult<UserAuthInfo>.Error(ErrorCode.INVALID_REQUEST, new[] { "unauthorized AAD issuer" });
+                    return OneFuzzResult<UserAuthInfo>.Error(ErrorCode.INVALID_REQUEST, new[] { "unauthorized AAD issuer. if multi-tenant auth is failing, make sure to include all tenant_ids in the `allowed_aad_tenants` list in the instance_config. to see the current instance_config, run `onefuzz instance_config get`. " });
```
mgreisen, line 95:
```suggestion
                    return OneFuzzResult<UserAuthInfo>.Error(ErrorCode.INVALID_REQUEST, new[] { "unauthorized AAD issuer. If multi-tenant auth is failing, make sure to include all tenant_ids in the `allowed_aad_tenants` list in the instance_config. To see the current instance_config, run `onefuzz instance_config get`. " });
```

PR #2777, opened 2023-01-25
Add cobertura output to coverage example

Porges:
This makes it very easy to generate Cobertura reports.

Replace `--source` flag argument with `--output={mod-off, source, cobertura}` (default is still `mod-off`).


src/agent/coverage/Cargo.toml
```diff
@@ -26,3 +26,4 @@ procfs = { version = "0.12", default-features = false, features=["flate2"] }
 clap = { version = "4.0", features = ["derive"] }
 env_logger = "0.10.0"
 pretty_assertions = "1.3.0"
+onefuzz-file-format  = { path = "../onefuzz-file-format" }
```
ranweiler, line 29: Instead of taking on this OneFuzz-specific crate as a dependency, can we instead:

1. Factor the `cobertura` module out of `onefuzz-file-format`, into its own crate
2. Make both `onefuzz-file-format` and `coverage` depend on the new `cobertura` crate

It's my fault for not doing it that way in the first place, since Cobertura is _not_ a OneFuzz-specific file format.

src/agent/coverage/examples/coverage.rs
```diff
@@ -19,15 +20,22 @@ struct Args {
     #[arg(short, long)]
     timeout: Option<u64>,
 
-    #[arg(short, long)]
-    source: bool,
+    #[arg(short, long, value_enum, default_value_t = OutputFormat::ModOff)]
```
ranweiler, line 23: Nice!

src/agent/cobertura/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "cobertura"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+
+[dependencies]
+anyhow = "1.0"
+coverage = { path = "../coverage" }
```
ranweiler, line 9: Ah, so there's a circular dependency now. Can we break the cycle by removing this dependency on `coverage`, and instead move the impl for `From<SourceCoverage> for CoberturaCoverage` into the `coverage` crate?

ranweiler (approved):

PR #2781, opened 2023-01-27
Rust 1.67

Porges:
Mostly automated `clippy` fixes.

### DIFF
--- a/src/agent/reqwest-retry/Cargo.toml
+++ b/src/agent/reqwest-retry/Cargo.toml
@@ -2,7 +2,7 @@
 name = "reqwest-retry"
 version = "0.1.0"
 authors = ["fuzzing@microsoft.com"]
-edition = "2018"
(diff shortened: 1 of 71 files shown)


src/agent/reqwest-retry/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "reqwest-retry"
 version = "0.1.0"
 authors = ["fuzzing@microsoft.com"]
-edition = "2018"
```
Porges, line 5: This change to fix a lint about a 2018→21 breaking change.

`reqwest-retry` contained the incorrect code:

```rs
assert!(resp.is_err(), "{resp:?}");
```

The second argument is _not_ formatted, but would be in Rust 2021. Simple fix: change to 2021.

chkeita (approved):

PR #2785, opened 2023-01-27
Apply allowlist to all blocks within a function

ranweiler:
Before this change, we only applied source allowlists to a function's entry offset. However, inlined blocks could come from other functions, including some that would be filtered out by deny rules. As a result, the allowlist could incompletely exclude coverage recording for such inlinees.

Instead, apply the source allowlist to each block entrypoint. This assumes that all instructions within a basic block come from the same function. The result is that block-level expansion and source-line mapping correctly inherits allowlist rule application at the binary level.

### DIFF
--- a/src/agent/coverage/src/binary.rs
+++ b/src/agent/coverage/src/binary.rs
@@ -131,12 +131,20 @@ pub fn find_coverage_sites(
                 if let Some(file) = location.file() {
                     let path = file.full_path();
 
-                    // Apply allowlists per block to account for inlining.
-                    if allowlist.source_files.is_allowed(&path) {
-                        offsets.insert(block.offset);
+                    // Apply allowlists per block, to account for inlining. The `location` values
+                    // here describe the top of the inline-inclusive call stack.
+                    if !allowlist.functions.is_allowed(&path) {
+                        continue;
                     }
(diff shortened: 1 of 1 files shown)


tevoinea (approved):

src/agent/coverage/src/binary.rs
```diff
@@ -124,14 +124,16 @@ pub fn find_coverage_sites(
...
-                        block::sweep_region(module, &debuginfo, function.offset, function.size)?;
-                    offsets.extend(blocks.iter().map(|b| b.offset));
+        let blocks = block::sweep_region(module, &debuginfo, function.offset, function.size)?;
+
+        for block in &blocks {
+            if let Some(location) = symcache.lookup(block.offset.0).next() {
+                if let Some(file) = location.file() {
+                    let path = file.full_path();
+
+                    if allowlist.source_files.is_allowed(&path) {
```
ranweiler, line 134: Add a comment, and apply the function allowlist here, too.

This may result in reporting artifacts around inlined calls (the `call` lines may appear unreachable). However, we don't have easy access to the full inline stack, so this is best for now (to avoid the worse artifact of allowlists "not working").

PR #2787, opened 2023-01-30
Scriban to jinja migration script

### LINKED ISSUE #2766: Scriban migration script
Soon, we will be enforcing accepting only scriban templates. Before that change we need to do one final migration of jinja templates to scriban and persist those modifications to the database. To accomplish that, we need a 1 time migration script that will iterate over every notification config, migrate the templates to scriban and store the changes.

AB#43087141

tevoinea:

_What is this about?_

Adds a database migration script for notification configs from jinja to scriban.

Closes #2766 

## PR Checklist
* [x] Tests added/passed

## Info on Pull Request

_What does this include?_

* Creates a new endpoint, when called will execute the migration script over the database
* The endpoint is protected by requiring admin privileges to execute.
* Will create a follow up PR with a CLI command for calling the endpoint
* You can specify a `dryRun` parameter which will only return the list of notifications which _would_ be updated

## Validation Steps Performed

I added the following integration tests:
* Executing with `dryRun=True` does not make any changes to DB
* The default executing mode is `dryRun=False` and changes are persisted
* All necessary fields are migrated when migrating Ado templates
* All necessary fields are migrated when migrating Github templates
* Teams templates are ignores as they do not required migration
* Multiple notification configs can be migrated
* Admin is required to execute the endpoint


Porges (approved):
Looks good, nice work on the tests 👍

PR #2793, opened 2023-02-02
Capture crash dumps from libfuzzer, when provided 

Porges:
Enables capturing crashdumps generated by ASAN at point of failure.

This helps in several ways:
- provides a crash dump in the case that we cannot repro a failure later
- some people have stated that crash dumps would be more useful to their team than the repro VM
- we should be able to use these for automated submission to Watson or similar

---

Crash dumps are automatically collected (for libfuzzer) jobs, if we find any. They should be activated by enabling crash dumps in ASAN, via:

- On Linux: `ASAN_OPTIONS=disable_coredump=0:abort_on_error=1:unmap_shadow_on_exit=1`
  - OneFuzz will override the (Ubuntu) crash dump options to generate core dumps instead and then upload them.
- On Windows: `ASAN_SAVE_DUMPS=my_dump.dmp`
  - OneFuzz will look for any `*.dmp` files in the working directory and then upload them.

In both cases, the crash dump will be renamed to match the crashing input, if possible, and uploaded to a new `crashdumps` container.

---

Also updated: the “simple” LibFuzzer test has been updated to be compiled with `cl.exe` instead of `clang` on Windows, so that we are exercising the MSVC implementation of ASAN/LibFuzzer, and the CI image has been updated to `windows-2022`. The restriction to an old version of the Windows SDK has been removed.


src/runtime-tools/linux/setup.sh
```diff
@@ -159,6 +159,12 @@ if type apt > /dev/null 2> /dev/null; then
     popd
 fi
 
+# enable crash dumps; by default on Ubuntu this is done by apport
+# so we need to disable that
+sudo sysctl -w 'kernel.core_pattern=core' # write dumps to file named 'core'
```
chkeita, line 164: this is equivalent to what is done in run.sh right ?
``` bash 
    # use core files, not external crash handler
    echo core | sudo tee /proc/sys/kernel/core_pattern
```

Also, this won't work inside a docker container. 
If you want to keep it in this file you can check for `DOCKER_BUILD` before executing this

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
```diff
@@ -291,28 +328,99 @@ where
...
+        } else {
+            warn!("no PID found for libfuzzer process");
+        }
+
+        // check for crash dumps on Windows:
+        #[cfg(target_os = "windows")]
+        {
+            let dumpfile_extension = Some(std::ffi::OsStr::new("dmp"));
+
+            let mut working_dir = tokio::fs::read_dir(".").await?;
```
tevoinea, line 387: Is this `.` because we're running the libfuzzer target from `.` (because we don't specify a working directory when running the command)?

tevoinea: Also, we should log a warning like we do in the linux impl if we don't find any dump files?

tevoinea (approved):

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
Porges: > Is this `.` because we're running the libfuzzer target from `.` (because we don't specify a working directory when running the command)?

Yes, that's right. The dump will appear in the current working directory.

src/integration-tests/integration-test.py
```diff
@@ -106,12 +107,21 @@ class Integration(BaseModel):
             ContainerType.unique_reports: 1,
             ContainerType.coverage: 1,
             ContainerType.inputs: 2,
+            # TODO: crashdumps are intermittently not captured
```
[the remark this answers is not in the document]
Porges: I've thought about this as well, but I think that writing the dump is synchronous: the ASAN runtime does it before the process exits. Dirwatcher is a possible solution but it would be hard to match up dumps and crashes, at the moment we make the assumption that if we see a dump then it matches the (single) crash that was just found, so we can rename them to match. If we're finding the dumps later on we don't know what crash to associate them with.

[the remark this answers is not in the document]
tevoinea: Ah that's a good point, if there were a way for us to control where the dump gets dropped, that could be an option. If we knew if it's placed in the working directory, or always adjacent to the location of the binary, we could create a folder per libfuzzer invocation (and a DirWatcher per folder) so we know which dump is associated with which crash.

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
Porges: Added the warning, reduced both of the log messages to info.

src/integration-tests/integration-test.py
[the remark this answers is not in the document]
Porges: I think I'll open a separate workitem for this. I think running the fuzzer in a temporary folder each time is probably a good idea even in general, since it means that we start with a fresh slate each time.

PR #2811, opened 2023-02-07
Resolving bug in config refactor.

nharper285:

_What is this about?_
A bug was introduced with the recent config refactor. `list` commands were not working properly. Resolving with this PR. 
## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -346,7 +346,12 @@ class Backend:
 
         url = endpoint + "/api/" + path
 
-        self.config_params()
+        if (
+            self.config.client_id == ""
+            and self.config.authority == ""
+            and self.config.tenant_domain == ""
+        ):
+            self.config_params()
         headers = self.headers()
         json_data = serialize(json_data)
 
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -112,7 +112,7 @@ class Backend:
         self.client_secret = client_secret
         self.config = config
         self.token_cache: Optional[msal.SerializableTokenCache] = None
-        # self.init_cache()
+        self.init_cache()
         self.app: Optional[msal.ClientApplication] = None
         self.token_expires = 0
         self.load_config()
--- a/src/cli/onefuzz/backend.py
+++ b/src/cli/onefuzz/backend.py
@@ -25,13 +25,13 @@ from typing import (
     TypeVar,
     cast,
 )
-from onefuzztypes import responses
 from urllib.parse import urlparse, urlunparse
 from uuid import UUID
 
 import msal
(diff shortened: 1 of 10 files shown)


src/cli/onefuzz/backend.py
```diff
@@ -322,6 +345,8 @@ def request(
             raise Exception("endpoint not configured")
 
         url = endpoint + "/api/" + path
+
+        self.config_params()
```
chkeita, line 349: Why do we update the config on each request ? 
It does not have to be done in this PR but if the config is already set there is no need fetch the config again from the server

nharper285: The goal is to not store the config values locally.

chkeita (approved):

src/cli/onefuzz/backend.py
Porges: Let's have `config_params` check  if it has the local values first before issuing the request.

nharper285: I think this is a good idea. I'll get that going.

src/cli/onefuzz/backend.py
```diff
@@ -322,6 +345,13 @@ def request(
             raise Exception("endpoint not configured")
 
         url = endpoint + "/api/" + path
+
+        if (
```
nharper285, line 349: @chkeita @Porges

Porges: I think checking client_id and authority is enough, tenant_domain isn't always required?

nharper285: Actually it is now. `tenant_domain` will always be set and used.

chkeita: if all of these value are required i would use `or` instead of `and` here

src/cli/onefuzz/backend.py
chkeita: > Let's have `config_params` check if it has the local values first before issuing the request, if it has all of them there's no need to go fetch them from the server. This makes it okay to invoke `config_params` on each request; the first time it will fetch the values if needed and then they will be used for the rest of the lifetime of the current CLI command.

i think this is good first step but ultimatelyy i think we still need to store the config in a file to avoid having to fetch the same config for every commands. This will be use full for apps that uses the cli for scripting.

src/cli/onefuzz/backend.py
nharper285: It actually needs to be the following: 
`if self.config.client_id == "" or (
            self.config.authority == "" and self.config.tenant_domain == ""
        ):`
 
 This allows service principals to be used when the user specifies a `client_id`.

PR #2831, opened 2023-02-14
Fix coverage example build

ranweiler:
The `coverage` crate does not build on its own unless the `namedpipeapi` feature from `winapi` is enabled.

In the larger OneFuzz workspace, this feature is enabled for `winapi` elsewhere, which is why the build generally is not broken.

A note: `winapi` hasn't had a commit since November 2021.

### DIFF
--- a/src/agent/debugger/Cargo.toml
+++ b/src/agent/debugger/Cargo.toml
@@ -23,6 +23,7 @@ features = [
     "debugapi",
     "handleapi",
     "memoryapi",
+    "namedpipeapi",
     "processthreadsapi",
     "securitybaseapi",
     "shellapi",


tevoinea (approved):

tevoinea:
> A note: winapi hasn't had a commit since November 2021.

Do you think it'd be worth looking into moving to [`windows`](https://crates.io/crates/windows)?

ranweiler:
> > A note: winapi hasn't had a commit since November 2021.
> 
> Do you think it'd be worth looking into moving to [`windows`](https://crates.io/crates/windows)?

Definitely. It's not urgent, but I'll need to do a port of the debugger backend eventually.

PR #2832, opened 2023-02-14
Defer coverage breakpoints

ranweiler:
For each loaded module, defer setting coverage breakpoints until exit from certain known initialization functions (when present). Crucially, this lets us avoid breaking Address Sanitizer interceptor initialization, which will otherwise divergently fail execution by [raising its own debug break exception](https://github.com/llvm/llvm-project/blob/llvmorg-14.0.6/compiler-rt/lib/interception/interception_win.cpp#L655-L660).

With this PR, neither OneFuzz itself nor locally-debugging end users will need to use an allowlist to exclude ASan-intercepted functions from coverage recording.

This PR also fixes a bug in breakpoint setting, which would would prevent setting breakpoints whose offset collided with any other breakpoint's offset (_ignoring_ base module). This would manifest in false negative coverage counts for colliding breakpoints.

### DIFF
--- a/src/agent/coverage/src/record/windows.rs
+++ b/src/agent/coverage/src/record/windows.rs
@@ -83,7 +83,7 @@ impl<'data> WindowsRecorder<'data> {
                     let thread_id = dbg.get_current_thread_id();
                     let state = DeferralState::PendingReturn { thread_id };
                     self.deferred_breakpoints.insert(id, (trigger, state));
-                },
+                }
                 DeferralState::PendingReturn { thread_id } => {
                     if dbg.get_current_thread_id() == thread_id {
                         // We've returned from the trigger function, and on the same thread.
@@ -99,15 +99,13 @@ impl<'data> WindowsRecorder<'data> {
                         let id = trigger.set(dbg)?;
                         self.deferred_breakpoints.insert(id, (trigger, state));
                     }
-                },
+                }
             }
 
             return Ok(());
         }
 
-        let breakpoint = self
-            .breakpoints
-            .remove(id);
+        let breakpoint = self.breakpoints.remove(id);
 
         let Some(breakpoint) = breakpoint else {
             let stack = dbg.get_current_stack()?;
@@ -153,7 +151,11 @@ impl<'data> WindowsRecorder<'data> {
         Ok(())
     }
 
-    fn set_or_defer_module_breakpoints(&mut self, dbg: &mut Debugger, path: FilePath) -> Result<()> {
+    fn set_or_defer_module_breakpoints(
+        &mut self,
+        dbg: &mut Debugger,
+        path: FilePath,
+    ) -> Result<()> {
         let (_module, debuginfo) = &self.modules[&path];
 
         // For borrowck.
@@ -177,7 +179,10 @@ impl<'data> WindowsRecorder<'data> {
             debug!("deferring coverage breakpoints for module {}", path);
             self.defer_module_breakpoints(dbg, path, trigger)
         } else {
-            debug!("immediately setting coverage breakpoints for module {}", path);
+            debug!(
+                "immediately setting coverage breakpoints for module {}",
+                path
+            );
             self.set_module_breakpoints(dbg, path)
         }
     }
@@ -192,7 +197,8 @@ impl<'data> WindowsRecorder<'data> {
         let entry_breakpoint = Breakpoint::new(path, trigger.offset);
         let id = entry_breakpoint.set(dbg)?;
 
-        self.deferred_breakpoints.insert(id, (entry_breakpoint, state));
+        self.deferred_breakpoints
+            .insert(id, (entry_breakpoint, state));
 
         Ok(())
     }
--- a/src/agent/coverage/src/record/windows.rs
+++ b/src/agent/coverage/src/record/windows.rs
@@ -174,6 +174,12 @@ impl Breakpoint {
     pub fn new(module: FilePath, offset: Offset) -> Self {
         Self { module, offset }
     }
+
+    pub fn set(&self, dbg: &mut Debugger) -> Result<BreakpointId> {
+        let name = Path::new(self.module.base_name());
+        let id = dbg.new_rva_breakpoint(name, self.offset.0, BreakpointType::OneTime)?;
+        Ok(id)
+    }
 }
 
 impl<'data> DebugEventHandler for WindowsRecorder<'data> {
--- a/src/agent/coverage/src/record/windows.rs
+++ b/src/agent/coverage/src/record/windows.rs
@@ -166,8 +166,8 @@ impl Breakpoints {
 
 #[derive(Clone, Debug)]
 struct Breakpoint {
-    module: FilePath,
-    offset: Offset,
+    pub module: FilePath,
+    pub offset: Offset,
 }
 
 impl Breakpoint {
--- a/src/agent/coverage/src/record/windows.rs
+++ b/src/agent/coverage/src/record/windows.rs
@@ -110,7 +110,8 @@ impl<'data> WindowsRecorder<'data> {
             self.breakpoints.set(dbg, breakpoint)?;
         }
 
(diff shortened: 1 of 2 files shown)


src/agent/coverage/src/record/windows.rs
```diff
@@ -66,11 +70,47 @@ impl<'data> WindowsRecorder<'data> {
...
+                        //
+                        // It's safe to set coverage breakpoints.
+                        self.set_module_breakpoints(dbg, trigger.module)?;
+                    } else {
+                        // Hit a ret breakpoint, but on the wrong thread. Reset it so the correct
+                        // thread has a chance to see it.
+                        //
+                        // We only defer breakpoints in image initialization code, so we don't
+                        // expect to reach this code in practice.
+                        let id = trigger.set(dbg)?;
```
ranweiler, line 99: Add a warning here.

src/agent/coverage/src/record/windows.rs
```diff
@@ -66,11 +70,47 @@ impl<'data> WindowsRecorder<'data> {
...
-    fn try_on_breakpoint(&mut self, _dbg: &mut Debugger, id: BreakpointId) -> Result<()> {
-        let breakpoint = self
-            .breakpoints
-            .remove(id)
-            .ok_or_else(|| anyhow!("stopped on dangling breakpoint"))?;
+    fn try_on_breakpoint(&mut self, dbg: &mut Debugger, id: BreakpointId) -> Result<()> {
+        if let Some((trigger, state)) = self.deferred_breakpoints.remove(&id) {
+            match state {
+                DeferralState::NotEntered => {
+                    // Find the return address.
```
ranweiler, line 77: Add some debug logging.

src/agent/coverage/src/record/windows.rs
```diff
@@ -103,72 +143,121 @@ impl<'data> WindowsRecorder<'data> {
...
+        path: FilePath,
+    ) -> Result<()> {
+        let (_module, debuginfo) = &self.modules[&path];
+
+        // For borrowck.
+        let mut trigger = None;
+
+        for function in debuginfo.functions() {
+            // Called on process startup.
+            if function.name.contains("__asan::AsanInitInternal") {
```
ranweiler, line 166: Add constants for these.

ranweiler: And use `starts_with()`.

src/agent/coverage/src/record/windows.rs
```diff
@@ -103,72 +143,121 @@ impl<'data> WindowsRecorder<'data> {
...
+        dbg: &mut Debugger,
+        path: FilePath,
+        trigger: Function,
+    ) -> Result<()> {
+        let state = DeferralState::NotEntered;
+        let entry_breakpoint = Breakpoint::new(path, trigger.offset);
+        let id = entry_breakpoint.set(dbg)?;
+
+        self.deferred_breakpoints
+            .insert(id, (entry_breakpoint, state));
```
ranweiler, line 201: This rustfmt result is very silly.

src/agent/coverage/src/record/windows.rs
```diff
@@ -103,72 +143,121 @@ impl<'data> WindowsRecorder<'data> {
...
+    pub offset: Offset,
 }
 
 impl Breakpoint {
     pub fn new(module: FilePath, offset: Offset) -> Self {
         Self { module, offset }
     }
+
+    pub fn set(&self, dbg: &mut Debugger) -> Result<BreakpointId> {
+        let name = Path::new(self.module.base_name());
```
ranweiler, line 257: Copy over the previous comment about truly needing to use module _base name_, not path or file name.

src/agent/coverage/src/record/windows.rs
```diff
@@ -103,72 +143,121 @@ impl<'data> WindowsRecorder<'data> {
...
-        self.offset_to_breakpoint
-            .insert(breakpoint.offset, breakpoint);
-
+        let id = breakpoint.set(dbg)?;
+        self.id_to_breakpoint.insert(id, breakpoint);
         Ok(())
     }
 
-    pub fn is_set(&self, breakpoint: &Breakpoint) -> bool {
-        self.offset_to_breakpoint.contains_key(&breakpoint.offset)
```
ranweiler, line 153: Bug: this check ignores modules. So if we had a breakpoint set in `foo+123`, this would return `true` for `bar+123`, even if no breakpoint was set.

src/agent/coverage/src/record/windows.rs
```diff
@@ -103,72 +143,121 @@ impl<'data> WindowsRecorder<'data> {
...
 #[derive(Debug, Default)]
 struct Breakpoints {
-    id_to_offset: BTreeMap<BreakpointId, Offset>,
-    offset_to_breakpoint: BTreeMap<Offset, Breakpoint>,
+    id_to_breakpoint: BTreeMap<BreakpointId, Breakpoint>,
 }
 
 impl Breakpoints {
     pub fn set(&mut self, dbg: &mut Debugger, breakpoint: Breakpoint) -> Result<()> {
-        if self.is_set(&breakpoint) {
```
ranweiler, line 129: Bug: since `is_set()` ignored modules, if we had a breakpoint set in `foo+123`, we would not set a breakpoint for `bar+123`.

The redundancy check is actually unnecessary wrt the underlying debugger engine (it handles overlapping breakpoints just fine), and our current coverage counting is idempotent (max hit count is 1). To avoid future double-counting issues, we should add a corrected redundancy check anyway. It'd be reasonable to just do this at the module level.

src/agent/coverage/src/record/windows.rs
ranweiler: 2fa4f9331fd819fa045818e811b679fb63e12a3c

src/agent/coverage/src/record/windows.rs
ranweiler: 2fa4f9331fd819fa045818e811b679fb63e12a3c

src/agent/coverage/src/record/windows.rs
```diff
@@ -66,11 +70,59 @@ impl<'data> WindowsRecorder<'data> {
...
+            }
+
+            return Ok(());
+        }
+
+        let breakpoint = self.breakpoints.remove(id);
+
+        let Some(breakpoint) = breakpoint else {
+            let stack = dbg.get_current_stack()?;
+            bail!("stopped on dangling breakpoint, debuggee stack:\n{}", stack);
```
ranweiler, line 124: I added the call stack to the error here so make it easier to debug / identify the interception bug if it pops up again.

Interceptor init-breaking errors will look something like:
```
Error: stopped on dangling breakpoint, debuggee stack:
fuzz!__interception::InterceptionFailed+0x0
fuzz!__interception::GetInstructionSize+0x5d7
fuzz!__interception::OverrideFunctionWithHotPatch+0x22
fuzz!__interception::OverrideFunction+0x27
fuzz!InitializeCommonInterceptors+0x1a
fuzz!__asan::InitializeAsanInterceptors+0x32
fuzz!__asan::AsanInitInternal+0x14e
ntdll!RtlActivateActivationContextUnsafeFast+0x11d
ntdll!RtlActivateActivationContextUnsafeFast+0x1ff
ntdll!LdrInitShimEngineDynamic+0x3b88
ntdll!LdrInitializeThunk+0x1db
ntdll!LdrInitializeThunk+0x63
ntdll!LdrInitializeThunk+0xe
```

src/agent/coverage/src/record/windows.rs
ranweiler: 1a45c51edb2f025d033b5bc2b9907478d5ec87a0

src/agent/coverage/src/record/windows.rs
ranweiler: Added a module-level guard: c69edb9ecb9f7a8b6749f57eb2f94de5a1d51147

chkeita (approved):

Issue #2773, opened 2023-01-25
CLI Config Refactor


nharper285:
Currently, a number of parameters and arguments must be specified when a user configures their endpoint to an instance. 
ex. 
`onefuzz config --endpoint https://[endpoint].azurewebsites.net/ --authority https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47 --client_id 72f1562a-8c0c-41ea-beb9-fa2b71c80134`

These parameters - `authority`, `client_id`, `tenant_domain`, etc. - are used to authenticate to the instance when a user runs a CLI command. They are stored on the local machine, as well.

**Proposal**
Refactor the CLI config command such that it does not require these parameters and only asks the user to specify the `endpoint`. Instead, store the additional config params on the instance - to be set at deployment time - and pull them down every time a user makes a request.  

AB#43100314

PR #2835, opened 2023-02-14
Config Refactor Part 2 - Change Opt Param Names & Set File Expiry

nharper285:

_What is this about?_
This is the final step of #2773. In this PR, we change the names of the optional `config` params. This will cause a breaking change for the CLI. 
We have also added an expiry the local config value cache. Now, if a file is older than 24hrs, we delete and reset. 

NOTE: For SP use, users will need to update their `--client_id` flag to `--override_client_id`

## PR Checklist
* [ ] Applies to work item: #2773 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


chkeita (approved):

mgreisen (approved):
:shipit:

PR #2847, opened 2023-02-16
Adding extra container to tasks

### LINKED ISSUE #2827: Allow extra container to be used in tasks
Currently tasks have access to a limited set of containers. Each of these containers has a specific purpose. This makes it hard for users to customize the input to their task if they need additional info.

This proposal if for adding an additional parameter to each task for a `extra` input container, and a way to access it via command line injection

AB#43318315

chkeita:

- Adds an `extra` readonly container to all tasks
- The container will be synced by the agent, same as the `setup_dir`
- Adds a new variable substitution to reference it 
- Updates the documentation to add the new `{extra}` variable substitution
- Updated some of the integration test to reference the `extra` container. The content is ignored but it allows us to test the variable substitution.

closes #2827


src/agent/onefuzz-task/src/tasks/report/generic.rs
```diff
@@ -201,13 +204,16 @@ impl<'a> GenericReportProcessor<'a> {
             try_resolve_setup_relative_path(&self.config.common.setup_dir, &self.config.target_exe)
                 .await?;
 
+        //let extra_dir = self.config.common.extra_dir.map(|x| x.as_path());
```
tevoinea, line 207:
```suggestion
```

tevoinea (approved):

PR #2856, opened 2023-02-17
Update clap to remove suppressions

### LINKED ISSUE #2277: Replace uses of `ansi_term`
The `ansi_term` crate is [unmaintained](https://github.com/ogham/rust-ansi-term/issues/72), resulting in a [RUSTSEC advisory](https://rustsec.org/advisories/RUSTSEC-2021-0139.html) that we must ignore.

Look for maintained alternatives, once they're ready, as suggested in the linked `ansi_term` issue.

AB#40995758

Porges:
A Friday afternoon jaunt.

To be merged after #2855; does the remaining work to close #2277 (and closes off some things that Dependabot was trying to upgrade).

Tested by manually running the commands; we don't have good coverage for this kind of stuff. OTOH, most of these commands are for the experimental local fuzzing mode, which is not fully supported yet. I did specifically test the `onefuzz-task managed` command which is the one used in production.

## Details

- Bump `clap` to 4.1.6
  - Remove `structopt` as this is subsumed by clap now
- Bump `envlogger` to 0.10 (removes problematic dependency)
- Set `default-features=false` on `proc-maps` (removes a feature which is only needed to support FreeBSD), and bump it to 0.3

The main changes migrating `clap` are:

- `value_t!` is gone; now use `matches.get_one::<T>`. If `T` is not `String` then a parser must have been registered on the `Arg` when it was created, with `arg.value_parser(value_parser!(T))`.
- `Command::with_name` and `Arg::with_name` are now called `new`.
- `Command` and `Subcommand` were unified, and `App` is removed.
- `arg.takes_value(true)` is gone; it is the default. For flags use `arg.action(ArgAction::SetTrue)` and then retrieve the flag value with `matches.get_flag`.

This code would be simplified a lot by using the `clap::Parser` on structs, but that requires reworking the code significantly as we cannot dynamically add/remove arguments the way that this is currently done.

## Also found

Found one bug while manually testing the `onefuzz-task local` commands; see comment below.

### DIFF
--- a/src/agent/onefuzz-task/src/local/libfuzzer_regression.rs
+++ b/src/agent/onefuzz-task/src/local/libfuzzer_regression.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use std::path::PathBuf;
+
 use crate::{
     local::common::{
         build_local_context, get_cmd_arg, get_cmd_env, get_cmd_exe, get_synced_dir, CmdType,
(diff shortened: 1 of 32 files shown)


src/agent/onefuzz-task/src/local/libfuzzer_regression.rs
```diff
@@ -73,75 +76,63 @@ pub fn build_regression_config(
...
     args
 }
 
-pub fn args(name: &'static str) -> App<'static, 'static> {
-    SubCommand::with_name(name)
+pub fn args(name: &'static str) -> Command {
+    Command::new(name)
         .about("execute a local-only libfuzzer regression task")
-        .args(&build_shared_args(false))
+        .args(&build_shared_args(true))
```
Porges, line 137: Seems to have been a bug in existing code.

src/deny.toml
```diff
@@ -16,8 +16,6 @@ unsound = "deny"
 yanked = "deny"
 ignore = [
     "RUSTSEC-2022-0048", # xml-rs is unmaintained
-    "RUSTSEC-2021-0139", # ansi_term is unmaintained
-    "RUSTSEC-2021-0145", # waiting for clap upgrade; we are unaffected by the bug (no custom allocator)
```
Porges, line 20: This is the main goal.

chkeita (approved):

Porges:
Passed `check-pr`.

PR #2866, opened 2023-02-19
Fix `{target_harness}` expansion in template

Porges:
This seems to have been inadvertently broken by 41f973184ec7653c7048872362e47e2ca9999f89.

The missing variable expansion was exposed by #2789, so tasks using `{target_harness}` (which is not a valid expansion) fail.


chkeita (approved):

ranweiler:
This is a real bug, but it only lingered because that template has been superseded by what we currently call `libfuzzer dotnet_dll`.

Since a major release is coming up, now would be a great time to just remove the old POC implementation entirely. PR for that here: #2875

Porges:
Closing in favour of #2875

PR #2879, opened 2023-02-21
Fix .NET integration tests

Porges:
This was missed in #2875.


Porges:
Will verify with check-pr first.

mgreisen (approved):

PR #2886, opened 2023-02-24
Create 2 way IPC connection between agent and task

### LINKED ISSUE #2768: Create a communication channel between onefuzz-agent and onefuzz-task
This should be a bi-directional, IPC-like channel between the onefuzz-agent and all onefuzz-tasks that are spawned from it. For now, this channel will be used for orchestrating the graceful shutdown of a task. In the future, we will want to funnel telemetry from the onefuzz-task through this channel.

AB#43087255

### LINKED ISSUE #2769: Graceful shutdown
Use the communication channel described in #2768 to orchestrate a graceful shutdown of the tasks which are being run by an agent. These shutdowns are expected (such as a newer job being submitted) and should not create alerts.

AB#43087288

tevoinea:

_What is this about?_

Closes #2769
Closes #2768 

Creates a 2 way communication channel between the onefuzz-agent and the onefuzz-task binaries.

## Info on Pull Request

_What does this include?_

* When the agent starts a task, it includes paths for the task to connect 2 IPC channels:
    * agent -> task
    * task -> agent
* Each task invocation will bootstrap a separate set of channels so the collocated task scenario is supported
* The agent waits up to 30 seconds for the task to complete the IPC channel bootstrap
* When it comes time to shutdown, the task has 90 seconds to terminate after the shutdown message is sent, otherwise we kill it

## Validation Steps Performed

* Updated tests and tested locally using unmanaged nodes feature


src/agent/onefuzz-agent/src/worker.rs
```diff
@@ -130,10 +133,36 @@ impl<C: Context> State<C> {
...
+        let child = runner
+            .run(
+                &self.ctx.setup_dir,
+                &self.work,
+                from_agent_to_task_endpoint,
+                from_task_to_agent_endpoint,
+            )
+            .await?;
+
+        // Accept is a blocking call:
```
Porges, line 148: If it's blocking we should use `spawn_blocking`: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

src/agent/onefuzz-agent/src/worker.rs
```diff
@@ -151,6 +214,10 @@ impl State<Ready> {
 
 impl State<Running> {
     pub fn wait(mut self) -> Result<Waited> {
+        while let Ok(res) = self.ctx.from_task_to_agent.try_recv() {
+            info!("received message from server_receiver: {:?}", res);
```
tevoinea, line 218: Eventually, we plan on doing things with the messages received from the onefuzz-task. For now though, this will be a noop.

src/agent/onefuzz-task/src/managed/cmd.rs
```diff
@@ -55,6 +96,10 @@ pub async fn run(args: &clap::ArgMatches) -> Result<()> {
             let err = format_err!("out of memory: {} bytes available, {} required", oom.available_bytes, oom.min_bytes);
             Err(err)
         },
+
+        _shutdown = shutdown_listener => {
```
tevoinea, line 100: When we receive the `Shutdown` message, the `shutdown_listener` future will complete.

src/agent/onefuzz-agent/src/worker.rs
```diff
@@ -159,13 +226,51 @@ impl State<Running> {
...
+                    Ok(Some(_)) => {
+                        info!("Agent was gracefully shut down");
+                        // Since the agent has exited, drain one more time before returning
+                        while let Ok(res) = self.ctx.from_task_to_agent.try_recv() {
+                            info!("received message from server_receiver: {:?}", res);
+                        }
+                        return Ok(());
+                    }
+                    Ok(None) => {
+                        if i == 89 {
```
chkeita, line 255: Why not use the `time::timeout` for the logic of waiting for the process to return like you did in `State<Ready>::run`  ?
If you have to do it this way at least make the end condition dependent on the loop counter. 
``` rust 
    if i == MAX_COUNT - 1 
```

PR #2911, opened 2023-03-06
Add additional filter check for reports and regressions

nharper285:

_What is this about?_
Add additional filter condition to catch `source-coverage.json` files. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Reports.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Reports.cs
@@ -28,7 +28,7 @@ public class Reports : IReports {
 
     public async Async.Task<IReport?> GetReportOrRegression(Container container, string fileName, bool expectReports = false, params string[] args) {
         var filePath = string.Join("/", new[] { container.String, fileName });
-        if (!fileName.EndsWith(".json", StringComparison.Ordinal) || fileName.Contains("source-coverage", StringComparison.Ordinal)) {
+        if (!fileName.EndsWith(".json", StringComparison.Ordinal) || fileName.Contains("source-coverage", StringComparison.InvariantCultureIgnoreCase)) {
             if (expectReports) {
                 _log.Error($"get_report invalid extension or filename: {filePath:Tag:FilePath}");
             }
--- a/src/ApiService/ApiService/onefuzzlib/Reports.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Reports.cs
@@ -28,9 +28,9 @@ public class Reports : IReports {
 
     public async Async.Task<IReport?> GetReportOrRegression(Container container, string fileName, bool expectReports = false, params string[] args) {
         var filePath = string.Join("/", new[] { container.String, fileName });
-        if (!fileName.EndsWith(".json", StringComparison.Ordinal)) {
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/Reports.cs
```diff
@@ -28,9 +28,9 @@ public class Reports : IReports {
 
     public async Async.Task<IReport?> GetReportOrRegression(Container container, string fileName, bool expectReports = false, params string[] args) {
         var filePath = string.Join("/", new[] { container.String, fileName });
-        if (!fileName.EndsWith(".json", StringComparison.Ordinal)) {
+        if (!fileName.EndsWith(".json", StringComparison.Ordinal) || fileName.Contains("source-coverage", StringComparison.Ordinal)) {
```
tevoinea, line 31: I think `StringComparison.InvariantCultureIgnoreCase` is the recommended option: https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#specifying-string-comparisons-explicitly

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/Reports.cs
nharper285: Ah good to know, thank you!

Porges: For machine-readable stuff like this filename `OrdinalIgnoreCase` is more suited. The culture-sensitive versions can be _very_ weird if you're not expecting it:

```csharp
"source-coverage".Contains("C­", StringComparison.InvariantCultureIgnoreCase)
```

src/ApiService/ApiService/onefuzzlib/Reports.cs
```diff
@@ -28,9 +28,9 @@ public class Reports : IReports {
 
     public async Async.Task<IReport?> GetReportOrRegression(Container container, string fileName, bool expectReports = false, params string[] args) {
         var filePath = string.Join("/", new[] { container.String, fileName });
-        if (!fileName.EndsWith(".json", StringComparison.Ordinal)) {
+        if (!fileName.EndsWith(".json", StringComparison.Ordinal) || fileName.Contains("source-coverage", StringComparison.InvariantCultureIgnoreCase)) {
```
Porges, line 31: Should this not be:
```suggestion
        if (!fileName.EndsWith(".json", StringComparison.Ordinal) || !fileName.Contains("source-coverage", StringComparison.InvariantCultureIgnoreCase)) {
```

Porges:
We should also be able to write a test with this using the existing implementations we have for Azurite stuff... let me know if you want any help putting one together.

src/ApiService/ApiService/onefuzzlib/Reports.cs
nharper285: @tevoinea can confirm, but it was my understanding that we don't want to parse the file if its name starts `source-config`.

tevoinea: That's right, we don't want to parse the file is the name _ends with_ `source-coverage.json`

PR #2912, opened 2023-03-07
Ipc between agent and task

tevoinea:

_What is this about?_

This is a re-attempt of #2910 which adds a 2 way IPC connection between the agent and the task, and also implements a graceful shutdown on top of that connection.

What is different this time is that the graceful shutdown is implicitly tied to the lifetime of the IPC connection. In other words, when the IPC connection drops and the task sees the disconnect, the task will gracefully shut down.


mgreisen (approved):
:shipit:

PR #2940, opened 2023-03-23
Better logging of failure in the task_logger

chkeita:

Add better logging when the task_logger fails


src/agent/onefuzz-task/src/tasks/task_logger.rs
```diff
@@ -440,7 +440,10 @@ pub struct SpawnedLogger {
 
 impl SpawnedLogger {
     pub async fn flush_and_stop(self, timeout: Duration) -> Result<()> {
-        let _ = tokio::time::timeout(timeout, self.logger_handle).await;
+        if let Ok(Err(e)) = tokio::time::timeout(timeout, self.logger_handle).await {
+            error!("task logger failure {}", e);
```
tevoinea, line 444: NIT: maybe choose some different text to log so that when we see it it'll be easy to know exactly where it's coming from?

ex:
```suggestion
            error!("failed to flush and stop task logger {}", e);
```

tevoinea (approved):

src/agent/onefuzz-agent/src/validations.rs
```diff
@@ -64,10 +64,11 @@ async fn validate_libfuzzer(config: ValidationConfig) -> Result<()> {
     let libfuzzer = LibFuzzer::new(
         &config.target_exe,
         config.target_options.clone(),
-        config.target_env.into_iter().collect(),
+        config.target_env.iter().cloned().collect(),
         config
             .setup_folder
-            .unwrap_or(config.target_exe.parent().unwrap().to_path_buf()),
+            .clone()
+            .unwrap_or_else(|| config.target_exe.parent().unwrap().to_path_buf()),
```
Porges, line 71: Can we use `expect` instead of `unwrap` here to provide a message in case this ever fails?

Porges (approved):

PR #2943, opened 2023-03-24
Install v17 redistributables

Porges:
The latest VC redistributables [support VS2015 through VS2022](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#visual-studio-2015-2017-2019-and-2022), so we only need to install one version.


chkeita (approved):

Porges:
Validated in check-pr

PR #2960, opened 2023-04-03
Rename EventGrid subscription

Porges:
The EventGrid subscription name is invalid according to the published schema. Rename it, and add a deployment step to delete any old instances of this subscription.


Porges:
Deployment works fine

PR #2962, opened 2023-04-04
Use minimized stack for crash site if no ASAN logs available

### LINKED ISSUE #841: use the top of the filtered stack for `crash_site`
Currently, `report.crash_site` uses the top function of the call stack (or whatever ASAN provided in the crash line).  Instead, we should use the top of the filtered stack if that list is available. 

AB#35984

Porges:
Closes #841. 

If all we have is a call stack, use the top of the "minimized stack" as the crash site, not the top of the full stack.

This only happens if we don't have ASAN logs available for the crash.

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/report/dotnet/generic.rs
+++ b/src/agent/onefuzz-task/src/tasks/report/dotnet/generic.rs
@@ -208,7 +208,7 @@ impl AsanProcessor {
                     input_blob,
                     executable,
                     crash_type: exception.exception,
-                    crash_site: exception.call_stack[0].clone(),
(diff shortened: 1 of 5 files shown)


src/agent/onefuzz-task/src/tasks/report/dotnet/generic.rs
```diff
@@ -208,7 +208,7 @@ impl AsanProcessor {
                     input_blob,
                     executable,
                     crash_type: exception.exception,
-                    crash_site: exception.call_stack[0].clone(),
```
Porges, line 211: just fixing a potential issue here

tevoinea (approved):

Issue #2984, opened 2023-04-05
Missing `UniqueFields` key in `AdoFields` notification config is not handled. 


nharper285:
## Information

- Onefuzz version: 7.0.0

## Provide detailed reproduction steps (if any)

1. Create notification config.json file with list of `UniqueFields` that are not present in `AdoFields`. 
Ex. 
```
{
    "config": {
        "base_url": "https://dev.azure.com/fuzzfuzz",
        "type": "Bug",
        "project": "Onefuzz",
        "ado_fields": {
            "System.AreaPath": "Onefuzz",
            "Microsoft.VSTS.Scheduling.StoryPoints": "1",
            "System.Title": "[C#ANARY - TESTING STUFF] {{ report.crash_site }} - {{ report.executable }}",
            "Microsoft.VSTS.TCM.ReproSteps": "This is my call stack: <ul> {% for item in report.call_stack %} <li> {{ item }} </li> {% endfor %} </ul>"
        },
        "comment": "This is my comment. {{ report.input_sha256 }} {{ input_url }} <br> <pre>{{ repro_cmd }}</pre>",
        "unique_fields": [
            **"Microsoft.VSTS.Common.CustomString01",
            "Microsoft.VSTS.Common.CustomString02",
            "Microsoft.VSTS.Common.CustomString03",**
            "System.Title",
            "System.AreaPath"
        ],
        "on_duplicate": {
            "comment": "Another <a href='{{ input_url }}'>POC</a> was found in <a href='{{ target_url }}'>target</a>. <br> <pre>{{ repro_cmd }}</pre>",
            "set_state": {
                "Resolved": "Active"
            },
            "ado_fields": {},
            "increment": [
                "Microsoft.VSTS.Scheduling.StoryPoints"
            ]
        }
    }
}

```
### Expected result

_What is the expected result of the above steps?_
Ado bugs are properly filed. 

### Actual result

_What is the actual result of the above steps?_
Bugs are not filed. 

[AB#45444](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/45444)

PR #2986, opened 2023-04-05
Adding handle for missing unique field key in `AdoFields`

nharper285:

_What is this about?_

## PR Checklist
* [ ] Applies to work item: #2984 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
Resolving bug #2984 by adding exception handling. 

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
+++ b/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
@@ -146,15 +146,20 @@ public class Ado : NotificationsBase, IAdo {
             }
         }
 
-        public async IAsyncEnumerable<WorkItem> ExistingWorkItems() {
+        public async IAsyncEnumerable<WorkItem> ExistingWorkItems((string, string)[] notificationInfo) {
             var filters = new Dictionary<string, string>();
             foreach (var key in _config.UniqueFields) {
                 var filter = string.Empty;
                 if (string.Equals("System.TeamProject", key)) {
                     filter = await Render(_config.Project);
                 } else {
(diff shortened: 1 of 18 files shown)


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -146,15 +146,20 @@ sealed class AdoConnector {
...
             var filters = new Dictionary<string, string>();
             foreach (var key in _config.UniqueFields) {
                 var filter = string.Empty;
                 if (string.Equals("System.TeamProject", key)) {
                     filter = await Render(_config.Project);
                 } else {
-                    filter = await Render(_config.AdoFields[key]);
+                    try {
+                        filter = await Render(_config.AdoFields[key]);
+                    } catch (Exception e) {
```
chkeita, line 158: is there a more specific exception type we can use here ?

tevoinea: Or alternatively:

```csharp
if (string.Equals("System.TeamProject", key)) {
   filter = await Render(_config.Project);
} else if (_config.AdoFields.ContainsKey(key)) {
   filter = await Render(_config.AdoFields[key]);
} else {
   _logTracer.WithTags(notificationInfo).Error($"Failed to check for existing work items using the UniqueField Key: {key}. Value is not present in config field AdoFields.");
   continue;
}
```

The `continue` in the else branch is important because with the PRs current state, we would still be adding a filter for the erroring key but the filter would be empty.

tevoinea (approved):

PR #2992, opened 2023-04-06
Cap recursion in ORM

### LINKED ISSUE #2900: Set a recursion limit in ORM when deserializing from DB
This issue is created as a result of the investigation in #2890. Relevant info: https://github.com/microsoft/onefuzz/issues/2890#issuecomment-1451982878

The OOM Exception describe there was caused by misconfigured task entity which resulted in a recursive loop in the ORM.

Inside `GetFieldValue` we can call it recursively: https://github.com/microsoft/onefuzz/blob/503fc7cd9effa410cf68174f22581700f65602a5/src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs#L288

It's not that we _shouldn't_ do this recursively, just that we should limit the depth to something reasonable so that we don't waste resources when we already know that the operation won't be successful (say depth = 100).

[AB#36108](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/36108)

### LINKED ISSUE #2901: Try to log at least the PartitionKey/RowKey when deserialization exceptions happen
This issue is created as a result of the investigation in #2890. Relevant info: https://github.com/microsoft/onefuzz/issues/2890#issuecomment-1451982878

The EntityConverter can throw exceptions when failing to deserialize table entities. We should log the PartitionKey/RowKey when deserializations fail to make it easier to identify the culprit entity later.

[AB#36109](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/36109)

### LINKED ISSUE #2902: Try to dump the column information when deserialization exceptions happen
This issue is created as a result of the investigation in #2890. Relevant info: https://github.com/microsoft/onefuzz/issues/2890#issuecomment-1451982878

The EntityConverter can throw exceptions when failing to deserialize table entities. Even if we log the PartitionKey/RowKey, that entity may still be changed later so that it ends up in a valid state. We should dump the column information we have on the entity when exceptions are thrown to help debug invalid entity states.

[AB#36110](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/36110)

tevoinea:

_What is this about?_

Closes #2900
Closes #2901 
Closes #2902

## #2900

Considers 2 new cases in our Entity Conversion code:

1. If the discriminator field name is the field we're currently trying to deserialize, we're guaranteed to end up in an infinite loop
2. If for any other reason, we iterate more than 100 times while trying to convert an entity to a type

In both cases, we throw a new `OrmShortCircuitInfiniteLoopException` and stop deserialization.

## #2901 & #2902

* Add a `LogTracer` to the `EntityConverter`
* When throwing an exception, include as many tags and information as possible to help pin down the faulting entity/fields

## Validation Steps Performed

Added a test that demonstrates the issue where the discriminator field `EventType` is self referential.

I looked through our codebase and didn't find any situation where we have this self referential situation but it's possible for the issue to be coming from entities in our database -- which would be much harder to check for this self referential behavior. Though, if/when this issue happens again, because of the improved logging changes, we should be able to track down the offending etities.


src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -285,7 +288,25 @@ public class EntityConverter {
...
                 if (ef.discriminator != null) {
                     var (attr, typeProvider) = ef.discriminator.Value;
-                    var v = GetFieldValue(info, attr.FieldName, entity) ?? throw new Exception($"No value for {attr.FieldName}");
+                    if (iterationCount > MAX_DESERIALIZATION_RECURSION_DEPTH) {
+                        var tags = GenerateTableEntityTags(entity);
+                        tags.AddRange(new (string, string)[] {
+                            ("outputType", outputType?.Name ?? string.Empty),
+                            ("fieldName", fieldName)
+                        });
+                        _logTracer.WithTags(tags).Error($"Too many iterations deserializing {info.type}");
```
chkeita, line 297: do we need to pass a logger here to this class. We already throw an exception when we have an issue

chkeita: Instead you can use a more specific exception type if you want to collect more information about the failure. This will reduce the coupling in general

tevoinea: I understand the coupling concern. My thinking was we risk losing that information if downstream callers forget to log it -- like if we do `catch (Exception)`.

chkeita: We already log exceptions when they occur at this level right. You were able to identify the issue from the logs already. 
```
Unable to extract properties from TableEntity for Microsoft.OneFuzz.Service.Task Unable to get value for property 'Config' (entity field 'config') Exception of type 'System.OutOfMemoryException' was thrown.
```

In this case if we had a more specific exception type you would get all the information you need in the logs.

tevoinea: Updated to remove logtracer from entityconverter and include all the necessary information in the exception.

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -124,8 +125,8 @@ public class EntityConverter {
     }
 
     private static IEnumerable<EntityProperty> GetEntityProperties<T>(ParameterInfo parameterInfo) {
-        var name = parameterInfo.Name.EnsureNotNull($"Invalid paramter {parameterInfo}");
-        var parameterType = parameterInfo.ParameterType.EnsureNotNull($"Invalid paramter {parameterInfo}");
+        var name = parameterInfo.Name.EnsureNotNull($"Invalid paramater {parameterInfo}");
```
Porges, line 128:
```suggestion
        var name = parameterInfo.Name.EnsureNotNull($"Invalid parameter {parameterInfo}");
```

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -124,8 +125,8 @@ public class EntityConverter {
...
+        var parameterType = parameterInfo.ParameterType.EnsureNotNull($"Invalid paramater {parameterInfo}");
```
Porges, line 129:
```suggestion
        var parameterType = parameterInfo.ParameterType.EnsureNotNull($"Invalid parameter {parameterInfo}");
```

src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs
```diff
@@ -222,7 +223,7 @@ public class EntityConverter {
     }
 
 
-    private object? GetFieldValue(EntityInfo info, string name, TableEntity entity) {
+    private object? GetFieldValue(EntityInfo info, string name, TableEntity entity, int iterationCount = 0) {
```
Porges, line 226: I'd make the parameter required so that it doesn't get accidentally omitted upon a recursive call

Porges (changes requested):
A few small typos to fix

Porges (approved):

PR #3010, opened 2023-04-11
Unify some crate versions

Porges:
In Component Governance we have a bunch of crates which are consumed twice under different versions in the `agent` and `proxy-manager` components. Try to unify some of these to reduce the overall dependency count (and potentially help with build times).


tevoinea (approved):
I'm all for staying on top of the latest versions for our dependencies and reducing build times but we don't really need to worry about dependency count in CG.

Issue #2750, opened 2023-01-18
Replace Desired Traces & Custom Events w/ Custom Metrics to Avoid Sampling


nharper285:
Currently, Custom Events and Traces are sampled outside our control. 

Custom Metrics are not sampled, so we are proposing that we'd replace logging with metrics.

PR #3018, opened 2023-04-12
Refactor logging to use  and new function queue

nharper285:

_What is this about?_
In order to improve the reliability of OneFuzz's telemetry - and therefore the Grafana data dashboards that rely on it - we are pursuing a refactor that makes use of App Insight's CustomMetrics type. These metrics are, importantly, not sampled and will offer us a more consistent monitoring type without fear of sampling. 

In order to centralize and standardize the CustomMetric format, we are making use of a new set of 'metrics' types and functions. Funneling all of the logging through this one source will allow us to keep better track of the creation of the actual metric _and_, again, standardize the metric's format - name, value, custom dimensions. 

To start, I have implemented custom metrics for the node_heartbeat event and the task_heartbeat event. 

## PR Checklist
* [ ] Applies to work item: #2750 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
* New 'metrics' types
* node and task heartbeat metrics

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/Metrics.cs
```diff
@@ -0,0 +1,57 @@
...
+        public Metrics(ILogTracer log, IOnefuzzContext context) {
+            _context = context;
+            _log = log;
+            _options = new JsonSerializerOptions(EntityConverter.GetJsonSerializerOptions()) {
+                DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull
+            };
+            _options.Converters.Add(new RemoveUserInfo());
+        }
+
+        public async Async.Task QueueCustomMetric(MetricMessage message) {
```
chkeita, line 36: Can you explain the decision of queueing the metrics instead of logging it directly

src/ApiService/ApiService/OneFuzzTypes/Metrics.cs
```diff
@@ -0,0 +1,376 @@
...
+[MetricType(MetricType.TaskStopped)]
+public record MetricTaskStopped(
+    Guid JobId,
+    Guid TaskId,
+    UserInfo? UserInfo,
+    TaskConfig Config
+) : BaseMetric();
+
+[MetricType(MetricType.TaskFailed)]
+public record MetricTaskFailed(
```
chkeita, line 127: These definitions look like the Events. Do we need new records for those? 
We could just use the event and derive the Metric form them

src/ApiService/ApiService/onefuzzlib/Metrics.cs
nharper285: Just added a PR description that answers this!

In order to centralize and standardize the CustomMetric format, we are making use of a new storage queue custom-metrics and a new function QueueCustomMetrics in the app function. Funneling all of the logging through this one source will allow us to keep better track of the creation of the actual metric and, again, standardize the metric's format - name, value, custom dimensions.

src/ApiService/ApiService/OneFuzzTypes/Metrics.cs
nharper285: They are essentially duplicates right now, but that was just meant to be a starting point. I'm thinking that the actual fields may change as we go along to include different info. For example, I updated the TaskHeartbeat _metric_ record to be different from the event one.

nharper285: For ease, I just duplicated all existing events with the idea being that many of the events will be metrics at some point.

src/ApiService/ApiService/onefuzzlib/Metrics.cs
nharper285: Added: 
In addition, this function app may act as a starting point for eventually _storing_ metrics and logging as we consider moving away from app insights entirely.

Porges, line 36: I think we should treat these metrics as best-effort here, so if we try to enqueue the metric and it fails we should capture and swallow the exception, in order that we not disrupt the ordinary flow of the code sending the metric.

A more advanced implementation would be that this synchronously adds the event to an in-memory queue which is added to the storage queue by a background task, so that "sending" a metric is instantaneous and cannot fail from the POV of the calling code. But this could be left for later.

PR #3032, opened 2023-04-18
Allow libfuzzer verification to retry

### LINKED ISSUE #2862: When a DLL is excluded in setup.ps1 it takes an unknown time before Windows responds to the setting
## Information

- Onefuzz version: 6.4
- OS: Windows 10 and 11

1. Create a DLL in the setup container that should be used instead of the system one
2. Include setup.ps1 in your setup container
3.The known DLL exclusion is not immediate, but takes some time to take effect. 

This can cause fuzzers to fail. 

The powershell code that we provide for customers includes this function.
```pwsh
function Exclude-Library {
    Param(
        [Parameter(Position=0,
                   Mandatory,
                   ValueFromPipeline,
                   ValueFromRemainingArguments)]
        [String[]] $Libraries)
    Begin {
        $Path = 'HKLM:System\CurrentControlSet\Control\Session Manager'
        $Name = 'ExcludeFromKnownDLLs'
    }
    Process {
        # Get existing excluded libraries
        $ExistingProperty = Get-ItemProperty -Path $Path -Name $Name -ErrorAction SilentlyContinue
        $ExistingExclusions = @()

        # Normalize DLL name to lowercase for comparison
        if ($ExistingProperty -ne $null) {
            $ExistingExclusions = $ExistingProperty.$Name.ForEach("ToLower")
        }

        # Normalize DLL name to lowercase for comparison, remove duplicates
        $Libs = $Libraries.ForEach("ToLower") | Select-Object -Unique
        Write-Log "Excluding libraries $Libs"

        # Discard empty strings and libraries already excluded
        $Libs = $Libs.Where({$_.Length -gt 0 -and !($ExistingExclusions.Contains($_))})

        # If anything remains either add or update registry key
        if ($Libs.Length -gt 0) {
            if ($ExistingProperty -eq $null) {
                # Create registry key to exclude DLLs
                New-ItemProperty -Path $Path -Name $Name -PropertyType MultiString -Value $Libs
                Write-Log "Created known DLLs exclusions with $Libs"
            } else {
                # Update registry key to exclude DLLs
                Set-ItemProperty -Path $Path -Name $Name -Value ($ExistingProperty.$Name + $Libs)
                Write-Log "Updated known DLLs exclusions with $Libs"
            }
        } else {
            # DLLs already excluded
            Write-Log "Known DLL exclusions already exist for $Libraries"
        }
    }
    End {}
}
```
And it's executed like this
```
gci -filter '*.dll' | Exclude-Library
```

### Expected result

The known DLL exclusion to be in place before the fuzzer is called. 

### Actual result

The fuzzer runs and produces unexpected results because the wrong DLL is loaded. 

AB#35870

Porges:
Closes #2862 by brute force.

We’ve seen cases where it takes some time for an executable with a KnownDlls override to actually work after the node is restarted; thus, we now allow verification to take some time (currently up to a minute) to validate that a libfuzzer-based fuzzer actually works.

### DIFF
--- a/src/agent/onefuzz/src/libfuzzer.rs
+++ b/src/agent/onefuzz/src/libfuzzer.rs
@@ -226,6 +226,7 @@ impl LibFuzzer {
                 Ok(()) => return Ok(()),
                 Err(e) => {
                     if (Instant::now() + sleep_time) <= (started_at + max_verification_time) {
+                        warn!("libfuzzer verification failed, will retry: {e:?}");
                         tokio::time::sleep(sleep_time).await;
                     } else {
                         return Err(e.context(format!(
--- a/src/agent/onefuzz/src/libfuzzer.rs
+++ b/src/agent/onefuzz/src/libfuzzer.rs
@@ -16,6 +16,7 @@ use std::{
     ffi::{OsStr, OsString},
     path::{Path, PathBuf},
     process::Stdio,
+    time::{Duration, Instant},
 };
 use tempfile::tempdir;
 use tokio::process::{Child, Command};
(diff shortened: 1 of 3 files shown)


Porges:
Will be deploying and testing this before marking ready for review.

src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -212,6 +208,38 @@ impl LibFuzzer {
...
+        let mut attempts = 1;
+        loop {
+            let result = self
+                .verify_inner(check_fuzzer_help, inputs.unwrap_or_default())
+                .await;
+
+            match result {
+                Ok(()) => return Ok(()),
+                Err(e) => {
+                    if (Instant::now() + sleep_time) <= (started_at + max_verification_time) {
```
chkeita, line 228: you might want to gate on the number of attempts instead of the running time. The current default running time is [10min](https://github.com/chkeita/onefuzz/blob/256f261f2c1a5fa98fe48a39a14507fa3236b7f6/src/agent/onefuzz/src/libfuzzer.rs#L23)

chkeita (approved):

PR #3045, opened 2023-04-20
Expand valid scaleset names

### LINKED ISSUE #2189: Changing the way scalesets are named
Currently when scaleset resources are created they are named using a guid - which is converted into a string. This makes managing the scalesets in the azure portal difficult because we typically want to find the scalesets that are related to a pool. 

It would be helpful if the scaleset name could include some portion of the pool name. 
Azure resource names need to adhere to certain rules. https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#:~:text=virtualmachinescalesets

The total number of characters allowed for a virtual machine scaleset resource name is 64 characters. 
The length of a guid including dashes is 36 characters. 
If the first 28 characters of the pool name were prepended to the exiting way we name a scaleset now, that would make identifying the scalesets much easier and it would also group multiple scalesets in the same pool together when sorted by name.

AB#36023

Porges:
Closes #2189.

Scaleset names are now permitted to be any (valid) strings, instead of only GUIDs. When we generate a scaleset name it is now based upon the pool name; for example the pool `pool` might get a scaleset named `pool-3b24ba211cad4b078655914754485838`.

This should be backwards-compatible since GUIDs are [already serialized to table storage as strings](https://github.com/microsoft/onefuzz/blob/dddcfa49490e44834afc9a6910c350698022792c/src/ApiService/ApiService/onefuzzlib/orm/EntityConverter.cs#L190-L191), so this simply loosens the restrictions placed upon them.

Scaleset IDs now have a strong type in the same way as other IDs; this helps to avoid mixing them up with other strings. Because of this I found one bug in the scaleset search query logic due to Pool ID/VMSS ID confusion. As part of fixing this I've changed the scaleset search query to only return nodes from the table rather than querying Azure to find a list; this seems to be sufficient for the CLI.


Porges:
Passed check-pr.

chkeita (approved):

PR #3048, opened 2023-04-20
Improve agent perf at the cost of compile time

tevoinea:
Setting codegen units to 1 may improve performance. It _will_ increase compile times so I'm opening this PR to see how much.

https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

### DIFF
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -22,3 +22,4 @@ members = [
 
 [profile.release]
 lto = "thin"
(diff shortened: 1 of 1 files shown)


Porges:
I think you'll need to bust the cache to see the effects of this

Porges:
I'd support merging this since it also reduces binary size, if we can somehow only enable it for release builds?

PR #3050, opened 2023-04-20
Migrate from winapi to windows-rs

Porges:
`windows-rs` is the newer, Microsoft-supported version of the API bindings; `winapi` hasn't been updated in some time. This allows us to remove some code, as `windows-rs` includes the `Sym*` functions that we had to previously defined ourselves in `dbghelp`.

Rather than port the `jobs` and `com` code I removed it, as it was unused.

The `check_hr` and `check_winapi` macros have been replaced by use of `.ok()?` on `BOOL` and `?` on `HRESULT` which now support standard error handling facilities.


Porges:
I've checked coverage output with check-pr, which is the main thing that I would expect to be affected by this change.

Porges:
> Just cloned this repo and tried to build with `winapi v0.3.9`. The build (of `src\agent\coverage`) is broken as `synchapi` and `impl-default` aren't enabled. So I guess this PR would be nice as it should happen to fix the build too :)

The issue here is that it depends on what level in the repo you build at. If you build `src/agent` then all the features of `winapi-rs` which we need are enabled, since cargo will union them all together. However, some of the subcrates use features without enabling them (for example, I just fixed one over [here](https://github.com/microsoft/onefuzz/pull/3195/files#diff-60907d5f6523d425ee5f228780ef1d990d0f0e2585618547d71028ae71a9f240)), so you might get failures if you build an individual crate. This is annoying to detect in the CI build since we'd have to build each crate individually rather than just building the whole workspace...

chkeita (approved):

PR #3059, opened 2023-04-24
Add New Endpoint to update the pool authentication

### LINKED ISSUE #2995: Unmanaged node receive multiple stop message after it shuts down
When using the unmanaged nodes. The service sends multiple stop messages after the node is shutdown. 
This causes new instances of the node to exit soon after launch

[AB#45531](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/45531)

chkeita:

- This PR adds a new `PATCH` endpoint to the `Pool` function to handle updates to an existing pool
- Fixes #2995


src/ApiService/ApiService/OneFuzzTypes/Requests.cs
```diff
@@ -264,6 +264,13 @@ public record PoolCreate(
     [property: Required] Os Os,
     [property: Required] Architecture Arch,
     [property: Required] bool Managed,
+    Guid? ObjectId = null,
+    bool Update = false
```
tevoinea, line 268: Why do we need this?

tevoinea (approved):

src/ApiService/ApiService/OneFuzzTypes/Requests.cs
chkeita: This was just missing from the data returned from the server. It is the object id of the principal used by an unmanaged node

PR #3061, opened 2023-04-25
Fixup Dockerfile syntax

Porges:
The unquoted ampersands seem to be causing problems in other tooling. Quote all the string values.


chkeita (approved):

Porges:
Actually the issue here is that Pinned-Dependencies is assuming all commands are valid `sh`. Closing this, as I filed a bug here: https://github.com/ossf/scorecard/issues/2911

PR #3064, opened 2023-04-26
Add `extra_output` container, rename `extra` container

### LINKED ISSUE #3113: Failures happening before the task channel initialization  are not reported
When an issue occurs in the task before the bootstrapping of the communication channel with the agent. The failure is not reported. Instead, the agent reports a timeout issue.

This happens when an environment variable is missing for example.

[AB#152289](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/152289)

Porges:

- **Breaking** (but as far as I know this feature is not yet in use): rename the `extra_container` to `extra_setup_container`.
- **Add**: the `extra_output_container`, which pushes its outputs continually.
  - We may also want a type of container which both pushes & pulls? See discussion below.
- **Improved**: if `onefuzz-task` fails upon launch, we will log its output for diagnosis (might close #3113)

---

Some thoughts for the future:

We might want to redesign the containers so that we have something like the following which is passed to the agent, and the agent doesn't need to know the specifics of the containers supplied:

```jsonc
{
    // ...
    "containers": {
        "extra_setup_dir": {
            "mode": "pull",
            "container_name": "yyy",
        },
        "extra_output_dir": {
            "mode": "push",
            "continuous": true, // keep pushing while job is running
            "container_name": "xxx"
        }
    }
}
```

At the moment the agent needs to know what each container is for, for each task type. A more generic and flexible method might be simpler overall.


Porges:
Code changes are ready for looking at but this needs its own integration test to be fully complete.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -917,7 +917,7 @@ public record VmDefinition(
     long Value
 );
 
-public record TaskDefinition(
+public readonly record struct TaskDefinition(
```
tevoinea, line 920: General question and not a comment: What's the benefit of changing this to a struct?

From what I read [here](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct), structs should be smaller than 16 bytes.

src/agent/onefuzz/src/expand.rs
```diff
@@ -75,7 +76,8 @@ impl PlaceHolder {
             Self::SupervisorExe => "{supervisor_exe}",
             Self::SupervisorOptions => "{supervisor_options}",
             Self::SetupDir => "{setup_dir}",
-            Self::ExtraDir => "{extra_dir}",
+            Self::ExtraSetupDir => "{extra_setup_dir}",
```
tevoinea, line 79: Is there a migration plan for this breaking change?

tevoinea (approved):

src/ApiService/ApiService/OneFuzzTypes/Model.cs
Porges: It was so I can make it `readonly`, but `record` (classes) are also readonly for positional parameters, so I can undo this. 🙂

src/agent/onefuzz/src/expand.rs
Porges: No. To my knowledge the feature is not yet used by anyone (the person who requested it has not started using the feature).

src/agent/onefuzz-agent/src/worker.rs
```diff
@@ -193,10 +197,24 @@ impl State<Ready> {
...
+
+                error!(
+                    "timeout waiting for client_sender_server.accept(): child status: {}",
+                    child_output,
+                );
+
                 return Err(format_err!(
-                    "timeout waiting for client_sender_server.accept(): {:?}",
-                    e
+                    "timeout waiting for client_sender_server.accept()"
```
chkeita, line 217: we need include the child output here so it will appear in the error field of the task

PR #3079, opened 2023-05-01
Collect additional report field in an ExtensionData property

### LINKED ISSUE #3078: Allow ADO work item to reference any properties from the crash report
User Story description:  As a creator of a custom generic-analyzer task I want to add extra data to a crash report, and then add that extra analysis data to the subsequent work item.

Impact: There are a few extra fields we particularly want to add - data about the register states, exploitability. There are enough fields in the report that we could improvise and repurpose certain fields, but that's not ideal.

[AB#141952](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/141952)

chkeita:
closes #3078

### DIFF
--- a/src/ApiService/Tests/ReportTests.cs
+++ b/src/ApiService/Tests/ReportTests.cs
@@ -36,7 +36,9 @@ public class ReportTests {
     },
     "tool_name": "libfuzzer",
     "tool_version": "1.2.3",
-    "onefuzz_version": "1.2.3"
+    "onefuzz_version": "1.2.3",
+    "extra_porperty1": "test",
+    "extra_porperty2": 5
 }
 """;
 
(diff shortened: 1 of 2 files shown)


src/ApiService/Tests/ReportTests.cs
```diff
@@ -36,7 +36,9 @@ public class ReportTests {
     },
     "tool_name": "libfuzzer",
     "tool_version": "1.2.3",
-    "onefuzz_version": "1.2.3"
+    "onefuzz_version": "1.2.3",
+    "extra_porperty1": "test",
```
mgreisen, line 40:
```suggestion
    "extra_property1": "test",
```

mgreisen: I see what you did there.... :)

tevoinea (approved):

PR #3082, opened 2023-05-03
Initial Custom Metrics - Node & Task Heartbeat

nharper285:

_What is this about?_
In order to improve the reliability of OneFuzz's telemetry - and therefore the Grafana data dashboards that rely on it - we are pursuing a refactor that makes use of App Insight's CustomMetrics type. These metrics are, importantly, not sampled and will offer us a more consistent monitoring type without fear of sampling. 

In order to centralize and standardize the CustomMetric format, we are making use of a set of 'metrics' types and functions. Funneling all of the logging through this one source will allow us to keep better track of the creation of the actual metric _and_, again, standardize the metric's format - name, value, custom dimensions. 

To start, I have implemented custom metrics for the node_heartbeat event and the task_heartbeat event. 

## PR Checklist
* [ ] Applies to work item: #2750 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_
* New 'metrics' types
* node and task heartbeat metrics

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/Metrics.cs
```diff
@@ -0,0 +1,53 @@
...
+            var metricTypeSnakeCase = _options.PropertyNamingPolicy.ConvertName($"{metricType}");
+
+            var dimensionString = JsonSerializer.Serialize(customDimensions, customDimensions.GetType(), _options);
+            var dimensionDict = JsonSerializer.Deserialize<Dictionary<string, string>>(dimensionString);
+
+            _log.Metric($"{metricTypeSnakeCase}", metricValue, dimensionDict);
+            LogMetric(customDimensions);
+        }
+
+        public void LogMetric(BaseMetric metric) {
```
chkeita, line 48: Do we still need to write metrics as logs if we can do it as metrics

src/ApiService/ApiService/OneFuzzTypes/Metrics.cs
```diff
@@ -0,0 +1,376 @@
...
+public class MetricTypeAttribute : Attribute {
+    public MetricTypeAttribute(MetricType metricType) {
+        this.MetricType = metricType;
+    }
+
+    public MetricType MetricType { get; }
+}
+
+
+public enum MetricType {
```
chkeita, line 21: Same comment as the other PR. Instead of creating a new type of metrics. We can just reuse the Event and modify the Event type as needed.

nharper285: I'm conflicted about this. I don't like the ambiguity of records called 'events' being used as metrics when we log both. I think it makes sense from an organizational perspective to keep them separate. 

A middle ground could be removing the 'metric types that I haven't yet used and only create them on a case by case basis as needed.

src/ApiService/ApiService/onefuzzlib/Metrics.cs
```diff
@@ -0,0 +1,53 @@
...
+            _ = _options.PropertyNamingPolicy ?? throw new ArgumentException("Serializer _options not available.");
+
+            var metricTypeSnakeCase = _options.PropertyNamingPolicy.ConvertName($"{metricType}");
+
+            var dimensionString = JsonSerializer.Serialize(customDimensions, customDimensions.GetType(), _options);
+            var dimensionDict = JsonSerializer.Deserialize<Dictionary<string, string>>(dimensionString);
+
+            _log.Metric($"{metricTypeSnakeCase}", metricValue, dimensionDict);
+            LogMetric(customDimensions);
+        }
```
chkeita, line 46: try this instead. It should save you the round trip of serialize deserialize
```C#
var dimensionDict = JsonSerializer.SerializeToNode(customDimensions, customDimensions.GetType(), _options ).AsObject().ToDictionary(kvp => kvp.Key, kvp =>  kvp.Value.ToString());
```

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/Metrics.cs
nharper285: Oh wow this is incredibly helpful. Thank you Cheick.

src/ApiService/ApiService/onefuzzlib/Metrics.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void SendMetric(int metricValue, BaseEvent customDimensions) {
+            var metricType = customDimensions.GetEventType();
+
+            _ = _options.PropertyNamingPolicy ?? throw new ArgumentException("Serializer _options not available.");
+
+            var metricTypeSnakeCase = _options.PropertyNamingPolicy.ConvertName($"{metricType}");
+
+            var dimensionNode = JsonSerializer.SerializeToNode(customDimensions, customDimensions.GetType(), _options);
+            _ = dimensionNode ?? throw new JsonException("Was not able to properly serialize the custom dimensions.");
+            var dimensionDict = dimensionNode.AsObject().ToDictionary(kvp => kvp.Key.ToString(), kvp => kvp.Value is not null ? kvp.Value.ToString() : "");
```
chkeita, line 43: You can filter out the null values before creating the dictionar. 

``` suggestion
var dimensionDict = dimensionNode.AsObject().Where(kvp => kvp.value is not null).ToDictionary(kvp => kvp.Key.ToString(), kvp =>  kvp.Value.ToString())
```

tevoinea (approved):

PR #3089, opened 2023-05-08
Deserialize user info when pulling from blob

tevoinea:

_What is this about?_

Our user info serializer fails when reading. We know though, that reading from the events container (which is controlled by onefuzz and only written to by onefuzz_ is safe. So this PR adds a special case when deserializing events from the events container.

## Validation Steps Performed

Added integration test that specifically exercises the failing code path.

### DIFF
--- a/src/ApiService/IntegrationTests/EventsTests.cs
+++ b/src/ApiService/IntegrationTests/EventsTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using Azure.Storage.Blobs;
 using FluentAssertions;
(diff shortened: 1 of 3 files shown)


src/ApiService/IntegrationTests/EventsTests.cs
```diff
@@ -62,4 +63,52 @@ public EventsTestBase(ITestOutputHelper output, IStorage storage)
         var eventData = await sr.ReadToEndAsync(); // read to make sure the SAS URL works
         eventData.Should().Contain(ping.PingId.ToString());
     }
+
+    [Fact]
+    public async Async.Task UserInfoIsDeserialized() {
```
tevoinea, line 68: If you undo the `_deserializingFromBlobOptions` change, this test fails with " reading UserInfo is not supported..." as expected.

mgreisen (approved):

PR #3102, opened 2023-05-12
Add more context to the deletion of nodes

chkeita:

Add a reason when deleting nodes. This will help us understand circumstances that lead to task being stopped while running


src/ApiService/ApiService/onefuzzlib/ScalesetOperations.cs
```diff
@@ -636,7 +636,7 @@ where x.State.ReadyForReset()
 
         if (scaleset.State == ScalesetState.Shutdown) {
             _log.Info($"scaleset shutting down, deleting rather than reimaging nodes {scaleset.ScalesetId:Tag:ScalesetId}");
-            return await DeleteNodes(scaleset, nodes);
+            return await DeleteNodes(scaleset, nodes, "node is shutting down");
```
tevoinea, line 639:
```suggestion
            return await DeleteNodes(scaleset, nodes, "scaleset is shutting down");
```

tevoinea (approved):

PR #3103, opened 2023-05-15
Fix bug when truncating certain reports

tevoinea:

_What is this about?_

I missed `ExecutableName` when I added the `..Math.Min` code. Took the chance to factor that out into a function as well.

### DIFF
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -468,23 +468,23 @@ public record Report(
     [JsonExtensionData] public Dictionary<string, JsonElement>? ExtensionData { get; set; }
     public Report Truncate(int maxLength) {
         return this with {
-            Executable = Executable[..maxLength],
-            CrashType = CrashType[..Math.Min(maxLength, CrashType.Length)],
-            CrashSite = CrashSite[..Math.Min(maxLength, CrashSite.Length)],
+            Executable = TruncateUtils.TruncateString(Executable, maxLength),
+            CrashType = TruncateUtils.TruncateString(CrashType, maxLength),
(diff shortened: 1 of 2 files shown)


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -468,23 +468,23 @@ public record Report(
     [JsonExtensionData] public Dictionary<string, JsonElement>? ExtensionData { get; set; }
     public Report Truncate(int maxLength) {
         return this with {
-            Executable = Executable[..maxLength],
-            CrashType = CrashType[..Math.Min(maxLength, CrashType.Length)],
-            CrashSite = CrashSite[..Math.Min(maxLength, CrashSite.Length)],
+            Executable = TruncateUtils.TruncateString(Executable, maxLength),
```
tevoinea, line 471: This is the fix

stishkin (approved):

chkeita (approved):

src/ApiService/ApiService/onefuzzlib/Utils.cs
```diff
@@ -45,4 +45,12 @@ public static class TruncateUtils {
...
         return data.TakeWhile(curr => (currentLength += curr.Length) <= maxLength).ToList();
     }
+
+    public static string TruncateString(string data, int maxLength) {
+        return data[..Math.Min(data.Length, maxLength)];
+    }
+
+    public static string? TruncateNullableString(string? data, int maxLength) {
+        return data?[..Math.Min(data.Length, maxLength)];
+    }
```
Porges, line 55: You could unify these with:

```suggestion
    [return: NotNullIfNotNull("data")]
    public static string? TruncateString(string? data, int maxLength) {
        return data?[..Math.Min(data.Length, maxLength)];
    }
```

PR #3127, opened 2023-05-23
Store authentication info in keyvault

chkeita:

Convert the Authentication fields to ISecret


chkeita:
Check-pr passed

src/ApiService/ApiService/onefuzzlib/Extension.cs
```diff
@@ -239,7 +239,8 @@ private sealed class Settings {
         var sep = pool.Os == Os.Windows ? "\r\n" : "\n";
 
         if (pool.Os == Os.Windows && scaleSet.Auth is not null) {
-            var sshKey = scaleSet.Auth.PublicKey.Trim();
+            var auth = await _context.SecretsOperations.GetSecretValue<Authentication>(scaleSet.Auth);
+            var sshKey = auth?.PublicKey.Trim();
```
tevoinea, line 243: Thoughts on using `!` instead of `?` here? I'm thinking we'd probably want to faily early and loud if we're expecting auth to be present for the scaleset but we don't find it.

Otherwise, it'll happily serialize to "null" and be a lot harder to debug

src/ApiService/ApiService/onefuzzlib/Secrets.cs
```diff
@@ -28,41 +30,51 @@ public class SecretsOperations : ISecretsOperations {
...
-            } else {
-                secretValue = JsonSerializer.Serialize(sValue.Value, EntityConverter.GetJsonSerializerOptions());
-            }
-
-            var kv = await StoreInKeyvault(GetKeyvaultAddress(), secretName.ToString(), secretValue);
-            return new SecretData<T>(new SecretAddress<T>(kv.Id));
         }
+        var kv = await StoreSecret(secretData.Secret);
+        return new SecretData<T>(new SecretAddress<T>(kv));
+    }
```
tevoinea, line 46: We don't need this because of the default implementation in `ISecretsOperations` right?

tevoinea (approved):
Non-blocking comments

src/ApiService/ApiService/onefuzzlib/Extension.cs
chkeita: added a null check

PR #3130, opened 2023-05-24
fix logic to set check_fuzzer_help

chkeita:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/cli/onefuzz/templates/libfuzzer.py
+++ b/src/cli/onefuzz/templates/libfuzzer.py
@@ -66,7 +66,6 @@ class Libfuzzer(Command):
         colocate_all_tasks: bool = False,
         colocate_secondary_tasks: bool = True,
         check_fuzzer_help: bool = False,
-        no_check_fuzzer_help: bool = False,
         expect_crash_on_failure: bool = False,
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -66,7 +66,6 @@ def _create_tasks(
         colocate_all_tasks: bool = False,
         colocate_secondary_tasks: bool = True,
         check_fuzzer_help: bool = False,
-        no_check_fuzzer_help: bool = False,
```
stishkin, line 69: no_check_fuzzer_help is the new one, and that's the one that we set


check_fuzzer_help <- is the one that should be removed, no ?

chkeita: This is the internal `_create_tasks` function. The bug was that the value of no_check_fuzzer_help ignored here. the template function (including q_emu) pass the value of `not no_check_fuzzer_help` to this function

stishkin (approved):

PR #3131, opened 2023-05-24
Don’t validate error codes on client side

Porges:
Stop `pydantic` from validating error codes returned from the service. This prevents the problem in #3129 from happening.

Include the error code name in the JSON returned from the service.


Porges:
Passed check-pr.

tevoinea (approved):

chkeita (approved):

PR #3133, opened 2023-05-25
Move auth into middleware

### LINKED ISSUE #2098: Move "CallIf" into Azure Function middleware
Utilize middleware to check whether azure function is allowed to be executed by a request initiating client.

[AB#36014](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/36014)

Porges:
Closes #2098.

This cleans up the authentication a bit; after this change we have two stages in the middleware pipeline:

- `AuthenticationMiddleware` reads the JWT token (it does not validate it, this is done by the Azure Functions service) and stores it in `FunctionContext.Items["ONEFUZZ_USER_INFO"]`
- `AuthorizationMiddleware` checks the user info against the `[Authorize]` attribute to see if the user has the required permissions
- Functions can read the user info from the `FunctionContext` if needed

The authorize attribute can be `[Authorize(Allow.User)]` or `Allow.Agent` or `Allow.Admin`. The `Admin` case is new and allows this to be declaratively specified rather than being checked in code. We have several functions which could be changed to use this (e.g. Pool POST/DELETE/PATCH, Scaleset POST/DELETE/PATCH), but I have only changed one so far (JinjaToScriban).

One of the benefits here is that this simplifies the test code a lot: we can set the desired user info directly onto our `(Test)FunctionContext` rather than having to supply a fake that pretends to parse the token from the HTTP request. This will also have benefits when running the service locally for testing purposes (refer to internal issue).

The other benefit is the ability to programmatically read the required authentication for each function, which may help with Swagger generation.


Porges:
Passes check-pr.

src/ApiService/ApiService/Auth/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    public AuthorizationMiddleware(IEndpointAuthorization auth, ILogTracer log) {
+        _auth = auth;
+        _log = log;
+    }
+
+    public async Async.Task Invoke(FunctionContext context, FunctionExecutionDelegate next) {
+        var attribute = GetAuthorizeAttribute(context);
+        if (attribute is not null) {
+            var req = await context.GetHttpRequestDataAsync() ?? throw new NotSupportedException("no HTTP request data found");
```
tevoinea, line 23: How does the middleware work with queue/event grid triggers? Do they even go through the auth flow in the first place?

src/ApiService/ApiService/Functions/Pool.cs
```diff
@@ -17,22 +18,27 @@ public class Pool {
...
+    private async Task<HttpResponseData> Delete(HttpRequestData r, FunctionContext context) {
         var request = await RequestHandling.ParseRequest<PoolStop>(r);
         if (!request.IsOk) {
             return await _context.RequestHandling.NotOk(r, request.ErrorV, "PoolDelete");
         }
 
-        var answer = await _auth.CheckRequireAdmins(r);
+        var user = context.GetUserAuthInfo();
+        var answer = await _auth.CheckRequireAdmins(user);
         if (!answer.IsOk) {
```
tevoinea, line 42: What do you think of:

1. Adding `CheckRequireAdmins` to the `Authorize` attribute
1. Splitting the endpoint into 2 definitions: `Get` and `Post, Delete, Patch`

tevoinea: The benefit being that we'd be able to include that information in swagger docs.

tevoinea (approved):

src/ApiService/ApiService/Auth/AuthorizationMiddleware.cs
Porges: Yes, they go through the same pipeline (in that case the `HttpRequestDataAsync` will return null). We don't use the Authorize attribute on any of those functions.

src/ApiService/ApiService/Functions/Pool.cs
Porges: Yes, there is `[Authorize(Allow.Admin)]`. I didn't split any existing functions to reduce the changes here, but it's probably better to do this as part of the PR.

Porges: Done this now.

src/ApiService/ApiService/Auth/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,103 @@
...
+    private static AuthorizeAttribute? GetAuthorizeAttribute(FunctionContext context) {
+        // fully-qualified name of the method
+        var entryPoint = context.FunctionDefinition.EntryPoint;
+        if (_authorizeCache.TryGetValue(entryPoint, out var cached)) {
+            return cached;
+        }
+
+        var lastDot = entryPoint.LastIndexOf('.');
+        var (typeName, methodName) = (entryPoint[..lastDot], entryPoint[(lastDot + 1)..]);
+        var assemblyPath = context.FunctionDefinition.PathToAssembly;
```
chkeita, line 92: Non-blocking comment.
We can fill the dictionary once by collecting all methods with the FunctionAttribute in the current assembly at the startup. IT looks like the context is only used to get the assembly info

chkeita (approved):

src/ApiService/ApiService/Auth/AuthorizeAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+public sealed class AuthorizeAttribute : Attribute {
+    public AuthorizeAttribute(Allow allow) {
+        Allow = allow;
+    }
+
+    public Allow Allow { get; set; }
+}
+
+public enum Allow {
```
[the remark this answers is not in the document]
Porges: I did start with a flags enum but at the moment all functions are for one of the three values, so I switched it to this to reduce complexity. We can always change it in future.

Porges:
Did some manual validation after check-pr. All looks good.

PR #3139, opened 2023-05-28
Use --locked when installing Rust pre-reqs

Porges:
CI builds have been broken by https://github.com/bluejekyll/enum-as-inner/issues/98. Try using `--locked` when setting up our pre-reqs so that we get the versions of the dependencies which the executables have been released (and presumably built) with.


drtimwright (approved):
LGTM

theunrepentantgeek (approved):

PR #3144, opened 2023-05-30
Fix bug for scale-in protection

tevoinea:

_What is this about?_

Fixes a bug where we return `OneFuzzResult.Ok` even if applying scale in protection wasn't OK.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

* Only return OK if acquiring scale-in protection was successful
* Add metrics

## Validation Steps Performed

Will be validating in check-pr

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
+++ b/src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
@@ -91,6 +91,9 @@ public class NodeOperations : StatefulOrm<Node, NodeState, NodeOperations>, INod
         if (node.ScalesetId is ScalesetId scalesetId &&
             await TryGetNodeInfo(node) is NodeInfo nodeInfo) {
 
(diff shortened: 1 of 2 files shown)


src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -110,17 +113,25 @@ public NodeOperations(ILogTracer log, IOnefuzzContext context)
             var r = await _context.VmssOperations.UpdateScaleInProtection(nodeInfo.Scaleset, instanceId, protectFromScaleIn: true);
             if (!r.IsOk) {
                 _logTracer.Error(r.ErrorV);
+                _logTracer.Metric($"FailedAcquiringScaleInProtection", 1, metricDimensions);
+                return r.ErrorV;
```
tevoinea, line 117: This is the fix

chkeita (approved):

Porges (approved):

PR #3145, opened 2023-05-30
Update requests version in CLI requirements to fix CVE

AdamL-Microsoft:

Updates Python requests version from 2.27.1 to 2.31.0 to fix CVE-2023-32681

### DIFF
--- a/src/cli/requirements.txt
+++ b/src/cli/requirements.txt
@@ -1,5 +1,6 @@
 msal==1.20.0
-requests~=2.27.1
+# install requests 2.31.0 to fix CVE-2023-32681
+requests==2.31.0
 jmespath~=0.10.0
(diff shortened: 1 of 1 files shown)


src/cli/requirements.txt
```diff
@@ -1,5 +1,6 @@
 msal==1.20.0
-requests~=2.27.1
+# install requests 2.31.0 to fix CVE-2023-32681
+requests==2.31.0
```
Porges, line 3: Do we want this to be a min bound?

AdamL-Microsoft: yeah good point

AdamL-Microsoft: @Porges - updated to be the minimum version allowed now

PR #3160, opened 2023-06-05
Downloading files to locally repro crashes

kananb:

Add a command to the CLI for downloading the files to locally repro a crash
Command:
`onefuzz repro get_files <report_container> <report_name> [--output_dir <path>] [--include-setup]`

## PR Checklist
* [ ] Tests added/passed

## Validation Steps Performed

Ran the command with correct and invalid arguments.

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -551,7 +551,7 @@ class Repro(Endpoint):
         report_container: primitives.Container,
         report_name: str,
         include_setup: bool = False,
-        output_dir: primitives.Directory = "."
+        output_dir: primitives.Directory = ".",
     ) -> None:
         """downloads the files necessary to locally repro the crash from a given report"""
         report_bytes = self.onefuzz.containers.files.get(report_container, report_name)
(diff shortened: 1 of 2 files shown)


src/cli/onefuzz/api.py
```diff
@@ -546,6 +546,38 @@ def get(self, vm_id: UUID_EXPANSION) -> models.Repro:
...
+            report["input_blob"]["name"],
+            os.path.join(output_dir, report["input_blob"]["name"]),
+        )
+
+        if include_setup:
+            setup_container = list(
+                self.onefuzz.jobs.containers.list(
+                    report["job_id"], enums.ContainerType.setup
+                )
+            ).pop()
```
Porges, line 576: Is there an unusual reason we wrap this in a list and then pop it out?

src/cli/onefuzz/api.py
```diff
@@ -546,6 +546,38 @@ def get(self, vm_id: UUID_EXPANSION) -> models.Repro:
...
+        output_dir: primitives.Directory = primitives.Directory("."),
+    ) -> None:
+        """downloads the files necessary to locally repro the crash from a given report"""
+        report_bytes = self.onefuzz.containers.files.get(report_container, report_name)
+        report = json.loads(report_bytes)
+
+        self.logger.debug(
+            "downloading files necessary to locally repro crash %s",
+            report["input_blob"]["name"],
+        )
```
Porges, line 563: This might be useful even in normal use, could raise it to INFO level?

src/cli/onefuzz/api.py
kananb: Nope, the goal here was just to get the lone key out of the returned dictionary. Besides using `[0]` instead of `pop()` are there any other better ways to do this?

chkeita (approved):

src/cli/onefuzz/api.py
kananb:
```py
def list(
        self,
        job_id: UUID_EXPANSION,
        container_type: enums.ContainerType,
    ) -> Dict[str, List[str]]:
        """
        List the files for all of the containers of a given container type
        for the specified job
        """
        ...
```
Despite the method name, it actually returns a dictionary.

Issue #3156, opened 2023-06-05
FindContainer should cache the storage account to avoid generating spurious 404s and for performance reasons


Porges:
https://github.com/microsoft/onefuzz/blob/f1c37d861eee3426319e3a03923f27863f0c76b1/src/ApiService/ApiService/onefuzzlib/Containers.cs#L126

[#AB152645](https://fuzzfest.visualstudio.com/Onefuzz/_workitems/edit/152645)

Porges:
Also related to #1419 where we currently do not store any container metadata.

PR #3162, opened 2023-06-06
Port logging to ILogger

stishkin:
image

Moving to ILogger, as discussed in meetings several times

### DIFF
--- a/src/ApiService/ApiService/Log.cs
+++ b/src/ApiService/ApiService/Log.cs
@@ -1,418 +1,435 @@
-﻿using System.Diagnostics;
-using System.Net;
-using System.Runtime.CompilerServices;
-using System.Text;
+﻿using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Globalization;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.OneFuzz.Service;
 
-//See: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/interpolated-string-handler
-//by ref struct works, but Moq does not support by-ref struct value so all the tests break... https://github.com/moq/moq4/issues/829
-[InterpolatedStringHandler]
-public struct LogStringHandler {
-
-    private readonly StringBuilder _builder;
-    private Dictionary<string, string>? _tags;
-
-    public LogStringHandler(int literalLength, int formattedCount) {
-        _builder = new StringBuilder(literalLength);
-        _tags = null;
-    }
-
-    public void AppendLiteral(string message) {
-        _builder.Append(message);
-    }
-
-    public void AppendFormatted<T>(T message) {
-        if (message is not null) {
-            _builder.Append(message.ToString());
-        } else {
-            _builder.Append("<null>");
-        }
-    }
-
-    public void AppendFormatted<T>(T message, string? format) {
-        if (format is not null && format.StartsWith("Tag:")) {
-            var tag = format["Tag:".Length..];
-            if (_tags is null) {
-                _tags = new Dictionary<string, string>();
-            }
-            _tags[tag] = $"{message}";
-            _builder.Append('{').Append(tag).Append('}');
-        } else if (message is IFormattable msg) {
-            _builder.Append(msg?.ToString(format, null));
-        } else {
-            _builder.Append(message?.ToString()).Append(':').Append(format);
-        }
-    }
-
-    private bool HasData => _builder is not null && _builder.Length > 0;
-
-    public override string ToString() => this.HasData ? _builder.ToString() : "<null>";
-    public IReadOnlyDictionary<string, string>? Tags => _tags;
+/// <summary>
+/// 
+/// </summary>
+public enum Telemetry {
+    /// <summary>
+    /// 
+    /// </summary>
+    Trace,
+    /// <summary>
+    /// 
+    /// </summary>
+    Exception,
+    /// <summary>
+    /// 
+    /// </summary>
+    Request,
+    /// <summary>
+    /// 
+    /// </summary>
+    Dependency,
+    /// <summary>
+    /// 
+    /// </summary>
+    PageView,
+    /// <summary>
+    /// 
+    /// </summary>
+    Availability,
+    /// <summary>
+    /// 
+    /// </summary>
+    Metric,
+    /// <summary>
+    /// 
+    /// </summary>
+    Event
 }
 
+/// <summary>
+/// 
+/// </summary>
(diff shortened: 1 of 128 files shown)


chkeita:
@stishkin Can you please add to the description the reason why we are moving to ILogger.

src/ApiService/ApiService/Log.cs
```diff
@@ -1,418 +1,435 @@
...
-        } else {
-            _builder.Append(message?.ToString()).Append(':').Append(format);
-        }
-    }
-
-    private bool HasData => _builder is not null && _builder.Length > 0;
-
-    public override string ToString() => this.HasData ? _builder.ToString() : "<null>";
-    public IReadOnlyDictionary<string, string>? Tags => _tags;
+/// <summary>
```
chkeita, line 10: if there are no decriptions can we remove the <summary> comments.

src/ApiService/ApiService/Program.cs
```diff
@@ -12,43 +12,66 @@
...
+            _ = activity.Start();
+            string correlationId = Guid.NewGuid().ToString();
+
+            if (await context.GetHttpRequestDataAsync() is HttpRequestData requestData) {
+                //if header has 1f-CorrelationId then use that
+                //otherwise check if message can be deserialized to {"correlationId": "SOME-GUID"}, then use that
+
+                if (requestData.Headers.TryGetValues("CorrelationId", out var values1f)) {
+                    correlationId = values1f.First();
+                } else if (requestData.Headers.TryGetValues("x-correlationId", out var values)) {
```
Porges, line 51: I think a more standard format for HTTP headers would look like "X-Correlation-ID"?

src/ApiService/ApiService/Program.cs
```diff
@@ -130,19 +145,33 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
-                builder.UseAzureAppConfiguration();
                 builder.UseMiddleware<LoggingMiddleware>();
+                builder.UseAzureAppConfiguration();
+
+                //this is a must, to tell the host that worker logging is done by us
+                builder.Services.Configure<WorkerOptions>(workerOptions => workerOptions.Capabilities["WorkerApplicationInsightsLoggingEnabled"] = bool.TrueString);
                 builder.AddApplicationInsights(options => {
-                    options.ConnectionString = $"InstrumentationKey={configuration.ApplicationInsightsInstrumentationKey}";
+                    options.DeveloperMode = false;
+                    options.EnableDependencyTrackingTelemetryModule = false;
```
Porges, line 171: Disabling dependency tracking?

src/ApiService/ApiService/Program.cs
stishkin: yeah 

https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

X-Correlation-ID,[[43]](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-44) 
Correlation-ID[[44]](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#cite_note-45)

src/ApiService/ApiService/Log.cs
stishkin: those are auto-populated/auto-created by Visual Studio

src/ApiService/ApiService/Log.cs
```diff
@@ -1,418 +1,435 @@
...
+            } else {
+                if ((this as ILogger).IsEnabled(logLevel)) {
+                    if (exception is null) {
+                        if (config.EnabledTelemetry is null || config.EnabledTelemetry.Contains(Telemetry.Trace)) {
+                            TraceTelemetry traceTelemetry = new TraceTelemetry(
+                                formatter(state, exception),
+                                OneFuzzLogger.GetSeverityLevel(logLevel));
+
+                            traceTelemetry.Context.Operation.Id = Activity.RootId;
+                            traceTelemetry.Context.Operation.ParentId = Activity.SpanId.ToString();
```
Porges, line 161: I'm wondering if this should be: 
```suggestion
                            traceTelemetry.Context.Operation.Id = Activity.Id;
                            traceTelemetry.Context.Operation.ParentId = Activity.Parent?.Id;
```
?

src/ApiService/ApiService/Log.cs
```diff
@@ -1,418 +1,435 @@
...
+/// </summary>
+public static class OneFuzzLoggerExt {
+    private static EventId EmptyEventId = new EventId(0);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <returns></returns>
+    public static string? GetCorrelationId(this ILogger _) {
+        foreach (var tag in OneFuzzLogger.Activity.Tags) {
```
Porges, line 275: TryGetValue?

src/ApiService/ApiService/Log.cs
Porges: Across all files, please 🙂

src/ApiService/ApiService/Program.cs
stishkin: we can enable - but it's sooooo noisy 

I will enable, but it will be a lot

src/ApiService/ApiService/Log.cs
stishkin: Span ID is correct

image

Root ID <- is most parent ID, which also should be correct since the most parent ID

src/ApiService/ApiService/Log.cs
stishkin: Tags are IEnumerable  of KeyValue pairs,.

It is allowed to add multiple values with the same key, and they will all add all of them.

There is no method TryGetValue at all in the Tags collection

src/ApiService/ApiService/Program.cs
Porges: We already have it enabled in the production deployment (which is how we noticed #3156, for example)

Porges (approved):

src/ApiService/ApiService/host.json
```diff
@@ -3,8 +3,8 @@
     "logging": {
         "applicationInsights": {
             "samplingSettings": {
-                "isEnabled": false,
-                "excludedTypes": "Request;Trace;Dependency;Event;Exception"
+                "isEnabled": true,
+                "excludedTypes": "Request"
```
tevoinea, line 7: I think these were changed recently by Noah to improve our sampling issues, does this PR negate the need for it in the first place?

stishkin: if you turn this off <- then there will be no logs at all

stishkin: in the previous implementation this was "extra" logs that were coming from "lower layers". In this implementation this controls all of logs

tevoinea (approved):

PR #3166, opened 2023-06-07
Make coverage-recording errors non-fatal

### LINKED ISSUE #3041: If an input crashes the fuzzer, the coverage task exits
## Information

- Onefuzz version: 7.0.0
- OS: Windows

I have a set of inputs, some of which crash my fuzzer when run in the context of a coverage task. When this happens, the coverage recording stops, and no other inputs get processed for coverage.

```
[2023-04-18 22:00:32.426955100 UTC] INFO: agent ready, dispatching task
[2023-04-18 22:00:32.427126300 UTC] task_start: event_type coverage
[2023-04-18 22:00:32.427161700 UTC] INFO: starting coverage task
[2023-04-18 22:01:08.593162400 UTC] WARN: error recording coverage for input = task_readonlyinputs_0\000d6e2ba72cb88fac1da4b61252fb6a90169178: coverage recording failed, child status = exit code: 0xc0000005
[2023-04-18 22:01:11.649161700 UTC] ERROR: error running task: failed to record coverage for input = task_readonlyinputs_0\000d6e2ba72cb88fac1da4b61252fb6a90169178 after 2 attempts

 

Caused by:
    coverage recording failed, child status = exit code: 0xc0000005
```

1) Can the error message be changed from "child status" to "target.exe status" or similar, it took me a while to figure out what the error meant.
2) After failing to record coverage for this input, it should continue on to processing the next input, not exit.

[AB#140609](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/140609)

Porges:
When we record coverage, if any of the files fail (e.g. due to timeouts during coverage recording), then we fail the whole process.

Instead, make coverage recording best-effort: if any files fail then we continue to record coverage using the other files. A warning is printed that we can monitor for any ongoing problems.

Closes #3041

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
@@ -373,12 +373,19 @@ impl<'a> TaskContext<'a> {
             match entry {
                 Ok(entry) => {
                     if entry.file_type().await?.is_file() {
-                        self.record_input(&entry.path()).await?;
-                        count += 1;
+                        if let Err(e) = self.record_input(&entry.path()).await {
+                            warn!(
+                                "ignoring error recording coverage for input: {}, error: {}",
+                                entry.path().display(),
+                                e
+                            );
+                        } else {
+                            count += 1;
 
-                        // make sure we save & sync coverage every 10 inputs
-                        if count % 10 == 0 {
-                            self.save_and_sync_coverage().await?;
+                            // make sure we save & sync coverage every 10 inputs
+                            if count % 10 == 0 {
+                                self.save_and_sync_coverage().await?;
+                            }
                         }
                     } else {
                         warn!("skipping non-file dir entry: {}", entry.path().display());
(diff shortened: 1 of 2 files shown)


src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -373,12 +373,19 @@ impl<'a> TaskContext<'a> {
...
+                            warn!(
+                                "ignoring error recording coverage for input: {}, error: {}",
+                                entry.path().display(),
+                                e
+                            );
+                        } else {
+                            count += 1;
+
+                            // make sure we save & sync coverage every 10 inputs
+                            if count % 10 == 0 {
```
mgreisen, line 386: Is this an expensive operation? It seems like an odd performance enhancement that might leave us with missing data if we don't wind up with an even 10 inputs.

Porges: We do a final sync at the end; this is to ensure we have some data while coverage is running.

tevoinea (approved):

chkeita (approved):

PR #3172, opened 2023-06-07
Skip entire function if entry offset excluded

### LINKED ISSUE #3170: Apply source allowlists to function entrypoint before disassembly
When computing coverage offsets for the functions in a debuggable module, if the entrypoint of a function is excluded by the source allowlist, skip it entirely. This enables reduction of analysis time for functions we explicitly don't care about. Right now, the allowlist only impacts what software breakpoints get set and used in coverage measurements.

ranweiler:
Closes #3170.


chkeita (approved):

tevoinea (approved):

ranweiler:
Locally tested that source allowlist exclusion now short-circuits function analysis. Integration tests pass.

PR #3177, opened 2023-06-07
Try to kill debuggee if Linux recording times out

### LINKED ISSUE #3174: Linux coverage tracing thread continues after timeout
When recording coverage on Linux using the `timed` timeout combinator , the worker thread (and tracee child process) may continue to run after timeout. Instead, expose the PID of the tracee process, and on timeout, kill it.

ranweiler:
Closes #3174.


ranweiler:
Manual testing:
- Modified `coverage` example to loop instead of exiting on recording errors (to allow observing dangling tracee child processes)
- Ran `coverage` against a target tracee modified to hang on any input
- Red/green testing: before changes, observed a dangling tracee subprocess on timeout. After changes, tracee was terminated across a wide range of timeout values.

ranweiler:
Integration tests passed, confirmed that reports look as expected.

PR #3205, opened 2023-06-20
Require input marker in args when given corpus dir

ranweiler:
Fail fast if the user passes an input corpus directory, but doesn't include an input marker (`@@`) in the command to execute. Before, this resulted in either an immediate spurious exit or, in the case of libFuzzer targets, a guaranteed timeout.


ranweiler:
Tested all combinations of `-d` present/absent and `@@` present/absent in command args.

tevoinea (approved):

PR #3208, opened 2023-06-21
Apply allowlist to source conversion

### LINKED ISSUE #3207: Excluded source files included in source coverage if inlined
A single module offset can represent many source-line coverage locations via a virtual stack of inlined functions. When converting binary coverage to source coverage, we initialize source locations for every inlinee. This does not take into account the original allowlist. As a result, excluded source files end up in the source coverage output.

While this doesn't interfere with recording, and never _excludes_ source files, we care about, it produces confusing and overly verbose output. This becomes really acute for Rust binaries, where debug builds include detailed debuginfo for the many statically-linked dependencies. Instead, plumb source allowlists through to the binary-to-source conversion functions, and only emit source-line coverage locations for inlinees if the allowlist includes them.

ranweiler:
Closes #3207.

### DIFF
--- a/src/agent/coverage/examples/record.rs
+++ b/src/agent/coverage/examples/record.rs
@@ -2,7 +2,7 @@ use std::process::Command;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use clap::Parser;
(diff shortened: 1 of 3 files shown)


tevoinea (approved):

ranweiler:
Manually inspected source coverage reports in integration tests.

PR #3218, opened 2023-06-22
Add cache for coverage locations and debuginfo

### LINKED ISSUE #3216: Enable caching debuginfo analysis
When recording coverage, debuginfo for a module is analyzed on-demand during coverage recording. It is not cached, and when testing multiple inputs for a single target, we reanalyze each module once for each input.

Instead, provide a mechanism for caching module coverage location and reusing that analysis it across runs. Note that the cache must be allowlist-sensitive.

[AB#154069](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/154069)

ranweiler:
Closes #3216.

This PR does _not_ implement caching or analysis reuse in the OneFuzz `coverage` task, but is a preqreq for doing so.

### DIFF
--- a/src/agent/coverage/examples/record.rs
+++ b/src/agent/coverage/examples/record.rs
@@ -93,7 +93,7 @@ fn main() -> Result<()> {
         let recorded = CoverageRecorder::new(cmd)
             .allowlist(allowlist.clone())
             .loader(loader)
-            .debuginfo_cache(cache.clone())
+            .debuginfo_cache(cache)
             .timeout(timeout)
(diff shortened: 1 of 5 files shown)


ranweiler:
Manually tested using the updated `record` example to ensure the cache is being hit when testing multiple inputs against the same target using the `-d` option.

ranweiler:
Integration tests passed, manually inspected coverage output files.

tevoinea (approved):

PR #3227, opened 2023-06-26
Add optional condition when updating/re-opening Work Items

tevoinea:

_What is this about?_

Work Items marked as "Won't Fix" or similar shouldn't be re-opened. Our AdoTemplate didn't have an affordance for that capability.

This PR adds an "Unless" field to on_duplicate. When _all_ of the fields and values in "Unless" match the work item, we will skip updating it.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -302,6 +305,17 @@ sealed class AdoConnector {
...
+        private async Async.Task<bool> MatchesUnlessCase(WorkItem workItem) =>
+            _config.OnDuplicate.Unless != null &&
+            await _config.OnDuplicate.Unless.ToAsyncEnumerable()
+                // Any condition from the list may match
+                .AnyAwaitAsync(async condition =>
+                    await condition.ToAsyncEnumerable()
+                        // All fields within the condition must match
+                        .AllAwaitAsync(async kvp =>
+                            workItem.Fields.TryGetValue<string>(kvp.Key, out var value) &&
+                            string.Equals(await Render(kvp.Value), value, StringComparison.OrdinalIgnoreCase)));
```
Porges, line 317: We should probably make the `Render` funciton non-async. Looking at the implementation, we invoke `Template.RenderAsync`, but we aren't benefiting from any async IO here (we are working entirely in-memory), we could invoke `Template.Render` instead, and this would simplify all the callers.

tevoinea: That's a good point, I can do that in a follow-up PR since I think it would cascade-remove a bunch async in our code.

Porges (approved):

PR #3238, opened 2023-06-27
log exception as "error" since we are retrying anyways

stishkin:


src/ApiService/ApiService/onefuzzlib/WebhookOperations.cs
```diff
@@ -269,7 +269,7 @@ public WebhookMessageLogOperations(ILogger<WebhookMessageLogOperations> log, IOn
             }
             return sendResult.IsOk;
         } catch (Exception exc) {
-            _logTracer.LogError(exc, "Send Webhook");
+            _logTracer.LogError("Send Webhook: {exception}", exc);
```
mgreisen, line 272: Does the retry code eventually throw an exception if we exhaust retries?

stishkin: No, we just log a warning:

_logTracer.LogWarning("sending webhook event failed, re-queued {WebhookId}:{EventId}", newMessage.WebhookId, newMessage.EventId);

stishkin: Good point though, we can log an exception so alerting picks this up

mgreisen (approved):
:shipit:

PR #3269, opened 2023-07-04
Set up .local file before invoking LibFuzzer

### LINKED ISSUE #3180: Improve DLL-redirection by using `.local` files
In #3032 we attempted to address a timing issue with `KnownDll` overrides. There might be a better way by using `.local` files: https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-redirection#how-to-redirect-dlls-for-unpackaged-apps

We should be able to do this unconditionally on Windows (I can't think of any cases where you'd want a system DLL to override an explicitly user-provided one), which would remove the need for the example `setup.ps1` file as well.

[#AB152694](https://fuzzfest.visualstudio.com/Onefuzz/_workitems/edit/152694)

Porges:
Closes #3180.


src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/generic.rs
```diff
@@ -37,7 +39,67 @@ impl common::LibFuzzerType for GenericLibFuzzer {
             config.common.machine_identity.clone(),
         ))
     }
+
+    async fn extra_setup(config: &common::Config<Self>) -> Result<()> {
```
mgreisen, line 43: How does this loop for all the DLL files that might be in the directory?

chkeita (approved):

src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/generic.rs
Porges: It's only needed for the entry executable

PR #3280, opened 2023-07-07
Use debuginfo cache

### LINKED ISSUE #3217: Eagerly analyze debuggable modules in coverage task
After the debuginfo analysis cache is implemented:
- Eagerly analyze `target_exe`
- Ideally, analyze any statically-observable dependencies with debuginfo
- Do this _before_ coverage recording, so analysis time does not count against the recording timeout (or does so minimally)

[AB#154070](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/154070)

tevoinea:

_What is this about?_

Closes #3217

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
@@ -121,7 +121,7 @@ impl CoverageTask {
                 .await?;
         let target_exe = target_exe_path
             .to_str()
-            .expect("Failed to resolve target_exe relative path");
+            .ok_or_else(|| anyhow::format_err!("target_exe path is not valid unicode"))?;
 
         let mut context = TaskContext::new(
             &self.config,
@@ -225,18 +225,20 @@ impl<'a> TaskContext<'a> {
         let cache = DebugInfoCache::new(allowlist.source_files.clone());
         let loader = Loader::new();
 
+        // Preload the cache with the target executable, to avoid counting debuginfo analysis
+        // time against the exeuction timeout for the first iteration.
         let module: Box<dyn Module> = LoadModule::load(&loader, FilePath::new(target_exe)?)?;
 
-        cache.get_or_insert(&*module)?;
-
-        let arccache = Arc::new(cache);
+        cache
+            .get_or_insert(&*module)
+            .context("Failed to load debuginfo for target_exe when populating DebugInfoCache")?;
 
         Ok(Self {
             config,
             coverage,
             allowlist,
             heartbeat,
-            cache: arccache,
+            cache: Arc::new(cache),
         })
     }
 
--- a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
@@ -5,15 +5,20 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
 use cobertura::CoberturaCoverage;
 use coverage::allowlist::{AllowList, TargetAllowList};
-use coverage::binary::BinaryCoverage;
+use coverage::binary::{BinaryCoverage, DebugInfoCache};
 use coverage::record::CoverageRecorder;
 use coverage::source::{binary_to_source_coverage, SourceCoverage};
+use debuggable_module::load_module::LoadModule;
+use debuggable_module::loader::Loader;
+use debuggable_module::path::FilePath;
+use debuggable_module::Module;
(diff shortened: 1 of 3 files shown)


mgreisen (approved):

mgreisen:
Do we have tests for these changes?

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -106,17 +111,32 @@ impl CoverageTask {
...
-        let mut context = TaskContext::new(&self.config, coverage, allowlist, heartbeat);
+
+        let mut seen_inputs = false;
+
+        let target_exe_path =
+            try_resolve_setup_relative_path(&self.config.common.setup_dir, &self.config.target_exe)
+                .await?;
+        let target_exe = target_exe_path
+            .to_str()
+            .expect("Failed to resolve target_exe relative path");
```
ranweiler, line 124: This should currently be unreachable, since we deserialized `target_exe` from a JSON file. However, in case things change, let's convert to a `Result` with an accurate error message:
```suggestion
            .ok_or_else(|| anyhow::format_err!("target exe path is not valid unicode"))?;
```

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -199,13 +220,24 @@ impl<'a> TaskContext<'a> {
...
         allowlist: TargetAllowList,
         heartbeat: Option<TaskHeartbeatClient>,
-    ) -> Self {
-        Self {
+        target_exe: String,
+    ) -> Result<Self> {
+        let cache = DebugInfoCache::new(allowlist.source_files.clone());
+        let loader = Loader::new();
+
+        let module: Box<dyn Module> = LoadModule::load(&loader, FilePath::new(target_exe)?)?;
```
ranweiler, line 228: Let's document the motivation for what's happening here:
```suggestion
        // Preload the cache with the target executable, to avoid counting debuginfo analysis 
        // time against the exeuction timeout for the first iteration.
        let module: Box<dyn Module> = LoadModule::load(&loader, FilePath::new(target_exe)?)?;
```
Note that in the future, it might actually make sense to do this for every executable module file with debuginfo in the setup container.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -199,13 +220,24 @@ impl<'a> TaskContext<'a> {
...
+
+        cache.get_or_insert(&*module)?;
```
ranweiler, line 230: By using `?` here, we bail on error. One type of error is that the module simply has no debuginfo. In almost all real life cases, this would signal an error in job submission. However, in general, we could imagine that the `target_exe` is a libfuzzer binary whose true target code is in some DLL that gets dynamically linked or loaded.

Question for the OneFuzz team to consider: is it a hard requirement that the `target_exe` include debuginfo?

I think it is probably fair to say yes, and it may even already be de facto true (I don't remember). But that should be understood and agreed upon.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -199,13 +220,24 @@ impl<'a> TaskContext<'a> {
...
+
+        let arccache = Arc::new(cache);
```
ranweiler, line 232: Style suggestion:
```suggestion
```

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -199,13 +220,24 @@ impl<'a> TaskContext<'a> {
...
+
+        Ok(Self {
             config,
             coverage,
             allowlist,
             heartbeat,
-        }
+            cache: arccache,
```
ranweiler, line 239: Style suggestion:
```suggestion
            cache: Arc::new(cache),
```

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
ranweiler: Regardless: consider adding a `context()` call here to note that this error is occurring in pre-caching, not during an iteration.

ranweiler (changes requested):

ranweiler:
Please document how you observed that the pre-cache step is successfully shortening the runtime of the first coverage iteration.

It would also be interesting to test the error path for when `target_exe` is missing debuginfo.

Finally, if there isn't already an issue for caching other debuggable modules in the `setup` container, I'd consider adding one. This would help "factor out" (perf-wise) the module analysis for any linked and loaded libraries that we actually care about.

tevoinea:
Using the rust example binary from onefuzz-samples.

No cache `record_input` elapsed time first run: 13 s | After first run average time: 930 ms
Cache time for first run: 1.2s                                      | After first run average time: 930 ms

From 13 seconds down to 1.2, over 10 times faster

ranweiler (approved):

Issue #3015, opened 2023-04-12
Trim work item title if longer than what is accepted by ADO


tevoinea:
## Information

- Onefuzz version: 7.0.0
- OS: N/A

## Provide detailed reproduction steps (if any)

1. Saw error `TF401324: Value for field 'System.Title' is too long.`

Max name length is 128 characters.

We should look over the naming restrictions linked in this comment to see if there's anything else we should be trimming:

https://github.com/MicrosoftDocs/azure-devops-docs/issues/5890#issuecomment-539632059

[AB#46654](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/46654)

PR #3284, opened 2023-07-11
Trim System.Title if length is > 128

kananb:
Trim System.Title when longer than 128 characters. Trims some extra characters to append the first 8 chars of the title hash for collision avoidance.

We probably want to write the full System.Title value somewhere in the bug so that the trimmed information isn't just lost.

_What is this about?_
https://github.com/microsoft/onefuzz/issues/3015

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
+++ b/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
@@ -401,12 +401,9 @@ public class Ado : NotificationsBase, IAdo {
                     _config.AdoFields[reproStepsKey] = $"# {systemTitle}<br>{_config.AdoFields[reproStepsKey]}";
                 }
 
-                var systemTitleHashString = System.Security.Cryptography.SHA256.HashData(Encoding.UTF8.GetBytes(systemTitle))
-                    .Aggregate(
-                        new StringBuilder(),
-                        (builder, next) => builder.Append(next.ToString("x2")),
-                        builder => builder.ToString()
-                    );
+                var systemTitleHashString = Convert.ToHexString(
+                    System.Security.Cryptography.SHA256.HashData(Encoding.UTF8.GetBytes(systemTitle))
+                );
                 // try to avoid naming collisions caused by the trim by appending the first 8 characters of the title's hash at the end
                 _config.AdoFields["System.Title"] = $"{systemTitle[..(MAX_SYSTEM_TITLE_LENGTH - 14)]}... [{systemTitleHashString[..8]}]";
             }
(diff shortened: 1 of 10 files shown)


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -396,6 +400,15 @@ public sealed class AdoConnector {
 
                 if (string.Equals(field, "System.Tags")) {
                     value += ";Onefuzz";
+                } else if (string.Equals(field, "System.Title") && value.Length > MAX_SYSTEM_TITLE_LENGTH) {
+                    var titleHashString = System.Security.Cryptography.SHA256.HashData(Encoding.UTF8.GetBytes(value))
+                        .Aggregate(
```
Porges, line 405: You can use `Convert.ToHexString` here

Porges:
Agree about this:
> We probably want to write the full System.Title value somewhere in the bug so that the trimmed information isn't just lost.

Could we add an additional field to the template for the full title and always include it?

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: To replace the hash? Wouldn't that risk collision for titles starting with the same first 128 chars?

kananb:
> Could we add an additional field to the template for the full title and always include it?
Something like a unique field with the full title in it? Where are these and the other non-custom fields visible?

chkeita (approved):

Porges:
> > Could we add an additional field to the template for the full title and always include it?
> 
> We can probably put it at the top of the bug's "Repro Steps". I don't think we have much control over the fields since they're org specific.

Sorry, I was imprecise here, I meant a variable for the Scriban template to render, not the ADO fields.

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
Porges: To replace the conversion to hex:

```cs
hash.Aggregate(
  new StringBuilder(),
  (builder, next) => builder.Append(next.ToString("x2")),
  builder => builder.ToString())
```
  
Is equivalent to (aside from being in a different case):
```cs
Convert.ToHexString(hash)
```

Porges (approved):

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: Thanks for the clarification! I probably should have been able to figure that `ToHexString` would be for the hex string conversion part.

mgreisen:
@Kananb how did we decide on the 128 char limit?

kananb:
> @kananb how did we decide on the 128 char limit?

https://github.com/MicrosoftDocs/azure-devops-docs/issues/5890#issuecomment-539632059

src/ApiService/ApiService/onefuzzlib/notifications/NotificationsBase.cs
```diff
@@ -85,6 +87,7 @@ public class Renderer {
             return new Renderer(
                 container,
                 filename,
+                issueTitle,
```
tevoinea, line 90: Let's render the title, untrimmed.

Issue #1419, opened 2021-11-01
Store container data as a resource in the database


chkeita:
There is currently no backing resources for the container in the onefuzz database. 
By storing the information, we can gate access to specific resource with the permissions defined in the #1413   

AB#35951

PR #3304, opened 2023-07-19
Cache FindContainer

### LINKED ISSUE #3156: FindContainer should cache the storage account to avoid generating spurious 404s and for performance reasons

Porges:
Closes #3156.

From now on, we will store additional information about containers in a table (`ContainerInformation`). This is populated when a container is created, or lazily when looking for a container we have not yet stored in the table.

Reads from this table are also cached in memory for a short period of time (currently 10 minutes).

The table is currently used to speed up lookups for containers across storage accounts; we round-robin creation of containers across multiple storage accounts to spread load, so when we read from a container we need to find what storage account it lives in.

The `ContainerInformation` table could also be used in future to store information about access control (see, e.g. #1419).

### DIFF
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -173,7 +173,7 @@ public record Error(ErrorCode Code, List<string>? Errors) {
         => new(code, errors.ToList());
 
     public sealed override string ToString() {
-        var errorsString = Errors != null ? string.Concat("; ", Errors) : string.Empty;
+        var errorsString = Errors != null ? string.Join("; ", Errors) : string.Empty;
         return $"Error {{ Code = {Code}, Errors = {errorsString} }}";
     }
 };
--- a/src/ApiService/ApiService/OneFuzzTypes/Model.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Model.cs
@@ -767,6 +767,11 @@ public record ClientCredentials
     string ClientSecret
 );
 
+public record ContainerInformation(
+    [PartitionKey] StorageType Type,
+    [RowKey] Container Name,
+    string ResourceId // full ARM resource ID for the container
+) : EntityBase;
 
 public record AgentConfig(
     ClientCredentials? ClientCredentials,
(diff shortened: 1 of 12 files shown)


src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -173,7 +173,7 @@ public static Error Create(ErrorCode code, params string[] errors)
         => new(code, errors.ToList());
 
     public sealed override string ToString() {
-        var errorsString = Errors != null ? string.Concat("; ", Errors) : string.Empty;
+        var errorsString = Errors != null ? string.Join("; ", Errors) : string.Empty;
```
Porges, line 176: Fixed a bug with error display here

src/ApiService/ApiService/OneFuzzTypes/ReturnTypes.cs
```diff
@@ -2,6 +2,16 @@
...
 namespace Microsoft.OneFuzz.Service {
 
+    public static class Result {
+        public readonly record struct OkPlaceholder();
+        public readonly record struct OkPlaceholder<T>(T Value);
+        public readonly record struct ErrPlaceholder<T>(T Value);
+
+        public static OkPlaceholder Ok() => new();
+        public static OkPlaceholder<T> Ok<T>(T value) => new(value);
+        public static ErrPlaceholder<T> Error<T>(T value) => new(value);
```
Porges, line 12: This is an improvement-in-passing to improve the ergonomics of using the result types. See the changes in `Orm.cs` below.

src/ApiService/IntegrationTests/TestLogTracer.cs
```diff
@@ -146,36 +146,30 @@ public class OneFuzzLogger : ILogger {
...
-                if (exception is null) {
-                    TraceTelemetry traceTelemetry = new TraceTelemetry(
-                        formatter(state, exception),
-                        OneFuzzLogger.GetSeverityLevel(logLevel));
-
-
-                    traceTelemetry.Context.Operation.Id = Activity.RootId;
-                    traceTelemetry.Context.Operation.ParentId = Activity.SpanId.ToString();
-                    this.PopulateTelemetry(traceTelemetry, state, eventId);
-                    _output.WriteLine($"[Trace] {traceTelemetry}");
```
Porges, line 160: fixing an issue here where the test log would only show the type name of TraceTelemetry

PR #3306, opened 2023-07-19
Add mariner support

### LINKED ISSUE #1955: Implement startup profiles for different Linux distros
Functionally, we let users create pools with _any_ OS image in Azure. However, the OneFuzz agents will only work if the startup scripts in `runtime-tools` are well-defined, and all dependencies are available. On Linux, that is only currently ensured for Ubuntu 18.04 and 20.04, and probably various recent Debian variants.

Instead, support a notion of distro _profiles_. These could have short names with varying levels of specificity. For example:
- `ubuntu-based`
- `ubuntu-1804`
- `apt`
- `fedora-35`
- `custom`
- `cbl-mariner`
Et cetera.

The `custom` profile could require users to include their own pre-init scripts or sources of certain packages, but would at least allow truly using arbitrary images. 

AB#35922

### LINKED ISSUE #3090: [Meta] Mariner host support
This is a meta issue tracking the work required for adding support for mariner.

- [ ] #1955 
- [ ] User should be able to specify profile from CLI when creating a pool

* The Azure VM Extensions we're currently using for ubuntu are already supported by Mariner 2.0 ([security extension](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#requirements), [azure monitor](https://learn.microsoft.com/en-us/azure/aks/use-mariner#limitations))
* We do not need to recompile the agent on a mariner host for distribution.

[AB#151950](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/151950)

### LINKED ISSUE #3372: Add logging support for mariner
Our current OMS extension for linux doesn't support mariner. We need to update to the Azure Monitor logging agent for linux extension and make sure that the logs are still accessible.

[AB#163007](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/163007)

tevoinea:

_What is this about?_

Closes #3372 
Closes #3090 
Closes #1955

This PR enables us to create a mariner linux pool and run fuzzing tasks on it.

## Info on Pull Request

_What does this include?_

* Updated `setup.sh` to support mariner
* Updated `integration-test.py` to test libfuzzer on mariner
* Removed `OmsAgentForLinux` extension since it's not supported on mariner

## Validation Steps Performed

Ran check-pr with mariner, ubuntu and windows test and it's all green.


AdamL-Microsoft (approved):

src/ApiService/ApiService/onefuzzlib/Extension.cs
```diff
@@ -329,37 +331,21 @@ private sealed class Settings {
...
-                TypePropertiesType = "MicrosoftMonitoringAgent",
-                Publisher = "Microsoft.EnterpriseCloud.Monitoring",
-                TypeHandlerVersion = "1.0",
-                AutoUpgradeMinorVersion = true,
-                Settings = new BinaryData(extensionSettings),
-                ProtectedSettings = new BinaryData(protectedExtensionSettings),
-                EnableAutomaticUpgrade = false
-            };
-        } else if (vmOs == Os.Linux) {
-            return new VMExtensionWrapper {
```
chkeita, line 349: isn't this a required extension?
cc: @nharper285

src/integration-tests/integration-test.py
```diff
@@ -97,6 +98,7 @@ class Integration(BaseModel):
         target_exe="fuzz.exe",
         inputs="seeds",
         wait_for_files={ContainerType.unique_reports: 1},
+        pool="linux",
```
chkeita, line 101: The pool should not be hardcoded here. the test should be independent of the pool as long as it has the right OS. This will allow us to run all linux tests on mariner

src/ApiService/ApiService/onefuzzlib/Extension.cs
tevoinea: OMS extension is going to be deprecated in 2024, the Azure Monitor Agent for Linux is the recommended path now (which we have installed) https://learn.microsoft.com/en-us/azure/azure-monitor/agents/azure-monitor-agent-migration

src/integration-tests/integration-test.py
tevoinea: We can still run all linux tests on mariner, I just need to update the integration tests lists to add those tests. How are you imagining that we select which OS a test should run on?

chkeita: > We can still run all linux tests on mariner, I just need to update the integration tests lists to add those tests

looks like we will end up diplicating the linux definitions for the mariner.  Ideally we shoudl use the exact same definition

> How are you imagining that we select which OS a test should run on?

I was thinking of having an extra paramter in the target definition which would be the list of OS flabors where this test can run. Something like 

```python
"linux-trivial-crash-afl": Integration(
        template=TemplateType.afl,
        os=OS.linux,
        target_exe="fuzz.exe",
        inputs="seeds",
        wait_for_files={ContainerType.unique_reports: 1},
        os_flavors=["mariner", "ubuntu"],
)
```
and use this information to select the pool where it can run. But ultimately this might be a bigger change than i anticipated. So, take this as just a non-blocking suggestion.

chkeita (approved):

PR #3322, opened 2023-07-21
Fuzz coverage recording

tevoinea:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/coverage/fuzz/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[profile.release]
+debug = 1
+
+[[bin]]
+name = "fuzz_target_1"
```
mgreisen, line 27: I think you have to change the name and target values to match your actual filename.

tevoinea: Good catch!

src/agent/coverage/fuzz/fuzz_targets/fuzz_target_record_coverage.rs
```diff
@@ -0,0 +1,51 @@
...
+        return;
+    }
+
+    // Write mutated bytes to a file
+    let mut file = NamedTempFile::new_in(env::current_dir().unwrap()).unwrap();
+    file.write_all(data);
+    let path = String::from(file.path().to_str().unwrap());
+
+    // Make sure the file is executable
+    Command::new("chmod").args(["+wrx", &path]).spawn().unwrap();
```
Porges, line 32: This should be waited upon to finish:
```suggestion
    Command::new("chmod").args(["+wrx", &path]).spawn().unwrap().wait();
```

I think also that once this is done you don't need the `file.keep` or `fs::remove_file` calls; the file should be kept until the end of the function.

tevoinea: The reason I had `file.keep`/`fs::remove_file` was that tempfile kept a handle open on the file (and deleting the file on `drop`) which was causing issues with our binary inspection.

mgreisen (approved):

PR #3325, opened 2023-07-23
Fix resolution of sibling .NET DLLs

### LINKED ISSUE #3326: Loading .NET DLL with dependencies fails
[AB#162846](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/162846)

Porges:
When `LibFuzzerDotNetLoader` loads DLLs that have dependencies, they are not currently resolved correctly, as sibling DLLs in the same directory are not considered. In order to do this, we need to provide a custom `AssemblyLoadContext`. This PR does so, based on the [dotnet core example code](https://learn.microsoft.com/en-us/dotnet/core/tutorials/creating-app-with-plugin-support).

Validated with custom fuzzer implementation; we don't yet have a full test for this in check-pr.

Closes #3326.


mgreisen (approved):
:shipit:

tevoinea (approved):

PR #3329, opened 2023-07-24
Render System.Title before trying to trim

kananb:

The [last fix for trimming System.Title](https://github.com/microsoft/onefuzz/pull/3284) wasn't much of a fix as it checked the pre-rendered System.Title length.
This PR fixes that issue by rendering the title in ConstructRenderer() and giving the rendered title a public accessor.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

System.Title trim fix and some logging that happens when the title is too long.

## Validation Steps Performed

Deployed changes to a manual instance and filed a bug that triggers the updated code.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/notifications/GithubIssues.cs
+++ b/src/ApiService/ApiService/onefuzzlib/notifications/GithubIssues.cs
@@ -3,7 +3,7 @@ using Octokit;
 namespace Microsoft.OneFuzz.Service;
 
 public interface IGithubIssues {
-    Async.Task GithubIssue(GithubIssuesTemplate config, Container container, IReport reportable, Guid notificationId, Uri instanceUrl);
+    Async.Task GithubIssue(GithubIssuesTemplate config, Container container, IReport reportable, Guid notificationId);
 }
 
 public class GithubIssues : NotificationsBase, IGithubIssues {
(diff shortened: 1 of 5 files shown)


src/ApiService/ApiService/onefuzzlib/notifications/GithubIssues.cs
```diff
@@ -3,15 +3,15 @@
...
+    Async.Task GithubIssue(GithubIssuesTemplate config, Container container, IReport reportable, Guid notificationId, Uri instanceUrl);
 }
 
 public class GithubIssues : NotificationsBase, IGithubIssues {
 
     public GithubIssues(ILogger<GithubIssues> logTracer, IOnefuzzContext context)
     : base(logTracer, context) { }
 
-    public async Async.Task GithubIssue(GithubIssuesTemplate config, Container container, IReport reportable, Guid notificationId) {
+    public async Async.Task GithubIssue(GithubIssuesTemplate config, Container container, IReport reportable, Guid notificationId, Uri instanceUrl) {
```
tevoinea, line 14: You have access to `_context` in this function so we don't need to change the signature. You can get the isntanceUrl the same way you got it in notificationoperations by doing `_context.Creds.GetInstanceUrl()`

src/ApiService/ApiService/onefuzzlib/notifications/NotificationsBase.cs
```diff
@@ -37,6 +37,7 @@ public class Renderer {
         private readonly Report _report;
         private readonly Container _container;
         private readonly string _filename;
+        public string IssueTitle => _issueTitle;
```
tevoinea, line 40: Make sure this is public get, private set. We don't want it to be modifiable.

kananb: This declaration only expands to `public string IssueTitle { get; }`, keeping the value immutable from the outside.
image

src/ApiService/ApiService/onefuzzlib/notifications/GithubIssues.cs
kananb: Thanks for pointing this out :)

tevoinea (approved):

PR #3334, opened 2023-07-24
Fix #3309

### LINKED ISSUE #3309: `target_timeout` not set when creating task with libfuzzer template
## Information

- Onefuzz version: 8.5
- OS: N/A

1. Create a new job with `onefuzz libfuzzer template basic .... --target_timeout 360`
2. Get the a task that was created `onefuzz tasks get {task id}`
3. Notice the `target_timeout` isn't present in the response

It's not in the storage table either so we never store it

### Expected result

### Actual result

[AB#162552](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/162552)

Porges:
LibFuzzer basic template was missing `target_timeout`

### DIFF
--- a/src/cli/onefuzz/templates/libfuzzer.py
+++ b/src/cli/onefuzz/templates/libfuzzer.py
@@ -500,6 +500,7 @@ class Libfuzzer(Command):
             analyzer_env=analyzer_env,
             tools=tools,
             task_env=task_env,
+            target_timeout=target_timeout,
         )
 
         self.logger.info("done creating tasks")


mgreisen (approved):

src/cli/onefuzz/templates/libfuzzer.py
```diff
@@ -500,6 +500,7 @@ def basic(
             analyzer_env=analyzer_env,
             tools=tools,
             task_env=task_env,
+            target_timeout=target_timeout,
```
chkeita, line 503: you also need to fix the invocation on line 163. `target_timeout` is missing from the list of argument there too

Porges: I think this is okay: the LibFuzzerFuzz task definition doesn't support timeout (it doesn't have the required feature).

Issue #3335, opened 2023-07-25
report.crash_site returns frame 0 of the stack. Sometimes its not relevant


mgreisen:
When returning the `crash_site` we should do some analysis and walk the frame down to what's relevant removing `asan` code.

Porges:
What we call `crash_site` is really (in most cases) currently the `SUMMARY` line given to us by ASan.

This generally has the format `"AddressSanitizer: {fault_type} ({top_frame})"`.

If we want to point to a relevant stack frame instead of the one that ASan reports, we could instead generate our own in a very similar format: `"{crash_finder}: {fault_type} ({top_frame_of_minimized_stack})"`. This should reduce the amount of churn during changeover between the old and new formats.

PR #3336, opened 2023-07-25
Exclude ASAN runtime from minimized stacktrace

Porges:
Starts to address #3335. 

On Linux, the ASAN runtime was being excluded via module name; on Windows this was not done, but ASAN functions were excluded by function name. However, when symbols are not available this doesn't function properly, so also exclude by module name.

### DIFF
--- a/src/agent/libclusterfuzz/src/generated.rs
+++ b/src/agent/libclusterfuzz/src/generated.rs
@@ -210,4 +210,7 @@ pub const STACK_FRAME_IGNORE_REGEXES: &[&str] = &[
     r".*libc\+\+_shared\.so",
     r".*libstdc\+\+\.so",
     r".*libc-.*\.so",
+    // OneFuzz added: ignore ASAN runtime on Windows
+    r"clang_rt.asan_dynamic-x86_64.dll$",
+    r"clang_rt.asan_dbg_dynamic-x86_64.dll$",
 ];
--- a/src/agent/libclusterfuzz/src/lib.rs
+++ b/src/agent/libclusterfuzz/src/lib.rs
@@ -18,8 +18,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[macro_use]
-extern crate lazy_static;
+use std::sync::OnceLock;
 
 use regex::RegexSet;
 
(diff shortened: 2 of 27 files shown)


src/agent/libclusterfuzz/src/lib.rs
```diff
@@ -28,11 +27,12 @@ mod generated;
 pub fn get_stack_filter() -> &'static RegexSet {
     // NOTE: this is build upon first use, but we've verified these will compile
     // using unit tests
-    lazy_static! {
-        static ref RE: RegexSet = RegexSet::new(generated::STACK_FRAME_IGNORE_REGEXES)
-            .expect("libclusterfuzz regex compile error");
-    }
-    &RE
+    static ONCE: OnceLock<RegexSet> = OnceLock::new();
```
Porges, line 30: Updated this since `OnceLock` is now part of the stdlib.

src/agent/libclusterfuzz/src/generated.rs
```diff
@@ -210,4 +210,7 @@ pub const STACK_FRAME_IGNORE_REGEXES: &[&str] = &[
     r".*libc\+\+_shared\.so",
     r".*libstdc\+\+\.so",
     r".*libc-.*\.so",
+    // OneFuzz added: ignore ASAN runtime on Windows
+    r"clang_rt.asan_dynamic-x86_64.dll$",
+    r"clang_rt.asan_dbg_dynamic-x86_64.dll$",
```
Porges, line 215: This is the key fix for the `asan-missing-symbols` test below.

tevoinea (approved):
Is this going to make debugging asan bugs harder since we're going to filter it out of the stack trace? Or do we still keep the whole unfiltered stack trace anyway?

mgreisen (approved):

Porges:
> @tevoinea Is this going to make debugging asan bugs harder since we're going to filter it out of the stack trace? Or do we still keep the whole unfiltered stack trace anyway?

We keep both around. At the moment I believe we don't actually _use_ the minimized stack trace anywhere. For bug filing deduplication we always use the full stack trace hash. However, I'll start using the top stack frame in #3337

Issue #3338, opened 2023-07-25
`get_files` in the CLI needs to support regression reports


tevoinea:
## Information

- Onefuzz version: 8.5.0
- OS: N/A

## Provide detailed reproduction steps (if any)

1. Execute `onefuzz repro get_files` on a regression report

```
*snip*
ERROR:cli:traceback: Traceback (most recent call last):
ERROR:cli:traceback:   File "onefuzz\cli.py", line 602, in execute_api
ERROR:cli:traceback:   File "onefuzz\cli.py", line 77, in call_func
ERROR:cli:traceback:   File "onefuzz\api.py", line 562, in get_files
ERROR:cli:traceback: KeyError: 'input_blob'
ERROR:cli:traceback:
ERROR:cli:command failed: input_blob
```

[AB#162907](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/162907)

kananb:
https://github.com/microsoft/onefuzz/pull/3340

PR #3340, opened 2023-07-25
Update `repro get_files` to handle regression reports

kananb:

https://github.com/microsoft/onefuzz/issues/3338
`onefuzz repro get_files` only handles `reports` and not `regression_reports`.

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

Adds logic for checking that the fields expected in a normal report exist before falling back on fields expected from a regression report. An error is logged otherwise.

## Validation Steps Performed

### DIFF
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -612,6 +612,7 @@ class Repro(Endpoint):
                     crash_info["job_id"], enums.ContainerType.setup
                 )
             )[0]
+
             self.onefuzz.containers.files.download_dir(
                 primitives.Container(setup_container), output_dir
             )
--- a/src/cli/onefuzz/api.py
+++ b/src/cli/onefuzz/api.py
@@ -562,20 +562,19 @@ class Repro(Endpoint):
             "input_blob_name": None,
             "job_id": None,
         }
-        if "input_blob" not in report:
-            if "original_crash_test_result" in report:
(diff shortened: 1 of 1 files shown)


src/cli/onefuzz/api.py
```diff
@@ -557,21 +557,51 @@ def get_files(
...
+        crash_info = {
+            "input_blob_container": None,
+            "input_blob_name": None,
+            "job_id": None,
+        }
+        if "input_blob" in report:
+            crash_info["input_blob_container"] = report["input_blob"]["container"]
+            crash_info["input_blob_name"] = report["input_blob"]["name"]
+            crash_info["job_id"] = report["job_id"]
+        elif "original_crash_test_result" in report:
```
tevoinea, line 569: `original_crash_test_result` is optional while `crash_test_result` is required for a `RegressionReport`, see the model here:
https://github.com/microsoft/onefuzz/blob/62c3f079af465316f1f6e360142774847918ad29/src/ApiService/ApiService/OneFuzzTypes/Model.cs#L544-L548

But, since we need the input blob from the original report, we should just check that both crash_test_result and original_crash_test_result are present before continuing and reporting an error otherwise.

Ex:

```python
elif "original_crash_test_result" in report and "crash_test_result" in report:
    # assume both are present
else:
    # log error
```

tevoinea (approved):

PR #3341, opened 2023-07-25
Update redirect URI if deployment uses a custom domain

### LINKED ISSUE #3228: Update the redirect uri if using a custom domain
## Information

- Onefuzz version: 8.4
- OS:

The redirect URI on the app registration, Authentication tab does not have the custom domain name. 
It should be added here https://github.com/microsoft/onefuzz/blob/512bb078b76aae36b700d340ba9e9ee8673cc3f4/src/deployment/deploy.py#L563 during deployment. 

We should add ` https://<customDomain>/.auth/login/aad/callback` as a redirect URI if the custom domain exists. 

[AB#155293](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/155293)

AdamL-Microsoft:

Deployments with a custom domain should have that custom domain as a URI in the associated App Registration.

Fixes #3228

## Info on Pull Request

updated:
* src/deployment/deploy.py

## Validation Steps Performed

verified deployment has correct URI on app registration in manual OneFuzz deployments


mgreisen (approved):

src/deployment/deploy.py
```diff
@@ -536,6 +536,13 @@ def create_new_app_registration(
     ) -> Dict[str, Any]:
         logger.info("creating Application registration")
 
+        appRegRedirectUris = [f"{self.get_instance_url()}/.auth/login/aad/callback"]
+
+        if self.custom_domain:
+            appRegRedirectUris.extend(
```
Porges, line 542: as a stylistic thing I'd use `append` instead of `extend` with one item

Porges (approved):

Issue #3234, opened 2023-06-27
[CLI] Incorrectly-set AZCOPY env variable causes confusing errors


Porges:
If the `AZCOPY` environment variable is set to a directory containing the AzCopy executable instead of the executable itself, then the CLI produces the error:

```console
ERROR:cli:command failed: 13 Access is denied None 5 None
```

This is confusing and not helpful. I propose that if the `AZCOPY` variable is set to a directory, then we look in that directory for the executable.

[AB#156895](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/156895)

kananb:
https://github.com/microsoft/onefuzz/pull/3344

PR #3344, opened 2023-07-25
Look for azcopy.exe in AZCOPY if it's a directory

kananb:

https://github.com/microsoft/onefuzz/issues/3234

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

Adds code to look for azcopy.exe inside of AZCOPY if it's a directory and prints a nice error if azcopy.exe can't be found.

## Validation Steps Performed

Set AZCOPY to a directory containing azcopy.exe and observe that commands work as expected.
Set AZCOPY to a directory _not_ containing azcopy.exe and observe that an informative error is printed out.
Set AZCOPY to a path pointing to azcopy.exe and observe that commands work as expected.
Set AZCOPY to a path pointing to an azcopy.exe that _doesn't exist_ and observe that an informative error is printed out.


chkeita (approved):

src/cli/onefuzz/azcopy.py
```diff
@@ -9,6 +9,17 @@ def find_azcopy() -> str:
     if azcopy:
         if not os.path.exists(azcopy):
             raise Exception(f"AZCOPY environment variable is invalid: {azcopy}")
+        elif os.path.isdir(azcopy):
+            contains_azcopy = "azcopy.exe" in [
+                f for f in os.listdir(azcopy) if os.path.isfile(os.path.join(azcopy, f))
+            ]
```
Porges, line 15: Suggest simplifying this:
```suggestion
            contains_azcopy = os.path.isfile(os.path.join(azcopy, "azcopy.exe"))
```

Porges (approved):

src/cli/onefuzz/azcopy.py
Porges: Actually, this should check for both `azcopy` and `azcopy.exe` to ensure that it works on Linux.

Porges: (This may or may not need an `exists` check first…)

PR #3345, opened 2023-07-25
Replacing 'Upgrade' flag with 'Skip AAD'.

nharper285:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/deployment/deploy.py
```diff
@@ -465,12 +465,13 @@ def setup_rbac(self) -> None:
...
-                onefuzz_cli_app["displayName"],
-                self.get_subscription_id(),
-                OnefuzzAppRole.ManagedNode,
-            )
+            if not self.skip_aad_setup:
+                assign_instance_app_role(
+                    self.application_name,
+                    onefuzz_cli_app["displayName"],
+                    self.get_subscription_id(),
+                    OnefuzzAppRole.ManagedNode,
```
chkeita, line 473: the role here should be `CliClient.CliClient`

nharper285: I didn't actually update the role there, just added the condition. I'm not sure how long it's been ManagedNode, but I wanted to be clear that there was no change in this PR.

chkeita: It should be changed then. We are currently assigning the wrong role to the cli application. We are not enforcing this currently but when we start implementing the access control it might prevent the cli from accessing the service. 
If you don't want to address it in this PR, you can open an issue to fix it later.

chkeita (approved):

src/deployment/deploy.py
```diff
@@ -1275,9 +1276,9 @@ def main() -> None:
         "password for the pool agent",
     )
     parser.add_argument(
-        "--upgrade",
+        "--skip_aad_setup",
         action="store_true",
-        help="Indicates that the instance is being upgraded",
+        help="Assumes existing aad resources that do not need to be upgraded.",
```
Porges, line 1281:
```suggestion
        help="Assumes that AAD resources already exist, and they do not need to be upgraded.",
```

Porges (approved):

src/deployment/deploy.py
nharper285: I'll address here!

PR #3366, opened 2023-07-31
Update log message to accurately represent action being take

tevoinea:

_What is this about?_

We log a message "sending event" when we are actually only logging an event.

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/Events.cs
+++ b/src/ApiService/ApiService/onefuzzlib/Events.cs
@@ -89,7 +89,7 @@ namespace Microsoft.OneFuzz.Service {
 
         public virtual void LogEvent(BaseEvent anEvent) {
             var serializedEvent = JsonSerializer.Serialize(anEvent, anEvent.GetType(), _options);
-            _log.LogInformation("sending event: {EventType} - {serializedEvent}", anEvent.GetEventType(), serializedEvent);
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/Events.cs
```diff
@@ -89,7 +89,7 @@ public class Events : IEvents {
 
         public virtual void LogEvent(BaseEvent anEvent) {
             var serializedEvent = JsonSerializer.Serialize(anEvent, anEvent.GetType(), _options);
-            _log.LogInformation("sending event: {EventType} - {serializedEvent}", anEvent.GetEventType(), serializedEvent);
+            _log.LogInformation("logging event: {EventType} - {serializedEvent}", anEvent.GetEventType(), serializedEvent);
```
chkeita, line 92: That was meant to describe what is happening on line 86.

tevoinea: Ah I see, I think we could remove `LogEvent` since it's only used by `SendEvent`. It'll make things a little clearer

chkeita: I agree but I think this function is used elsewhere. If that is the case, I suggest you take the log message string (or just a prefix) as parameter to this function. That wat you keep the intent close to where it is being used.

PR #3370, opened 2023-08-01
Refactor AdoTemplate rendering

### LINKED ISSUE #1692: Enable debugging WIQL generated from ADO notification configs
When creating an ADO notification, the user-provided ADO notification config is [used to generate a WIQL query](https://github.com/microsoft/onefuzz/blob/c9f7dd51f7860249f6081b6b230a7c9100493da4/src/api-service/__app__/onefuzzlib/notifications/ado.py#L80-L136) for existing work items.

ADO has various [limits and restrictions on WIQL queries](https://docs.microsoft.com/en-us/azure/devops/user-guide/service-limits?view=azure-devops#queries). For example, a query cannot result in more than 20,000 results, but [WIQL lacks a `LIMIT` keyword](https://docs.microsoft.com/en-us/azure/devops/boards/queries/wiql-syntax?view=azure-devops#query-language-overview). Our query should typically be filtering work item results to _much_ less than this limit, but certain user-provided notification configs might map to excessively broad queries with limit-violating result sizes.

To debug possible conflicts, we need to be able to inspect the [query submitted to ADO](https://github.com/microsoft/onefuzz/blob/c9f7dd51f7860249f6081b6b230a7c9100493da4/src/api-service/__app__/onefuzzlib/notifications/ado.py#L125).

1. Extract the inline WIQL query generation code from the actual querying
2. Unit test WIQL query generation 

[AB#35938](https://fuzzfest.visualstudio.com/Onefuzz/_workitems/edit/35938)

### LINKED ISSUE #3369: Bug when checking for existing work items with truncated title
## Information

- Onefuzz version: 8.6.0
- OS: N/A

1. We create a work item with a truncated title
2. Later, we get a crash which would result in a duplicate work item, but when searching for existing work items we don't search for the truncated title
3. Since we don't find the existing work item (with the truncated title), we create a new work item

### Expected result

We should update the existing work item

[AB#162996](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/162996)

tevoinea:

_What is this about?_

Closes #3369 
Closes #1692

## Info on Pull Request

_What does this include?_

* A new model `RenderedAdoTemplate`
* Once per notification attempt, we render our `AdoTemplate` into a `RenderedAdoTemplate`. During that process, we can enforce limitations imposed by ADO such as a max title length
* Later, when using the fields from the config, we don't need to remember/worry about re-rendering or making sure the rendered results meets the ADO limitations
* Refactor `AdoConnector` to only work with a `RenderedAdoTemplate` to enforce the behavior from the previous 2 bullet points 
* `AdoConnector` no longer has the ability to render fields in a config
* Split `ExistingWorkItems` into 2 parts: - Allowing us to test the query we generate
    1. Creating the ADO query
    2. Executing the query and returning results
* A test to validate the ADO query we're generating
* A test to validate that the title is truncated when rendering `AdoTemplate` to `RenderedAdoTemplate`

## Validation Steps Performed

Validated in test instance:

1. Start a new job
2. Attach a notification config to the job and make the config so that it's guaranteed to truncate the title (I just repeated report.crash_site over and over)
3. Create a debug notification for the job from step 1
4. Verify a bug was created with the appropriate template and that the title was truncated
5. Repeat step 3 with a new crash name
6. Verify the bug from step 4 was updated with a comment containing the new crash name
7. Verify a new bug was not created


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -171,56 +174,124 @@ public class Ado : NotificationsBase, IAdo {
...
+                })) {
+                    continue;
+                }
+
+                yield return item;
             }
         }
 
-        public async IAsyncEnumerable<WorkItem> ExistingWorkItems(IList<(string, string)> notificationInfo) {
+        public (Wiql, Dictionary<string, string>) CreateExistingWorkItemsQuery(IList<(string, string)> notificationInfo) {
```
tevoinea, line 287: This is the query part of the `ExistingWorkItems` split

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -171,56 +174,124 @@ public class Ado : NotificationsBase, IAdo {
...
+    public sealed class AdoConnector {
+        private readonly RenderedAdoTemplate _config;
+        private readonly string _project;
+        private readonly WorkItemTrackingHttpClient _client;
+        private readonly ILogger _logTracer;
+        private readonly Dictionary<string, WorkItemField2> _validFields;
+
+        public AdoConnector(RenderedAdoTemplate config, string project, WorkItemTrackingHttpClient client, Uri instanceUrl, ILogger logTracer, Dictionary<string, WorkItemField2> validFields) {
             _config = config;
-            _renderer = renderer;
```
tevoinea, line 200: `AdoConnector` can no longer render

nharper285 (approved):

PR #3374, opened 2023-08-01
Update default windows image for windows 11

AdamL-Microsoft:

Updating default OneFuzz windows 10 images to windows 11

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


Porges:
I believe for production and other deployments we also have these values set in InstanceConfig so we'll need to either clear or update those values as well.

AdamL-Microsoft:
> I believe for production and other deployments we also have these values set in InstanceConfig so we'll need to either clear or update those values as well.

good call - PR for that ADO update is testing now

tevoinea (approved):

Porges (approved):

PR #3378, opened 2023-08-02
Forcefully shut down the runtime before existing the main task

### LINKED ISSUE #3377: Tasks hangs when shutting down
The onefuzz task hangs while waiting for the tokio runtime to close. We should force a shutdown of the tokio environment when we exit the task. 

[AB#163024](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/163024)

chkeita:

The default behavior or the runtime is [wait for all spawned work](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#shutdown) to be done before exiting. This was causing the task to hang when  exiting. 
This change adds a force shutdown at the end of the main function of the task.

closes #3377

### DIFF
--- a/src/agent/onefuzz-task/src/main.rs
+++ b/src/agent/onefuzz-task/src/main.rs
@@ -40,6 +40,7 @@ fn main() -> Result<()> {
     let rt = tokio::runtime::Runtime::new()?;
     let result = rt.block_on(run(matches));
     atexit::execute();
+    rt.shutdown_background();
     result
 }
 


mgreisen (approved):

tevoinea (approved):
Just to help me better understand:

* We'll never call `rt.shutdown_background()` until `let result = rt.block_on(run(matches));` completes
* `run(matches)` will only complete when one of the futures complete from this `select!` macro https://github.com/microsoft/onefuzz/blob/bcf3cf941a6b0e229e57ab79d3087d23f7c29c14/src/agent/onefuzz-task/src/managed/cmd.rs#L101-L114
* `select!` [claims](https://docs.rs/tokio/latest/tokio/macro.select.html) to cancel the remaining branches
> Waits on multiple concurrent branches, returning when the first branch completes, cancelling the remaining branches.

Does that mean the cancellation from `select!` happens in the background? And that prior to this PR, we wait for the cancellation to complete when we drop the runtime?

chkeita:
> Does that mean the cancellation from `select!` happens in the background? And that prior to this PR, we wait for the cancellation to complete when we drop the runtime?

This is not necessarily a problem with `select!`. The tokio docs says it waits on all the *spawned* work. Select will cancel everything related the task  itself  But something was spawned inside that task of it becomes an independent branch that is not waited on.

PR #3380, opened 2023-08-02
Job Result Round #2

nharper285:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-result/src/job_result.rs
```diff
@@ -0,0 +1,152 @@
...
+    RuntimeStats,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+struct JobResult {
+    task_id: Uuid,
+    job_id: Uuid,
+    machine_id: Uuid,
+    machine_name: String,
+    data: Vec<JobResultData>,
```
tevoinea, line 36: I'm not sure I understand why `data` is a `Vec` here. We only expose it through `send_direct` which takes a single `JobResultData`.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -892,6 +914,21 @@ public record SecretAddress<T>(Uri Url) : ISecret<T> {
 public record SecretData<T>(ISecret<T> Secret) {
 }
 
+public record JobResult(
+    [PartitionKey] Guid JobId,
+    [RowKey] string Project,
```
tevoinea, line 919: I don't think there's ever a case where we can have one job for multiple projects right? I'm just thinking the `RowKey` might not be necessary but I don't know if there's any pros/cons to including it anyway.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -0,0 +1,128 @@
...
+
+    public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext context)
+        : base(log, context) {
+    }
+
+    public async Async.Task<JobResult?> GetJobResult(Guid jobId) {
+        return await SearchByPartitionKeys(new[] { jobId.ToString() }).SingleOrDefaultAsync();
+    }
+
+    public JobResult UpdateResult(JobResult result, JobResultType type, Dictionary<string, double> resultValue) {
```
tevoinea, line 24: It looks like `UpdateResult` is only used internally by `JobResultOperations` so we can make it private and remove it from the interface.

tevoinea: Same goes for `TryUpdate`

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -0,0 +1,128 @@
...
+                _logTracer.LogError("failed to update job result {JobId}", jobResult.JobId);
+                throw new InvalidOperationException($"failed to insert job result {jobResult.JobId}");
+            }
+            _logTracer.LogInformation("updated job result {JobId}", jobResult.JobId);
+        }
+
+        return true;
+    }
+
+    public async Async.Task<OneFuzzResult<bool>> CreateOrUpdate(Guid jobId, JobResultType resultType, Dictionary<string, double> resultValue) {
```
tevoinea, line 104: I don't think it's possible for us to return `OneFuzzResult<bool>.Ok(false)` here, maybe this should return a `OneFuzzResultVoid`?

src/agent/onefuzz-result/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "onefuzz-result"
+version = "0.2.0"
+authors = ["fuzzing@microsoft.com"]
+edition = "2018"
```
tevoinea, line 5: We just updated to 2021

src/agent/onefuzz-result/src/job_result.rs
```diff
@@ -0,0 +1,152 @@
...
+    task_id: Uuid,
+    job_id: Uuid,
+    machine_id: Uuid,
+    machine_name: String,
+}
+
+pub struct JobResultContext<TContext, T> {
+    pub state: TContext,
+    pub queue_client: QueueClient,
+    pub pending_messages: Mutex<HashSet<T>>,
```
tevoinea, line 51: Do we ever use this? It looks like we directly queue the messages and never put them in this hashset.

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -305,6 +330,29 @@ impl SyncedDir {
 
                 event!(event.clone(); EventData::Path = file_name_event_str);
```
tevoinea, line 331: Have you looked into hooking these `send_direct` calls right into the `event!` or `metric!` macros?

src/deployment/bicep-templates/storageAccounts.bicep
```diff
@@ -33,7 +33,7 @@ var storageAccountFuncQueuesParams = [
   'update-queue'
   'webhooks'
   'signalr-events'
-  'custom-metrics'
```
tevoinea, line 36: Is there any other code we need to remove by removing this queue?

src/agent/onefuzz-result/src/job_result.rs
nharper285: Leftover from a previous design. I based this off the heartbeat which can send messages with 'cumulative' events. Agreed that it should be removed here.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
nharper285: Correct. I suppose we could just double up the partition/row key on the job id?

src/agent/onefuzz-result/src/job_result.rs
nharper285: Nope, another leftover. Removing.

src/agent/onefuzz-result/src/job_result.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+#[async_trait]
+pub trait JobResultSender {
+    async fn send_direct(&self, data: JobResultData, value: HashMap<String, f64>) -> Result<()>;
+}
+
+#[async_trait]
+impl JobResultSender for TaskJobResultClient {
+    async fn send_direct(&self, data: JobResultData, value: HashMap<String, f64>) -> Result<()> {
```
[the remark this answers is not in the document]
nharper285: Done!

src/agent/onefuzz/src/syncdir.rs
nharper285: Yes. I'd tried and ran into a few cyclic dependencies that I was having trouble getting around. 

More importantly, going forward I don't think metrics will always coincide with send_direct calls. Currently, it just so happens that we are only making metric calls when reporting job data, but I see a future where a metric is used more broadly.

src/deployment/bicep-templates/storageAccounts.bicep
nharper285: No. This was lingering from a few months ago and I just noticed it.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
tevoinea: Yup, you could do `[PartitionKey][RowKey] Guid JobId` like we do here: https://github.com/microsoft/onefuzz/blob/8980988fe083cc1aa79b1895c6b40e97763a10e6/src/ApiService/ApiService/OneFuzzTypes/Model.cs#L728

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -0,0 +1,126 @@
...
+        var jobId = job.JobId;
+        _logTracer.LogInformation($"Inside try, but before.");
+
+        var jobResult = await GetJobResult(jobId);
+
+        _logTracer.LogInformation($"Inside try.");
+        if (jobResult == null) {
+            _logTracer.LogInformation("Creating new JobResult for Job {JobId}", jobId);
+
+            var entry = new JobResult(JobId: jobId, Project: job.Config.Project, Name: job.Config.Name, 0, 0, 0, 0, 0, 0, 0, 0, 0);
```
Porges, line 74: To simplify the use of JobResult, let's have a constructor that initializes all the values to 0

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -0,0 +1,126 @@
...
+                _logTracer.LogInformation($"Invalid Field {type}.");
+                break;
+        }
+        _logTracer.LogInformation($"Attempting to log new result: {newResult}");
+        return newResult;
+    }
+
+    private async Async.Task<bool> TryUpdate(Job job, JobResultType resultType, Dictionary<string, double> resultValue) {
+        var jobId = job.JobId;
+        _logTracer.LogInformation($"Inside try, but before.");
```
Porges, line 66: Remove before merge

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -0,0 +1,126 @@
...
+                break;
+            case JobResultType.RuntimeStats:
+                _logTracer.LogInformation("Attempting update to iterations.");
+                double newTotalIterations = resultValue["total_count"];
+                _logTracer.LogInformation($"Attempting update to iterations {newTotalIterations}.");
+                // int newExecsSec = resultValue["execs_sec"];
+                newResult = result with { IterationCount = newTotalIterations };
+                break;
+            default:
+                _logTracer.LogInformation($"Invalid Field {type}.");
```
Porges, line 57: This should be warning or error level

src/agent/onefuzz-result/src/job_result.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone)]
+pub struct TaskContext {
+    task_id: Uuid,
+    job_id: Uuid,
+    machine_id: Uuid,
+    machine_name: String,
+}
+
+pub struct JobResultContext<TContext> {
+    pub state: TContext,
```
Porges, line 46: If we only ever use this with `TaskContext` I'd say make it `pub task: TaskContext`

nharper285: Updated - let me know if the change is what you meant.

src/agent/onefuzz/src/syncdir.rs
```diff
@@ -265,9 +268,39 @@ impl SyncedDir {
...
+                        Event::new_crashdump => {
+                            jr_client
+                                .send_direct(
+                                    JobResultData::NewCrashDump,
+                                    HashMap::from([("count".to_string(), 1.0)]),
+                                )
+                                .await;
+                        }
+                        _ => {
+                            debug!("Unhandled job result!");
```
tevoinea, line 300: We should upgrade this to at least the info level since we won't see debug logs in prod and we might otherwise miss the message (in case we forget to add a new case here)

tevoinea: Same thing on line 360

tevoinea (approved):

PR #3383, opened 2023-08-03
Handle another duplicate field

### LINKED ISSUE #3382: Handle when duplicate reason is stored in `Microsoft.VSTS.Common.ResolvedReason` field
## Information

- Onefuzz version: 8.6.2
- OS: N/A

1. Some orgs use nonstandard fields for indicating a work item is duplicate

### Expected result

As a quick fix we should add support for this new `Microsoft.VSTS.Common.ResolvedReason` field.

[AB#163772](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/163772)

tevoinea:

_What is this about?_

Closes #3382

Adds duplicate checking for `"Microsoft.VSTS.Common.ResolvedReason"`

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
+++ b/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
@@ -528,8 +528,8 @@ public class Ado : NotificationsBase, IAdo {
             // OR it could have System.State == Closed && System.Reason == Duplicate
             // I haven't found any other combinations where System.Reason could be duplicate but just to be safe
             // we're explicitly _not_ checking the state of the work item to determine if it's duplicate
-            return wi.Fields.ContainsKey("System.Reason") && string.Equals(wi.Fields["System.Reason"].ToString(), "Duplicate")
-            || wi.Fields.ContainsKey("Microsoft.VSTS.Common.ResolvedReason") && string.Equals(wi.Fields["Microsoft.VSTS.Common.ResolvedReason"].ToString(), "Duplicate")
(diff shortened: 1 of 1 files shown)


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -529,6 +529,7 @@ public sealed class AdoConnector {
             // I haven't found any other combinations where System.Reason could be duplicate but just to be safe
             // we're explicitly _not_ checking the state of the work item to determine if it's duplicate
             return wi.Fields.ContainsKey("System.Reason") && string.Equals(wi.Fields["System.Reason"].ToString(), "Duplicate")
+            || wi.Fields.ContainsKey("Microsoft.VSTS.Common.ResolvedReason") && string.Equals(wi.Fields["Microsoft.VSTS.Common.ResolvedReason"].ToString(), "Duplicate")
```
mgreisen, line 532: I can't remember if the .Equals is case sensitive or not.

mgreisen: And we're sure these are fields that would apply regardless of the ADO org?

tevoinea: It is case sensitive by default, I could make those insensitive.

It's ok if the fields aren't present in the work item, it'll just evaluate to false. This is just a quick fix until we can implement the proper fix described here: https://github.com/microsoft/onefuzz/issues/3384

tevoinea: Made the comparisons case insensitive

AdamL-Microsoft (approved):

Issue #3220, opened 2023-06-22
`debug notification test_template` returns "success" when AreaPath is invalid


kananb:
## Information

- Onefuzz version: 8.3.0
- OS: Windows

## Provide detailed reproduction steps (if any)
```
onefuzz debug notification test_template <task_id> "{ \"config\": { \"base_url\": \"https://dev.azure.com/<org>\", \"project\": \"<project>\", \"type\": \"Bug\", \"unique_fields\": [ \"System.Title\" ], \"comment\": \"This comment was intentionally left blank.\", \"ado_fields\": { \"System.Tags\": \"<tag1>, <tag2>\", \"System.Title\": \"[Bug Filing Test]\", \"Microsoft.VSTS.TCM.ReproSteps\": \"This field was intentionally left blank.\", \"System.AssignedTo\": \"<alias>@domain.com\", \"System.AreaPath\": \"Invalid\\AreaPath\", \"System.IterationPath\": \"<iteration_path>\" }, \"on_duplicate\": { \"increment\": [], \"comment\": \"This comment was intentionally left blank.\", \"set_state\": { \"Resolved\": \"Active\", \"Closed\": \"Active\" }, \"ado_fields\": {} }, \"auth_token\": \"<auth_token>\" }, \"enabled\": true }"
```
Raw ADO config JSON:
```json
{
  "config": {
    "base_url": "https://dev.azure.com/<org>",
    "project": "<project>",
    "type": "Bug",
    "unique_fields": [
      "System.Title"
    ],
    "comment": "This comment was intentionally left blank.",
    "ado_fields": {
      "System.Tags": "<tag1>, <tag2>",
      "System.Title": "[Bug Filing Test]",
      "Microsoft.VSTS.TCM.ReproSteps": "This field was intentionally left blank.",
      "System.AssignedTo": "alias@domain.com",
      "System.AreaPath": "Invalid\\AreaPath",                 //              <-----------------
      "System.IterationPath": "<iteration_path>"
    },
    "on_duplicate": {
      "increment": [],
      "comment": "This comment was intentionally left blank.",
      "set_state": {
        "Resolved": "Active",
        "Closed": "Active"
      },
      "ado_fields": {}
    },
    "auth_token": "<auth_token>"
  },
  "enabled": true
}
```

### Expected result

An error message is returned indicating a failure to create the bug, preferably indicating the AreaPath as the source of the failure.

### Actual result

A success message is returned:
```
{
    "success": true
}
```
[...]

Issue #2985, opened 2023-04-05
Notification `config.json` Schema Validation


nharper285:
We need to validate notification config files during job submission. Ideally, a user would attempt to deploy a job with a malformed json file and OneFuzz would immediately fail the job submission and return the failure to the user. 

One potential solution would be to detail some kind of schema that can be used to check incoming configs against.

mgreisen:
I think what your proposing here is sematic validation, right?

PR #3386, opened 2023-08-03
Kanan/semantic validation

kananb:

Adds validation for Iteration/AreaPath when a job is submitted with a notification config and for `onefuzz debug notification test_template`.

## PR Checklist
* [x] Applies to work items: #2985, #3220
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/NotificationOperations.cs
```diff
@@ -134,6 +134,7 @@ public NotificationOperations(ILogger<NotificationOperations> log, IOnefuzzConte
         if (await _context.FeatureManagerSnapshot.IsEnabledAsync(FeatureFlagConstants.SemanticNotificationConfigValidation)) {
             var validConfig = await config.Validate();
             if (!validConfig.IsOk) {
+                _logTracer.LogError("Error(s) ocurred during template validation: {title}\r\n{errors}", validConfig.ErrorV.Title, string.Join("\r\n", validConfig.ErrorV.Errors ?? new()));
```
tevoinea, line 137: You could use Environment.NewLine to get the right line endings all the time
https://learn.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-7.0

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,6 +89,37 @@ public class Ado : NotificationsBase, IAdo {
         return errorCodes.Any(errorStr.Contains);
     }
 
+    private static async Async.Task<OneFuzzResultVoid> ValidatePath(string project, string path, TreeStructureGroup structureGroup, WorkItemTrackingHttpClient client) {
+        var pathParts = path.Split('\\');
+        if (pathParts[0] != project) {
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PATH, new string[] {
+                $"Path \"{path}\" is invalid. It must start with the project name, \"{project}\".",
```
tevoinea, line 96: We know at this point if we're validating an area of iteration path based on the structureGroup so we could include that in the error message. Ex: `Area Path \"{path}\" is invalid` or `Iteration Path ...`

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,6 +89,37 @@ public class Ado : NotificationsBase, IAdo {
         return errorCodes.Any(errorStr.Contains);
     }
 
+    private static async Async.Task<OneFuzzResultVoid> ValidatePath(string project, string path, TreeStructureGroup structureGroup, WorkItemTrackingHttpClient client) {
+        var pathParts = path.Split('\\');
+        if (pathParts[0] != project) {
```
tevoinea, line 94: We should probably do a case insensitive comparison

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,6 +89,37 @@ public class Ado : NotificationsBase, IAdo {
...
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PATH, new string[] {
+                $"Path \"{path}\" is invalid. It must start with the project name, \"{project}\".",
+                $"Example: \"{project}\\{path}\".",
+            });
+        }
+
+        var current = await client.GetClassificationNodeAsync(project, structureGroup, depth: pathParts.Length - 1);
```
tevoinea, line 101: Just to check my understanding: for depth, we're doing parts.Length - 1 because we're considering the last part in the path as a leaf and we don't care about its children (if it has any) right?

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,6 +89,37 @@ public class Ado : NotificationsBase, IAdo {
...
+        if (current == null) {
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PATH, new string[] {
+                $"Path \"{path}\" is invalid. \"{project}\" is not a valid project.",
+            });
+        }
+
+        foreach (var part in pathParts.Skip(1)) {
+            var child = current.Children?.FirstOrDefault(x => x.Name == part);
```
tevoinea, line 109: Case insensitive comparison again just to be safe

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -162,6 +192,61 @@ public class Ado : NotificationsBase, IAdo {
...
+        try {
+            if (config.AdoFields.TryGetValue("System.AreaPath", out var areaPathString)) {
+                var validateAreaPath = await ValidatePath(config.Project, areaPathString, TreeStructureGroup.Areas, witClient);
+                if (!validateAreaPath.IsOk) {
+                    return validateAreaPath;
+                }
+            }
+        } catch (VssUnauthorizedException e) {
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PAT, new string[] {
+                "The provided PAT may be missing scopes. We were able to connect with it but unable to validate the fields.",
```
tevoinea, line 205: Do we need additional scopes on a PAT (other than work item list/create) to use this 'get classification nodes' endpoint?

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -162,6 +192,61 @@ public class Ado : NotificationsBase, IAdo {
...
+        }
+
+        // Validate IterationPath exists
+        try {
+            if (config.AdoFields.TryGetValue("System.IterationPath", out var iterationPathString)) {
+                var validateIterationPath = await ValidatePath(config.Project, iterationPathString, TreeStructureGroup.Iterations, witClient);
+                if (!validateIterationPath.IsOk) {
+                    return validateIterationPath;
+                }
+            }
```
tevoinea, line 229: We could merge this and the AreaPath validation into the same try block

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: `parts.Length - 1` because the project is the root node, so we don't need to count it for the child-depth.

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: [Nope](https://learn.microsoft.com/en-us/rest/api/azure/devops/wit/classification-nodes/get-classification-nodes?view=azure-devops-rest-7.1&tabs=HTTP), good point

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: Yep, you're right. I was under the impression that area and iteration paths were case sensitive for some reason.

src/ApiService/ApiService/Functions/NotificationsTest.cs
```diff
@@ -24,6 +24,11 @@ public class NotificationsTest {
         }
 
         var notificationTest = request.OkV;
+        var validConfig = await notificationTest.Notification.Config.Validate();
+        if (!validConfig.IsOk) {
+            return await _context.RequestHandling.NotOk(req, validConfig.ErrorV, context: "notification create");
```
tevoinea, line 29:
```suggestion
            return await _context.RequestHandling.NotOk(req, validConfig.ErrorV, context: "notification test");
```

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,6 +89,38 @@ public class Ado : NotificationsBase, IAdo {
         return errorCodes.Any(errorStr.Contains);
     }
 
+    private static async Async.Task<OneFuzzResultVoid> ValidatePath(string project, string path, TreeStructureGroup structureGroup, WorkItemTrackingHttpClient client) {
+        var pathType = (structureGroup == TreeStructureGroup.Areas) ? "Area" : "Iteration";
```
tevoinea, line 93: Just a note, not expecting you to make this change.

I know the `TreeStructureGroup` enum only has 2 variants so this if statement is sufficient, but, by using a switch expression we can guard against incorrect behavior in case a future update adds a new variant. For example if they add `TreeStructureGroup.Option3`, then this code will happily produce `pathType = "Iteration"`.

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -362,7 +414,7 @@ public sealed class AdoConnector {
                 return false;
             }
 
-            if (_config.OnDuplicate.Comment != null) {
+            if (!string.IsNullOrEmpty(_config.OnDuplicate.Comment)) {
```
tevoinea, line 417: 👍

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
kananb: Thanks for pointing this out!

PR #3402, opened 2023-08-09
Switch to NSubstitute

Porges:


src/ApiService/IntegrationTests/packages.lock.json
```diff
@@ -225,8 +216,8 @@
       },
       "Castle.Core": {
         "type": "Transitive",
-        "resolved": "5.1.1",
-        "contentHash": "rpYtIczkzGpf+EkZgDr9CClTdemhsrwA/W5hMoPjLkRFnXzH44zDLoovXeKtmxb1ykXK9aJVODSpiJml8CTw2g==",
+        "resolved": "5.0.0",
```
theunrepentantgeek, line 219: Is the downgrade of Castle.Core needed?

Porges: Nuget did it. I don't even try to understand nuget resolution any more, not until it grows a real package resolver.

tevoinea (approved):

PR #3409, opened 2023-08-10
Make crashdumps optional in config

Porges:
Make crashdumps optional in task definition so that old CLI clients are not rejected.


Porges:
Validated with check-pr

stishkin (approved):

PR #3410, opened 2023-08-11
Deserialize the coverage files directly into the output files

### LINKED ISSUE #3420: Deserialize the coverage files directly into the output files
The Coverage is loading the output of the coverage in memory before writing it to the output file.
This could cause an increase in memory consumption. 

AB#164560

chkeita:

The Coverage file was loading the output of the coverage in memory before writing it to the output file. 
This could cause an increase in memory consumption. This change updates the logic to serialize the output directly in the output file without loading it in the memory.

closes #3420

### DIFF
--- a/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
+++ b/src/agent/onefuzz-task/src/tasks/coverage/generic.rs
@@ -10,7 +10,7 @@ use std::time::Duration;
 
 use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
-use cobertura::CoberturaCoverage;
+use cobertura::{CoberturaCoverage, WriteXml};
 use coverage::allowlist::AllowList;
 use coverage::binary::{BinaryCoverage, DebugInfoCache};
 use coverage::record::CoverageRecorder;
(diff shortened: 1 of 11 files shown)


src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -457,49 +457,78 @@ impl<'a> TaskContext<'a> {
...
+            &copbertura_file_path,
+        )
+        .await?;
         self.config.coverage.sync_push().await?;
+        Ok(())
+    }
 
+    async fn save_cobertura_xml(source: &SourceCoverage, path: &Path) -> Result<(), anyhow::Error> {
+        let cobertura = CoberturaCoverage::from(source);
+        let cobertura_coverage_file = std::fs::File::create(path)
```
tevoinea, line 501: We could use the tokio file create here

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -457,49 +457,78 @@ impl<'a> TaskContext<'a> {
...
+            .with_context(|| format!("creating cobertura coverage file {}", path.display()))?;
+        let cobertura_coverage_file_writer = std::io::BufWriter::new(cobertura_coverage_file);
+        cobertura
+            .write_xml(cobertura_coverage_file_writer)
```
tevoinea, line 505: It might be worthwhile (as a forever backlog feature) to add `AsyncWrite` support for `write_xml`, that way we can use tokio's `BufWriter`. I'm not sure how much time we spend blocking the thread to write these coverage files but I guess if they get big enough there's some value there. Especially since we write the file every time we merge the coverage which I think is on every input processed?

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
tevoinea: Same for `save_source_coverage` and `save_binary_coverage`

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -457,49 +457,78 @@ impl<'a> TaskContext<'a> {
...
+    pub async fn save_coverage(
+        coverage: &BinaryCoverage,
+        source_allowlist: &AllowList,
+        binary_coverage_path: &Path,
+        source_coverage_path: &Path,
+        copbertura_file_path: &Path,
+    ) -> Result<()> {
+        Self::save_binary_coverage(coverage, binary_coverage_path)?;
+
+        let source = binary_to_source_coverage(coverage, source_allowlist.clone())?;
```
tevoinea, line 469: We used to do this in a `spawn_blocking` call, was that intentionally removed? I think we're still doing the same amount of blocking I/O as before

tevoinea (approved):

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
chkeita: this would require a change in `write_xml` to make it work with the tokio buWriter

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
chkeita: #3419

PR #3431, opened 2023-08-18
Impl AsyncWrite for cobertura output

tevoinea:
So far, this PR is just me exploring with implementing AsyncWrite. The current 'blocker' is that the Writer implementation that uses async doesn't have 'high level' functions like `create_element` which we use from the sync impl. Next time, or the next person that picks up this PR should figure out how to have async versions of those high level functions.

### DIFF
--- a/src/agent/cobertura/Cargo.toml
+++ b/src/agent/cobertura/Cargo.toml
@@ -7,3 +7,5 @@ license = "MIT"
 [dependencies]
 anyhow = "1.0"
 quick-xml = "0.30"
+tokio = "1"
+async-trait = "0.1"
--- a/src/agent/cobertura/src/lib.rs
+++ b/src/agent/cobertura/src/lib.rs
@@ -1,9 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::io::{Cursor, Write};
+use std::{
+    io::{Cursor, Write},
+    sync::Arc,
+};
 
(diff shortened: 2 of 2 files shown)


src/agent/cobertura/Cargo.toml
```diff
@@ -6,4 +6,8 @@ license = "MIT"
 
 [dependencies]
 anyhow = "1.0"
-quick-xml = "0.30"
+quick-xml = { git = "https://github.com/tevoinea/quick-xml/", branch = "tevoinea/AsyncElementWriter", features = [
```
tevoinea, line 9: This is only temporary until my PR gets reviewed and merged. https://github.com/tafia/quick-xml/pull/635

src/agent/cobertura/src/lib.rs
```diff
@@ -17,6 +19,16 @@ impl CoberturaCoverage {
...
+
+    pub async fn to_string_async(&self) -> anyhow::Result<String> {
+        let mut buffer = Vec::new();
+        let mut bufwriter = tokio::io::BufWriter::new(&mut buffer);
+
+        self.write_xml_async(&mut bufwriter).await?;
+        bufwriter.flush().await?;
+        let text = String::from_utf8(buffer)?;
+        Ok(text)
+    }
```
Porges, line 31: I wouldn't add this as it's a bit misleading, all the operations are on an in-memory so it doesn't need to be async.

PR #3438, opened 2023-08-21
Starting integration tests

### LINKED ISSUE #3422: Add integration tests for the local fuzzing
Add integration tests for the local fuzzing
[AB#164584](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/164584)

tevoinea:

Closes #3422 

Adds an integration test for the local template fuzzing feature.

## Info on Pull Request

_What does this include?_

* Updates to CI
    * Build simple-libfuzzer
    * Place the binary into the integration test directory
    * Run integration tests in CI
* Bump `pete` to version 0.12, this fixes a bug on linux where it was calling `waitpid` on all subprocesses instead of just the ones we're trying to debug
* Expose a `onefuzz_task_lib` crate, allowing us to write integration tests using rust's testing framework
* An integration test for a basic libfuzzer template including fuzzing, coverage, and crash reporting
* Fix a bug in path preparation on Windows when absolute paths are given
* 

## Validation Steps Performed


src/agent/onefuzz-task/src/tasks/fuzz/libfuzzer/common.rs
```diff
@@ -272,7 +272,7 @@ where
         info!("config is: {:?}", self.config);
 
         let fuzzer = L::from_config(&self.config).await?;
-        let mut running = fuzzer.fuzz(crash_dir.path(), local_inputs, &inputs).await?;
+        let mut running = fuzzer.fuzz(crash_dir.path(), local_inputs, &inputs)?;
```
tevoinea, line 275: This function didn't actually use any async/await

src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -352,8 +352,7 @@ impl LibFuzzer {
         // specify that a new file `crash-<digest>` should be written to a
         // _directory_ `<corpus_dir>`, we must ensure that the prefix includes a
         // trailing path separator.
-        let artifact_prefix: OsString =
-            format!("-artifact_prefix={}/", fault_dir.as_ref().display()).into();
```
tevoinea, line 356: `/` is not a valid windows path separator when using absolute (not URI) paths.

chkeita (approved):

src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -441,6 +441,20 @@ impl LibFuzzer {
     }
 }
 
+#[cfg(target_os = "windows")]
+fn artifact_prefix(fault_dir: impl AsRef<Path>) -> OsString {
```
Porges, line 445: suggest just using `&Path`, caller is already doing `.as_ref()`

```suggestion
fn artifact_prefix(fault_dir: &Path) -> OsString {
```

src/agent/onefuzz-task/tests/template_integration.rs
```diff
@@ -0,0 +1,210 @@
...
+async fn assert_directory_is_not_empty(dir: &Path) {
+    assert!(
+        dir_is_empty(dir).await,
+        "Expected directory to not be empty. dir = {:?}",
+        dir
+    );
+}
+
+async fn create_test_directory(config: &Path, target_exe: &Path) -> Result<TestLayout> {
+    let mut test_directory = PathBuf::from(".").join(uuid::Uuid::new_v4().to_string());
```
Porges, line 127: it would be best to use TempDir here rather than the test directory

Porges (approved):

src/agent/onefuzz-task/tests/template_integration.rs
tevoinea: I did initially use TempDir, but I found it difficult to debug failures since it gets cleaned up on drop. I figure in CI, it doesn't really matter if the directories don't get cleaned up. I could add a directory cleanup step at the end that someone could comment out if they're debugging the test.

PR #3440, opened 2023-08-21
Add a start time to the job and task

### LINKED ISSUE #3357: Add a start_time to the job and tasks
We need to capture the time the tasks was effectively started (picked up by a node)

[AB#162978](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/162978)

chkeita:

Add a start time to the job and task
closes #3357

### DIFF
--- a/src/ApiService/ApiService/OneFuzzTypes/Responses.cs
+++ b/src/ApiService/ApiService/OneFuzzTypes/Responses.cs
@@ -52,7 +52,9 @@ public record TaskSearchResult(
     List<NodeAssignment> Nodes,
     [property: JsonPropertyName("Timestamp")] // must retain capital T for backcompat
     DateTimeOffset? Timestamp
-) : BaseResponse();
+) : BaseResponse() {
+    public DateTimeOffset? StartTime => EndTime is DateTimeOffset endTime ? endTime.Subtract(TimeSpan.FromHours(Config.Task.Duration)) : null;
+}
 
 public record BoolResult(
     bool Result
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/OneFuzzTypes/Responses.cs
```diff
@@ -52,7 +52,9 @@ public record TaskSearchResult(
     List<NodeAssignment> Nodes,
     [property: JsonPropertyName("Timestamp")] // must retain capital T for backcompat
     DateTimeOffset? Timestamp
-) : BaseResponse();
+) : BaseResponse() {
+    public DateTimeOffset? StartTime => EndTime is DateTimeOffset endTime ? endTime.Subtract(TimeSpan.FromHours(Config.Task.Duration)) : null;
```
tevoinea, line 56: In the issue, we say we want the start time to be when the task gets picked up by a node but this calculation doesn't cover for example a task `B` that depends on a prerequisite task `A`. Or, if the workset message stays in the pool queue while waiting for auto scale to kick in.

I think we'd get a more accurate picture by tracking StartTime around here:
https://github.com/microsoft/onefuzz/blob/6d8be1a83473c941565071d8f7988ea341e0b544/src/ApiService/ApiService/Functions/AgentCanSchedule.cs#L62-L72

We could store it as a column in the Tasks table.

chkeita: we set the `EndTime` at the when we receive the `OneStart` event. By making the `StartTime` dependent on the `EndTime` we capture the effective start

tevoinea (approved):

PR #3481, opened 2023-09-05
Add a new test for `report.extension_data`

DrChat:

Adds a new (failing) test for using `report.extension_data` in a notification template.

## PR Checklist
* [ ] Tests added/passed


DrChat:
Looks like this is going to be a massive PITA to fix.
Basically, Scriban doesn't support `JsonElement` in any capacity for their little `ScriptObject` helper. So we can't pass `JsonElement` in any capacity to them.

See here for more context: https://github.com/scriban/scriban/issues/148

PR #3489, opened 2023-09-07
Improve area/iteration path validation

kananb:

Adds validation against the path naming restrictions defined in https://learn.microsoft.com/en-us/azure/devops/organizations/settings/about-areas-iterations?view=azure-devops#naming-restrictions

The newly added checks each have their own error message for their violation, so, for example, users that delimit their paths with a forward slash instead of a backslash will get a message saying:
"Area Path "Foo/Bar" is invalid. "Foo/Bar" contains an invalid character (\ / : * ? " < > | ; # $ * { } , + = [ ])."
"Make sure that the path is separated by backslashes (\\) and not forward slashes (/)."

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
+++ b/src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
@@ -89,7 +89,14 @@ public class Ado : NotificationsBase, IAdo {
         return errorCodes.Any(errorStr.Contains);
     }
 
-    public static OneFuzzResultVoid ValidateTreePath(IEnumerable<string> path, WorkItemClassificationNode root) {
+    public static OneFuzzResultVoid ValidateTreePath(IEnumerable<string> path, WorkItemClassificationNode? root) {
+        if (root is null) {
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PROJECT, new string[] {
+                $"Path \"{string.Join('\\', path)}\" is invalid. The specified ADO project doesn't exist.",
+                "Double check the 'project' field in your ADO config.",
+            });
+        }
+
         string treeNodeTypeName;
         switch (root.StructureType) {
             case TreeNodeStructureType.Area:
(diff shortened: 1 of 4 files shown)


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,23 +89,66 @@ public class Ado : NotificationsBase, IAdo {
...
+        var pathParts = path.Split('\\');
+        var maxLength = 255;
+        var maxDepth = 13;
+        var invalidChars = new char[] { '/', '.', '~', '[', ']', ':', '$', '?', '*', '<', '>', '|', '+', '=', ';', ',' };
+
+        // Ensure that none of the path parts are too long
+        var erroneous = pathParts.FirstOrDefault(part => part.Length > maxLength);
+        if (erroneous != null) {
+            return OneFuzzResultVoid.Error(ErrorCode.ADO_VALIDATION_INVALID_PATH, new string[] {
+                $"{pathType} Path \"{path}\" is invalid. \"{erroneous}\" is too long. It must be less than {maxLength} characters.",
```
tevoinea, line 104: We could include the link to the docs in each each of these error messages.

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -89,23 +89,66 @@ public class Ado : NotificationsBase, IAdo {
         return errorCodes.Any(errorStr.Contains);
     }
 
+    private static OneFuzzResultVoid ValidatePathNaming(string path, string? pathType) {
+        // Validate path based on https://learn.microsoft.com/en-us/azure/devops/organizations/settings/about-areas-iterations?view=azure-devops#naming-restrictions
+
+        var pathParts = path.Split('\\');
+        var maxLength = 255;
```
tevoinea, line 96: This could be confused with the max path length which is 4000. Maybe a better name would be `maxNodeLength`.

tevoinea (approved):

PR #3509, opened 2023-09-17
Include all loader errors when failing

### LINKED ISSUE #2736: Include missing exports, loader transcript when available
On Windows, when we enable loader snaps in `dynamic-linker` in the fuzzer help check, we get an extensive loader transcript. When a specific export is missing from the DLL that was resolved, we also get information about which export is missing.

1. When a DLL is found, but one or more expected exports are missing, report them. The wrong DLL was probably found and loaded.
2. On error on Windows, include the entire loader transcript in the task error logs.

[AB#36071](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/36071)

Porges:
Closes #2736.

Include all errors from the loader transcript when looking for missing DLLs. This will help users to analyze the problem, as this information can include missing symbols, etc.


src/agent/onefuzz/src/libfuzzer.rs
```diff
@@ -306,25 +306,26 @@ impl LibFuzzer {
...
             // To provide user-actionable errors, try to identify any missing shared libraries.
             match self.find_missing_libraries().await {
-                Ok(missing) => {
+                Ok((missing, errors)) => {
                     if missing.is_empty() {
-                        bail!("fuzzer does not respond to '-help=1'. no missing shared libraries detected. output: {:?}", result);
+                        bail!("fuzzer does not respond to '-help=1'. no missing shared libraries detected. output: {result:?}");
                     } else {
                         let missing = missing.join(", ");
+                        let all_errors = errors.join("\n");
```
tevoinea, line 314: We should put the correct line ending depending on the platform

tevoinea (approved):

src/agent/onefuzz/src/libfuzzer.rs
Porges: I just checked and `println` in Rust _always_ uses `\n` and never `\r\n`, so I’ll leave this as-is.

PR #3527, opened 2023-09-25
Make modules case insenstive on windows

### LINKED ISSUE #3525: Code coverage allow list and modules should be case insensitive on Windows
## Information

- Onefuzz version: 8.8
- OS: Windows

Paths on Windows are not case sensitive, thus we should not be case sensitive when when evaluating coverage allow list for source or modules.

On Linux, the behavior should not change.

[AB#166580](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/166580)

tevoinea:

_What is this about?_

Closes #3525

## Info on Pull Request

_What does this include?_

* Binary and source coverage allowlists are now case insensitive on Windows
* Modules that are loaded on Windows are tracked internally in a case insensitive way
* Tests

## Validation Steps Performed

* Added a test that a source allowlist of `inlinee.exe` will pick up a binary compiled as `Inlinee.exe`
* Added a test that the modules loaded from `Inlinee.exe` are all only lowercase


src/agent/coverage/tests/snapshot.rs
```diff
@@ -52,6 +53,13 @@ fn windows_snapshot_tests() {
             .record()
             .unwrap();
 
+        // Windows modules should be case insensitive
```
ranweiler, line 56: IIUC: this test is just asserting that the names OneFuzz uses to track modules / report coverage are _normalized_ to be all-lowercase.

That's good, but maybe update the comment, because you really need something else to be true to test that the comparisons _based on this normalization_ are also effectively case-insensitive.

src/agent/debugger/src/module.rs
```diff
@@ -47,6 +47,8 @@ impl Module {
             "???".into()
         });
 
+        let path = PathBuf::from(path.as_os_str().to_ascii_lowercase());
```
ranweiler, line 50: Note that this change impacts Linux, too.

src/agent/coverage/src/allowlist.rs
```diff
@@ -144,7 +144,13 @@ fn glob_to_regex(expr: &str) -> Result<Regex> {
     // Anchor to line start and end.
     let expr = format!("^{expr}$");
 
-    Ok(Regex::new(&expr)?)
+    let mut rb = RegexBuilder::new(&expr);
+
+    if cfg!(windows) {
+        rb.case_insensitive(true);
```
ranweiler, line 150: The allowlist matching is unit-tested. I suggest adding at least one test decorated with `[cfg(target_os = "windows")]` which exercises the new matching logic. Maybe also a Linux-only test that demonstrates case-sensitivity on that platform.

src/agent/coverage/tests/snapshot.rs
ranweiler: I think your file renaming is meant to accomplish this semi-indirectly, but it would be good to document that somewhere if possible.

ranweiler (changes requested):

src/agent/debugger/src/module.rs
tevoinea: I thought it wouldn't since this function takes a `HANDLE` which comes from the `windows` crate.

ranweiler: Wow, my coffee hasn't fully hit yet. You're absolutely right, I totally misread the file path. 🥴

So, a question, instead: this `debugger` crate is also used by the Windows `input-tester` crate. Does normalizing here change the behavior on that end? For this PR, maybe it would be safer to do the normalization explicitly in the `coverage` code for now, where the `debugger` crate's `Module` gets used.

src/agent/coverage/src/allowlist.rs
tevoinea: That unit test caught a bug! Setting the case_inensitive flag in the RegexBuilder didn't get passed along when we convert from Regex to RegexSet in a different part of the code.

src/agent/debugger/src/module.rs
tevoinea: I updated the code to only become case-insensitive when we convert from binary to source coverage. That way we get what we're ultimately looking for (a cleaner/more organized cobertura file) with no risk to changing the behavior for the rest of the program.

src/agent/coverage/tests/snapshot.rs
tevoinea: Updated the comment to call out the importance of the name of the test file

src/agent/coverage/src/source.rs
```diff
@@ -104,13 +104,18 @@ pub fn binary_to_source_coverage(
                         };
 
                         if let Some(file) = location.file() {
+                            let file_path = if cfg!(windows) {
+                                // Windows paths are case insensitive.
+                                FilePath::new(file.full_path().to_ascii_lowercase())?
+                            } else {
+                                FilePath::new(file.full_path())?
+                            };
```
Porges, line 112: We should do this once up-front rather than every time we examine a location

tevoinea: I used the `Transformer` available in the symbolic-symcache crate so we only do the modification once.

PR #3529, opened 2023-09-26
Terminate process on timeout in windows for the coverage task

### LINKED ISSUE #3513: Not terminating process on timeout in Windows for the coverage task
## Information

- Onefuzz version: 8.8
- OS: Windows 11

If this timeout case hits, we never terminate the child process.
https://github.com/microsoft/onefuzz/blob/c7a982748e13cfd1d8bf2e39dd8b36ee28980e46/src/agent/coverage/src/record.rs#L128-L148

### Expected result

We should call `quit_debugging` from the `Debugger` module as well as `kill` and/or `wait` on the returned `Child.

https://github.com/microsoft/onefuzz/blob/c7a982748e13cfd1d8bf2e39dd8b36ee28980e46/src/agent/debugger/src/debugger.rs#L306
https://doc.rust-lang.org/std/process/struct.Child.html#method.kill
https://doc.rust-lang.org/std/process/struct.Child.html#method.wait

[AB#166281](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/166281)

chkeita:

use [process_control](https://docs.rs/crate/process_control/latest) to automatically terminate the process when it is timed out

closes #3513


src/agent/coverage/src/record.rs
```diff
@@ -121,31 +121,34 @@ impl CoverageRecorder {
...
     pub fn record(self) -> Result<Recorded> {
         use debugger::Debugger;
+        use process_control::{ChildExt, Control};
         use windows::WindowsRecorder;
 
         let loader = self.loader.clone();
+        let mut recorder =
+            WindowsRecorder::new(&loader, self.module_allowlist, self.cache.as_ref());
+        let (mut dbg, child) = Debugger::init(self.cmd, &mut recorder)?;
+        dbg.run(&mut recorder)?;
```
tevoinea, line 131: `run` here won't terminate until the target has exited. Is there a way for us to put the time limit on the child before we call `run`?

tevoinea (approved):

tevoinea (approved):

PR #3532, opened 2023-09-28
Ignore regression update when the work item is in some states

### LINKED ISSUE #3523: Regression task should not comment on bugs if they are not in a Closed state.
## Information

- Onefuzz version: 8.8
- OS: n/a

When the regression task runs it will generate a comment on the matching bug if a crash is found. We're finding that we're generating thousands of comments on bugs because it's taking a while before the bug is fixed. This is causing us to run into ADO limits. 

We can mitigate this by the regression task only commenting bugs when we need to reopen the bug from a Closed state. We should use Closed and not Resolved because it may take a while for a fix to propagate completely. 

[AB#166526](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/166526)

### LINKED ISSUE #3530: Do not hide messages going to the poison queue
The [current logic](https://github.com/microsoft/onefuzz/blob/f3b7e2059c33475260d1093a3be244455a5f59b9/src/ApiService/ApiService/Functions/QueueFileChanges.cs#L131C1-L145C6) is setting a viability timeout when a message is added to the poison queue. These messages should be visible.

[AB#167021](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/167021)

chkeita:

closes #3523 
closes #3530


src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
```diff
@@ -19,6 +19,7 @@ public class Ado : NotificationsBase, IAdo {
     // https://github.com/MicrosoftDocs/azure-devops-docs/issues/5890#issuecomment-539632059
     private const int MAX_SYSTEM_TITLE_LENGTH = 128;
     private const string TITLE_FIELD = "System.Title";
+    private static List<string> DEFAULT_REGRESSION_IGNORE_STATES = new() { "New", "Commited", "Active" };
```
tevoinea, line 22: I was thinking about this from the opposite perspective: we only care about regression reports if the WI state is 'Closed', any other state we can ignore. I'm not asking you to change it since both ways achieve the same result, just floating the thought!

tevoinea (approved):

src/ApiService/ApiService/onefuzzlib/notifications/Ado.cs
chkeita: yeah they are equivalent to me. Even if we invert the condition, we will still need to care about more than one state.
The example below we will need to care about Resolved, Closed, and Removed
image.

In our ado project we don't even have the closed state 
image

tevoinea: Ah that's a good point, in my head there was only one "done" state but it looks like that's not true

Issue #3512, opened 2023-09-18
RuntimeStats are not agregated per nodes


chkeita:
The current [aggregation method](https://github.com/microsoft/onefuzz/blob/main/src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs) of the `RuntimeStats` does not take into account the fact that multiple machine can run a fuzzer. This causes us to loose half of the iteration count data.

[AB#166280](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/166280)

PR #3539, opened 2023-10-02
JobResult Table Re-design - Improved task- and machine- granularity

nharper285:

_What is this about?_

## PR Checklist
* [ ] Applies to work item: #3512
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -15,86 +15,56 @@ public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext con
...
-
-            var r = await Insert(jobResult);
+            _logTracer.LogInformation($"attempt to insert new job result {taskId} and machineId+metricType {machineIdMetric}");
+            var entry = new JobResult(TaskId: taskId, MachineIdMetric: machineIdMetric, JobId: jobId, Project: job.Config.Project, Name: job.Config.Name, resultType, resultValue);
+            var r = await Insert(entry);
             if (!r.IsOk) {
-                throw new InvalidOperationException($"failed to insert job result {jobResult.JobId}");
+                throw new InvalidOperationException($"failed to insert job result with taskId {taskId} and machineId+metricType {machineIdMetric}");
+            }
+        } else if (resultType.Equals("CoverageData") && jobResult.MetricValue["rate"] < resultValue["rate"]) {
```
chkeita, line 37: For readability can you factor out these conditions as individual method or inner methods.

chkeita, line 37: another readability suggestion is to use a switch expression instead of if else here.

nharper285: I'd actually originally used a switch(), but changed it to the condition so I could include the value comparisoin.

chkeita: you can use a [case guard](https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/operators/switch-expression#case-guards) to include a condition in you switch expression

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -15,86 +16,52 @@ public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext con
...
+
+        ResultVoid<(HttpStatusCode Status, string Reason)> r;
+        switch (resultType) {
+            case "CoverageData" when oldEntry.MetricValue["rate"] < newEntry.MetricValue["rate"]:
+                r = await Replace(newEntry);
                 break;
-            case JobResultType.RuntimeStats:
-                double newTotalIterations = resultValue["total_count"];
-                newResult = result with { IterationCount = newTotalIterations };
+            case "RuntimeStats" when oldEntry.MetricValue["total_count"] < newEntry.MetricValue["total_count"]:
```
chkeita, line 45: Can we add a type that represent the values instead of a dictionary.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -15,86 +16,52 @@ public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext con
...
+            var result = await Insert(newEntry);
+            if (!result.IsOk) {
+                throw new InvalidOperationException($"failed to insert job result with taskId {taskId} and machineId+metricType {machineIdMetric}");
+            }
+            return true;
+        }
+
+        ResultVoid<(HttpStatusCode Status, string Reason)> r;
+        switch (resultType) {
+            case "CoverageData" when oldEntry.MetricValue["rate"] < newEntry.MetricValue["rate"]:
```
chkeita, line 42: use constants or an enum type for the type name.

Fell free to create an issue if you prefer to address it later.

nharper285: I originally did this, but reverted it to string types so that we wouldn't have to hardcode a bunch of metric names. This makes them metrics generalized for future use.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: Like the below, we need the value types to be generic. We can't have a struct MetricValue { count = 0, } because some of the metrics don't use 'count' as there aggregate. RuntimeStats and CoverageData both use different values. Using a dictionary allows us to not be restrained by hard-coded values when considering future metrics to include.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -918,25 +905,50 @@ public record SecretAddress<T>(Uri Url) : ISecret<T> {
 public record SecretData<T>(ISecret<T> Secret) {
 }
 
+[SkipRename]
+public enum JobResultType {
+    CoverageData,
+    RuntimeStats,
+}
+
 public record JobResult(
```
Porges, line 914: How are we going to migrate the existing data, or will we just drop the existing table?

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,69 @@
...
 
-    public async Async.Task<JobResult?> GetJobResult(Guid jobId) {
-        return await SearchByPartitionKeys(new[] { jobId.ToString() }).SingleOrDefaultAsync();
+    public async Async.Task<JobResult?> GetJobResult(Guid taskId, string machineIdMetric) {
+        var data = QueryAsync(Query.SingleEntity(taskId.ToString(), machineIdMetric));
+        return await data.FirstOrDefaultAsync();
     }
 
-    private JobResult UpdateResult(JobResult result, JobResultType type, Dictionary<string, double> resultValue) {
+    private async Async.Task<bool> TryUpdate(Job job, Guid taskId, Guid machineId, string resultType, Dictionary<string, double> resultValue) {
```
Porges, line 27: I'd make this take just the jobId since we don't need the whole job object

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,69 @@
...
-                double newTotalIterations = resultValue["total_count"];
-                newResult = result with { IterationCount = newTotalIterations };
+            case RUNTIME_STATS when oldEntry.MetricValue["total_count"] < newEntry.MetricValue["total_count"]:
+                r = await Replace(newEntry);
                 break;
             default:
-                _logTracer.LogWarning($"Invalid Field {type}.");
+                _logTracer.LogInformation($"attempt to update job result {taskId} and machineId+metricType {machineIdMetric}");
+                oldEntry.MetricValue["count"]++;
+                var newResult = oldEntry with { MetricValue = oldEntry.MetricValue };
```
Porges, line 54: this is a no-op

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,69 @@
...
+            var result = await Insert(newEntry);
+            if (!result.IsOk) {
+                throw new InvalidOperationException($"failed to insert job result with taskId {taskId} and machineId+metricType {machineIdMetric}");
+            }
+            return true;
+        }
+
+        ResultVoid<(HttpStatusCode Status, string Reason)> r;
+        switch (resultType) {
+            case COVERAGE_DATA when oldEntry.MetricValue["rate"] < newEntry.MetricValue["rate"]:
```
Porges, line 45: I think you want to nest the conditionals here, place an if statement inside the case - if you get a new entry with a lower "rate" value than the existing one you don't want to go to the default case

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,69 @@
...
+                r = await Replace(newEntry);
```
Porges, line 46: we should use update instead of replace

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,69 @@
...
+    const string COVERAGE_DATA = "CoverageData";
+    const string RUNTIME_STATS = "RuntimeStats";
+
     public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext context)
         : base(log, context) {
     }
 
-    public async Async.Task<JobResult?> GetJobResult(Guid jobId) {
-        return await SearchByPartitionKeys(new[] { jobId.ToString() }).SingleOrDefaultAsync();
+    public async Async.Task<JobResult?> GetJobResult(Guid taskId, string machineIdMetric) {
```
Porges, line 22: I'd make this method take machine ID + metric type separately - the way the data is stored in the table is a concern of this class only

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -918,25 +905,50 @@ public record SecretAddress<T>(Uri Url) : ISecret<T> {
...
-) : EntityBase() {
-    public JobResult(Guid JobId, string Project, string Name) : this(
-        JobId: JobId,
-        Project: Project,
-        Name: Name, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0) { }
+    string Type,
+    Dictionary<string, double> MetricValue
+) : EntityBase();
+
+public record CoverageMetricValue(IDictionary<string, double> inner) {
```
Porges, line 924: these classes not used yet, planned for future?

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges, line 22: We'll also want some functionality that fetches all results for a particular job. I'm wondering if the jobID should be the partition and then the rowkey is task/machine/type? This way you can query a whole partition by passing the jobID, which would solve the problem nicely.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
nharper285: Dropping the table will likely be the path forward. We don't currently retain data past 30 days (bc of app insights) and haven't yet exposed the data from the original design.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: Hm not sure what you mean. I increment the 'count' key's value and then update. I've tested and this works as expected.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: Good call out. I've updated.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: I purposefully used replace over update because it significantly decreases the collisions inherent to updating the table. Because we only actually need the max, most recent value from the metric (iteration count, coverage rate, etc.), we can overwrite the existing entry without having to worry about updating in the right order.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
nharper285: Working with Cheick to see if this is something we want to use instead of having to key into the dictionaries during the metric update in JobOperations

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: I like that logic. Updated.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges: I mean you might as well use the `oldEntry` object, rather than making a new one with the same values.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges: We _shouldn't_ get collisions here with Update since we're expecting this value to be updated in a "single-threaded" manner with only one writer. But it protects us in the case that there has been a problem and two updates come in at once:

1. "Thread" A gets entity, rate = 0
2. "Thread" B gets entity, rate = 0 
3. B compares with rate 2, it's bigger, so writes entity
4. A compares with rate 1, it's bigger, so writes entity - rate has been decreased from 2 to 1

----

Alternately, if we really don't care about checking for collisions on this entity (we can make a strong assumption that the updates are "single-threaded"), then we can do a simple Replace without fetching the entity first, since we don't actually need to compare with the existing value. This would half the number of table operations performed.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,85 @@
...
-        return newResult;
+    public async Async.Task<JobResult?> GetJobResults(Guid jobId) {
+        var data = QueryAsync(Query.PartitionKey(jobId.ToString()));
+        return await data.FirstOrDefaultAsync();
     }
 
-    private async Async.Task<bool> TryUpdate(Job job, JobResultType resultType, Dictionary<string, double> resultValue) {
+    private async Async.Task<bool> TryUpdate(Job job, Guid taskId, Guid machineId, string resultType, Dictionary<string, double> resultValue) {
         var jobId = job.JobId;
+        var taskIdMachineIdMetric = string.Concat(taskId, machineId, resultType);
```
Porges, line 35: For readability in the table it would probably be good to separate these values with `|` or something.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: There can be multiple instances of the same function running at the same time, so I don't think we can treat the function as one writer. Your example makes a good case for Update, though. I was just trying to reduce collisions, which definitely seem to be happening - it's why I initially wrote the `polly` policy. 

The latter scenarios would not be possible, though. It's entirely possible for files that get uploaded to the queue by the agent get out of order, no? Furthermore, with multiple function workers, items can get out of order as they are de-queued.

Porges: The "one writer" that I'm thinking is per job/task/machine/resultType. But yes if this isn't coming from the agent itself but instead from one of the queues then we can't assume that.

tevoinea: It's impossible to avoid collisions when writing to the table unless we use a lock. I don't really understand why ordering is an issue here, but to solve it you have 2 options (assuming we're also using a lock):

1. Include a UTC timestamp in the message and only update if the timestamp is newer. We know a collision in a row of the JobResult table can only happen if the messages come from the same machine. Since the messages come from the same machine, the timestamps from the messages come from the same clock and we can assume the clock is internally consistent so we don't need to worry about any clock sync issues.
2. Don't store order sensitive data. Instead of storing a 'rate', and needing to keep track of the latest value; store 'iteration count' which only requires incrementing the value (which is safe under a lock) and a start time

nharper285: @tevoinea, ordering is an issue for both the `CoverageData` and the `RuntimeStats` metrics. `CoverageData` has three metrics that we want to track (rate, covered, & features). `RuntimeStats` has iteration_count. 

All of these metric fields require we take the most recent max. Iteration count sends the total iterations up to time of emission, so your point about 2 does not hold. This is also why ordering matters. If a bunch of `RuntimeStats` end up in the queue out of order, then we need to take only take the max value up to that point. 

Wrt to your point on 1., this had already been suggested, but I considered the value comparison (only update when newRuntimeStats > oldRuntimeStats) to be equivalent. If we don't think this is sufficient then I can just include the timestamp and run with that.

tevoinea: @nharper285 makes sense, the benefit I see with using timestamp is that we can use the same comparison method for all event types meaning we don't need to check `rate` for coverage, `total_count` for runtime_stats, etc. Also, it will still work if we ever produce event types that don't contain ordinal data.

nharper285: That's a good point! I'll make the update.

nharper285: @tevoinea, let me know when you get a chance to review updates.

tevoinea (approved):
LGTM! If you want, you can update the return type of `TryUpdate` to `Async.Task<OnefuzzResultVoid>` to communicate the fallibility of the function instead of throwing but don't consider this a blocking comment.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
 using Microsoft.Extensions.Logging;
 using Polly;
 namespace Microsoft.OneFuzz.Service;
+using System.Net;
 
 public interface IJobResultOperations : IOrm<JobResult> {
 
-    Async.Task<JobResult?> GetJobResult(Guid jobId);
-    Async.Task<OneFuzzResultVoid> CreateOrUpdate(Guid jobId, JobResultType resultType, Dictionary<string, double> resultValue);
+    Async.Task<JobResult?> GetJobResult(Guid jobId, Guid taskId, Guid machineId, string metricType);
```
stishkin, line 9: are you reverting the change here ?

src/ApiService/ApiService/OneFuzzTypes/Model.cs
```diff
@@ -918,26 +906,23 @@ public record SecretAddress<T>(Uri Url) : ISecret<T> {
...
 }
 
+[SkipRename]
+public enum JobResultType {
+    CoverageData,
+    RuntimeStats,
+}
+
 public record JobResult(
-    [PartitionKey][RowKey] Guid JobId,
```
stishkin, line 922: changing partition key and row key to a new one - that means that all previous metrics will be pretty much impossible to retrieve. 

Is that what you want here ? Starting from clean slate after this PR is deployed ?

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
...
+        return await data.FirstOrDefaultAsync();
     }
 
-    private async Async.Task<bool> TryUpdate(Job job, JobResultType resultType, Dictionary<string, double> resultValue) {
+    private async Async.Task<bool> TryUpdate(Job job, Guid taskId, Guid machineId, DateTime createdAt, string resultType, Dictionary<string, double> resultValue) {
         var jobId = job.JobId;
+        var taskIdMachineIdMetric = string.Concat(taskId, machineId, resultType);
 
-        var jobResult = await GetJobResult(jobId);
+        var oldEntry = await GetJobResult(jobId, taskId, machineId, resultType);
```
stishkin, line 37: what happens to entries with old schema ? The job is already running and you are swapping the schema

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: Changing the GetJobResult params? Yes, all four are needed.

src/ApiService/ApiService/OneFuzzTypes/Model.cs
nharper285: Yes. We are going to start from a clean slate. Data is already only kept for 30 days right now, so we will just let data build up for thirty days or so and then switch over.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: We are starting from a clean slate.

stishkin (approved):

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges: > I don't really understand why ordering is an issue here

It's an issue because these updates are apparently driven by queue messages. We don't know how long they've been on the queue or how many other processors of queue messages there are. 

If this was an HTTP API that was being directly invoked from the agent machine then we could simply do the update without checking (this is the "one writer" scenario).

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
...
-            if (!r.IsOk) {
-                throw new InvalidOperationException($"failed to insert job result {jobResult.JobId}");
-            }
-            _logTracer.LogInformation("updated job result {JobId}", jobResult.JobId);
+        ResultVoid<(HttpStatusCode Status, string Reason)> r;
+        switch (resultType) {
+            case COVERAGE_DATA:
+            case RUNTIME_STATS:
+                if (oldEntry.CreatedAt < createdAt) {
+                    oldEntry = oldEntry with { MetricValue = resultValue };
```
Porges, line 54: You'll also need to update `CreatedAt`

nharper285: What's the issue here?

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
...
+    const string RUNTIME_STATS = "RuntimeStats";
+
     public JobResultOperations(ILogger<JobResultOperations> log, IOnefuzzContext context)
         : base(log, context) {
     }
 
-    public async Async.Task<JobResult?> GetJobResult(Guid jobId) {
-        return await SearchByPartitionKeys(new[] { jobId.ToString() }).SingleOrDefaultAsync();
+    public async Async.Task<JobResult?> GetJobResult(Guid jobId, Guid taskId, Guid machineId, string metricType) {
+        var data = QueryAsync(Query.SingleEntity(jobId.ToString(), string.Concat(taskId, machineId, metricType)));
```
Porges, line 24: I still think the RowKey should have some kind of delimiter in it :)

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
...
-                break;
-            default:
-                _logTracer.LogWarning($"Invalid Field {type}.");
-                break;
-        }
-        _logTracer.LogInformation($"Attempting to log new result: {newResult}");
-        return newResult;
+    public async Async.Task<JobResult?> GetJobResults(Guid jobId) {
+        var data = QueryAsync(Query.PartitionKey(jobId.ToString()));
+        return await data.FirstOrDefaultAsync();
```
Porges, line 30: This should return an IAsyncEnumerable, right? there's a result per task/machine/metric

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
```diff
@@ -2,99 +2,79 @@
...
+        switch (resultType) {
+            case COVERAGE_DATA:
+            case RUNTIME_STATS:
+                if (oldEntry.CreatedAt < createdAt) {
+                    oldEntry = oldEntry with { MetricValue = resultValue };
+                    r = await Update(oldEntry);
+                    if (!r.IsOk) {
+                        throw new InvalidOperationException($"failed to replace job result with taskId {taskId} and machineId+metricType {taskIdMachineIdMetric}");
+                    }
+                }
```
Porges, line 59: It might be useful to log if we have an out-of-date metric

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges: If never updated, CreatedAt will always be the original timestamp of the metric, not the latest one. It also means this check doesn't stop an earlier value overwriting a later one

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: What would your suggestion be - you can't insert non-characters (-, |, etc.) into the names of these values.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: I'm deleting for now, as the function isn't used anywhere.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: I'm not sure I see what's wrong with "createdAt will always be the original timestamp of the metric" - we _want_ the table entry to only reflect the last time it was updated.

nharper285: Oh I see. Of course, good catch.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
nharper285: DOne.

src/ApiService/ApiService/onefuzzlib/JobResultOperations.cs
Porges: I would join them together with `|` just to make it easier to separate the GUIDs from each other in case someone is visually inspecting the data.

PR #3549, opened 2023-10-05
Fix default windows interceptor list

tevoinea:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/agent/coverage/src/allowlist.rs
+++ b/src/agent/coverage/src/allowlist.rs
@@ -75,11 +75,12 @@ impl AllowList {
     }
 
     /// Build a new `Allowlist` that adds the allow and deny rules of `other` to `self`.
-    pub fn extend(&self, other: &Self) -> Self {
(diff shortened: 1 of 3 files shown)


src/agent/coverage/src/allowlist.rs
```diff
@@ -75,11 +75,12 @@ impl AllowList {
     }
 
     /// Build a new `Allowlist` that adds the allow and deny rules of `other` to `self`.
-    pub fn extend(&self, other: &Self) -> Self {
+    pub fn extend(&mut self, other: &Self) {
```
chkeita, line 78: you are changing he semantic of this function here. i suggest you create a new one called `extend_in_place`. To indicate that the AllowList is updated instead of copied.

chkeita: can you also update the description

chkeita (approved):

Issue #3024, opened 2023-04-17
Some parameters are not available for expansion.


chkeita:
The variable expander in the agent is not initialized with `crashes` value. 

All available parameters should be available for expansion.

PR #3553, opened 2023-10-05
Add a `GetExpand` trait for initializing `Expand` objects in our tasks

kananb:

Implements a slightly more standardized way for initializing `Expand` objects by adding a `GetExpand` trait that defines a similarly named function. Each config then implements it to generate the `Expand` object.

Property-based unit tests are also added to ensure that all of the variables that are expected to be available for expansion, in fact, are.

A few things that could use further consideration:
1. Expand object parameters sometimes come from the function's arguments, not always from just the config fields.
    1. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-641f936686cd3425c7b3548eb94e578cfc75f0c47e2c8a4eb1bf631d4167cebeR362) in the generic coverage task.
    2. [`generated_inputs`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-6078c3382462b84cb0e327e615d619ba66792220906b831784f550df10b44015R189) and [`input_corpus`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-6078c3382462b84cb0e327e615d619ba66792220906b831784f550df10b44015R190) in the generator task.
    3. [`runtime_dir`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR294) and [`crashes`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR295) in the supervisor task.
    4. [`generated_inputs`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-739859a740727a2fe41beb162d86d4e2f880e1017dc05227e3bc5112c9cb6907R148) in the merge task.
5. Sometimes a value is calculated asynchronously and then passed to the `Expand` object. I'm not sure how to accommodate this from the trait implementation when the value is calculated using only config fields.
    1. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-22a0a626e92341671e0dc52ea872d4075f7acfa48f3317b462f7d12bc9eda91dR243) in the analysis task.
    2. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-ea6d900dee86018caa510bd9e157bf57009a902c8bf158e0d667d7dd239e0d67R310) in the dotnet coverage task.
    3. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR301) in the supervisor task.
    4. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-739859a740727a2fe41beb162d86d4e2f880e1017dc05227e3bc5112c9cb6907R149) in the merge task.
6. In a few places, a function argument is passed to the `Expand` object that overrides the value that was passed from the config.
    1. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-22a0a626e92341671e0dc52ea872d4075f7acfa48f3317b462f7d12bc9eda91dR243) and [`reports_dir`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-22a0a626e92341671e0dc52ea872d4075f7acfa48f3317b462f7d12bc9eda91dR244) in the analysis task.
    2. [`target_exe`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-ea6d900dee86018caa510bd9e157bf57009a902c8bf158e0d667d7dd239e0d67R310) in the dotnet coverage task.
    5. [`input_corpus`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR296), [`reports_dir`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR297), and [`crashdumps`](https://github.com/microsoft/onefuzz/pull/3553/files/1e5e37aa32a7b46f763def22c4d28042faed8f57#diff-c7803bcb6de2085d1be07c3b4ccdba290a85e477ad5334ea61f5a18ee716a64fR298-R299) in the supervisor task.
7. I'm still very new to Rust, so I ~~might~~ probably have done some silly things without realizing.

## PR Checklist
* [ ] Applies to work item: #3024 
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-task/src/tasks/config.rs
```diff
@@ -364,3 +384,79 @@ impl Config {
...
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn test_get_expand_values_match_config(
+            config in any::<CommonConfig>(),
+        ) {
+            // This function implementation is repeated across all config tests
+            // There might be a way to share it by taking advantage of the `GetExpandFields` trait, but I'm not sure how
```
tevoinea, line 449: You could use a macro that takes the config type as an input (in this case: `CommonConfig`) and spits out this `proptest!` macro as an output.

src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
```diff
@@ -56,6 +56,18 @@ impl Config {
     }
 }
 
+impl GetExpand for Config {
+    fn get_expand(&self) -> Result<Expand<'_>> {
+        Ok(self
+            .common
+            .get_expand()?
+            .target_exe(&self.target_exe)
```
tevoinea, line 64: Do we need to expand target_exe if we're going to override it later here?

```rust
        let expand = self
            .config
            .get_expand()?
            .input_path(input)
            .target_exe(&target_exe); <---- here
```

src/agent/onefuzz-task/src/tasks/fuzz/supervisor.rs
```diff
@@ -252,57 +290,19 @@ async fn start_supervisor(
...
 
-    let expand = Expand::new(&config.common.machine_identity)
-        .machine_id()
-        .supervisor_exe(&config.supervisor_exe)
-        .supervisor_options(&config.supervisor_options)
+    let expand = config
+        .get_expand()?
         .runtime_dir(&runtime_dir)
         .crashes(&crashes.local_path)
+        .input_corpus(&inputs.local_path) // Why isn't this value in the config? It's not super clear to me from looking at the calling code.
```
tevoinea, line 297: You could create an issue to look into that after this PR merges. I have a similar gut feeling that we _shouldn't_ need it and it would clean up the code by removing it.

tevoinea: If you create the issue, make sure to include all of the locations in the code that we know about that have this potential issue.

src/agent/onefuzz-task/src/tasks/merge/generic.rs
```diff
@@ -33,13 +36,28 @@ pub struct Config {
     pub target_options_merge: bool,
     pub tools: SyncedDir,
     pub input_queue: Url,
-    pub inputs: SyncedDir,
+    pub inputs: SyncedDir, // Is this something we can pass to the expander?
```
tevoinea, line 39: Not sure if this is an old comment but: it's part of `Config`, so yes when you `impl GetExpand for Config` and define a function that takes `&self`, you have access to the properties of the struct referenced by `self`.

kananb: Sorry, that's a bad comment. What I should have written is: is there / what is the expander variable that that field corresponds to?

src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
kananb: My intention for setting it in both places was to have a sort of insurance policy in case new code is ever added that initializes an `Expand` object and doesn't / forgets to set the `target_exe`.

You're right though, it's effectively dead code right now. I'd be very easily convinced to remove it.

src/agent/onefuzz-task/src/tasks/fuzz/supervisor.rs
kananb: https://github.com/microsoft/onefuzz/issues/3565

chkeita (approved):

src/agent/onefuzz-task/Cargo.toml
```diff
@@ -85,4 +85,6 @@ schemars = { version = "0.8.12", features = ["uuid1"] }
 
 [dev-dependencies]
 pretty_assertions = "1.4"
+proptest = "1.2.0"
```
Porges, line 88: Why not use the latest version? 1.3.1

src/agent/onefuzz-task/src/tasks/fuzz/generator.rs
```diff
@@ -239,95 +288,99 @@ mod tests {
...
+                    extra_output: Default::default(),
+                    min_available_memory_mb: Default::default(),
+                    machine_identity: onefuzz::machine_id::MachineIdentity {
+                        machine_id: uuid::Uuid::new_v4(),
+                        machine_name: "test".to_string(),
+                        scaleset_name: None,
+                    },
+                    tags: Default::default(),
+                    from_agent_to_task_endpoint: "/".to_string(),
+                    from_task_to_agent_endpoint: "/".to_string(),
```
Porges, line 372: If `CommonConfig` implements `Default` then this can all be replaced by `CommonConfig { some_field: X, ..CommonConfig::default() }`

src/agent/onefuzz/src/expand.rs
```diff
@@ -93,6 +93,10 @@ impl PlaceHolder {
     }
 }
 
+pub trait GetExpand {
+    fn get_expand(&self) -> Result<Expand<'_>>;
+}
```
Porges, line 98: I would say don't add the trait, since it is never used to abstract over implementations (we never use `dyn GetExpand` or `T: GetExpand` or `impl GetExpand` anywhere). Just add the `get_expand` function to the `impl T`.

src/agent/onefuzz-task/Cargo.toml
kananb: There's no reason, thanks for catching that.

src/agent/onefuzz/src/expand.rs
kananb: I think this is the right move, I don't envision any cases where it could be useful for an abstraction.

src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
tevoinea: It's not the same `target_exe` though right? Setting it in `get_expand` is purely from the config but later, we populate it from `self.target_exe`

src/agent/onefuzz-task/src/tasks/merge/generic.rs
tevoinea: Might be this one? https://github.com/microsoft/onefuzz/blob/679309a87a27251ffabf698fdd9eae2c643467b6/docs/command-replacements.md?plain=1#L14

src/agent/onefuzz-task/src/tasks/coverage/dotnet.rs
kananb: The difference between the two is that the value gotten from `self.target_exe()` is the expanded version of what's in the config and it's guaranteed to exist in the filesystem.
It doesn't make sense to me to try and fit that validation into the config instantiation nor does it make sense to do it just in `get_expand()`. So, I think you're right, I'll remove it from `get_expand()` and just let it be initialized in that function.

src/agent/onefuzz-task/src/tasks/merge/generic.rs
kananb: The only thing that gives me pause is that `supervisor.rs` has a similarly named and typed field in it's config that is used to expand the `input_corpus`, but this config (`merge/generic.rs`) is already providing a value for that (`input_corpus`) expansion variable.
But in the context of being in the merge task, `inputs` corresponding to `generated_inputs` does make sense I think.

PR #3555, opened 2023-10-06
Return the failure from NotifyAdo as result  instead of an exception

### LINKED ISSUE #3554: Notification test endpoint should return the validation failure if the template is invalid

1. use  `onefuzz debug notification test_template` with an invalid template
2.
3.

### Expected result

A response containing the validation failure

### Actual result

500 status code 

[AB#167404](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/167404)

chkeita:

closes #3554


src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -58,13 +58,7 @@ public class QueueFileChanges {
...
         try {
-            // Setting isLastRetryAttempt to false will rethrow any exceptions
-            // With the intention that the azure functions runtime will handle requeing
-            // the message for us. The difference is for the poison queue, we're handling the
-            // requeuing ourselves because azure functions doesn't support retry policies
-            // for queue based functions.
-
-            var result = await FileAdded(storageAccount, fileChangeEvent, isLastRetryAttempt: false);
+            var result = await FileAdded(storageAccount, fileChangeEvent);
             if (!result.IsOk && result.ErrorV.Code == ErrorCode.ADO_WORKITEM_PROCESSING_DISABLED) {
```
tevoinea, line 62: We're expecting FileAdded not to throw anymore so you need to update this if condition.

ADO_WORKITEM_PROCESSING_DISABLED -> Requeue 1 day
Other Errors -> Requeue with default visibility timeout

tevoinea: Also I found a bug unrelated to your change but when we receive `ADO_WORKITEM_PROCESSING_DISABLED` and we requeue for a day, we still increment the retry count, which we shouldn't be doing since the work item creation should be 'paused'.
I'll make an issue for that.

src/ApiService/ApiService/Functions/QueueFileChanges.cs
```diff
@@ -58,23 +58,19 @@ public class QueueFileChanges {
...
                 await RequeueMessage(msg, TimeSpan.FromDays(1));
+            } else {
+                await RequeueMessage(msg);
```
tevoinea, line 65: This will requeue even if result.IsOk == true

tevoinea (approved):

PR #3564, opened 2023-10-11
Add middleware for optional strict CLI version checking

kananb:

Adds middleware to the ApiService that checks for two headers: `cli-version` and `strict-version`
If both headers are present _and_ the `strict-version` is set to true, then the middleware will compare the supplied version with the service version and immediately send back a 400 response to requests with out of date versions.

On the CLI end, this feature is opted into by setting the `ONEFUZZ_STRICT_VERSIONING` environment variable to true (case insensitive).

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

Added unit tests for the following cases:
* [x] neither header provided
* [x] one header provided but not the other
* [x] both headers, but `strict-version` != true
* [x] both headers and `strict-version` == true

(Both valid and invalid values are tested for each case, where applicable)

### DIFF
--- a/src/ApiService/ApiService/Program.cs
+++ b/src/ApiService/ApiService/Program.cs
@@ -58,6 +58,54 @@ public class Program {
         }
     }
 
+    /// <summary>
+    /// 
+    /// </summary>
+    public class VersionCheckingMiddleware : IFunctionsWorkerMiddleware {
+        private readonly Version _oneFuzzServiceVersion;
+        private readonly IRequestHandling _requestHandling;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="oneFuzzServiceVersion"></param>
+        public VersionCheckingMiddleware(IServiceConfig config, IRequestHandling requestHandling) {
+            _oneFuzzServiceVersion = Version.Parse(config.OneFuzzVersion);
+            _requestHandling = requestHandling;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="next"></param>
+        /// <returns></returns>
+        public async Async.Task Invoke(FunctionContext context, FunctionExecutionDelegate next) {
+            var requestData = await context.GetHttpRequestDataAsync();
+            if (requestData is not null) {
+                var doStrictVersionCheck =
+                    requestData.Headers.TryGetValues("Cli-Version", out var cliVersion) &&
+                    requestData.Headers.TryGetValues("Strict-Version", out var strictVersion)
+                    && strictVersion?.FirstOrDefault()?.Equals("true", StringComparison.InvariantCultureIgnoreCase) == true; // "== true" necessary here to avoid implicit null -> bool casting
+
+                if (doStrictVersionCheck) {
+                    if (!Version.TryParse(cliVersion?.FirstOrDefault() ?? "", out var version)) {
+                        var response = await _requestHandling.NotOk(requestData, Error.Create(ErrorCode.INVALID_REQUEST, "unable to parse version string in 'Cli-Version' header"), "version middleware");
+                        context.GetInvocationResult().Value = response;
+                        return;
+                    }
+                    if (version < _oneFuzzServiceVersion) {
+                        var response = await _requestHandling.NotOk(requestData, Error.Create(ErrorCode.INVALID_CLI_VERSION, "cli is out of date"), "version middleware");
+                        context.GetInvocationResult().Value = response;
+                        return;
+                    }
+                }
+            }
+
+            await next(context);
+        }
+    }
+
 
     //Move out expensive resources into separate class, and add those as Singleton
     // ArmClient, Table Client(s), Queue Client(s), HttpClient, etc.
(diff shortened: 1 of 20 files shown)


src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,54 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+        /// 
+        /// </summary>
+        /// <param name="oneFuzzServiceVersion"></param>
+        public VersionCheckingMiddleware(IServiceConfig config, IRequestHandling requestHandling) {
+            _oneFuzzServiceVersion = Version.Parse(config.OneFuzzVersion);
+            _requestHandling = requestHandling;
+        }
+
+        /// <summary>
+        ///
```
chkeita, line 78: can you remove those empty comments if you are not filling the description

src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,54 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+        /// 
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="next"></param>
+        /// <returns></returns>
+        public async Async.Task Invoke(FunctionContext context, FunctionExecutionDelegate next) {
+            var requestData = await context.GetHttpRequestDataAsync();
+            if (requestData is not null) {
+                var doStrictVersionCheck =
+                    requestData.Headers.TryGetValues("Cli-Version", out var cliVersion) &&
```
chkeita, line 87: Move the headers names to constants

src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,54 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
         }
     }
 
+    /// <summary>
+    /// 
+    /// </summary>
+    public class VersionCheckingMiddleware : IFunctionsWorkerMiddleware {
```
mgreisen, line 64: Do we have unit tests for these changes?

kananb: Unit tests have been added.

src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,67 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+        /// Initializes an instance of <see cref="VersionCheckingMiddleware"/> with the provided config and request handling objects.
+        /// </summary>
+        /// <param name="config">The service config containing the service version.</param>
+        /// <param name="requestHandling">The request handling object to create HTTP responses with.</param>
+        public VersionCheckingMiddleware(IServiceConfig config, IRequestHandling requestHandling) {
+            _oneFuzzServiceVersion = Version.Parse(config.OneFuzzVersion);
+            _requestHandling = requestHandling;
+        }
+
+        public Error? TestCliVersion(Azure.Functions.Worker.Http.HttpHeadersCollection headers) {
```
chkeita, line 80: i would rename this to `CheckCliVersion` instead `TestCliVersion` makes it sounds like this is only used for unit tests.

chkeita (approved):

src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,67 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+        /// Initializes an instance of <see cref="VersionCheckingMiddleware"/> with the provided config and request handling objects.
+        /// </summary>
+        /// <param name="config">The service config containing the service version.</param>
+        /// <param name="requestHandling">The request handling object to create HTTP responses with.</param>
+        public VersionCheckingMiddleware(IServiceConfig config, IRequestHandling requestHandling) {
+            _oneFuzzServiceVersion = Version.Parse(config.OneFuzzVersion);
+            _requestHandling = requestHandling;
+        }
+
+        public Error? CheckCliVersion(Azure.Functions.Worker.Http.HttpHeadersCollection headers) {
```
tevoinea, line 80: This could be a `OnefuzzResultVoid` so we don't have to encode null as success, we can return `Ok` instead.

src/ApiService/ApiService/Program.cs
```diff
@@ -58,6 +58,67 @@ public class LoggingMiddleware : IFunctionsWorkerMiddleware {
...
+            var doStrictVersionCheck =
+                headers.TryGetValues(StrictVersionHeader, out var strictVersion)
+                && strictVersion?.FirstOrDefault()?.Equals("true", StringComparison.InvariantCultureIgnoreCase) == true; // "== true" necessary here to avoid implicit null -> bool casting
+
+            if (doStrictVersionCheck) {
+                if (!headers.TryGetValues(CliVersionHeader, out var cliVersion)) {
+                    return Error.Create(ErrorCode.INVALID_REQUEST, $"'{StrictVersionHeader}' is set to true without a corresponding '{CliVersionHeader}' header");
+                }
+                if (!Version.TryParse(cliVersion?.FirstOrDefault() ?? "", out var version)) {
```
tevoinea, line 89: `System.Version` doesn't support our dev versioning scheme. 

>... All defined components must be integers greater than or equal to 0. ...
https://learn.microsoft.com/en-us/dotnet/api/system.version?view=net-7.0#remarks

Take a look at this code sample I made with the current version in my dev deployment. https://dotnetfiddle.net/44IZLZ

We need to use a library that supports semver for version parsing.

kananb: Thanks for catching that! I went for [Semver](https://github.com/maxhauser/semver) for parsing to replace System.Version.

kananb:
> Semver parsing of dev deploys is blocking this PR

Changed version parsing to use [the Semver library](https://github.com/maxhauser/semver) and included versions with metadata and prerelease components to the unit tests.

tevoinea (approved):

Issue #3566, opened 2023-10-11
New `UniqueReport` Metric Type - Represents all reports that we attempt to upload


nharper285:
We would like to know how often we attempt to upload a 'unqiue report', right now we only emit metrics when new reports are successfully uploaded.

PR #3567, opened 2023-10-11
Create new `CrashReported` metric that track report upload attempts

nharper285:

_What is this about?_
We would like to know how often we attempt to upload a 'unqiue report', right now we only emit metrics when new reports are successfully uploaded

## PR Checklist
* [ ] Applies to work item: #3566
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-task/src/tasks/report/crash_report.rs
```diff
@@ -168,6 +168,14 @@ impl CrashTestResult {
                 // Use SHA-256 of call stack as dedupe key.
                 if let Some(unique_reports) = unique_reports {
                     let name = report.unique_blob_name();
+                    if let Some(jr_client) = jr_client {
```
chkeita, line 171: let's move this logic outside of the `unique_report` container check.
and rename to `crash_reported`

chkeita (approved):

Issue #3572, opened 2023-10-13
`debug logs get` should always download all in-scope log files regardless of `--all` and `--last`


kananb:
Currenlty, the default behavior is to download the single most recent file for a single machine or for each task or machine depending on the other arguments.
There are two options: `--all` to download all files and `--last n` to download only the most recent `n` files where `n` defaults to `1`.

Our machines will only ever have a single log file associated with them now and a user providing a `--task_id` with no other options will likely want/expect to receive the most recent logs associated with the task. These "most recent logs" might be split across multiple machines.

So, the default and, I argue, only behavior should be to download all of the logs associated with the most specific option provided. In other words:
If `--machine_id` is provided, download the single log file associated with that machine.
If `--task_id` is provided, download the log file for every machine associated with that task.
If `--job_id` is provided, download all log files belonging to the machines associated with every task that is associated with that job.

[AB#167800](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/167800)

Issue #3490, opened 2023-09-07
Update message when downloading logs


mgreisen:
## Information

- Onefuzz version: 8.8
- OS: n/a

## Provide detailed reproduction steps (if any)

1. download the logs using the command "onefuzz debug logs get --job_id xxxx"

The statement that is returned to the console is:
"INFO:onefuzz:Downloading only the 1 most recent files"

It is unclear that is the most recent file for each task. 

Using alternate phrasing easily clears this up by adding "for each task" to the end. 

[AB#165341](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/165341)

stishkin:
There is no "most recent" anymore, since we should be downloading "all" logs by default

kananb:
From what I understand, unless the `--all` flag is used only the `--last n` most recent logs are downloaded where `n` defaults to `1`.

PR #3573, opened 2023-10-13
Make `debug logs get` behave more intuitively

kananb:

A few facets to this PR:
1. Add an optional `--out_dir` parameter to specify where to download the logs.
2. Deprecate the `--last` and `--all` options, making them no-ops.
3. Update the behavior to now always download all available logs (see #3572).
4. Update the logs to the console to reflect the new behavior.

## PR Checklist
* [x] Applies to work item: #3490, #3572
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

Local validation

### DIFF
--- a/src/cli/onefuzz/debug.py
+++ b/src/cli/onefuzz/debug.py
@@ -712,7 +712,7 @@ class DebugLog(Command):
 
         if not all:
             # Don't want to say we're downloading 2 files right after saying only 1 was found
-            last = min(last, len(files))
+            last = min(last, len(files)) if last == None else len(files)
             files = files[:last]
             if granularity == "job":
                 self.logger.info(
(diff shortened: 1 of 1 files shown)


mgreisen (approved):

src/cli/onefuzz/debug.py
```diff
@@ -707,14 +716,23 @@ def __init__(self, name: str, creation_time: datetime):
...
             return None
 
-        if not all:
-            self.logger.info(f"Downloading only the {last} most recent files")
-            files = files[:last]
+        if granularity == "job":
+            self.logger.info(
+                f"Downloading all of the log files for each task associated with the job with id {job_id}"
+            )
+        elif granularity == "task":
```
chkeita, line 723: these are branches are no mutually exclusive. 
We can specify job, task and machine at the same time

kananb: I think we should always download according to the most specific parameter provided. Otherwise, we could never download the log file for just a single machine.
Do you have something else in mind?

chkeita: oh i see. I missed the logic to set the granularity

chkeita (approved):

Issue #3518, opened 2023-09-20
Allow the min_available_memory_mb  to be configurable per task


chkeita:
the value of  [min_available_memory_mb ](https://github.com/chkeita/onefuzz/blob/21374b36e675aa34ac0ceafdb4944166afe98982/src/agent/onefuzz-task/src/tasks/config.rs#L60C9-L60C32) is currently hardcoded. it should be configurable per tasks.

[AB#166395](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/166395)

PR #3577, opened 2023-10-17
Make `min_available_memory_mb` configurable

kananb:

Makes `min_available_memory_mb` configurable per task from the CLI and service.

## PR Checklist
* [x] Applies to work item: #3518
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed

### DIFF
--- a/src/agent/onefuzz-task/src/managed/cmd.rs
+++ b/src/agent/onefuzz-task/src/managed/cmd.rs
@@ -100,7 +100,7 @@ pub async fn run(args: &clap::ArgMatches) -> Result<()> {
         0 => {
             log::info!("memory watchdog is disabled: this task may fail suddenly if it runs out of memory.");
             config.run().await
-        },
+        }
         _ => {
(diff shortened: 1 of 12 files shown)


src/agent/onefuzz-task/src/managed/cmd.rs
```diff
@@ -96,7 +96,7 @@ pub async fn run(args: &clap::ArgMatches) -> Result<()> {
 
     let min_available_memory_bytes = 1_000_000 * config.common().min_available_memory_mb;
 
-    // If the memory limit is 0, this will resolve immediately with an error.
+    // If the memory limit is 0, this will never return.
     let check_oom = out_of_memory(min_available_memory_bytes);
```
chkeita, line 100: let's disable the check altogether if the memory limit is not >0 instead of launching a task that will loop forever

chkeita (approved):

PR #3579, opened 2023-10-18
Scoped notification pause

### LINKED ISSUE #3556: When requeuing a message because of ADO_WORKITEM_PROCESSING_DISABLED, we should not increment the retry count.
Also I found a bug unrelated to your change but when we receive `ADO_WORKITEM_PROCESSING_DISABLED` and we requeue for a day, we still increment the retry count, which we shouldn't be doing since the work item creation should be 'paused'.
I'll make an issue for that.

_Originally posted by @tevoinea in https://github.com/microsoft/onefuzz/pull/3555#discussion_r1348715529_
            

[AB#167419](https://fuzzfest.visualstudio.com/5eaacf23-2084-4d62-8fbe-7d37326a8b74/_workitems/edit/167419)

tevoinea:

_What is this about?_

* Adds capability to pause notification creation at the container level
* Closes #3556 

## Info on Pull Request

_What does this include?_

* New containers endpoint for modifying container metadata
* New CLI command for modifying container metadata (`onefuzz containers update {container name} {metadata key}={metadata value}`)
* Fixes a bug where messages requeued due to `ADO_WORKITEM_PROCESSING_DISABLED` have their retry count incremented. We should only increment the retry count when we actually _try_ to process the notification
* Do not attempt to process notifications for containers that are tagged with `pauseNotifications=true`. If that's the case, requeue the message with a 1 day timeout

## Validation Steps Performed

Tested in dev deployment:

* Create a file in a container
* Check that a message was added in file-changes queue
* Using the cli, set pauseNotifications=true metadata tag on the container
* Create another file in the container
* Check that a message was added in file-changes queue
* Check that file-changes queue shows "0 of 1 messages" meaning that the message was requeued with a visibility timeout


src/ApiService/ApiService/Functions/Containers.cs
```diff
@@ -16,11 +16,12 @@ public class ContainersFunction {
 
     [Function("Containers")]
     [Authorize(Allow.User)]
-    public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "GET", "POST", "DELETE")] HttpRequestData req)
+    public Async.Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "GET", "POST", "PUT", "DELETE")] HttpRequestData req)
```
chkeita, line 19: PATCH might more appropriate here since the container is partially updated instead of Replaced

chkeita (approved):

PR #3594, opened 2023-10-23
Improve slow Task queries

Porges:
Querying Tasks only by TaskID is slow as it requires querying many partitions. Querying Task by JobID and TaskID is fast as it is a point lookup.

1. rename existing `GetTaskById` to `GetTaskByIdSlow`
2. include JobID in event data from agent (new properties are optional for back-compat)
3. include JobID in NodeTask entity (if this is missing upon lookup we will fall back to `GetTaskByIdSlow`)


src/ApiService/ApiService/Functions/AgentEvents.cs
```diff
@@ -110,18 +110,23 @@ public class AgentEvents {
...
                         $"setup without tasks.  machine_id: {machineId}"
                     );
                 }
 
-                foreach (var taskId in settingUpData.Tasks) {
-                    var task = await _context.TaskOperations.GetByTaskId(taskId);
+                var tasks =
+                    settingUpData.Tasks is not null
+                    ? settingUpData.Tasks.Select(t => (t, _context.TaskOperations.GetByTaskIdSlow(t)))
+                    : settingUpData.TaskData!.Select(t => (t.TaskId, _context.TaskOperations.GetByJobIdAndTaskId(t.JobId, t.TaskId)));
```
stishkin, line 122: TaskData is not null <- is that 100% ?

src/ApiService/ApiService/Functions/AgentEvents.cs
```diff
@@ -110,18 +110,23 @@ public class AgentEvents {
...
+
+                foreach (var (id, taskTask) in tasks) {
```
stishkin, line 124: taskTask 👍

stishkin (approved):

src/ApiService/ApiService/Functions/AgentCanSchedule.cs
```diff
@@ -40,7 +40,12 @@ public class AgentCanSchedule {
         var allowed = canProcessNewWork.IsAllowed;
         var reason = canProcessNewWork.Reason;
 
-        var task = await _context.TaskOperations.GetByTaskId(canScheduleRequest.TaskId);
+        var task = await
+            (canScheduleRequest.JobId.HasValue
+            ? _context.TaskOperations.GetByJobIdAndTaskId(canScheduleRequest.JobId.Value, canScheduleRequest.TaskId)
+            // old agent, fall back
```
tevoinea, line 46: We should also create an issue to remove the old agent case in the release after this one (9.1 I guess?)

src/ApiService/ApiService/onefuzzlib/NodeOperations.cs
```diff
@@ -701,10 +706,18 @@ sealed record NodeInfo(Node Node, Scaleset Scaleset, string InstanceId);
 
     /// returns True on stopping the node and False if this doesn't stop the node
     private async Task<bool> StopIfComplete(Node node, bool done = false) {
-        var nodeTaskIds = await _context.NodeTasksOperations.GetByMachineId(node.MachineId).Select(nt => nt.TaskId).ToArrayAsync();
```
tevoinea, line 704: Removing this `ToArrayAsync` is going to be nice for perf 👍

src/ApiService/ApiService/OneFuzzTypes/Requests.cs
```diff
@@ -81,8 +84,15 @@ public record NodeStateUpdate(
 [JsonConverter(typeof(SubclassConverter<NodeStateData>))]
 public abstract record NodeStateData;
 
+public record NodeSettingUpData(
+    [property: Required] Guid JobId,
```
tevoinea, line 88: There's a small chance of failure here if:

1. Node runs setup script and pulls an 'old' agent binary
2. onefuz service updates
3. onefuzz node emits old setting up event

We can always resubmit so it's not a huge deal, just we should be aware of it when deploying.

tevoinea (approved):

chkeita (approved):

src/ApiService/ApiService/Functions/AgentEvents.cs
Porges: We require either Tasks (old) or TaskData (new) to be set. It's not guaranteed by the schema but the agent code will be sending one or the other, we don't have external clients submitting this.

Porges:
Passed check-pr.

PR #3599, opened 2023-10-24
Increasing Azure App Function Timeout to 12 Hours

nharper285:

_What is this about?_
While we work on a potential redesign of the azure app functions `TimerWorkers` and `TimerTasks` we are increasing the function timeout. 

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


AdamL-Microsoft (approved):

tevoinea (approved):
Do we have a work item for redesigning TimerTasks? We should include undoing this change as part of the required work

Porges (approved):

Porges:
Is there a way to restrict this to certain functions? We have some that should finish in seconds, so ideally we don't want their maximum runtime to be 12 hrs.

nharper285:
> Is there a way to restrict this to certain functions? We have some that should finish in seconds, so ideally we don't want their maximum runtime to be 12 hrs.

Not that I'm aware of. I don't think it's necessary, though, as we will only have this a short time and have a good way of keeping an eye on timer run length via data dashboards.

PR #3600, opened 2023-10-24
Parallelize some timer work

Porges:
When we update distinct entities we can perform the updates in parallel. Do this in `TimerTasks` and `TimerWorkers`.

### DIFF
--- a/src/ApiService/ApiService/Functions/TimerTasks.cs
+++ b/src/ApiService/ApiService/Functions/TimerTasks.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Azure.Functions.Worker;
+﻿using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.Logging;
 namespace Microsoft.OneFuzz.Service.Functions;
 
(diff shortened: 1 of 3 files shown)


src/ApiService/ApiService/Functions/TimerTasks.cs
```diff
@@ -22,35 +23,39 @@ public class TimerTasks {
...
             _logger.LogInformation("stopping expired task. job_id:{JobId} task_id:{TaskId}", task.JobId, task.TaskId);
             await _taskOperations.MarkStopping(task, "task is expired");
         }
 
-
         var expiredJobs = _jobOperations.SearchExpired();
-
-        await foreach (var job in expiredJobs) {
+        // job updates are all distinct and only update tasks owned by that job, can be performed in parallel
+        await Parallel.ForEachAsync(expiredJobs, parallelOptions, async (job, _cancel) => {
```
chkeita, line 39: Another option here is to have an overload to Stopping that takes a list of jobs. We can update all of them in batch we would also save in the number of connection to the db.

chkeita (approved):

src/ApiService/ApiService/Functions/TimerTasks.cs
Porges: You can only do batch updates within a single partition. Unfortunately, for the jobs each job lives in a separate partition so we cannot do this.

PR #3609, opened 2023-10-30
Report when coverage is empty

nharper285:

_What is this about?_

## PR Checklist
* [ ] Tests added/passed

## Info on Pull Request

_What does this include?_

## Validation Steps Performed


src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -174,7 +175,7 @@ impl CoverageTask {
             context.save_and_sync_coverage().await?;
         }
 
-        info!("report coverage");
+        info!("report final coverage");
```
chkeita, line 178: this is not the final coverage.
the task can still report coverage in the loop that reads from the queue.

src/agent/onefuzz-task/src/tasks/coverage/generic.rs
```diff
@@ -447,6 +452,8 @@ impl<'a> TaskContext<'a> {
                             if count % 10 == 0 {
                                 self.save_and_sync_coverage().await?;
                             }
+                            info!("report coverage");
```
chkeita, line 455: don't emit this event here. At this point the coverage is is not really reported yet because we have no update the files.

